PR #3, opened 2017-01-17
Initial draft of stream coalescing for C++

ctiller:
Discussion at: https://groups.google.com/forum/#!topic/grpc-io/GfxB7nuV9GY

This is re-opening/re-creating #2 since GitHub seems to have lost the link between my fork and this repository.

### DIFF
--- a/L1.md
+++ b/L1.md
@@ -1,0 +1,154 @@
+# C++ streaming coalescing API's
+
+* Author(s): ctiller
+* Approver: a11r
+* Status: Draft
+* Implemented in: C++
+* Last updated: 2017/01/13
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/GfxB7nuV9GY
+
+## Abstract
+
+Provide C++ API's to coalesce:
+- initial metadata and first streaming message
+- final streaming message and trailing metadata
+into the same core batch.
+
+## Background
+
+Currently when making a streaming call using gRPC and sending a single message,
+three core batches are created: one for initial metadata, one for the message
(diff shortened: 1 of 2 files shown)


L1.md
```diff
@@ -0,0 +1,154 @@
...
+  //               a single step
+  // server-side:  hold the Write until the service handler returns (sync api)
+  //               or until Finish is called (async api)
+  WriteOptions& set_last_message();
+  WriteOptions& clear_last_message();
+  bool is_last_message() const;
+};
+```
+
+### Add a convenience WriteLast method to ClientWriterInterface, ClientReaderWriterInterface, ServerWriterInterface, ServerReaderWriterInterface
```
a11r: I am not sure we want all these additional convenience methods. The pro of course is that it these concepts are elevated to the top level in the API. But the con is that the API loses more orthogonality and there are to many ways of accomplishing the same thing.

L1.md
```diff
@@ -0,0 +1,154 @@
...
+void Write(const W& msg, WriteOptions options, void* tag);
+```
+
+### Add a WriteAndFinish method to ServerAsyncWriterInterface, ServerAsyncReaderWriterInterface
+```c++
+// Perform Write, Finish in a single step
+void WriteAndFinish(const W& msg, WriteOptions options, const Status& status void* tag);
+```
+
+### Expand client streaming stub signatures to add an overload that takes the first message
```
a11r: Per previous discussion, I would recommend adding an option either to the context or the initial metadata send that signals the same buffer-hint that we have in the write options.

PR #4, opened 2017-01-18
Proposal for HTTP CONNECT proxy support in gRPC.

markdroth:
Discussion thread: https://groups.google.com/d/topic/grpc-io/zG7iYpDw6lk/discussion


A1.md
```diff
@@ -0,0 +1,133 @@
+Title
```
ctiller: Title needs to be fixed

markdroth: Done.

A1.md
```diff
@@ -0,0 +1,133 @@
...
+## Rationale
+
+The proposed design allows two different ways of triggering use of the
+HTTP CONNECT handshaker code, which addresses the needs of both
+environments described above.
+
+There have also been discussions about how to do proxying for an
+environment where a service is accessible over the internet but its
+internal architecture (e.g., DNS names or server addresses) should not
+be exposed publicly.  In this case, it is impossible to support per-call
```
jboeuf: `impossible` is too strong a word. See above comment about the role of the proxy mapper with respect to the HTTP CONNECT request.

A1.md
```diff
@@ -0,0 +1,133 @@
...
+
+- For case 2, in the subchannel, before we connect to the target address,
+  we will call a new *proxy mapper* hook, which will allow selectively
+  requesting the use of a proxy based on the address to which the
+  subchannel is going to connect.  If a proxy is to be used, the mapper
+  will return the proxy address (which the subchannel will connect to
+  instead of the address returned by the resolver) and some additional
+  state (e.g., a channel argument in C-core) requesting that the
+  subchannel send an HTTP CONNECT request for the original subchannel
+  address.  Note that in this case, the server name used in the HTTP
```
jboeuf: The proxy mapper could decide what type of HTTP CONNECT request is going to be made: it doesn't have to be to the original address: it could be to the original name instead.

markdroth:
@jboeuf, see my reply in the discussion thread.

A1.md
```diff
@@ -3,14 +3,14 @@ HTTP CONNECT Proxy Support
...
-* Last updated: 2017-01-18
+* Implemented in: C-core (Java is in progress)
+* Last updated: 2017-01-25
 * Discussion at: https://groups.google.com/d/topic/grpc-io/zG7iYpDw6lk/discussion
 
 ## Abstract
 
 Describes what types of TCP-level proxy setups gRPC will support and
-how those setups interact with load balancing.
+how those setups interact with per-call load balancing.
```
markdroth: Server-side L7 load balancing would be the gRPC-level proxy approach described in the "Rationale" section below.  Since this document is mainly concerned with TCP-level proxies, I think it should be clear from context that that's not what's meant here.

That having been said, I am open to using different terminology here, assuming that we can find something that's unambiguous.  As I mentioned earlier, I dislike using the term "client-side" here, because grpclb is fundamentally not a client-side system -- it's a look-aside system.  I think using the term "client-side" may confuse readers who know about grpclb into thinking that we're referring only to round_robin and pick_first, not grpclb.  I think that's more likely to be confusing to readers than the potential implication that we're talking about gRPC-level load balancing.

I'd welcome alternative suggestions that avoid this problem.

ejona86: > Since this document is mainly concerned with TCP-level proxies, I think it should be clear from context that that's not what's meant here.

It is not at all to me. If I didn't know that you made the change in response to my comment, I honestly would have no clue what you were talking about. If I had to guess, it would have been the wrong thing, but that's sort of expected if I have to guess.

How about you just say "GRPCLB" here? The other sentences you could say "client-side" without concern, since you're already calling out GRPC specifically.

> As I mentioned earlier, I dislike using the term "client-side" here, because grpclb is fundamentally not a client-side system -- it's a look-aside system.

I find that virtually every client-side LB system would be look-aside, because the client still needs to be told the list of addresses to use from some system. Say, from DNS or Zookeeper.

I've been using client-side LB to include GRPCLB, for several years, without apparent trouble. I thought that is what others call them. Google search doesn't seem to directly disagree. Lookaside LB turns up nothing on Google. Same for per-call LB.

markdroth: I've added the term "client-side", but added a parenthetical to make it clear that this includes both grpclb and round_robin.

A1.md
```diff
@@ -0,0 +1,172 @@
...
+  the original server name (which will cause the proxy to establish a
+  connection to one of the load balancers).
+- When it sees any internal address (i.e., the addresses returned by
+  the load balancer), it will replace it with the proxy address and set
+  the HTTP CONNECT argument to the internal address.
+
+Note that this will work only if the `grpclb` load balancing policy is
+in use; it will not work with client-side policies like `round_robin`.
+
+Also, note that in case 1, the client will not have access to the
```
a11r: I think this assumption is overly restrictive. If the name resolution is happening via the proxy then we can allow the proxy to also return an appropriate config with the resolution response. Following this line of reasoning, if that config says we should do client side round-robin and the resolution returns a list of `k` addresses (possibly the same proxy address repeated k times) , then the client should respect that and create `k` connections to the proxy. This has two benefits:
1. Consistent behavior in other functionality unrelated to proxying.
2.  Ability to load-balance client-to-proxy load across multiple connections. Note that proxies may be scaled horizontally while presenting the same IP address to the clients.

ejona86: > If the name resolution is happening via the proxy then we can allow the proxy to also return an appropriate config with the resolution response. 

That's not how it works. There is no resolution response.

You ask to "CONNECT example.com" and then you get a response "200 OK", and now anything you write to the TCP connection will be proxied to example.com. While a proxy could add some metadata in a response header (along with the 200), this is not standard, and still happens _after_ you're creating connections already, so we're well past name resolution/LB policy.

markdroth: To expand on what Eric said, note that when this doc says that the proxy does the name resolution for us, that doesn't mean that the proxy actually provides the same result that we would get from the resolver; all it means is that we're sending the hostname to the proxy (as the argument to the CONNECT request) and depending on it do resolve that to one specific backend address when establishing the requested connection.

There's no reasonable way for the proxy to return the service config data.  Although it could conceivably send that information via a header in the CONNECT response, most proxies won't have any way of knowing to do that, because this type of proxy is not grpc-specific.  And even if it did, we don't have a reasonable way to make use of that information, because (as Eric points out) we would be obtaining that data in the subchannel connection code, which is lower in the stack than where we actually need the service config data.

mwitkow:
We kinda had exactly this problem in Go. We've built a tunnelling proxy that did `HTTP CONNECT` over HTTPS for gRPC-Go. We also built and open sourced a generic `net.Dialer` library that does HTTP CONNECT:  https://github.com/mwitkow/go-http-dialer

One thing that the proposal doesn't address, and that we encountered, is auth. What is the `Proxy-Authentication` API for this?

markdroth:
@mwitkow, let's keep feature discussion on the comment thread rather than here in the PR.  I'll answer your question there.

PR #7, opened 2017-01-25
A3: Initial draft of gRPC Channel Tracing

ncteisen:


ncteisen:
Discussion thread [here](https://groups.google.com/forum/#!topic/grpc-io/WFDj3KeHYTI)

A3.md
```diff
@@ -0,0 +1,93 @@
...
+* Author(s): Noah Eisen (ncteisen@google.com)
+* Approver: markdroth, ctiller
+* Status: Draft
+* Implemented in: N/A
+* Last updated: 01/25/17
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/WFDj3KeHYTI
+
+## Abstract
+
+The document proposes adding a method to the gRPC C-core Api called grpc_channel_get_trace() which returns a detailed trace for the channel in the form of a JSON string. Eventually this trace could be hooked up to a server with a clean UI, and will be instrumental in debugging channel related bugs.
```
markdroth: If this gRFC is intended to cover a mechanism that can be implemented in all languages, then let's remove references to C-core and to particular function names from everything except the "Implementation" section.

A3.md
```diff
@@ -111,14 +112,16 @@ grpc_json* grpc_channel_tracer_get_trace(grpc_channel_tracer* tracer);
 
 ### Format of Exported Data
```
markdroth: This subsection should probably be moved up into the "Proposal" section above, since it's not implementation-specific -- we should use the same data format in all implementations.

A3.md
```diff
@@ -87,7 +89,6 @@ The tracer files with expose several calls which with the C-core code can intera
 /* Initializes the tracing object with gpr_malloc. The caller has
    ownership over the returned tracing object */
 grpc_channel_tracer* grpc_channel_tracer_init_tracer();
```
markdroth: Should probably call this grpc_channel_tracer_create(), since it creates and returns a new object (as opposed to initializing a previously-allocated object).

A3.md
```diff
@@ -45,40 +45,42 @@ One downside to this approach is the potential for the tracer to bloat memory. F
...
-  gpr_mu tracer_mu;
-  gpr_refcount refs;
-  trace_node_list node_list;
   uint32_t num_subchannels_logged;
-  grpc_subchannel_tracer* next;
-  grpc_subchannel_tracer* prev;
+  gpr_timespec time_created;
+
+  // For the parent channel, these fields are the head and tail of the linked
+  // list of subchannels. For the subchannels, these fields are the next and
```
markdroth: It's not clear to me what the next and prev pointers would be used for in subchannels.  What's the use-case for this?

Keep in mind that in C-core, subchannels may be shared between parent channels.  So let's say that you have two parent channels, A and B.  Channel A has subchannels 1 and 2, and channel B has subchannels 1 and 3.  Where does the "next" pointer for subchannel 1 point to?

A3.md
```diff
@@ -30,13 +30,13 @@ Channel connectivity issues are the root cause of a significant portion of user
...
 
 ## Rationale
 
 This approach makes it easy to add trace, since each tracer is attached to its respective channel or subchannel. The tracer will also be passed down into the resolver and load balancer to record trace for interesting events that happen there. 
 
-One downside to this approach is the potential for the tracer to bloat memory. For this reason, the logs will be held in circular buffers that cannot grow in size. At a certain point, old trace will be overwritten. This is described in more detail in the [Garbage Collection](#garbage-collection) section.
+One downside to this approach is the potential for the tracer to bloat memory. For this reason, there will be a new channel argument that sets a limit for how many trace nodes each traced will hold. This is described in more detail in the [Garbage Collection](#garbage-collection) section.
```
markdroth: "each traced will hold" -> "each channel or subchannel will hold"

A3.md
```diff
@@ -45,40 +45,42 @@ One downside to this approach is the potential for the tracer to bloat memory. F
...
   grpc_slice* data;
   grpc_error* error;
   gpr_timespec time;
   grpc_connectivity_state connectivity_state;
   grpc_trace_node* next;
+
+  // this last field is used for efficient garbage collection of the
+  // subchannels. Once the node that traces a subchannel going inactive is
+  // overwritten, the tracing object for that subchannel is freed.
+  grpc_subchannel_tracer* free_this_subchannel;
```
markdroth: I would just call this "subchannel".

Also, there is no more grpc_subchannel_tracer type, so this should probably be a grpc_channel_tracer.

A3.md
```diff
@@ -45,40 +45,42 @@ One downside to this approach is the potential for the tracer to bloat memory. F
...
   uint32_t num_nodes_logged;
   trace_node_list node_list;
-  grpc_subchannel_tracer* head_subchannel;
-  grpc_subchannel_tracer* tail_subchannel;
-};
-
-/* the subchannel tracing object */
-struct grpc_subchannel_tracer {
-  gpr_mu tracer_mu;
-  gpr_refcount refs;
```
markdroth: Now that grpc_subchannel_tracer is going away, we need the refcount in the grpc_channel_tracer object.

A3.md
```diff
@@ -45,40 +45,42 @@ One downside to this approach is the potential for the tracer to bloat memory. F
...
-  trace_node_list node_list;
   uint32_t num_subchannels_logged;
-  grpc_subchannel_tracer* next;
-  grpc_subchannel_tracer* prev;
+  gpr_timespec time_created;
+
+  // For the parent channel, these fields are the head and tail of the linked
```
markdroth: If we're going to track subchannels via the pointer in the grpc_trace_node above, why do we need a list of subchannels here?  Can't we just get the list of subchannels by looking at all of the trace nodes?

To ask this another way, if we try to maintain the list of subchannels here, how do we know when to remove an entry from this list?  What happens if there's still a trace node referring to a subchannel that we remove from this list?  It seems like every time a node falls off of the list, we would need to check whether it was the last reference to an entry here, and if so, delete it.  It's not clear to me what the benefit is of doing that, though.

A3.md
```diff
@@ -45,40 +45,42 @@ One downside to this approach is the potential for the tracer to bloat memory. F
...
+  // list of subchannels. For the subchannels, these fields are the next and
+  // prev pointers.
+  grpc_subchannel_tracer* head_prev_subchannel;
```
markdroth: If these fields do stick around, they probably need be of type grpc_channel_tracer, not grpc_subchannel_tracer.

A3.md
```diff
@@ -96,10 +97,10 @@ void grpc_channel_tracer_add_trace(grpc_channel_tracer* tracer,
 
 /* Adds a subchannel to the tracing object */
 void grpc_channel_tracer_add_subchannel(grpc_channel_tracer* tracer,
```
markdroth: If we decide to track subchannels only via the references in the individual trace nodes, then this method may not be needed.

A3.md
```diff
@@ -96,10 +97,10 @@ void grpc_channel_tracer_add_trace(grpc_channel_tracer* tracer,
 
 /* Adds a subchannel to the tracing object */
 void grpc_channel_tracer_add_subchannel(grpc_channel_tracer* tracer, 
-    grpc_subchannel_tracer* subchannel);
+    grpc_channel_tracer* subchannel);
 
 /* Frees all of the resources held by the tracer object */
-void grpc_channel_tracer_destroy_tracer();
+void grpc_channel_tracer_destroy_tracer(grpc_channel_tracer* tracer);
```
markdroth: Probably don't need the "_tracer" suffix on this method name.

A3.md
```diff
@@ -141,7 +151,7 @@ message TraceNode {
   // error assosiated with the channel event
   string error = 2;
   // time the trace was logged
-  string time = 3;
+  Timestamp time = 3;
```
markdroth: This should be `google.protobuf.Timestamp`.

A3.md
```diff
@@ -111,14 +112,16 @@ grpc_json* grpc_channel_tracer_get_trace(grpc_channel_tracer* tracer);
...
+The trace data will be exported as JSON from the C-core. However, it will be JSON that is of the correct form to be transformed to protobuf and back again. The JSON schema will be as follows (note that all time strings will follow RFC 3339):
 
 ```
 {
-  "numSubchannelsLogged":,
+  "numSubchannelsLogged": number,
+  "time": string,
   "subchannels": [
     {
       "numNodesLogged": number,
```
markdroth: Don't we also want numNodesLogged for the parent channel?

A3.md
```diff
@@ -141,7 +151,7 @@ message TraceNode {
   // error assosiated with the channel event
```
markdroth: (This comment actually belongs a few lines above, but github won't let me put it there while looking at a specific commit.)

What happens when an individual trace node references a specific subchannel (for example, when adding or removing a subchannel to the parent channel)?  Don't we need some way to indicate which subchannel was affected?  Is that going to be indicated in the `data` field, or do we need some other way to reference the subchannel?

A3.md
```diff
@@ -165,21 +175,27 @@ message TraceNode {
 message SubchannelTrace {
```
markdroth: All of the fields in this message apply to both the parent channel and to each subchannel.  So how about something like this:

```
message TraceData {
  int32 num_nodes_logged = 1;
  google.protobuf.Timestamp start_time = 2;
  repeated TraceNode nodes = 3;
}

message ChannelTrace {
  TraceData data = 1;  // parent channel data
  int32 num_subchannels_seen = 2;
  repeated TraceData subchannel_data = 3;
}
```

A3.md
```diff
@@ -165,21 +175,27 @@ message TraceNode {
...
   // all of the subchannels for this channel
-  repeated SubchannelTrace subchannels = 2;
+  repeated SubchannelTrace subchannels = 4;
 };
 ```
 
 ### Garbage Collection
 
-In order to ensure that the tracer(s) do not use too much memory, both trace nodes and subchannel tracers will be discarded after a time. All trace nodes will be stored in circular buffers, the size of which can be set by toggling the channel tracer channel arg.
+In order to ensure that the tracer(s) do not use too much memory, both trace nodes and subchannel tracers will be discarded after a time. All tracers will only hold up to a hard limit of trace nodes, the size of which can be set by toggling a channel argument.
```
markdroth: "after a time" -> "when the number of trace nodes has gotten too high"
"hard limit" -> "configurable maximum number"
"size of which" -> "value of which"
"toggling" -> "setting"

A3.md
```diff
@@ -111,14 +112,16 @@ grpc_json* grpc_channel_tracer_get_trace(grpc_channel_tracer* tracer);
 
 ### Format of Exported Data
 
-The trace data will be exported as JSON from the C-core. However, it will be JSON that is of the correct form to be transformed to protobuf and back again. The JSON schema will be as follows:
+The trace data will be exported as JSON from the C-core. However, it will be JSON that is of the correct form to be transformed to protobuf and back again. The JSON schema will be as follows (note that all time strings will follow RFC 3339):
```
markdroth: Suggest adding a link to https://tools.ietf.org/html/rfc3339.

A3.md
```diff
@@ -165,21 +175,27 @@ message TraceNode {
 message SubchannelTrace {
   // tracks total trace nodes seen, since many may be overwritten
   int32 num_nodes_logged = 1;
+  // time this subchannel tracer was created
+  Timestamp time = 2;
```
markdroth: `google.protobuf.Timestamp`

A3.md
```diff
@@ -165,21 +175,27 @@ message TraceNode {
...
+  repeated TraceNode nodes = 3;
 }
 
 // trace for one channel
 message ChannelTrace {
   // tracks total subchannel seen, since some may be dropped as they go inactive
   int32 num_subchannels_logged = 1;
+  // all of the nodes for this channel
+  // time this channel tracer was created
+  Timestamp time = 2;
```
markdroth: `google.protobuf.Timestamp`

ncteisen:
Addressed all comments.

I added one design feature to solve the problem of how the trace nodes in the parent channel will refer to subchannels (if that particular trace dealt with one in particular). Each subchannel will be given a unique integer identifier. This could be as simple as adding a global subchannel counter to subchannel_index.*

Each subchannel would know its own id, and then the parent channel would include that id in any trace specifically about that subchannel, like if the subchannel went inactive.

I can add a section on this if you like the idea. Else we can discuss offline today sometime.

ncteisen:
Updated doc after discussion with @carl-mastrangelo.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+  };
+  // state the channel was in when the trace was logged
+  ChannelState state = 5;
+};
+
+// trace data for one channel or subchannel
+message TraceData {
+  // unique id of subchannel associated with this trace data
+  // set to zero if this is a channel's trace data
+  int64 subchannel_id = 1;
```
markdroth: How about generalizing this so that it's not specific to subchannels (i.e., even the parent channel should have an ID)?  In the parent, this can be set to the server URI; in the subchannels, it can be set to the URI of the server address to which the subchannel is connected.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+        "error": string,
+        "time": timestamp string,
+        // can only be one of "GRPC_CHANNEL_INIT", etc etc 
+        "state": enum string,
+        // id of referenced subchannel
+        "subchannelId": number
+      },
+    ]
+  },
+  "numSubchannelsSeen": number,
```
markdroth: This field doesn't seem to appear in the protobuf representation below.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+  * [Implementation](#implementation)
+     * [Basic Design](#basic-design)
+     * [Tracer Functions](#tracer-functions)
+     * [API Changes](#api-changes)
+     * [Format of Exported Data](#format-of-exported-data)
+     * [Garbage Collection](#garbage-collection)
+
+## Abstract
+
+The document proposes adding a dedicated, in-memory tracing object per channel and subchannel. The tracing object will record important steps in the life of a channel, like address resolution, subchannel creation, channel state changes, etc etc. This trace will be made available with a new gRPC API function call. The long term goal is to incorporate the trace into grpcz or a custom channel tracing UI.
```
markdroth: General comment: Can we be a little more precise with the terminology used in this doc?  In particular, it's not clear to me whether the term "trace" refers to a single trace node or to the collection of all trace nodes in a given tracer.  Let's define the terms and make sure that we're using them consistently throughout the document (and the code).

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+
+Channel connectivity issues are the root cause of a significant portion of user reported gRPC bugs (for example, [here](https://groups.google.com/a/google.com/forum/?utm_medium=email&utm_source=footer#!msg/grpc-users/gPFYbRAex1A/VM7h5FWAAgAJ) and [here](https://groups.google.com/a/google.com/forum/#!topic/grpc-users/0bsOqrxYvRc)). At the moment, gRPC has an extensive logging/tracing system in place. However, when juggling hundreds or thousands of connections, the logs can grow too huge to be quickly useful when debugging channel connectivity issues. Often times it is difficult to determine what events led to changes in connectivity.
+
+## Proposal
+
+To implement channel tracing, we will introduce a dedicated, in-memory tracing object that is owned by each channel and subchannel. We will add a new function to the grpc channel api that allows library users to retrieve all of the trace for the channel as a JSON or protobuf object. This returned trace will eventually be hooked into a UI, or exported through the ongoing grpcz project, where it could be used for debugging channel issues. The tracer will be enabled or disabled via a new channel argument, GRPC_ARG_CHANNEL_TRACING.
```
markdroth: > all of the trace for the channel

Shouldn't "trace" be plural here?  It seems like this document uses the singular form in plural context in several places, and it's not clear to me why.  (This might be related to my comment above about defining exactly what the term refers to.)

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+};
+```
+
+### Tracer Functions
+
+The tracer files with expose several calls which with the C-core code can interact with tracers:
+
+```C
+/* Initializes the tracing object with gpr_malloc. The caller has
+   ownership over the returned tracing object */
```
markdroth: Suggest alternative wording:

Creates a new tracer.  The caller owns a reference to the returned tracer.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+grpc_channel_tracer* grpc_channel_tracer_create(size_t max_nodes);
+
+/* Adds a new trace node to the tracing object */
+void grpc_channel_tracer_add_trace(grpc_channel_tracer* tracer, 
+    char* trace, struct grpc_error* error,
+    grpc_connectivity_state connectivity_state);
+
+/* Frees all of the resources held by the tracer object */
```
markdroth: Suggest just saying "Unrefs the tracer object."  (It doesn't actually free resources unless the refcount goes to 0.)

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+  gpr_timespec time_created;
+  grpc_connectivity_state connectivity_state;
+  grpc_trace_node* next;
+
+  // The tracer object that owns this trace node. This is used to ref and
+  // unref the tracing object as nodes are added or overwritten
+  grpc_channel_tracer* subchannel;
+};
+
+struct grpc_trace_node_list {
```
markdroth: Now that there's no separate subchannel tracer struct, there's no need for this to be a separate struct.  We can probably just put these fields directly into `grpc_channel_tracer`.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+    "startTime": timestamp string,
+    "nodes": [
+      {
+        "data": data,
+        "error": string,
+        "time": timestamp string,
+        // can only be one of "GRPC_CHANNEL_INIT", etc etc 
+        "state": enum string,
+        // id of referenced subchannel
+        "subchannelId": number
```
markdroth: If you take my suggestion below about giving all tracers an ID, then this field (minus the "subchannel" prefix) will be needed for subchannels below as well.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+      },
+    ]
+  },
+  "numSubchannelsSeen": number,
+  "subchannelData": [
+    {
+      "subchannelId": number,
```
markdroth: If you take my suggestion below about giving all tracers an ID, then this field (minus the "subchannel" prefix) will be needed for the parent channel above as well.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+message TraceNode {
+  // static string describing what event the channel went through
+  string data = 1;
+  // error assosiated with the channel event
+  string error = 2;
+  // time the trace was logged
+  google.protobuf.Timestamp time = 3;
+  // unique id of subchannel associated with this trace
+  // only used for trace present in the parent
+  int64 subchannel_id = 4;
```
markdroth: If you take my suggestion below for the ID to not be subchannel-specific, then this would just be a reference to some other tracer ID.  In practice, it may always point to a subchannel tracer, but in principle, the representation would not have that restriction.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+/* Frees all of the resources held by the tracer object */
+void grpc_channel_tracer_unref(grpc_channel_tracer* tracer);
+```
+### API Changes
+
+And finally, a new method will be added to the channel surface:
+
+```C
+/* returns the tracing data in the form of a grpc json object */
+grpc_json* grpc_channel_get_trace(grpc_channel* channel);
```
markdroth: I'm not sure that we want to expose the `grpc_json` type into the C-core public API.  It might be better to return a JSON string instead of the parsed JSON tree.

A3-channel-tracing.md
lukaszx0: Also, could we remove googlisms from it, like `grpcz` ?

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+     * [Format of Exported Data](#format-of-exported-data)
+     * [Garbage Collection](#garbage-collection)
+
+## Abstract
+
+The document proposes adding a dedicated, in-memory tracing object per channel and subchannel. The tracing object will record important steps in the life of a channel, like address resolution, subchannel creation, channel state changes, etc etc. This trace will be made available with a new gRPC API function call. The long term goal is to incorporate the trace into grpcz or a custom channel tracing UI.
+
+## Background
+
+Channel connectivity issues are the root cause of a significant portion of user reported gRPC bugs (for example, [here](https://groups.google.com/a/google.com/forum/?utm_medium=email&utm_source=footer#!msg/grpc-users/gPFYbRAex1A/VM7h5FWAAgAJ) and [here](https://groups.google.com/a/google.com/forum/#!topic/grpc-users/0bsOqrxYvRc)). At the moment, gRPC has an extensive logging/tracing system in place. However, when juggling hundreds or thousands of connections, the logs can grow too huge to be quickly useful when debugging channel connectivity issues. Often times it is difficult to determine what events led to changes in connectivity.
```
lukaszx0: The links to google groups are pointing to `google.com` org and are unavailable for the audience outside of Google.

A3-channel-tracing.md
lukaszx0: `grpcz` mentioned here again. Started wondering that maybe it's going to be a public project?

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Proposal
+
+To implement channel tracing, we will introduce a dedicated, in-memory tracing object that is owned by each channel and subchannel. We will add a new function to the grpc channel api that allows library users to retrieve all of the trace for the channel as a JSON or protobuf object. This returned trace will eventually be hooked into a UI, or exported through the ongoing grpcz project, where it could be used for debugging channel issues. The tracer will be enabled or disabled via a new channel argument, GRPC_ARG_CHANNEL_TRACING.
+
+### Format of Exported Data
+
+The trace data will be exported as JSON from the C-core. However, it will be JSON that is of the correct form to be transformed to protobuf and back again. The JSON schema will be as follows (note that all time strings will follow [RFC 3339](https://tools.ietf.org/html/rfc3339)):
```
lukaszx0: The proposal kind is `A`, why C-core is mentioned here explicitly? Will it be different for other langs? Will it be configurable to have custom marshallers and just store it as binary (protobuf) ?

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+
+```
+{
+  "channelData": {
```
lukaszx0: Hmm, maybe we could reformat this section and list protobuf source first. It has clear type definitions and full structure (enums etc) and then we can follow it with example JSON with sample data (as opposed to `"data": data,`).

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+    "numNodesLogged": number,
+    "startTime": timestamp string,
+    "nodes": [
+      {
+        "data": data,
```
lukaszx0: This is incorrect, it should be `"data": string` (or proto below is out of date).

In general I suggest restructuring this section, see my previous comment.

A3-channel-tracing.md
```diff
@@ -0,0 +1,204 @@
...
+And finally, a new method will be added to the channel surface:
+
+```C
+/* returns the tracing data in the form of a grpc json object */
+grpc_json* grpc_channel_get_trace(grpc_channel* channel);
+```
+
+### Garbage Collection
+
+In order to ensure that the tracer(s) do not use too much memory, both trace nodes and subchannel tracers will be discarded when the number of trace nodes has gotten too high. All tracers will only hold up to a configurable maximum number of trace nodes, the value of which can be set by setting a channel argument.
```
lukaszx0: This implies new API. Should it be described in the `Basic Design` with the example of planned new API in C-core?

ncteisen:
I just revived this doc. Now that #40 is close to being settled, I will begin working on the c++ implementation of channel tracing again

jtattermusch:
please ignore this comment: testing that CLA bot works.

jtattermusch:
please ignore this comment: testing that CLA bot works.

ncteisen:
This RFC is ready for review. The implementation is in grpc/grpc#13883

A14-channelz.md
```diff
@@ -701,16 +701,18 @@ message Subchannel {
   repeated SocketRef socket = 5;
 }
 
+enum ChannelConnectivityState {
```
carl-mastrangelo: Enums function oddly outside of a message.  If you want to move this, consider wrapping it in a separate message.

The issue is that enums are not namespaced in Go or CPP, so you end up with top level variables like UNKNOWN, but it isn't clear its a ChannelConnectivityState

A14-channelz.md
```diff
@@ -701,16 +701,18 @@ message Subchannel {
   repeated SocketRef socket = 5;
 }
```
carl-mastrangelo: Consider adding a simple doc comment pointing to the connectivity state API page here: https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md

A14-channelz.md
```diff
@@ -728,8 +730,27 @@ message ChannelData {
   google.protobuf.Timestamp last_call_started_timestamp = 7;
 }
 
+message ChannelTraceEvent {
```
carl-mastrangelo: Also here for adding a comment.   Protos are the defacto spec and documentation, so they are the source of truth.   Each message and field should have a comment usually.  (I may have not be as disciplined in my messages, but I should have been!)

A14-channelz.md
```diff
@@ -728,8 +730,27 @@ message ChannelData {
...
+  string description = 1;
+  string error = 2;
```
carl-mastrangelo: Should this be a Status?  https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto#L80

They can express the kinds of errors the application will see, and they have room for additional details.

A14-channelz.md
```diff
@@ -728,8 +730,27 @@ message ChannelData {
...
+  google.protobuf.Timestamp event_timestamp = 3;
+  ChannelConnectivityState state = 4;
+  // uuid of referenced subchannel.
+  // Optional, only present if this event refers to a child object.
+  string child_uuid = 5;
```
carl-mastrangelo: Channels ended up using int64's for the ids.   unless this is the id of the message itself.

A14-channelz.md
```diff
@@ -728,8 +730,27 @@ message ChannelData {
...
+message ChannelTraceData {
+  string uuid = 1;
+  int64 num_events_logged = 2;
+  google.protobuf.Timestamp channel_created_timestamp = 3;
+  repeated ChannelTraceEvent events = 4;
+}
+
 message ChannelTrace {
-  // TODO: fill this in.
+  ChannelTraceData channel_data = 1;
```
carl-mastrangelo: This should have a ChannelRef or Subchannel ref back to which channel it came from.

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
+Channel Tracing
+----
+* Author(s): Noah Eisen (ncteisen@google.com)
+* Approver: markdroth
+* Status: Draft
+* Implemented in: N/A
+* Last updated: 01/02/18
```
carl-mastrangelo: be sure to update this before submitting.

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
...
+  * [Background](#background)
+  * [Proposal](#proposal)
+  * [Format of Exported Data](#format-of-exported-data)
+
+## Terminology
+
+This design uses the same terminology discussed in the [channelz](A14-channelz.md) design.
+
+In addition, the following terminology will be used:
+1.  an "trace event" is an interesting thing that happens to a channel. Example include things like creation, address resolution, subchannel creation, connectivity state changes.
```
carl-mastrangelo: s/an/a/

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
...
+  * [Abstract](#abstract)
+  * [Background](#background)
+  * [Proposal](#proposal)
+  * [Format of Exported Data](#format-of-exported-data)
+
+## Terminology
+
+This design uses the same terminology discussed in the [channelz](A14-channelz.md) design.
+
+In addition, the following terminology will be used:
```
carl-mastrangelo: optional: add newline before list.

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
...
+1.  an "trace event" is an interesting thing that happens to a channel. Example include things like creation, address resolution, subchannel creation, connectivity state changes.
+2. "trace data" is all of the data that is contained for one channel. This includes a list of trace events, as well as metadata like the timestamp at which the channel was created.
```
carl-mastrangelo: Consider using either the message name, or the field name, or the capitalized Trace Data name.    For example:

> _Trace data_ is all of the data that [...]

carl-mastrangelo: Nit: Channel was split into channel and subchannel in channelz.   For correctness, please mention both, or state in your terms that channel means both channel and subchannel.

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
...
+3.  a "tracing object" is an in-memory data structure responsible for holding the trace data for a single channel.
+
+## Abstract
+
+The document proposes adding a dedicated, in-memory tracing object for every channel and subchannel. The tracing object will record important events in the life of a channel, like address resolution, subchannel creation, channel state changes, etc etc. The trace data from this tracing object will be made available through a new gRPC API as a JSON formatted string.
```
carl-mastrangelo: s/etc etc./etc./

carl-mastrangelo: s/JSON/Proto/

We want to pick one serialization and stick with it.  Having both would put undue stress on a client for reading the data.

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
...
+
+Channel tracing is one aspect of the wider design, [channelz](A14-channelz.md).
+
+## Background
+
+Channel connectivity issues are the root cause of a significant portion of user reported gRPC bugs. Channel tracing will be exceedingly helpful for getting live channel data from a misbehaving program.
```
carl-mastrangelo: s/exceedingly//

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Proposal
+
+To implement channel tracing, implementations will use a dedicated, in-memory tracing object that is owned by each channel and subchannel. We will add an API that allows library users to retrieve all of the tracing object's data for the channel as a JSON formatted string. Implementations must expose a toggle for this behavior (for example, C++ will add a new channel argument, GRPC_ARG_CHANNEL_TRACING).
```
carl-mastrangelo: For this gRFC, a specification is being described rather than an implementation being prescribed.   The difference is how programs should behave, rather than how they are written.

Here, is it really important that it is in memory?   I think the point you are trying to make is that it is ephemeral, rather than in memory.   For example, a gRPC implementation that logs channel trace events to an external database may want to not keep it in memory.

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
...
+
+Since the tracing object exists in memory, care must be taken to prevent the tracing object from using too much memory. Implementations must allow setting a _max_trace_events_ variable to control how many trace events are allowed in any tracing object's event list. Once the limit is reached, the oldest trace event will be overwritten by a new one (a circular buffer).
```
carl-mastrangelo: For code and variable names should use backticks:

```
`max_trace_events`
```

carl-mastrangelo: max trace events again is an implementation detail, not a spec detail.   Instead, consider rewording this like:

> Since the tracing objects may consume large amounts of space, care must be taken to prevent the tracing object from using too many resources.   gRPC implementations should allow limiting the number of trace events retained to a global maximum, as well as a per channel or subchannel maximum.

A3-channel-tracing.md
```diff
@@ -0,0 +1,74 @@
...
+
+At certain points a particular subchannel might stop being used. However, the tracing object for that subchannel may still be needed if any trace events from other tracing object reference the subchannel (as in the case of a parent's subchannel being destroyed). In order to efficiently discard old subchannels, implementations must free the tracing object for a subchannel when the trace event signaling that subchannel's destruction is overwritten in the last parent channel to reference it.
```
carl-mastrangelo: instead of dictating that objects must be freed, instead say that implementations are not responsible for making the objects available.   

An implementation may want to occasionally garbage collect such objects, and at a time that may be more convenient.   For example, you could decide to clean up resources when not serving latency sensitive traffic.

A14-channelz.md
ncteisen: Done

A14-channelz.md
ncteisen: Done

A14-channelz.md
ncteisen: Done

A14-channelz.md
ncteisen: I initially meant this to be `grpc_error`, but seeing as that is language specific, I think that Status is a better choice here

A14-channelz.md
ncteisen: Done

A14-channelz.md
ncteisen: Ok. ChannelTrace can be for a channel or subchannel, so I included both, and stipulated that only one may be set

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done

ncteisen: Done

A3-channel-tracing.md
ncteisen: Done

ncteisen: C core does not support proto and has no plans to. Since the new API will be added to the c core, the data has to be exported as a JSON string. The best we can do is ensure the JSON has a deterministic mapping to Proto. This is c core reads service config as JSON.

A3-channel-tracing.md
ncteisen: Acknowledged and corrected.

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done

ncteisen:
Thanks for the detailed review! I addressed all of the comments

A3-channel-tracing.md
carl-mastrangelo: Actually, Java doesn't have direct support of Proto either (and we also have no plans to).   You should still consider using proto to define the messages on the _API_ though, but pass such objects around internally as a struct/class.   

In java, outside of the core library we have a set of services (like health checking and reflection)  which can read the internal datastructures of core.  This allows decoupling of the implementation and the API.

You don't have to do the same approach, but you should consider it, as java team has found it works well and is maintainable.  We are doing the same thing for Channelz, which has both a shim proto service on top of out internal datastructures.

A14-channelz.md
```diff
@@ -728,8 +736,39 @@ message ChannelData {
   google.protobuf.Timestamp last_call_started_timestamp = 7;
 }
 
+// A trace event is an interesting thing that happened to a channel or
+// subchannel, such as creation, address resolution, subchannel creation, etc.
+message ChannelTraceEvent {
+  // High level description of the event.
+  string description = 1;
+  // Status/error associated with this event.
```
carl-mastrangelo: Can this be absent?   If present, will it ever be Status OK?

A14-channelz.md
```diff
@@ -728,8 +736,39 @@ message ChannelData {
...
+message ChannelTraceEvent {
+  // High level description of the event.
+  string description = 1;
+  // Status/error associated with this event. 
+  google.rpc.Status status = 2;
+  // When this event occurred.
+  google.protobuf.Timestamp event_timestamp = 3;
+  // The connectivity state the channel was in when this event occurred.
+  ChannelConnectivityState state = 4;
+  // uuid of referenced subchannel.
```
carl-mastrangelo: update comment

A14-channelz.md
```diff
@@ -728,8 +736,39 @@ message ChannelData {
...
+  // Optional, only present if this event refers to a child object. For example,
+  // this field would be filled if this trace event was for a subchannel being
+  // created.
+  int64 child_channel_id = 5; 
+}
+
+message ChannelTraceData {
+  int64 channel_id = 1;
```
carl-mastrangelo: We use the ChannelRef to refer to channels, which allows a name to be associated with the ID.  This would probbaly be better as:

```
oneof ref {
  ChannelRef channel_ref = 1;
  SubchannelRef subchannel_ref = 2;
}
```

A14-channelz.md
```diff
@@ -728,8 +736,39 @@ message ChannelData {
...
+  int64 num_events_logged = 2;
```
carl-mastrangelo: Is this the total number of events logged?   It might be worth clarifying that this may be different than than size of `events` below and why.

A14-channelz.md
```diff
@@ -728,8 +736,39 @@ message ChannelData {
...
+  google.protobuf.Timestamp channel_created_timestamp = 3;
+  repeated ChannelTraceEvent events = 4;
+}
+
 message ChannelTrace {
-  // TODO: fill this in.
+  // Ref to the channel or subchannel that this trace is associated with.
+  // Only one of "channel_ref" and "subchannel_ref" can be set.
```
carl-mastrangelo: use `oneof` as mentioned above.

A3-channel-tracing.md
```diff
@@ -0,0 +1,76 @@
...
+  * [Proposal](#proposal)
+  * [Format of Exported Data](#format-of-exported-data)
+
+## Terminology
+
+This design uses the same terminology discussed in the [channelz](A14-channelz.md) design. Channels and subchannels will be handled by the same channel tracing code, so for this design, a channel means either a channel or subchannel from the channelz terminology.
+
+In addition, the following terminology will be used:
+
+1.  a "Trace event" is an interesting thing that happens to a channel. Example include things like creation, address resolution, subchannel creation, connectivity state changes.
```
carl-mastrangelo: s/Example/Examples/

A3-channel-tracing.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Background
+
+Channel connectivity issues are the root cause of a significant portion of user reported gRPC bugs. Channel tracing will be helpful for getting live channel data from a misbehaving program.
+
+## Proposal
+
+To implement channel tracing, implementations will use a dedicated, in-memory _Tracing object_ that is owned by each channel and subchannel. We will add an API that allows library users to retrieve all of the _Tracing object's_ data for the channel as a JSON formatted string. Implementations must expose a toggle for this behavior (for example, C++ will add a new channel argument, GRPC_ARG_CHANNEL_TRACING).
+
+Since the tracing objects may consume large amounts of space, care must be taken to prevent the _Tracing object_ from using too many resources. gRPC implementations should allow limiting the number of _Trace events_ retained to a global maximum, as well as a per channel or subchannel maximum.
```
carl-mastrangelo: Optional: consider putting in some rough defaults that an implementer of this should use.  For example: "implementions should consider retaining at least the last 30 mins of events if possible."

A14-channelz.md
ncteisen: Done

A14-channelz.md
ncteisen: Done

A14-channelz.md
ncteisen: Done

A14-channelz.md
ncteisen: I removed this field, since it is present in the parent message

A14-channelz.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: I'd rather leave this completely up to implementations for now

A3-channel-tracing.md
ncteisen: Done

ncteisen:
Addressed all comments

A14-channelz.md
```diff
@@ -741,31 +741,38 @@ message ChannelData {
...
-  // Status/error associated with this event. 
+  // Status/error associated with this event.
+  // Optional, only present is Status != OK. 
   google.rpc.Status status = 2;
   // When this event occurred.
   google.protobuf.Timestamp event_timestamp = 3;
   // The connectivity state the channel was in when this event occurred.
   ChannelConnectivityState state = 4;
-  // uuid of referenced subchannel.
+  // channel_id of referenced subchannel.
```
carl-mastrangelo: as mentioned, this should probably be a ref.   Also, this could be a channel or subchannel.

A3-channel-tracing.md
```diff
@@ -0,0 +1,85 @@
...
+        // uuid of referenced subchannel.
+        // Optional, only present if this event refers to a child object.
+        // and example of a referenced child would be a trace event for a
+        // subchannel being created.
+        "child_channel_id": string
+      },
+    ]
+  },
+  // Optional, only present if this channel has children
+  "child_data": [
```
carl-mastrangelo: child data should be on the child, no?

A3-channel-tracing.md
```diff
@@ -0,0 +1,85 @@
...
+The data will be exported as JSON formatted string. The JSON must conform with the [protobuf-json mapping chart](https://developers.google.com/protocol-buffers/docs/proto3#json). The JSON schema will be as follows:
+
+```
+{
+  "channel_id": string,
+  "channel_data": {
+    // Number of events ever logged in this tracing object. This can differ from
+    // events.size() because events can be overwritten or garbage collected by
+    // implementations.
+    "num_events_logged": number,
```
carl-mastrangelo: does this mean this is a float?

A3-channel-tracing.md
```diff
@@ -0,0 +1,85 @@
...
+        // Optional, only present is Status != OK. 
+        "status": string,
+        "time": timestamp string,
+        // can only be one of the states in connectivity_state.h
+        "state": enum string,
+        // uuid of referenced subchannel.
+        // Optional, only present if this event refers to a child object.
+        // and example of a referenced child would be a trace event for a
+        // subchannel being created.
+        "child_channel_id": string
```
carl-mastrangelo: for consistency, this should be a ref.

A3-channel-tracing.md
```diff
@@ -0,0 +1,85 @@
...
+    "num_events_logged": number,
+    // Time that this channel was created.
+    "channel_created_timestamp": timestamp string,
+    // List of events that have occurred on this channel.
+    "events": [
+      {
+        // High level description of the event.
+        "description": string,
+        // Status/error associated with this event.
+        // Optional, only present is Status != OK.
```
carl-mastrangelo: Is this still a string?

A14-channelz.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: ref, channel_data, and child_data are all top level to the ChannelTrace proto, so they are also top level on the JSON object.

A3-channel-tracing.md
ncteisen: I clarified integer

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Corrected it to enum string (string that strictly complies with proto enum definition as per [this doc](https://developers.google.com/protocol-buffers/docs/proto3#json)

ncteisen:
Addressed all comments

ncteisen:
Ping @markdroth, following out offline 1:1

ncteisen:
Ping

ncteisen:
@markdroth are you comfortable if this is only reviewed by @carl-mastrangelo? If not can you get to this soon, or pass it along to someone else if you are too busy?

ncteisen:
I removed mention of JSON from the design, since that will be c core specific. If other languages want to surface the data as proto or other data struct, that is fine

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
...
+
+The document proposes adding a dedicated _Tracing object_ for every channel and subchannel. The _Tracing object_ will record important events in the life of a channel, like address resolution, subchannel creation, channel state changes, etc. The _Trace data_ from this _Tracing object_ will be made available through the [channelz service](A14-channelz.md).
+
+## Background
+
+Channel connectivity issues are the root cause of a significant portion of user reported gRPC bugs. Channel tracing will be helpful for getting live channel data from a misbehaving program.
+
+## Proposal
+
+To support channel tracing, implementations will use a dedicated _Tracing object_ that is owned by each channel and subchannel. We will add an API that allows library users to retrieve all of the _Tracing object's_ data for the channel in a language idiomatic way. Implementations must expose a toggle for this behavior (for example, C++ will add a new channel argument, GRPC_ARG_CHANNEL_TRACING).
```
markdroth: Given that we're going to expose this data via channelz, do we actually need to expose a language-specific API for it as well?  I realize that we need a C-core API to expose the data to the channelz service in each language, but it's not clear to me that we actually need to define or support local APIs in each language.

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
...
+
+Since the tracing objects may consume large amounts of space, care must be taken to prevent the _Tracing object_ from using too many resources. gRPC implementations should allow limiting the number of _Trace events_ retained to a global maximum, as well as a per channel or subchannel maximum.
```
markdroth: A global maximum sounds hard to implement, for two reasons.  First, given that the individual trace events will be added to different channels/subchannels at different times, I think a global maximum would require additional synchronization.  And second, it's not clear what the behavior should be when we hit the global maximum -- the spec needs to explicitly define the behavior (e.g., which entries get thrown out when that happens?).

I think it might be simpler to not support a global maximum.

A14-channelz.md
```diff
@@ -728,8 +736,49 @@ message ChannelData {
   google.protobuf.Timestamp last_call_started_timestamp = 7;
 }
 
+// A trace event is an interesting thing that happened to a channel or
+// subchannel, such as creation, address resolution, subchannel creation, etc.
+message ChannelTraceEvent {
```
markdroth: I suggest leaving these new message types out of this doc, since they weren't really part of the original channelz proposal.  Instead, let's just leave an empty `ChannelTrace` message with a comment that the details will be filled in by the channel tracing proposal (with a link to the gRFC).

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
...
+// A trace event is an interesting thing that happened to a channel or
+// subchannel, such as creation, address resolution, subchannel creation, etc.
+message ChannelTraceEvent {
+  // High level description of the event.
+  string description = 1;
+  // Status/error associated with this event.
+  // Optional, only present is Status != OK. 
+  google.rpc.Status status = 2;
+  // When this event occurred.
+  google.protobuf.Timestamp event_timestamp = 3;
```
markdroth: Suggest just calling this `timestamp`.  It's implicit from the context that this is the timestamp of this event.

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
...
+  }
+}
+
+message ChannelTraceData {
+  // Number of events ever logged in this tracing object. This can differ from
+  // events.size() because events can be overwritten or garbage collected by
+  // implementations.
+  int64 num_events_logged = 3;
+  // Time that this channel was created.
+  google.protobuf.Timestamp channel_created_timestamp = 4;
```
markdroth: Suggest calling this `creation_time`.

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
...
+  // Optional, only present if this event refers to a child object. For example,
+  // this field would be filled if this trace event was for a subchannel being
+  // created.
+  oneof child_ref {
+    ChannelRef channel_ref = 1;
+    SubchannelRef subchannel_ref = 2;
+  }
+}
+
+message ChannelTraceData {
```
markdroth: Why make this a separate message instead of including these fields directly in `ChannelTrace`?

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
...
+  // implementations.
+  int64 num_events_logged = 3;
+  // Time that this channel was created.
+  google.protobuf.Timestamp channel_created_timestamp = 4;
+  // List of events that have occurred on this channel.
+  repeated ChannelTraceEvent events = 5;
+}
+
+message ChannelTrace {
+  // Ref to the channel or subchannel this trace refers to.
```
markdroth: Why does the trace need a reference to the channel or subchannel, given that it's returned as part of the data for the channel or subchannel?

A3-channel-tracing.md
markdroth: We need to explicitly define the behavior when we hit the maximum.

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Terminology
+
+This design uses the same terminology discussed in the [channelz](A14-channelz.md) design. Channels and subchannels will be handled by the same channel tracing code, so for this design, a channel means either a channel or subchannel from the channelz terminology.
+
+In addition, the following terminology will be used:
+
+1.  a "Trace event" is an interesting thing that happens to a channel. Examples include things like creation, address resolution, subchannel creation, connectivity state changes.
+2. "Trace data" is all of the data that is contained for one channel. This includes a list of _Trace events_, as well as metadata like the timestamp at which the channel was created.
+3.  a "Tracing object" is an in-memory data structure responsible for holding the _Trace data_ for a single channel.
```
markdroth: If the only thing contained in the "Tracing object" is the "Trace data", then why is "Trace data" a separate object?  Why not simply put the fields from "Trace data" directly in the "Tracing object"?

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Proposal
+
+To support channel tracing, implementations will use a dedicated _Tracing object_ that is owned by each channel and subchannel. We will add an API that allows library users to retrieve all of the _Tracing object's_ data for the channel in a language idiomatic way. Implementations must expose a toggle for this behavior (for example, C++ will add a new channel argument, GRPC_ARG_CHANNEL_TRACING).
+
+Since the tracing objects may consume large amounts of space, care must be taken to prevent the _Tracing object_ from using too many resources. gRPC implementations should allow limiting the number of _Trace events_ retained to a global maximum, as well as a per channel or subchannel maximum.
+
+Implementations must keep the _Trace data_ for a subchannel until the _Trace event_ signaling that subchannel's destruction is overwritten or garbage collected in the last parent channel to reference it. After that point, implementations are free to use an appropriate garbage collection algorithm to reclaim the memory that the subchannel's _Tracing object_ was holding.
```
markdroth: The wording of this paragraph needs work; it's not clear enough about the specific requirements, and it's also too prescriptive in some areas (e.g., we should not mandate garbage collection; individual implementations can use whatever memory management techniques are appropriate).  I suggest something like the following:

The _Trace object_ for a given channel or subchannel must be maintained as long as there are any _Trace events_ that refer to the channel or subchannel.  After the last _Trace event_ that refers to the channel or subchannel is removed, the _Trace object_ for that channel or subchannel may be cleaned up.

A14-channelz.md
```diff
@@ -4,7 +4,7 @@ gRPC Channelz
 * Approver: a11r
 * Status: Approved
 * Implemented in: 
-* Last updated: 2/5/18
+* Last updated: 02/09/18
```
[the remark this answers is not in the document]
ncteisen: Done

A3-channel-tracing.md
```diff
@@ -0,0 +1,111 @@
+Channel Tracing
+----
+* Author(s): Noah Eisen (ncteisen@google.com)
+* Approver: markdroth
+* Status: Draft
+* Implemented in: N/A
+* Last updated: 02/09/18
```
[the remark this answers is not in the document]
ncteisen: Done

A3-channel-tracing.md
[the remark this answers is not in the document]
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done. I mentioned that the info is exposed via channelz. But I kept in the sentence stating implementations MAY still expose the data through a new API

A3-channel-tracing.md
ncteisen: Done

A14-channelz.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: I don't remember why I originally split these out... I have removed `ChannelTraceData` and inlined everything into channel trace

A3-channel-tracing.md
ncteisen: I think I had this in because I needed in the internal data structure I wrote for c core. But that information is definitely not needed in the proto, good point. I have removed it.

A3-channel-tracing.md
ncteisen: The tracing object also holds the child data. I will specify that

ncteisen: I also removed the notion of trace data, because it was uneeded.

A3-channel-tracing.md
ncteisen: Done

ncteisen:
Addressed all comments, PTAL

A3-channel-tracing.md
```diff
@@ -21,48 +21,35 @@ This design uses the same terminology discussed in the [channelz](A14-channelz.m
 In addition, the following terminology will be used:
 
 1.  a "Trace event" is an interesting thing that happens to a channel. Examples include things like creation, address resolution, subchannel creation, connectivity state changes.
-2. "Trace data" is all of the data that is contained for one channel. This includes a list of _Trace events_, as well as metadata like the timestamp at which the channel was created.
-3.  a "Tracing object" is an in-memory data structure responsible for holding the _Trace data_ for a single channel.
+2.  a "Channel trace" is a data structure responsible for holding all data for a single channel. This includes a list of _Trace events_, as well as metadata like the timestamp at which the channel was created. The _Channel trace_ also holds a list of _Channel trace_ object of its child channels (if it has any).
```
markdroth: The last sentence here is confusing, for two reasons.

First, I don't think you mean a list of objects; I think you mean a list of _pointers_ to objects.

Second, I thought the channel trace was going to contain references to the channelz data of its child channels, not to the channel trace objects in those channelz objects.

A3-channel-tracing.md
```diff
@@ -21,48 +21,35 @@ This design uses the same terminology discussed in the [channelz](A14-channelz.m
...
 
 ## Background
 
 Channel connectivity issues are the root cause of a significant portion of user reported gRPC bugs. Channel tracing will be helpful for getting live channel data from a misbehaving program.
 
 ## Proposal
 
-To support channel tracing, implementations will use a dedicated _Tracing object_ that is owned by each channel and subchannel. We will add an API that allows library users to retrieve all of the _Tracing object's_ data for the channel in a language idiomatic way. Implementations must expose a toggle for this behavior (for example, C++ will add a new channel argument, GRPC_ARG_CHANNEL_TRACING).
+To support channel tracing, implementations will use a dedicated _Channel trace_ that is owned by each channel and subchannel. All of the data will be exposed via the [channelz service](A14-channelz.md). Implementations MAY add an API that allows library users to retrieve all of the _Tracing object's_ data for the channel in a language idiomatic way. Implementations must expose a toggle for this behavior (for example, C++ will add a new channel argument, GRPC_ARG_CHANNEL_TRACING).
```
markdroth: Why provide a local API at all?  What's the use-case for that?

A3-channel-tracing.md
```diff
@@ -21,48 +21,35 @@ This design uses the same terminology discussed in the [channelz](A14-channelz.m
...
-3.  a "Tracing object" is an in-memory data structure responsible for holding the _Trace data_ for a single channel.
+2.  a "Channel trace" is a data structure responsible for holding all data for a single channel. This includes a list of _Trace events_, as well as metadata like the timestamp at which the channel was created. The _Channel trace_ also holds a list of _Channel trace_ object of its child channels (if it has any).
 
 ## Abstract
 
-The document proposes adding a dedicated _Tracing object_ for every channel and subchannel. The _Tracing object_ will record important events in the life of a channel, like address resolution, subchannel creation, channel state changes, etc. The _Trace data_ from this _Tracing object_ will be made available through the [channelz service](A14-channelz.md).
+The document proposes adding a dedicated _Channel trace_ for every channel and subchannel. The _Channel trace_ will record important events in the life of a channel, like address resolution, subchannel creation, channel state changes, etc. The data from this _Channel trace_ will be made available through the [channelz service](A14-channelz.md).
```
markdroth: In the "Proposal" section below, we should have some detailed guidance about what sorts of events should and should not be added to the channel trace.  In particular, I think we should state that no events associated with individual calls should be added, because that will very quickly fill up the buffer.

A3-channel-tracing.md
markdroth: I think the first sentence of this paragraph can be removed.  Its semantically not very clear (what does "use" mean in this context?), and it probably just duplicates what's already described in the "Abstract" section above.

markdroth: The wording of the last sentence needs to specify exactly what the knob does.  Currently, it just says "this behavior", but it's not clear exactly what that means.  I think the intent here is that this will inhibit collection of tracing data, but someone reading this would probably think it was referring to the exporting of the data, since that's what the rest of this paragraph talks about.

Also, the knob you've implemented in C-core is not just a toggle; it's actually a setting of how many trace events to keep per channel or subchannel, which disables collection if set to 0.  We should choose one of the following approaches:

1. Dictate the particular semantic of disabling collection when the max number of events is set to 0.  (In this case, we can probably move this to the end of the next paragraph.)

2. Require that implementations provide some API for disabling collection of trace events, but not dictate the exact semantics of how that API will work.  (In this case, the knob should probably be described in a paragraph of its own.)

A3-channel-tracing.md
```diff
@@ -21,48 +21,35 @@ This design uses the same terminology discussed in the [channelz](A14-channelz.m
...
+To support channel tracing, implementations will use a dedicated _Channel trace_ that is owned by each channel and subchannel. All of the data will be exposed via the [channelz service](A14-channelz.md). Implementations MAY add an API that allows library users to retrieve all of the _Tracing object's_ data for the channel in a language idiomatic way. Implementations must expose a toggle for this behavior (for example, C++ will add a new channel argument, GRPC_ARG_CHANNEL_TRACING).
 
-Since the tracing objects may consume large amounts of space, care must be taken to prevent the _Tracing object_ from using too many resources. gRPC implementations should allow limiting the number of _Trace events_ retained to a global maximum, as well as a per channel or subchannel maximum.
+Since the tracing objects may consume large amounts of space, care must be taken to prevent the _Channel trace_ from using too many resources. gRPC implementations must allow limiting the number of _Trace events_ retained to a per channel or subchannel maximum. Once the maximum is reached, a new _Trace event_ will overwrite the oldest _Trace event_.
```
markdroth: The term "overwrite" seems incorrect here.  We're not necessarily re-using the same memory of the previous event; we're just removing the oldest event when we add the new one.  An individual implementation could choose to actually re-use the memory, but that's an implementation detail, not something we should specify in the spec.

A3-channel-tracing.md
```diff
@@ -86,25 +73,15 @@ message ChannelTraceEvent {
...
-  // Ref to the channel or subchannel this trace refers to.
-  oneof ref {
-    ChannelRef channel_ref = 1;
-    SubchannelRef subchannel_ref = 2;
-  }
-  // All of the data for this channel.
-  ChannelTraceData channel_data = 3;
+  repeated ChannelTraceEvent events = 3;
   // Optional, only present if this channel has children
   repeated ChannelTrace child_data = 4;
```
markdroth: Why is this field needed?  In the channelz API, every node in the tree has its own `ChannelTrace`, so why do we need the `ChannelTrace` for an individual node to directly contain the `ChannelTrace` for its child nodes?

A3-channel-tracing.md
ncteisen: Sorry, this was another case of a leftover from writing the initial draft of this design before the channelz design was written. I have tried to be more careful, in this round of comments, of making sure the design fits well with channelz.

The Channel trace will hold a list of _references_ to the channels or subchannel mentioned in its list of trace events.

A3-channel-tracing.md
ncteisen: Debugging? But I guess that can be decided at a per-implementation level. I removed it from the design

A3-channel-tracing.md
ncteisen: I added an example of trace from a healthy channel

A3-channel-tracing.md
ncteisen: Done

ncteisen: I have chosen option 1.

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done.

ncteisen:
Thanks for the detailed review! The top level comment was very helpful for me, and I will keep it in mind for future designs. 

I think a lot of the problems in this design spawned because I wrote this, and a lot of the implementation, before channelz. I have attempted to do a more comprehensive job of ensuring this design fits into the channelz design.

A3-channel-tracing.md
```diff
@@ -20,8 +20,8 @@ This design uses the same terminology discussed in the [channelz](A14-channelz.m
...
 
-1.  a "Trace event" is an interesting thing that happens to a channel. Examples include things like creation, address resolution, subchannel creation, connectivity state changes.
-2.  a "Channel trace" is a data structure responsible for holding all data for a single channel. This includes a list of _Trace events_, as well as metadata like the timestamp at which the channel was created. The _Channel trace_ also holds a list of _Channel trace_ object of its child channels (if it has any).
+1.  a "Trace event" is an interesting thing that happens to a channel. Examples include things like creation, address resolution, subchannel creation, connectivity state changes. Some _Trace events_ (like a new subchannel being created), will refer to the ChannelData of the relevant channel or subchannel.
+2.  a "Channel trace" is a data structure responsible for holding all trace data for a single channel. This includes a list of _Trace events_, as well as metadata like the timestamp at which the channel was created. The _Channel trace_ also holds a list of references to the channels or subchannels that are mentioned in the list of _Trace events_.
```
markdroth: Why do we keep this list of references in the channel trace itself?  Isn't it sufficient that the references exist in the individual trace events that refer to those channels or subchannels?

Note that you removed this from the `ChannelTrace` proto message below, so I assume it should go away here too.

ncteisen:
Addressed the last comment!

A3-channel-tracing.md
```diff
@@ -0,0 +1,90 @@
...
+... Starting TCP connection
+... TCP connection established
+... Auth handshake complete
+... Entering idle mode
+```
+
+
+## Format of Exported Data
+
+The data will be accessed via the channelz service, which sends a ChannelTrace proto as part of a larger message concerning a channel or subchannel. The following is a relevant excerpt, taken directly from the proto definition in channelz](A14-channelz.md).
```
markdroth: Missing a `[` before `channelz` here.

ncteisen: Done

ncteisen:
Squashed all commits

PR #9, opened 2017-01-30
A4: Header extraction design proposal.

markdroth:
Discussion thread: https://groups.google.com/d/topic/grpc-io/oOzQi1cpNfM/discussion


A4.md
```diff
@@ -0,0 +1,254 @@
...
+There will also be a way to explicitly specify it via the gRPC client API.
+
+Note that for streaming RPCs, we extract headers only from the first
+request message on the stream.  However, in order to do this, we need to
+defer sending initial metadata until the client sends the first message,
+which is a change from the normal behavior.  This new behavior will only
+take effect if there is a header extraction config specified for the
+call.
+
+### Specification for Extracting Headers
```
ejona86: Could you define what happens for invalid configuration? Like, what if delimiterCharacter is two characters, or what if it is empty? Or what if the field specified payloadFieldName doesn't exist for the message? Or what if payloadFieldName is empty? What if numElementsToKeep is 0? What if headerName is an invalid header name?

I'd imagine the answer to almost all of those is going to be the same.

markdroth: Good idea.  Done.

A4.md
```diff
@@ -0,0 +1,266 @@
...
+
+```
+resource: {
+  id: "//foo/bar/baz"
+}
+user: "roth@quux@mumble@frotz"
+transaction_id: "some_big_opaque_string"
+```
+
+And let's say that we're using the following header extraction
```
carl-mastrangelo: Micro nit: The 'we' feels a little too informal for a specification.  Also, it seems to change between 'we' the gRPC team, and 'we' gRPC plus the end user.  I personally find it distracting to the actual content.

markdroth: I think the use of "we" to mean "gRPC team" is fine.  But I agree with you that this particular usage is different and inconsistent.  I've re-worded this sentence.

markdroth:
Withdrawing this gRFC in favor of the design @a11r is working on to set the header fields in the VTK layer.

PR #10, opened 2017-01-31
A5: Proposal for encoding grpclb data in DNS.

markdroth:
Discussion thread: https://groups.google.com/d/topic/grpc-io/6be1QsHyZkk/discussion


A5.md
```diff
@@ -0,0 +1,157 @@
...
+; grpclb for server.example.com goes to lb.example.com:1234
+_grpclb._tcp.server IN SRV 0 0 1234 lb
+; lb.example.com has 3 IP addresses
+lb                  IN A 10.0.0.1
+                    IN A 10.0.0.2
+                    IN A 10.0.0.3
+```
+
+With this data, a gRPC client will resolve the name `server.example.com`
+to the following addresses:
```
markdroth: Unless I'm misunderstanding something from the SRV RFC, I think that records with the same priority and weight are all equally likely to be chosen, which is exactly the behavior that we're supporting here.  So I don't think there's any surprise here.

That having been said, I have added a paragraph explaining that we do not support any priority or weighting.

A5.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Proposal
+
+This proposal makes use of DNS SRV records, as described in
+[RFC-2782](https://tools.ietf.org/html/rfc2782).  The SRV record for
+grpclb will use the following values:
+
+- service name: `grpclb`
+- protocol: `tcp`
+- priority: this will be `0` for all grpclb SRV records
```
markdroth: Added a note about this.

a11r:
The proposal content LGTM. ARe you leanning to fill out the Java and Go implementation sections soon ? It would be nice to get a sanity check that there is some path to an implementation before we merge this.

markdroth:
@carl-mastrangelo, can you please provide text for the java implementation section?  I know that we discussed several options for how this could be implemented, so it might be useful to outline all of the options and record why we chose the one that we're going to pursue.

@dfawley and/or @menghanl, can you please provide text for the go implementation section?  I suspect that this will rely on the work that the two of you are doing to restructure the grpc-go resolver and load balancer APIs (and this section can note that there will be a forthcoming gRFC about that).

Thanks!

A5-grpclb-in-dns.md
```diff
@@ -0,0 +1,170 @@
...
+found, it will then look up the name(s) that the SRV records point to
+and return those as balancer addresses.
+
+Note that, due to platform support issues, we will initially *not*
+support the c-ares resolver under Windows or for Node.  Alternatives
+will need to be found for these environments.
+
+### Java
+
+TODO(notcarl): Provide content for this section.
```
carl-mastrangelo: Java will depend on JNDI, Netty DNS, dnsjava, or a another DNS library to do SRV record resolution.  The existing name resolver, `DnsNameResolver` will be modified to resolve the additional records, and include them in the Attributes presented to the load balancer.

markdroth:
Thanks for the info, Carl.  I've added that to the doc.

@dfawley, @menghanl, and/or @lyuxuan, can any of you provide info for the Go implementation?  Thanks!

A5-grpclb-in-dns.md
```diff
@@ -0,0 +1,173 @@
...
+will need to be found for these environments.
+
+### Java
+
+Java will depend on JNDI, Netty DNS, dnsjava, or a another DNS library
+to do SRV record resolution. The existing name resolver, `DnsNameResolver`,
+will be modified to resolve the additional records and include them in
+the Attributes presented to the load balancer.
+
+### Go
```
lyuxuan: Go will use the LookupSRV function in package net to do SRV record resolution. We are going to implement a DNS name resolver and will include SRV record resolution.

markdroth: Thanks for the info.  I've updated the doc.

@a11r, I believe this is ready to be merged now.

PR #12, opened 2017-02-11
gRFC A6: Retries

ncteisen:
gRFC for retries

Authors: ncteisen and ericgribkoff


A6.md
```diff
@@ -0,0 +1,398 @@
+gRPC Retry Design
+----
+* Author(s): [Noah Eisen](https://github.com/ncteisen) and [Eric Gribkoff](https://github.com/ericgribkoff)
+* Approver: a11r
+* Status: Draft
+* Implemented in: All Languages
```
markdroth: I think this line is intended to indicate where this proposal is *already* implemented.  Given that we haven't yet implemented this in any language, I think this field should be blank.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+#### Retry Policy Capabilities
+
+Retry policies support configuring the maximum number of retries, the parameters for exponential backoff, and the set of retryable status codes. Each of these configuration options is detailed in its own section below.
+
+##### Maximum Number of Retries
+
+The maximum number of retry attempts per original RPC. This can be specified in the retry policy as follows:
+
+```
+'max_retry_attempts': 3
```
markdroth: The example for the hedging policy below shows the parameters wrapped in a 'hedging_policy' block.  For consistency, the retry examples should be wrapped in a 'retry_policy' block.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+  1. Retry policy: Don't retry, return failure to client application
+  2. Hedging policy: Cancel previous and pending hedges
+4. Pushback: Don't retry
+  1. Retry policy: Don't retry, return failure to client application
+  2. Hedging policy: Don’t send any more hedged requests.
+5. Pushback: Retry in *n* ms
+  1. Retry policy: Retry in *n* ms. If this attempt also fails, retry delay will reset to initial backoff for the following retry (if applicable)
+  2. Hedging policy: Send next hedged request in *n* ms. Subsequent hedged requests will resume at `n + hedging_delay`
+
+![State Diagram](A6_graphics/StateDiagram.png)
```
markdroth: It would probably be a good idea to include the SVG version of these diagrams as well as the PNG, so that we can edit them in the future if necessary.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+
+Similarly, if the RPC reaches the server, but has never been seen by the server application logic (the second bullet), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy.
+
+Since retry throttling is designed to prevent server application overload, and these transparent retries do not make it to the server application layer, they do not count as failures when deciding whether to throttle retry attempts.
+
+#### Exposed Retry Metadata
+Both client and server application logic will have access to data about retries via gRPC metadata. Upon seeing an RPC from the client, the server will know if it was a retry, and moreover, it will know the number of previously made attempts. Likewise, the client will receive the number of retry attempts made when receiving the results of an RPC.
+
+The new header names for exposing the metadata will be `"x-grpc-retry-attempts"` to give clients and servers access to the attempt count.
```
markdroth: I assume the value of this field will be a human-readable integer?  If so, please explicitly state that.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+```
+
+A value called `current_backoff` is initially set to `initial_backoff`. After every failed RPC, `current_backoff` is set to `min(current_backoff * multiplier, max_backoff)`. The failed RPCs will be retried after *x* seconds, where *x* is defined as `random(0, current_backoff)`.
+
+##### Retryable Status Codes
+
+When gRPC receives a non-OK response status from a server, this status is checked against the set of retryable status codes to determine if a retry attempt should be made.
+
+```
+'retryable_status_codes': {UNAVAILABLE}
```
markdroth: For consistency, let's make all examples in this doc be the exact JSON format that will be used in the service config.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+Service owners must choose between a retry policy or a hedging policy. Unless the service owner specifies a policy in the configuration, retries and hedging will not be enabled. The retry policy and hedging policy each have their own set of configuration options, detailed below. 
+
+The parameters for throttling retry attempts and hedged RPCs when failures exceed a certain threshold are also set in the service config. Throttling applies across methods and services on a particular server, and thus may only be configured per-server name.
+
+#### Retry Policy
+
+This is an example of a retry policy and its associated configuration. It implements exponential backoff with a maximum of three retry attempts, only retrying RPCs when an `UNAVAILABLE` status code is received.
+
+```
+'retry_policy': {
```
markdroth: The default conversion from protobuf will change a protobuf field of the form "foo_bar" to a JSON string of the form "fooBar".  So let's use that form for all fields in this doc (e.g., "retry_policy" should be "retryPolicy").

In fact, before finalizing this, I would encourage you to write the protobuf definition that will be used internally and then create a proto and see how it converts to JSON, to make sure it's compatible with what we're specifying here.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+
+#### Integration with Service Config
+
+The retry policy is transmitted to the client through the service config mechanism. The following is what the JSON configuration file would look like:
+
+```
+{
+  'loadBalancingPolicy': string,
+
+  # Throttling parameters for retry attempts and hedged RPCs.
```
markdroth: Please use `//` instead of `#` for comments in JSON.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+        'max_retry_attempts': number,
+        'exponential_backoff': {
+          # Integer
+          'initial_backoff_ms': number,
+          # Integer
+          'max_backoff_ms': number,
+          # Integer
+          'multiplier': number
+        },
+        # Array of strings. Supported values are the gRPC status codes, e.g.,
```
markdroth: I assume that the status codes should be encoded as JSON strings?  If so, let's state that explicitly.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+        # UNAVAILABLE.
+        'retryable_status_codes': []
+      },
+      'hedging_policy': {
+        # Integer
+        'max_requests' : number,
+        # Integer
+        'hedging_delay_ms' : number,
+        # Array of strings.
+        # Supported values are the gRPC status codes, e.g.,
```
markdroth: Same here.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+
+## Background
+
+Currently, gRPC does not retry failed RPCs. All failed RPCs are immediately returned to the application layer by the gRPC client library.
+
+Many teams have implemented their own retry logic wrapped around gRPC like [Veneer Toolkit](https://github.com/googleapis/toolkit) and [Cloud Bigtable](https://github.com/GoogleCloudPlatform/cloud-bigtable-client).
+
+## Proposal
+
+### Overview
```
markdroth: Oh, one more thing: This section should probably include a link to https://github.com/grpc/proposal/pull/5.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+
+The only RPCs that are counted as failures for the throttling policy are RPCs that fail with a status code that qualifies as a retryable or non-fatal status code (see [here](#retryable-status-codes) and [here](#hedging-policy)), or that receive a pushback response indicating not to retry. This avoids conflating server failure with responses to malformed requests (such as the `INVALID_ARGUMENT` status code).
+
+#### Pushback
+Servers may explicitly pushback by setting metadata in their response to the client. The pushback can either tell the client to retry after a given delay or to not retry at all. If the client has already exhausted its `max_retry_attempts`, the call will not be retried even if the server says to retry after a given delay.
+
+A new metadata key, `"x-grpc-retry-pushback-ms"`, will be added to support server pushback. If the value for pushback is set to -1, then it will be seen as the server asking the client not to retry at all.
```
lukaszx0: Before I got to "Throttling Configuration" below, it wasn't really clear to me what's the scope of pushback on the server side. I think it's worth mentioning here since this section is supposed to describe it in details, not the examples one below.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+
+Neither retry attempts or hedged RPCs block when `token_count` is less than or equal to the threshold. Retry attempts are canceled and the failure returned to the client application. The hedged request is cancelled, and if there are no other already-sent hedged RPCs the failure is returned to the client application.
+
+The only RPCs that are counted as failures for the throttling policy are RPCs that fail with a status code that qualifies as a retryable or non-fatal status code (see [here](#retryable-status-codes) and [here](#hedging-policy)), or that receive a pushback response indicating not to retry. This avoids conflating server failure with responses to malformed requests (such as the `INVALID_ARGUMENT` status code).
+
+#### Pushback
+Servers may explicitly pushback by setting metadata in their response to the client. The pushback can either tell the client to retry after a given delay or to not retry at all. If the client has already exhausted its `max_retry_attempts`, the call will not be retried even if the server says to retry after a given delay.
```
lukaszx0: I assume by the mention of `max_rety_attempts` that `Pushback` applies only to backoff policy. Probably should be stated explicitly?

A6.md
lukaszx0: Unrelated to this PR, but I think it'd be great to have a link to a _reference_ protobuf definition which reflects the JSON in the [main service config doc](https://github.com/grpc/grpc/blob/master/doc/service_config.md) in a form of Appendix at the bottom. I think a lot of people, due to DNS limitation will be using it, myself included.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+  1. Retry policy: Retry in *n* ms. If this attempt also fails, retry delay will reset to initial backoff for the following retry (if applicable)
+  2. Hedging policy: Send next hedged request in *n* ms. Subsequent hedged requests will resume at `n + hedging_delay`
+
+![State Diagram](A6_graphics/StateDiagram.png)
+
+### Retry Internals
+
+#### Where Retries Occur
+
+The retry policy will be implemented in-between the channel and the load balancing policy. That way every retry gets a chance to be sent out on a different subchannel than it originally failed on.
```
ericgribkoff: This is a good point, but currently the metadata only says if the RPC is a retry. We are considering adding information about the previously used subchannel(s), based on comments from the gRFC email thread. If/once that is added, then adding the link you suggest (based on the retry metadata...) makes sense.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+
+To clarify the second scenario, we define an *outgoing message* as everything the client sends on its connection to the server. For unary and server streaming calls, the outgoing message is a single message. For client and bidirectional streaming calls, the outgoing message is the entire message stream issued by the client after opening the connection. The gRPC client library buffers outgoing messages, and as long as the entirety of the outgoing message is in the buffer, it can be resent and retried. But as soon as the outgoing message grows too large to buffer, the gRPC client library cannot replay the entire stream of messages, and thus retries are not valid.
+
+#### Buffered RPCs
+
+The gRPC library will have a configurable amount of available memory, `retry_buffer_size`, to buffer outgoing retryable or hedged RPCs. There is also a per-RPC size limit ,`per_rpc_buffer_limit`. These limits are configured by the client, rather than coming from the service config.
```
ericgribkoff: Added `in_bytes` to the name.

A6.md
```diff
@@ -0,0 +1,398 @@
...
+#### Exposed Retry Metadata
+Both client and server application logic will have access to data about retries via gRPC metadata. Upon seeing an RPC from the client, the server will know if it was a retry, and moreover, it will know the number of previously made attempts. Likewise, the client will receive the number of retry attempts made when receiving the results of an RPC.
+
+The new header names for exposing the metadata will be `"x-grpc-retry-attempts"` to give clients and servers access to the attempt count.
+
+#### Disabling Retries
+
+Clients cannot override retry policy set by the service config. However, retry support can be disabled entirely within the gRPC client library. This is designed to enable existing libraries that wrap gRPC with their own retry implementation (such as Veneer Toolkit) to avoid having retries taking place at the gRPC layer and within their own libraries.
+
+Eventually, retry logic should be taken out of the wrapping libraries, and only exist in gRPC. But allowing the retries to be disabled entirely will make that rollout process easier.
```
ericgribkoff: I hesitate to say always...but it will always be supported, at least for now :-) We want clients in resource-constrained environments to also be able to disable retries, so this is intended as a feature of the API, not a temporary stopgap.

A6.md
ericgribkoff: Done

A6.md
ericgribkoff: Done

A6.md
ericgribkoff: Done

A6.md
ericgribkoff: Done.

A6.md
ericgribkoff: The (mostly finalized) JSON format of the service config with retry, hedging, and throttling policies is now included. I think it's a good idea to include the protobuf definition as well once its finalized.

A6.md
ericgribkoff: Done (aside: https://github.com/grpc/grpc/blob/master/doc/service_config.md uses `#` for the comments)

A6.md
ericgribkoff: Done.

A6.md
ericgribkoff: Done.

A6.md
ericgribkoff: Done

A6.md
ericgribkoff: @ncteisen - this request is buried on github after I pushed the latest commit. Do you have SVG versions of the diagrams we can add?

A6.md
```diff
@@ -0,0 +1,451 @@
...
+
+The gRPC library will have a configurable amount of available memory, `retry_buffer_size_in_bytes`, to buffer outgoing retryable or hedged RPCs. There is also a per-RPC size limit ,`per_rpc_buffer_limit_in_bytes`. These limits are configured by the client, rather than coming from the service config.
+
+RPCs may only be retried when they are contained in the buffer. 
+
+Hedged RPCs with a non-zero delay will only send out the subsequent hedged requests when the original RPC remains in the buffer. For hedge policies with a delay of zero milliseconds, the RPC does not need to be placed in the buffer at all and the outgoing RPCs can be sent immediately then discarded. 
+
+After the RPC response has been returned to the client application layer, the RPC is removed from the buffer.
+
+New RPCs which do not fit in the available buffer space (either due to the total available buffer space, or due to the per-RPC limit) will not be retryable. For client-side streaming RPCs, if additional outgoing messages would cause the buffered size to exceed the per-RPC limit, the entire message will be removed from the buffer.
```
lukaszx0: > New RPCs which do not fit in the available buffer space (either due to the total available buffer space, or due to the per-RPC limit) will not be retryable.

Does that mean that they'll fail immediately? Does that mean they will make original request and in case of failure won't qualify for retry and fail after first initial request? I assume it's the former, maybe worth clarifying?

ericgribkoff: Actually, it is the latter. We don't intend the buffer to limit the overall number or size of outgoing RPCs, so we still send the original request and essentially behave exactly as if there were no retry policy specified for RPCs sent when the buffer is full.

Updated the doc to state this explicitly.

A6.md
```diff
@@ -52,7 +52,7 @@ Many teams have implemented their own retry logic wrapped around gRPC like [Vene
 
 ### Overview
 
-gRPC will support two configurable retry policies. The service configuration may choose from a retry policy (retry failed RPCs) or a hedging policy (aggressively send the same RPC multiple times in parallel). An individual RPC may be governed by a retry policy or a hedge policy, but not both.
+gRPC will support two configurable retry policies. The [service configuration](https://github.com/grpc/proposal/pull/5) may choose from a retry policy (retry failed RPCs) or a hedging policy (aggressively send the same RPC multiple times in parallel). An individual RPC may be governed by a retry policy or a hedge policy, but not both.
```
markdroth: I suggest changing the link for the text "service configuration" to point to https://github.com/grpc/grpc/blob/master/doc/service_config.md, and adding the following after it:

(which may soon be [published via DNS](https://github.com/grpc/proposal/pull/5))

A6.md
```diff
@@ -214,15 +217,15 @@ To clarify the second scenario, we define an *outgoing message* as everything th
 
 #### Buffered RPCs
 
-The gRPC library will have a configurable amount of available memory, `retry_buffer_size`, to buffer outgoing retryable or hedged RPCs. There is also a per-RPC size limit ,`per_rpc_buffer_limit`. These limits are configured by the client, rather than coming from the service config.
+The gRPC library will have a configurable amount of available memory, `retry_buffer_size_in_bytes`, to buffer outgoing retryable or hedged RPCs. There is also a per-RPC size limit ,`per_rpc_buffer_limit_in_bytes`. These limits are configured by the client, rather than coming from the service config.
```
markdroth: I suggest changing the wording to make it clear that this part of the spec is optional and may differ based on how the client is implemented.  Suggested text (feel free to wordsmith):

The gRPC client library may support application-configured limits for the amount of memory used for retries.  It is suggested that the client limit both the total amount of memory used to buffer retryable or hedged RPCs as well as the amount of memory used by any one RPC (to prevent a single large RPC from using the whole buffer and preventing retries of subsequent smaller RPCs).

A6.md
```diff
@@ -214,15 +217,15 @@ To clarify the second scenario, we define an *outgoing message* as everything th
 
 #### Buffered RPCs
```
markdroth: Suggest changing this heading to something like "Memory Management".

A6.md
markdroth: @lukaszx0, can you please chime in on the comment thread for #5 (https://groups.google.com/d/topic/grpc-io/DkweyrWEXxU/discussion) and describe the limitations that would prevent you from using DNS for service configs?  I'd like more information on that, but I'd like to keep this discussion in that other thread, since it's not really directly related to the retry design.  Thanks!

A6.md
lukaszx0: "Memory Management (Buffering)" maybe?

A6.md
lukaszx0: Ah, I see. Makes sense, thanks for the clarification.

A6.md
```diff
@@ -0,0 +1,457 @@
...
+4. Pushback: Don't retry
+  1. Retry policy: Don't retry, return failure to client application
+  2. Hedging policy: Don’t send any more hedged requests.
+5. Pushback: Retry in *n* ms
+  1. Retry policy: Retry in *n* ms. If this attempt also fails, retry delay will reset to initial backoff for the following retry (if applicable)
+  2. Hedging policy: Send next hedged request in *n* ms. Subsequent hedged requests will resume at `n + hedgingDelayMs`
+
+![State Diagram](A6_graphics/StateDiagram.png)
+
+[Link to diagram on draw.io](https://drive.google.com/a/google.com/file/d/0Bwcp6fL6ikUvOGJWcExrNFhMZ0k/view?usp=sharing)
```
lukaszx0: This won't fly. I'm guessing you linked to your work drive at google. I think it would be nice if you could export SVGs and add them to repo alongside pngs ([looks like draw.io let's you do that](https://support.draw.io/display/DO/Exporting+Files))

ncteisen: Oops!! Thanks for the catch. I switched all of the diagrams to SVG format :)

lukaszx0: Cool, didn't know GH will render it. Nice, thank you!

lukaszx0: Um, apparently it won't render them inline - https://github.com/ncteisen/proposal/blob/3b24372020f0acc7017bbf2e1e559dcdc9279561/A6.md 😞

markdroth: I do think it would be better to include the SVG files in this PR instead of linking to an external source.  Otherwise, there's no guarantee that someone won't delete the external source without realizing that it's referenced from this document.

ncteisen: Haha oops again, reverted back to PNGs, and now there are links to the SVG files too.

lukaszx0: Looks good! Thank you :)

A6.md
lukaszx0: @markdroth sure, I had it opened in tab but didn't get to it.

A6.md
```diff
@@ -0,0 +1,465 @@
...
+
+When gRPC receives a non-OK response status from a server, this status is checked against the set of retryable status codes in `retryableStatusCodes` to determine if a retry attempt should be made.
+
+In general, only status codes that indicate the service did not process the request should be retried. However, a more aggressive set of parameters can be specified when the service owner knows that a method is idempotent, or safe to be processed more than once by the server. For example, if an RPC to a delete user method fails with an `INTERNAL` error code, it's possible that the user was already deleted before the failure occurred. If the method is idempotent then retrying the call should have no adverse effects. However, it is entirely possible to implement a non-idempotent delete method that has adverse side effects if called multiple times, and such a method should not be retried unless the error codes guarantee the original RPC was not processed. It is up to the service owner to pick the correct set of retryable status codes given the semantics of their service. gRPC retries do not provide a mechanism to specifically mark a method as idempotent.
+
+#### Hedging Policy
```
mwitkow: Can we rename hedging to something else? When I was first reading this I thought it meant one: 
send X RPCs immediately and return the call once the first one completes, cancelling the others.

A6.md
```diff
@@ -0,0 +1,465 @@
...
+
+#### Change to gRPC Wire Format
+
+If a server's response only consists of a status code, it is currently legal to send the status back in the trailers. Based on the previous section, this would mean that those RPCs would not be considered valid to retry, since the client would already have seen the initial metadata.
+
+While implementing retries we will make the wire protocol slightly stricter, and ensure that the status is sent back in the initial metadata if that is all the server is going to send, which will give the RPC a chance of being retried if it should.
+
+#### Memory Management (Buffering)
+
+The gRPC client library will support application-configured limits for the amount of memory used for retries. It is suggested that the client sets `retry_buffer_size_in_bytes` to limit the total amount of memory used to buffer retryable or hedged RPCs. The client should also set `per_rpc_buffer_limit_in_bytes` to limit the amount of memory used by any one RPC (to prevent a single large RPC from using the whole buffer and preventing retries of subsequent smaller RPCs). These limits are configured by the client, rather than coming from the service config.
```
mwitkow: I believe that this could be made significantly simpler by saying: number of buffered per RPC.
This makes it easier for users of the library to reason about it: they know that 5 unary RPCs will make becasue retries buffer 10.

A6.md
```diff
@@ -0,0 +1,465 @@
...
+
+#### Exposed Retry Metadata
+
+Both client and server application logic will have access to data about retries via gRPC metadata. Upon seeing an RPC from the client, the server will know if it was a retry, and moreover, it will know the number of previously made attempts. Likewise, the client will receive the number of retry attempts made when receiving the results of an RPC.
+
+The new header names for exposing the metadata will be `"x-grpc-retry-attempts"` to give clients and servers access to the attempt count. The value for this field will be a human-readable integer.
+
+#### Disabling Retries
+
+Clients cannot override retry policy set by the service config. However, retry support can be disabled entirely within the gRPC client library. This is designed to enable existing libraries that wrap gRPC with their own retry implementation (such as Veneer Toolkit) to avoid having retries taking place at the gRPC layer and within their own libraries.
```
mwitkow: This should be tunable per-call.

A6.md
ericgribkoff: Hedging seems like the appropriate name. It can mean what you say (with a hedging delay of 0ms) but also supports sending the RPCs on a fixed delay.

A6.md
ericgribkoff: This would work if we assume a fixed size for each RPCs. Since this also applies to streaming RPCs that are effectively unbounded, we need to specify the buffer size in memory rather than number of RPCs.

A6.md
ericgribkoff: Our intent is to limit the client-side API surface as much as possible. We have to allow disabling retries altogether, due to existing wrapper libraries and clients in resource-constrained environments, but otherwise we want to keep retry policy the domain of the service owner rather than individual clients.

ericgribkoff:
Hi Michal,

Thank you for the comments. We are trying to keep high-level discussion on the email thread ([see here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/grpc-io/zzHIICbwTZE/OPbO2E2VBQAJ)) but my responses to your points (b) and (c) are below.

> b) retry logic would benefit a lot from knowing whether the method is idempotent or not. II understand that this is supposed to be handled by "service configs", but realistically they're really hard to use. Few people would put their retry logic in DNS TXT entries, and even fewer people operate the gRPC LB protocol. Can we consider adding a .proto option (annotation) to Method definitions?

There are two concerns here. One is that saying a method is idempotent really just means "retry on status codes x y and z". If we pick a canonical set of idempotent status codes, we are forcing every service owner to obey these semantics if they want to use retries. However, the gRPC status code mechanism is very flexible (even allowing services to return UNAVAILABLE arbitrarily) so we'd prefer to force service owners to consider the semantics of their application and pick a concrete set of status codes rather than just flipping an "idempotent" switch. The second concern is around the ease of use of the service config. The intent is for the service config to be a universally useful mechanism, and we want to avoid just baking everything into the proto. Concerns about the delivery mechanism for service config shouldn't invalidate its use for encoding retry policy, and may be something we have to tackle separately.

> c) One thing I found out useful "in the wild" is the ability to limit the Deadline of the retriable call. For example, the "parent" RPC call (user invoked) has a deadline of 5s, but each retriable call only 1s. This allows you to skip a "deadlining" server and retry against one that works.

This is covered by our hedging policy. There doesn't seem to be any reason to cancel the first RPC in your scenario, as it may be just about to complete on the server and cancellation implies the work done so far is wasted. Instead, hedging allows you to send the first request, wait one second, send a second request, and accept the results of whichever completes first.

A6.md
```diff
@@ -215,9 +215,9 @@ The reasoning behind the first scenario is that as soon as a non-error response
...
 
-#### Buffered RPCs
+#### Memory Management (Buffering)
 
-The gRPC library will have a configurable amount of available memory, `retry_buffer_size_in_bytes`, to buffer outgoing retryable or hedged RPCs. There is also a per-RPC size limit ,`per_rpc_buffer_limit_in_bytes`. These limits are configured by the client, rather than coming from the service config.
+The gRPC client library will support application-configured limits for the amount of memory used for retries. It is suggested that the client sets `retry_buffer_size_in_bytes` to limit the total amount of memory used to buffer retryable or hedged RPCs. The client should also set `per_rpc_buffer_limit_in_bytes` to limit the amount of memory used by any one RPC (to prevent a single large RPC from using the whole buffer and preventing retries of subsequent smaller RPCs). These limits are configured by the client, rather than coming from the service config.
```
markdroth: I think this section is still a bit too specific (i.e., "will" instead of "may", and specifying specific names for the two pools).  Having discussed the C-core implementation with @ctiller, I think that we will not actually wind up supporting a separate total pool for retries; instead, that will come out of the existing pool used for all of gRPC.  However, we will still provide a per-RPC limit, since we want to avoid one large RPC from starving all other RPCs of space for retries.

I think this section should be re-worded to describe the types of constraints that an implementation may want to consider, but it should not mandate any particular mechanism or configuration parameters.

A6.md
```diff
@@ -140,7 +140,7 @@ If all instances of a hedged RPC fail, there are no additional retry attempts. E
 
 If server pushback that specifies not to retry is received in response to a hedged request, no further hedged requests should be issued for the call.
 
-Hedged requests should be sent to distinct backends, if possible.
+Hedged requests should be sent to distinct backends, if possible. To facilitate this, the gRPC client will maintain a list of previously used backend addresses for each hedged RPC. This list will be passed to the gRPC client's local load-balancing policy. Previously used addresses should be skipped when the load-balancing policy picks a server for the hedged RPC. If all available backend addresses have already been used, the load-balancing policy may pick any backend. This means that service owners who specify more hedged requests than they have available backends may still see hedged requests sent to the same backend.
```
markdroth: I think the language here should be less prescriptive and more about what functionality can be obtained if desired.  For example, instead of saying "Previously used addresses should be skipped", we can say something like "The load balancing policy may use this information to send retries to an address that was not previously used if possible".  In addition, the last couple of sentences can be changed to indicate that the load-balancing policy will need to decide what to do in the case where there are no addresses that have not already been tried.

a11r:
Thanks for the detailed design and everyone for the thoughtful feedback. I am approving and merging this PR now. Any follow up discussion can continue on the thread.

PR #13, opened 2017-02-14
L4: Add JRuby Support

JasonLunn:


blowmage:
Yes please. Many rubyists choose JRuby for their production environment.

L4.md
```diff
@@ -0,0 +1,77 @@
...
+to avoid creating a requirement on a unified repository that includes the C, Ruby, and Java source in a single project.
+That approach works fine (this is how Google Protobuf handles the same problem), and might even be more optimal, but
+would require repository-level refactoring. The `grpc-java` client already exposes releases via Maven so there would be
+no work required on their part to support this proposal.
+
+## Implementation
+
+1. Update the Gemspec to add JBundler to the list of development dependencies when run on JRuby 
+1. Create a [JarFile](https://github.com/torquebox/maven-tools/wiki/Jarfile) to create a dependency on the grpc-java jar 
+1. Create wrapper code that delegates the existing Ruby API calls to the Java client, as needed
```
apolcyn: A heads up about this, I would expect this shim layer over the java client to end up as about a full grpc-ruby reimplementation, with the requirement that it always present the same surface as the C-wrapping grpc-ruby gem. The majority of the current grpc-ruby code base is probably not immediately reusable - the pure ruby part is thin and a lot of it is tied to the C-extension ruby API, of which the semantics revolve around the grpc C-core library interface. Also the grpc-java libraries present slightly higher level and slightly different interfaces than the grpc C-core library that the ruby C-extensions wrap. For example, one trouble point is "channel args". Currently, users can pass a hash to grpc-ruby `Channel` constructors, of which the key-value pairs are passed straight down to the grpc C-core library - emulating this on top of grpc-java client, whose implementation does need to support these in the same way, would probably be difficult if doable at all.

L4.md
```diff
@@ -0,0 +1,77 @@
...
+1. Address any test regressions or expectation updates introduced by executing on JRuby
```
apolcyn: We probably wouldn't be able to drop the jruby target into the current tests without changes to the tests themselves. This is probably doable but somewhat involved.

The testing suite would have to be updated to add "jruby", as basically a separate language, to the continuously tested language-platform matrix. Also I suspect a lot of the current grpc-ruby tests would have to be refactored to be agnostic of the C-wrapping and java-wrapping implementations (higher level tests should work find but there's lower level stuff in C-extensions is used in tests).

JasonLunn: Is it fair to say that the high level tests are all that we care about in the context of JRuby because low-level tests will be exercised by unit tests with `grpc-java`, and that the rest could be marked to be skipped when running on JRuby?

L4.md
JasonLunn: Is the assumption correct that there was and continues to be either a performance reason and / or an ongoing maintenance reason not to build a pure Ruby implementation? If that assumption holds, do we agree that solutions that build on top of either the C client or the Java client are the only viable ways forward?

JasonLunn:
Should we move discussion to a Google groups thread? I think that phase is supposed to happen after a review is assigned - @apolcyn, should I update the PR to reflect you as a reviewer?

L4.md
apolcyn: Yeah, but just making clear that the grpc-java to grpc-ruby shim would be involved, with API copies probably not feasible. The other option I can see is a JNI wrapper of the C-core library, which would not be small task either but could be useful for other languages outside jruby..

L4.md
apolcyn: For the most part, yeah I think so.

apolcyn:
yeah @JasonLunn we can start a thread on https://groups.google.com/forum/#!forum/grpc-io

@ejona looped you in on this, for the question of writing on top of grpc-java

nicolasnoble:
Discussion is up there by the way: https://groups.google.com/forum/#!topic/grpc-io/4P6jg4ZYtfo

hsaliak:
Closing pull request due to inactivity. Feel free to re-open the comments and revive the PR when appropriate.

JasonLunn:
@hsaliak - I don't understand what your looking to "revive" this PR? I believe that I've replied to all the feedback that I've received on the discussion group thread. I replied to the last comment directly to the poster ( @nicolasnoble ) rather than the thread itself, but didn't receive a reply.

hsaliak:
Ah I did not know that there was a private thread. The public thread ended with a ping with no response.

JasonLunn:
@hsaliak - radio silence from @nicolasnoble since the exchange that I referenced in April.
@apolcyn - From your perspective, what's needed before this PR can be accepted?

hsaliak:
I am going to close this PR for now, as the discussion thread had a few unanswered questions and is now stale..

thelinuxfoundation:

* User @JasonLunn isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

Legogris:
@JasonLunn @hsaliak @apolcyn @ejona86 
I see this is long-dead - are there any blockers to consider this for merging or is it just a matter of CLA? Getting JRuby support is quite significant.

rdubya:
Is there any interest in making this happen at this point? We need this to be able to use google's published gems. I recently reintroduced jruby support to the protobuf gem: https://github.com/protocolbuffers/protobuf/pull/7923 but without grpc support, it doesn't help too much.

JasonLunn:
@rdubya - I'd be interested in reviving this effort, but having been based on the state-of-the-art of 2017, it's probably best if we start over with a fresh proposal. If you're interested in having a collaborator, my contact info is in my GH profile.

rdubya:
Hey @JasonLunn, I'll just be picking at this when I have some spare moments (which are few these days unfortunately.) I figured I'd go the approach that the protobuf gem took and wrap the java version's functionality in a ruby/java wrapper. Even if it doesn't get pulled into this repo, at least there will be a version that can be referenced.

PR #17, opened 2017-03-14
Add convention for naming files

lukaszx0:
`A6` or `P1` is cryptic, even if you understand what category it is not unique so I'm not sure how will store more that a one `A6`. A propose introducing  a file name convention to solve this problem.

cc @a11r @markdroth


markdroth:
This seems reasonable to me.  Any objections?

ejona86:
Oh, yes, please.

PR #18, opened 2017-03-14
Address grpc-io discusion page

ncteisen:


A6.md
```diff
@@ -171,26 +172,33 @@ Servers may explicitly pushback by setting metadata in their response to the cli
 
 Pushback may also be received to a hedged request. If the pushback says not to retry, no further hedged requests will be sent. If the pushback says to retry after a given delay, the next hedged request (if any) will be issued after the given delay has elapsed.
 
-A new metadata key, `"x-grpc-retry-pushback-ms"`, will be added to support server pushback. If the value for pushback is set to -1, then it will be seen as the server asking the client not to retry at all.
+A new metadata key, `"grpc-retry-pushback-ms"`, will be added to support server pushback. The value is to be an integer that represents how many milliseconds to wait before sending a retry. If the value for pushback is set to -1, then it will be seen as the server asking the client not to retry at all.
+
+#### Limits on Retries and Hedges
+Both retries and hedges will be subject to an overall limit that is not set by the service config, but rather exists in the client already. The limit will default to 5, but can be reset by the client explicitly through the use of channel args. This mitigates concerns over the security of the service config being transfered to the client via DNS.
```
markdroth: I think this wording makes it unclear as to how the service config value relates to this overall limit.  I suggest something like the following instead:

"""
The client will support a channel argument indicating the maximum value for the number of retries or hedged requests that may be set via the service config.  If the service config specifies a larger value, the client will ignore that value in favor of the limit specified by the channel arg.  The channel arg value will default to 5.
"""

ericgribkoff: How about:

> Both `maxRetryAttempts` in `retryPolicy` and `maxRequests` in `hedgingPolicy` have, by default, a client-side maximum value of `5`. This client-side maximum value can be changed by the client through the use of channel arguments. Service owners may specify a higher value for these parameters, but higher values will be treated as equal to the maximum value by the client implementation. This mitigates security concerns related to the service config being transferred to the client via DNS.

I think the current wording of 'an overall limit' can too easily be misinterpreted as a combined total limit.

markdroth: We should add a note indicating that this not only mitigates the security concern but also protects against the possibility of typos setting the number of retries or hedged requests too high.

A6.md
```diff
@@ -171,26 +172,33 @@ Servers may explicitly pushback by setting metadata in their response to the cli
...
 
 #### Summary of Retry and Hedging Logic
 There are five possible types of server responses. The list below enumerates the behavior of retry and hedging policies for each type of response. In all cases, if the maximum number of retry attempts or the maximum number of hedged requests is reached, no further RPCs are sent. Hedged RPCs are returned to the client application when all outstanding and pending requests have either received a response or been canceled.
 
 1. OK
-  1. Retry policy: Successful response, return success to client application
-  2. Hedging policy: Successful response, cancel previous and pending hedges
+  * Retry policy: Successful response, return success to client application
```
ericgribkoff: Good catch, I hadn't noticed that the current sublists don't format correctly at all. This is also an issue for the retry statistics section below. I don't particularly like the way this change renders either (it still doesn't come through as sublists). I found https://github.com/jgm/pandoc/issues/200, which suggests that the rendering as sublists might work properly with four spaces of indentation. If that works, can we keep this as ordered sublists?

A6.md
```diff
@@ -171,26 +172,33 @@ Servers may explicitly pushback by setting metadata in their response to the cli
 
 Pushback may also be received to a hedged request. If the pushback says not to retry, no further hedged requests will be sent. If the pushback says to retry after a given delay, the next hedged request (if any) will be issued after the given delay has elapsed.
 
-A new metadata key, `"x-grpc-retry-pushback-ms"`, will be added to support server pushback. If the value for pushback is set to -1, then it will be seen as the server asking the client not to retry at all.
+A new metadata key, `"grpc-retry-pushback-ms"`, will be added to support server pushback. The value is to be an integer that represents how many milliseconds to wait before sending a retry. If the value for pushback is set to -1, then it will be seen as the server asking the client not to retry at all.
```
ericgribkoff: Should probably specify that the integer should not have any leading zeros, unless it is set to `0`. (`...an integer with no unnecessary leading zeros`?)

A6.md
markdroth: Now that the security concerns have been addressed, we can remove the "security" section at the bottom of the document.

ncteisen: Should the design really account for a service owner having a typo in their config file? I agree that it is nice that these hard limits protect against that, but I don't think we need to explicitly call it out as a design consideration.

ncteisen: This is much better. Done

A6.md
ncteisen: https://en.wiktionary.org/wiki/bikeshedding

ncteisen: But I'll change it

ncteisen:
@markdroth, does the update look good to you?

PR #20, opened 2017-03-31
A7: Backward Compatibility Management Proposal

muxi:


muxi:
@a11r ping for review.

A7-backward-compatibility-management.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Client version identification
+gRPC server should identify the version of a client by analysing the `user-agent` header field on the server side. Current format of `user-agent` is well defined by the doc [PROTOCOL-HTTP2](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md). With this definition we can extract gRPC language and version from client's `user-agent` string and identify user's compatibility issues.
+
+### Server backward compatibility management
+gRPC server will maintain a list of server workarounds for backward compatibility issues. These workarounds are disabled by default for performance consideration. If a user cares about a backward compatibility issue, they should be able to enable the workaround. In this case gRPC server will identify the version of each client and determine whether the workaround should be applied.
+
+#### Workaround list
+The workaround list on gRPC server is a map from an integer `workaround id` to a `version comparing function`:
```
ctiller: Avoid talking implementation details here... we almost certainly won't implement it like this (since this API implies that we'd parse user agent strings N times - and that all workarounds would be triggered by user agent - maybe sometimes we'd want to do something else)

A7-backward-compatibility-management.md
```diff
@@ -0,0 +1,94 @@
...
+* The list should be compiled into gRPC server and shipped to users in releases
+* An item is add to the list when a backward compatibility issue is raised and a workaround is implemented on the server
+* Updated list will be pushed to users in the next release (or next import for Google3)
+* An item should be removed from the list after 6 months (roughly 4 minor version releases)
+
+The list should be implemented and maintained by the three core languages for interoperability. Wrapping languages should use surface APIs (below) to access the table.
+
+#### APIs
+```
+GRPCAPI void enable_workaround(uint32_t id);
```
ctiller: Skip discussing the core API's here: instead talk about what abstract API's an implementation must export.

Example:

When building a server, an implementation should support enabling/disabling workarounds by ID. Those ID's should be listed in an enumeration.

A7-backward-compatibility-management.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Rationale
+### User-agent field
+Since gRPC currently follows a 6-weeks release cycle, using current `user-agent` field to identify a client's version should allow users to be blocked by a backward compatibility issue for at most 6 weeks. This should be acceptable in most cases.
+
+### Version comparing function
+Instead of determining client's compatibility by a minimum version number, we decided to make it more general with `version comparing function`. This allows more flexibility in determining client's compatibility, such as including client's language into consideration.
+
+### Performance
+Current API requires a client's `user-agent` string to be checked against the version comparing functions of all items in the workaround list, involving a lot of (likely duplicated) string processing. Depends on how much performance we need, an alternative is to extract key information from `user-agent` (e.g. gRPC version, language, transport, etc.) and use them as parameter of `version comparing functions`. It is a tradeoff of flexibility to performance that we may want to consider.
```
ctiller: The proposed implementation is an absolute no in terms of performance.

That said, I don't think this RFC needs to consider the actual implementation, and instead concentrate on how this is exposed and how workarounds are to be curated.

hsaliak:
@muxi what's the status of this proposal? Can this be closed?

PR #22, opened 2017-04-24
A8: Client-side keepalive

ejona86:
grpc-io discussion: https://groups.google.com/d/topic/grpc-io/Yc5_vSUJgwQ/discussion


ejona86:
@a11r, could you approve?

PR #23, opened 2017-04-24
A9: Server-side connection management

ejona86:
grpc-io discussion: https://groups.google.com/d/topic/grpc-io/DIiIMiq9sV8/discussion


ejona86:
@a11r, could you approve?

PR #25, opened 2017-05-05
L7: Go Metadata API proposal

dfawley:


dfawley:
@a11r, could you please review this proposal?  I'd like to move as quickly as possible on this to avoid further problems caused by the behavior change that fixed a security issue, but caused correctness issues for a few use cases.  Thanks!

PR #26, opened 2017-06-15
P2 : Issue disposition proposal

makdharma:


P2-issue-disposal.md
```diff
@@ -0,0 +1,57 @@
...
+* Author(s): makdharma
+* Approver: a11r
+* Status: Draft
+* Implemented in: N/A
+* Last updated: 2017-06-15
+* Discussion at: N/A
+
+## Abstract
+
+There are hundreds of open issues in the 3 gRPC repos, some open for over 2 years. gRPC team periodically spends valuable time scrubbing the issue list. This proposal lays down concrete criteria for disposing of issues filed by OSS community.
```
jayantkolhe: >  down concrete criteria for disposing of issues filed by OSS community.

Should be: 
down concrete criteria for disposing of issues filed.

P2-issue-disposal.md
```diff
@@ -0,0 +1,57 @@
...
+There are hundreds of open issues in the 3 gRPC repos, some open for over 2 years. gRPC team periodically spends valuable time scrubbing the issue list. This proposal lays down concrete criteria for disposing of issues.
+
+## Background
+
+The github Issues database serves many purposes.
+
+* Capture bugs
+* Keep track of TODO
+* Submit feature requests
+* Ask questions (ad hoc documentation)
```
mehrdada: It is historically used for questions, but continuation of such use should be discourage IMO, especially in light of the new bug template.

P2-issue-disposal.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Proposal
+
+We propose the following.
+
+### Keep status updated
+
+Close following types of issues with explanation. Anyone can reopen if they disagree with the disposition.
+* Issues that are fixed, or should have been fixed
+* Issues that are waiting for more information and have not seen an update in over a month
```
mehrdada: Isn't a shorter time, as @hsaliak suggested, better?

P2-issue-disposal.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Correct classification
+
+Issues that start out as bugs can turn into feature requests or document requests. Use following *mutually exclusive* labels to identify the nature of an issue - Question, Enhancement, Bug.
+
+### Bring transparency to feature request handling
+
+* Old feature requests (pending for over a year)
+	* Make an honest assessment about feasibility and utility. Reject if appropriate with explanation. If the community feels strongly about the feature, anyone can open a new issue with added justification. This issue can then follow the New Issue path below.
+	* If the feature request makes sense but we have no immediate plans, then consolidate in a separate document and close the issue. This is equivalent of adding "Someday" label.
```
ejona86: Could this be expanded more? I don't get a strong grasp of how this would look.

Given how easy it is to filter issues with `Someday...` _milestone_ (or the Java-equivalent of Unscheduled milestone), what is this providing? It seems currently there are 100 open `Someday...` issues, and actually 97 are closed. Taking a small sampling of 6ish closed issues showed all were resolved, so maybe removing them from the issue tracker isn't a good idea, or maybe not everyone is using `Someday...` as strongly as would be done here.

Interestingly, Java is in a worse place here than grpc/grpc with 97 open Unscheduled issues.

P2-issue-disposal.md
```diff
@@ -0,0 +1,66 @@
...
+### Correctly classify issues
+
+Issues that start out as bugs can turn into feature requests or document requests. Use following **mutually exclusive** labels to identify the nature of an issue - Question, Enhancement, Bug.
+
+### Bring transparency to feature request handling
+
+* Old feature requests (pending for over a year)
+	* Make an honest assessment about feasibility and utility.
+	* Reject if appropriate with explanation. If the community feels strongly about the feature, anyone can reopen the issue with added justification.
+	* If the feature request makes sense but we have no immediate plans, then close the issue with "WONTFIX-P4".
```
hsaliak: P4 may not provide external users with the necessary context, can we change it to something more descriptive? "wontfix- not prioritized" or "wont fix, needs prioritization"

P2-issue-disposal.md
ejona86: If the item is closed, then it is out-of-sight, out-of-mind and is the equivalent of saying "no." I do think we should say "no" to things, but I don't think we should close items and say, "we may work on this eventually."

I feel like Java's "Unscheduled" or something similar fairly clearly says "not being worked on". "someday" is a milestone label that can be changed to something better in 15 seconds, if you knew what you wanted to change it to. But really, to me "someday" is already a clear description to me.

makdharma: "If the item is closed, then it is out-of-sight, out-of-mind" - I think keeping the issue open with no plans (Zombie issues) weighs down on productivity. I (personally) prefer to put zombie issues out-of-sight, out-of-mind and revisit periodically (Quarterly planning).

ejona86: And that's possible today. Any new issues are triaged. Any active issues are in a milestone/label that you look at. Closing it will seem to users like we're saying we're not going to do it, independent of what we claim the meaning is here. And based on "putting it in a document," I feel like I wouldn't end up revisiting it because that sounds horrible to dig through and figure out enough context/maintain the document.

hsaliak: I do not think this should go into a doc. Which is why these issues need to have a clear label associated with them that is specific for this purpose of triaging. 
Closing them is to exactly convey that we are not going to do it, until we get a better input to force up prioritization..

hsaliak:
I am closing this PR - we ultimately did not end up rolling this out.

PR #27, opened 2017-06-22
Clarify retry budgeting algorithm

lukaszx0:
Correct me if I'm wrong, but my understanding is that `token_count` should be incremented only if it's lower than `maxTokens`. Otherwise it'll grow indefinitely so if service will be stable for a day and generate 1M tokens, next day 1h outage won't trigger the algorithm.

cc @ericgribkoff @ncteisen @ejona86

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -134,7 +134,7 @@ When a non-error response is received (in response to any of the hedged requests
 
 If a non-fatal status code is received from a hedged request, then the next hedged request in line is sent immediately, shortcutting its hedging delay. If any other status code is received, all outstanding RPCs are canceled and the error is returned to the client application layer.
 
-If all instances of a hedged RPC fail, there are no additional retry attempts. Essentially, hedging can be seen as retrying the original RPC before a failure is even received. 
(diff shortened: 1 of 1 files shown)


lukaszx0:
Actually, I think that same clarification is need for decrementing `token_count` - _decrement if token_count > 0_

ncteisen:
Good catch! But rather than adding conditionals to the bullet points, how about we add a sentence like:

"the gRPC client maintains a `token_count` variable which is initially set to `maxTokens`, and is always clamped between 0 and `maxTokens`"

Do you think that is clear enough?

lukaszx0:
@ncteisen reworded a bit, what do you think?

lukaszx0:
:shipit:

lukaszx0:
@ericgribkoff extracted whitespace fixes to separate commit.

:shipit: ?

ericgribkoff:
Ah, sorry about that - I would have merged this as-is (the nit was surely overestimating the future value of an accurate blame on every line of this file...), but I didn't realize that the merge was waiting on me due to permissions. Thanks for the fix!

PR #29, opened 2017-07-24
Use server process to identify single instance of app

lukaszx0:
Server name is a bit vague and can be confusing.

I'm actually not sure if it's even meant to be single process? In any event, I think it's worth clarifying, few of my colleagues were confused when reading this doc too.

@ncteisen @ericgribkoff

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -142,7 +142,7 @@ Hedged requests should be sent to distinct backends, if possible. To facilitate 
 
 #### Throttling Retry Attempts and Hedged RPCs
 
-gRPC prevents server overload due to retries and hedged RPCs by disabling these policies when the client’s ratio of failures to successes passes a certain threshold. The throttling is done per server name. Retry throttling may be configured as follows:
+gRPC prevents server overload due to retries and hedged RPCs by disabling these policies when the client’s ratio of failures to successes passes a certain threshold. The throttling is done per server process. Retry throttling may be configured as follows:
 
 ```
 "retryThrottling": {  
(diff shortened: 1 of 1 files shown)


ericgribkoff:
I agree that the current text is overly vague. I don't think server process is the right change here though: the throttling mechanism is only on the gRPC client side, and each client process implements its throttling independently of other client processes. The intent was that each client tracks the failure rate for each "logical" server it is connected to, and if the failure rate is too high for a particular "logical" server, retries to that server are throttled. "Logical" server is what looks like a single server to the application layer - e.g., a particular host name, which may actually be a load-balancer distributing requests to multiple servers and/or resolve to multiple addresses for client-side load balancing. I think, to be precise, "server name" = authority portion of the URI.

This brings up something that I don't think we considered in the design phase. The specified, and desired, behavior is this: given a hostname h that resolves to IPs ip1 and ip2, failures on RPCs to either ip1 or ip2 will count "against" h in terms of throttling. But the client is also free to directly establish separate channels to ip1 and ip2 without using the shared hostname, and thus may end up in a situation where retries to ip1 are throttled but retries to ip2 are allowed. We can also have two separate hostnames h1 and h2 that both resolve to ip1, and again end up throttling on h1 and not on h2 even though they may be the same actual server. I don't think the spec needs to do anything to attempt to handle these edge cases separately, but maybe they serve as examples-by-contrast that better illustrate the intended throttling behavior.

hsaliak:
@ericgribkoff  /cc @ejona86 
What's the next step here? Could you please close this PR if we are not going to accept the change, could you please close? Also, please add the edge-case examples for the throttling behaviour as you mention in the comment in a separate PR..

PR #30, opened 2017-07-24
L9: new balancer and resolver APIs for gRPC-go

menghanl:
The new API will support:
* switching balancers based on service config
* get service config updates from resolver

Tracking issue: https://github.com/grpc/grpc-go/issues/1388


L9-go-resolver-balancer-API.md
```diff
@@ -0,0 +1,504 @@
...
+* Author(s): menghanl, dfawley, mmukhi
+* Approver: a11r
+* Status: In Review
+* Implemented in: <language, ...>
+* Last updated: 2017/07/19
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/rUDN7-cZvz8
+
+## Abstract
+
+The current balancer API in gRPC-go cannot suport balancer hot switching. There are also other issues related to balancer, resolver and connectivity state.
```
dmitris: typo: s/suport/support

L9-go-resolver-balancer-API.md
```diff
@@ -0,0 +1,504 @@
...
+
+Note that the updates in the current resolver API are delta, not the full address list. This is the reason of one of our issues.
+
+##### If balancer is not provided
+
+If balancer is not provided, the target name will be used directly to do net.Dial(). net.Dial() will do the resolving part and return a connection for use.
+
+#### Connection up and down
+
+gRPC notifies the balancer of addresses to which the connection is established or lost. So balancer can make load balancing decisions based on the connection status.
```
xiang90: how will this interact with health checking? if http2 or customized health checker fails, will the connection change state to lost, and will the balancer get notified?

menghanl: gRPC notices a connection is lost when it fails to read-from/write-to the connection. (We have [`keepalive`](https://github.com/grpc/grpc-go/blob/master/keepalive/keepalive.go) package for point-to-point healthcheck).

If there's another custom health checker reporting unhealthy but the connection is actually working, gRPC will not notice that.

xiang90: > If there's another custom health checker reporting unhealthy but the connection is actually working, gRPC will not notice that.

we probably need a hook for that too. Or we have to manually close that connection somehow. 

for quorum based systems, network partitioned server (partitioned from the majority of its peers) is not really usable for some types of RPC although the network between that server and client is OK.

menghanl: In this case, I think it would make more sense for the health checker to communicate to the balancer directly, not through gRPC.

This is what we do in `grpclb`. `grpclb` connects to a remote balancer, and the remote balancer decides which backends to connect and which to disconnect.

xiang90: what you suggest is what we are doing right now. but then down (or done) func will have multiple consumers, and i feel it is less cleaner. i will take a look at grpcLB code. 

thank you.

hsaliak:
@dfawley @menghanl please merge this gRFC

PR #32, opened 2017-08-29
Change license to Apache 2.0

mehrdada:


LICENSE
```diff
@@ -1,28 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
chwarr: Does this line need to be updated with the copyright holder details?

jtattermusch: I don't think so. It's the same as https://github.com/grpc/grpc/blob/master/LICENSE

jtattermusch:
Thanks @mehrdada! Can we also switch https://github.com/grpc/grpc.github.io to Apache 2.0?

jtattermusch:
There's also https://github.com/grpc/grpc-docker-library

PR #33, opened 2017-09-06
L10: A user friendly C++ gRPC API using Reactive Streams

per-gron:
Reviewing this proposal may take quite some time, especially if you do not know Reactive Streams, I am sorry about that. Hopefully it will be worth your time!

I would love to get any and all feedback, from anyone, about tiny details and the overall idea. Is it good? Bad? Irrelevant? Useless? Flames welcome.

### DIFF
--- a/L10-c++-reactive-streams.md
+++ b/L10-c++-reactive-streams.md
@@ -98,6 +98,7 @@ class TestServiceImpl : public TestService::ReactiveService {
 service UpstreamService {
   rpc GetUserInfo (UserInfoRequest) returns (UserInfoResponse) {}
   rpc ResolveUsername (ResolveUsernameRequest) returns (ResolveUsernameResponse) {}
+  rpc GetShoppingCart (GetShoppingCartRequest) returns (GetShoppingCartResponse) {}
 }
 
 message UserInfoRequest {
@@ -117,9 +118,19 @@ message ResolveUsernameResponse {
   string userid = 1;
 }
 
+message GetShoppingCartRequest {
+  string userid = 1;
+}
+
+message GetShoppingCartResponse {
+  repeated string items = 1;
+}
+
 service TestService {
   rpc CountryByUserid (CountryByUseridRequest) returns (CountryByUseridResponse) {}
   rpc CountryByUsername (CountryByUsernameRequest) returns (CountryByUsernameResponse) {}
+  rpc RenderShoppingCart (RenderShoppingCartRequest) returns (RenderShoppingCartResponse) {}
+  rpc CountryByUseridRace (CountryByUseridRequest) returns (CountryByUseridResponse) {}
 }
 
 message CountryByUseridRequest {
@@ -137,6 +148,15 @@ message CountryByUsernameRequest {
 message CountryByUsernameResponse {
   string country = 1;
 }
+
+message RenderShoppingCartRequest {
+  string userid = 1;
+}
+
+message RenderShoppingCartResponse {
+  string username = 1;
+  repeated string items = 2;
+}
 ```
 
 ```cpp
@@ -146,10 +166,23 @@ UserInfoRequest MakeUserInfoRequest(const std::string& userid) {
   return request;
 }
 
+GetShoppingCartRequest MakeGetShoppingCartRequest(const std::string& userid) {
+  GetShoppingCartRequest request;
+  request.set_userid(userid);
+  return request;
+}
+
+CountryByUseridResponse MakeCountryByUseridResponse(
+    const std::string& country) {
+  CountryByUseridResponse response;
+  response.set_country(country);
+  return response;
+}
+
 class TestServiceImpl : public TestService::ReactiveService {
  public:
   /// Request handler that makes one upstream request
-  AnyPublisher<CountryByUserIdResponse> CountryByUseridHandler(
+  AnyPublisher<CountryByUserIdResponse> CountryByUserid(
       CountryByUseridRequest request) override {
     return AnyPublisher<CountryByUserIdResponse>(Pipe(
         // Make RPC call to get userinfo
@@ -164,7 +197,7 @@ class TestServiceImpl : public TestService::ReactiveService {
   }
 
   /// Request handler that makes two upstream requests in a row
-  AnyPublisher<CountryByUsernameResponse> CountryByUsernameHandler(
+  AnyPublisher<CountryByUsernameResponse> CountryByUsername(
       CountryByUsernameRequest request) override {
     ResolveUsernameRequest resolve_username_request;
     resolve_username_request.set_username(request.username());
@@ -179,9 +212,53 @@ class TestServiceImpl : public TestService::ReactiveService {
         }),
         // Then take the resulting userinfo and construct a response
         Map([](UserInfoResponse response) {
-          CountryByUsernameResponse final_response;
-          final_response.set_country(response.country());
-          return final_response;
+          return MakeCountryByUseridResponse(response.country());
+        })));
+  }
+
+  /// Request handler that makes two upstream requests in parallel and combines
+  /// the results.
+  AnyPublisher<RenderShoppingCartResponse> RenderShoppingCart(
+      RenderShoppingCartRequest request) override {
+    auto user_info = upstream_service_.GetUserInfo(
+        MakeUserInfoRequest(request.userid()));
+    auto shopping_cart = upstream_service_.GetShoppingCartRequest(
+        MakeGetShoppingCartRequest(request.userid()));
+
+    return AnyPublisher<CountryByUsernameResponse>(Pipe(
+        // Wait for getting the user info and shopping cart together. This
+        // operator gives us a Publisher of
+        // std::tuple<UserInfoResponse, GetShoppingCartResponse>
+        Zip(user_info, shopping_cart),
+        // Here, it would be possible to use Map with a mapper function that
+        // takes a tuple but that's a bit inconvenient. Splat helps with that:
+        Map(Splat([](
+            UserInfoResponse user_info_response,
+            GetShoppingCartResponse get_shopping_cart_response) {
+          RenderShoppingCartResponse response;
+          response.set_username(user_info_response.username());
+          (*response.mutable_items()) = get_shopping_cart_response.items();
+          return response;
+        }))));
+  }
+
+  /// Request handler that makes two upstream requests in parallel and uses the
+  /// response that comes first. (The rpc that turned out to be slower is
+  /// cancelled.)
+  AnyPublisher<CountryByUserIdResponse> CountryByUseridRace(
+      CountryByUseridRequest request) override {
+    auto user_info = upstream_service_.GetUserInfo(
+        MakeUserInfoRequest(request.userid()));
+    return AnyPublisher<CountryByUserIdResponse>(Pipe(
+        // Make two requests to get the user info in pararallel. Publishers are
+        // lazy and can usually be used more than once. This could also be
+        // different requests.
+        Merge(user_info, user_info),
+        // We are only interested in one of the responses.
+        Take(1),
+        // Convert the result to a response of the proper type
+        Map([](UserInfoResponse response) {
+          return MakeCountryByUseridResponse(response.country());
         })));
   }
 
@@ -238,6 +315,7 @@ class TestServiceImpl : public TestService::ReactiveService {
 
 * Because this is a client streaming RPC, the parameter to `TestServiceImpl::Sum` is an `AnyPublisher<SumRequest>` instead of a `SumRequest` directly. This is the only difference between non-streaming and client streaming RPCs for application code.
 * `Sum()` as used here is a rather silly example, but it is possible for users of the API to write their own operators: [Here is the implementation of `Sum()`](https://github.com/per-gron/shuriken/blob/master/src/rs/include/rs/sum.h). Writing another operator is as easy as writing `Sum()`.
+* For more info on `Zip`, see its documentation [in rs](https://github.com/per-gron/shuriken/blob/master/src/rs/doc/reference.md#zippublisher) and [in ReactiveX](http://reactivex.io/documentation/operators/zip.html).
 
 
 #### Server streaming: FizzBuzz
@@ -299,7 +377,6 @@ class TestServiceImpl : public TestService::ReactiveService {
 More examples are possible, but this is already getting long. Please ask for more if you want to see what a particular use case would look like. Omitted examples include:
 
 * **Bidirectional streaming:** It is just client and server streaming combined. The handler takes a Publisher and returns a Publisher that can emit more than one element.
-* **Making two upstream calls in parallel:** Can be done with the [`Zip` operator](https://github.com/per-gron/shuriken/blob/master/src/rs/doc/reference.md#zippublisher).
 * **Cancellation:** Most of the time no explicit code related to cancellation is needed. For example, if an RPC has made two upstream requests in parallel and one fails, the other is automatically cancelled. This works with timeouts and streams, making two calls and using just the fastest result ([`Merge()`](https://github.com/per-gron/shuriken/blob/master/src/rs/doc/reference.md#mergepublisher)+[`Take(1)`](https://github.com/per-gron/shuriken/blob/master/src/rs/doc/reference.md#takecount)) etc as well.
 * **Error handling:** With errors, Publishers behave very much like futures/promises do in most languages/implementations. It is basically an async version of `try`/`catch`. Errors are automatically propagated through the chain of operators until they are caught, usually with the `Catch` operator.
 
--- a/L10-c++-reactive-streams.md
+++ b/L10-c++-reactive-streams.md
@@ -1,0 +1,387 @@
+A user friendly C++ gRPC API using Reactive Streams
+----
+* Author(s): Per Grön (pgron@google.com)
+* Approver: a11r
+* Status: Draft
+* Implemented in: C++ (but could be extended to more languages)
+* Last updated: 2017-09-06
+* Discussion at: <google group thread> (filled after thread exists)
+
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,387 @@
...
+* **Making two upstream calls in parallel:** Can be done with the [`Zip` operator](https://github.com/per-gron/shuriken/blob/master/src/rs/doc/reference.md#zippublisher).
+* **Cancellation:** Most of the time no explicit code related to cancellation is needed. For example, if an RPC has made two upstream requests in parallel and one fails, the other is automatically cancelled. This works with timeouts and streams, making two calls and using just the fastest result ([`Merge()`](https://github.com/per-gron/shuriken/blob/master/src/rs/doc/reference.md#mergepublisher)+[`Take(1)`](https://github.com/per-gron/shuriken/blob/master/src/rs/doc/reference.md#takecount)) etc as well.
+* **Error handling:** With errors, Publishers behave very much like futures/promises do in most languages/implementations. It is basically an async version of `try`/`catch`. Errors are automatically propagated through the chain of operators until they are caught, usually with the `Catch` operator.
+
+
+#### Threading model
+
+The default threading model for a server that uses this library is:
```
ctiller: Why not take an approach ala the ThreadManager used in the synchronous API to allow autosizing the thread count based upon active workload?

Such an approach would eliminate the need for line 313 in this document, and further simplify exposed API.

L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,387 @@
...
+This way, an application will be able to utilize multiple threads, yet no synchronization primitives are needed in the rs library, and they are needed in application code only if it has to do slow work or if requests need to interact.
+
+
+## Rationale
+
+Alternate approaches for a C++ for gRPC include
+
+* **The current synchronous API:** Too slow for many use cases.
+* **The current asynchronous API:** Too low level for many use cases.
+* **Futures, as suggested in [issue #7352](https://github.com/grpc/grpc/issues/7352):** Futures typically do not have a concept of cancellation and they do not map naturally to streams.
```
ctiller: We have a more fleshed out proposal for futures internally, and cancellation is fully supported. Hoping @vjpai can take some time to publish it soon.

L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,387 @@
...
+This document proposes a new C++ API for gRPC, in addition to the current synchronous and asynchronous APIs. It is based on rather than replaces the current async API. If the C++ version is successful, this proposal could be extended to cover more languages in the future.
+
+The proposal has two main parts: The Reactive Streams library and the new gRPC API.
+
+
+### A C++ Reactive Streams library
+
+[Reactive Streams](http://www.reactive-streams.org/) is currently defined only for Java and Javascript. This project needs a C++ version.
+
+In preparation for this RPC the author has written [a Reactive Streams library in C++ called rs](https://github.com/per-gron/shuriken/tree/master/src/rs). rs is well documented, well tested and implements many of the [ReactiveX operators](http://reactivex.io/documentation/operators.html). Reading rs' documentation may help explain parts of this RFC.
```
ctiller: Would accepting this proposal imply taking a dependency on rs?

Additional forced dependencies on users may be a deal breaker for us.

L10-c++-reactive-streams.md
per-gron: Thanks for the comment! The threading model section is the one I have put the least effort in. I will read and think more about this tomorrow and update the doc and/or reply here.

L10-c++-reactive-streams.md
per-gron: That sounds very interesting! I would love to read that.

L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+*Notes:*
+
+* `Pipe` is a function in rs that lets you "pipe" a Publisher through a series of operators. The first argument is a Publisher/stream, subsequent arguments are operators on that stream.
+* `Pipe(a, b, c)` is essentially the same as `c(b(a))`. Having `Pipe` like this isn't necessary. Another option is to let the operators be methods on a base Publisher type like RxJava does, but then it's impossible for user code to add their own first-class operators.
+* `ConcatMap` is similar to `Map` but it lets the mapper function return a Publisher instead of a value directly. I would personally call it `FlatMap` but I've stuck to ReactiveX's naming (there is a `FlatMap` in ReactiveX, but it behaves in an unintuitive way and its implementation requires an unbounded buffer).
+
+
+#### Client streaming: Sum
```
ctiller: Please provide an example whereby a client makes calls to multiple backends, with variants:
1. waits for all responses to come back, and then forms a coalesced response
2. waits for one response to come back, forwards it, and cancels all other requests

These patterns are incredibly common in real systems, and providing good support for them needs to be a requirement for any API addition.

L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,387 @@
...
+* If a request handler does heavy work, it is responsible for doing that somewhere else than on the threads created by the library; in order to be able to jump back the library provides a `Schedule` function that asynchronously executes a callback on the original `CompletionQueue` thread.
+
+This way, an application will be able to utilize multiple threads, yet no synchronization primitives are needed in the rs library, and they are needed in application code only if it has to do slow work or if requests need to interact.
+
+
+## Rationale
+
+Alternate approaches for a C++ for gRPC include
+
+* **The current synchronous API:** Too slow for many use cases.
```
ctiller: The synchronous API does have some performance issues right now... if these were addressed (they are addressable), would you advocate for a new API?

L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,387 @@
...
+* [rs](https://github.com/per-gron/shuriken/tree/master/src/rs) breaks Google's code style in the following ways. I am unsure about how exactly to fix this.
+  * **It uses C++14:** This is mostly for ease of implementation; it should be possible to backport to C++11.
+  * **It uses rvalue references:** These could mostly be changed to by-value instead, but I have a hard time justifying the performance penalty.
+  * **It uses `std::exception_ptr`:** Reactive Streams has built in error handling that is based on having a generic type for errors. `std::exception_ptr` is C++'s type for that. rs could use something else, but the chances of being able to convince the Reactive Streams people that rs should become the official C++ version of Reactive Streams is likely low without `std::exception_ptr`. Please note that rs never throws, it uses `std::exception_ptr` only as a container type.
```
ctiller: grpc also has a generic container for error handling (grpc::Status).

It's hard to justify introducing a wrinkle that forces conversion.

L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,387 @@
...
+
+
+## Open issues
+
+* The RPC handler methods might need to take one or more additional parameters to handle gRPC features like metadata.
+* This proposal glosses over many details; the API needs to be fleshed out more.
+* There are certain details in the current gRPC C++ API that seem to be missing: For example, the author could not find how get a signal in the server when a unary RPCs is cancelled.
+* The rs library is a fairly minimal implementation of Reactive Streams. One thing that many Rx libraries have that rs does not is a concept for a Publisher that emits only one value. This would increase the complexity of the library a bit, but it could be worth it for the additional type safety and performance benefits it would bring.
+* [rs](https://github.com/per-gron/shuriken/tree/master/src/rs) breaks Google's code style in the following ways. I am unsure about how exactly to fix this.
+  * **It uses C++14:** This is mostly for ease of implementation; it should be possible to backport to C++11.
```
ctiller: We'll need to stay with C++11 for some time.

L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,387 @@
...
+* Reactive Streams offers a clear and composable approach to writing asynchronous functions within a server as well: An asynchronous function is one that returns a Publisher. This can greatly help API consistency within a given programming language, not just on the wire protocol level, as well as making it easy to on a type level distinguish asynchronous code from synchronous code.
+* A concrete example of how code can benefit from Reactive Streams is [Netflix's Hystrix library](https://github.com/Netflix/Hystrix/wiki/How-it-Works), which helps provide latency and fault tolerance for microservices.
+
+The main drawbacks of this proposal are:
+
+* It is easy to use only after you have spent some time learning Reactive Streams' concepts: You must change mindset to thinking about data flows.
+* There are features in gRPC that don't obviously fit Reactive Streams, for example metadata.
```
ctiller: There needs to be a story for metadata.

L10-c++-reactive-streams.md
per-gron: I can see a few ways to do this:

1. Since rs is not used anywhere else and was made basically for this purpose, rs could be made a part of this and included with the rest of the code.

2. If rs is deemed to heavy of a dependency (the implementation + test suite is currently ~7.5kLOC counted with `wc -l`), then we could include only the interface part of rs and leave out all of the operators (they are most of the code). This seems a bit awkward to me but is certainly doable.

3. Given that Publisher is a "concept" (as in forward iterator etc), it is probably technically possible to not have a dependency to rs at all. However, in practice you would have to depend on rs to get things done.

L10-c++-reactive-streams.md
per-gron: I will update with examples demonstrating these use cases soon.

L10-c++-reactive-streams.md
per-gron: Interesting. Without the performance aspect the case for a new API certainly becomes weaker.

Personally I would still prefer this style of API because I like the model of thinking it promotes. It seems like this style of API could be more powerful in terms of expressing concurrency as well but I am not entirely sure.

What do you think?

L10-c++-reactive-streams.md
per-gron: I understand that. I am not opposed to changing this part of rs.

Perhaps it would be possible to change it so that the error type is configurable via template or something. But if it becomes too complicated I think it could be okay to tie it to gRPC's types.

L10-c++-reactive-streams.md
per-gron: I understand. If people actually are interested in this I will definitely consider backporting it. It's just a bit of a pain to do and parts of the implementation would be uglier/harder to read.

L10-c++-reactive-streams.md
per-gron: Ok. I will read and think about this more. This might take some time for me to fix however; this is more work than of your other comments so far.

L10-c++-reactive-streams.md
per-gron: I have added a bullet point under the "Implementation" section to change rs so that `grpc::Status` can be used as the error type.

With this I consider this issue resolved, let me know if you disagree.

L10-c++-reactive-streams.md
per-gron: I have added a bullet point under the "Implementation" section to backport to C++11.

With this I consider this issue resolved, let me know if you disagree.

L10-c++-reactive-streams.md
ctiller: It might be worth looking at how grpc-java, grpc-objective-c deal with this, as they each have reactive stream allowances in their API's.

@ejona86, @jcanizales care to provide pointers?

per-gron: I will have a look. I don't think this will necessarily be a hard problem to solve. I have simply assumed that it should be possible somehow and not thought about it.

L10-c++-reactive-streams.md
per-gron: I have added those examples now. For 1, see `RenderShoppingCart`. For 2, see `CountryByUseridRace`.

Notes:

If you don't like `Zip` returning a tuple and the `Splat` helper, `Zip` could be changed so that its last parameter is a function with one parameter per input stream that gets to combine the values (RxJava does this). Personally I can't decide which I prefer.

rs does not currently have variants of `Zip` and `Merge` that take a list of input streams; they are simply variadic functions. This is not a fundamental limitation, I just haven't written them yet.

It's a very contrived example, but you might be interested in knowing that the pattern of making one request twice and letting them race can be implemented as an operator (I'm cheating with C++14 here but it can be done in C++11 as well with `AnyPublisher` (with some runtime overhead) or more verbosity):

```cpp
template <typename Input>
auto TwinRace(const Input &input) {
  return Pipe(
      Merge(input, input),
      Take(1));
}
```

which could then be used as an operator like any other. With the addition of a sleep/delay operator (not in the current implementation), you could do optimistic retries:

```cpp
template <typename Input>
auto TwinRace(const Input &input) {
  using namespace std::chrono_literals;
  return Pipe(
      Merge(input, Delay(100ms, input)),
      Take(1));
}
```

Perhaps even better than `Take(1)` would be to use the [race operator](http://rxmarbles.com/#race) but rs doesn't have that currently.

---

With this I consider this issue resolved, let me know if you disagree.

L10-c++-reactive-streams.md
ejona86: grpc-java doesn't expose Metadata on the stubs; Metadata is only exposed to interceptors.

L10-c++-reactive-streams.md
```diff
@@ -0,0 +1,465 @@
...
+
+## Implementation
+
+A prototype of what is suggested in this RFC can be found [here](https://github.com/per-gron/shuriken/tree/master/src/rs-grpc). It implements the most important features, but it is still quite incomplete.
+
+Implementation of this RFC would involve:
+
+* Prepare the [rs](https://github.com/per-gron/shuriken/tree/master/src/rs) library so that it can be accepted as gRPC code:
+  * Backport to C++11.
+  * Remove usage of `std::exception_ptr` and make it possible to use `grpc::Status` instead.
```
per-gron: @ctiller you said (starting a new thread because the other one is hidden by the Github interface)

> grpc also has a generic container for error handling (grpc::Status). It's hard to justify introducing a wrinkle that forces conversion.

Would you use `grpc::Status` for non-gRPC errors also, like HTTP errors or Cassandra errors? A strength of Reactive Streams is that it provides a uniform API style. Unless you would advocate for wrapping non-gRPC errors in `grpc::Status`, I think there might be a good case for wrapping errors, to be able to handle different kinds.

The syntactic overhead does not have to be that large. In the current version of rs, you usually catch errors like this: `Catch([](std::exception_ptr error) { ... })`. Instead of this, `Catch` could allow you to specify what you want to catch:

```
Catch<grpc::Status>([](grpc::Status status) { ... })
```

That way, you could also catch your `CassandraError`s or `HttpError`s or whatever, like you would with normal `try`/`catch` statements. The only extra code the user has to write is `<grpc::Status>`. Since the benefit of doing it like this is to not force users to encode non-gRPC errors in `grpc::Status`, it could be worth it.

This type of `try`/`catch` style with type matching could be done using `std::exception_ptr` but it could also without too much effort be done with an rs wrapper type.

L10-c++-reactive-streams.md
per-gron: Are you sure that such an approach would not have a negative performance impact? Every test and benchmark I have ever seen has clearly showed that having much more than 1 thread per core causes performance to degrade significantly when the server is under heavy load due to context switches.

I am pretty sure you know more about this stuff than I do so if you say so I will take your word for it but it does surprise me.

Or is the point that doing everything async is best, but if you do have blocking operations, you might as well do it on the same thread pool?

L10-c++-reactive-streams.md
per-gron: @ctiller I have read a little bit about metadata and how interceptors work in grpc-java. I can see a few different directions that I could go designing this, and would appreciate advice on which I should try first. We could do more than one or even all of them at once, but for the sake of simplicity and MVP I would prefer trying only one at first if possible. The main directions I can think of are

1. Expose metadata directly to individual rpc handlers (aka what grpc-java does **not** do): This is probably the most conceptually lightweight approach and is appropriate if metadata is often used for RPC-specific purposes.
2. Introduce some kind of Reactive Streams style Interceptor API. This is good for doing middleware-style things such as auth, logging, metrics, tracing etc.
3. Add an Interceptor API to the async C++ API (perhaps it can even be done strictly as a layer outside of it). I think this could be done without adding too much complexity to the current async API. It might lead to lower level code in middleware than option 2. However, middlewares wouldn't need to depend on if the application uses Reactive Streams or not. It might even be possible to make the same Interceptor interface work with the sync API.

I like option number 3 even though it seems like the most amount of work, because I think middleware writers would benefit from it. What do you think?

jcanizales: On Objective-C, request and response metadata are exposed as properties of the "call" object. This call object acts as an Rx-style source of response messages and a sink of request messages.

per-gron:
@a11r what is the next step? Based on how the process is described, it seems like I should be waiting for this RFC to be assigned an APPROVER, no?

per-gron:
@a11r @ctiller As I am awaiting a response to the question of what the next step forward is for this RFC, I have been hacking on the prototype implementation. I have fixed the most serious bugs and written a protoc plugin. It is now possible to write real services using the prototype even though it does not support all gRPC features. See for example [this test](https://github.com/per-gron/shuriken/blob/master/src/rs-grpc/test/unary_test.cpp) and [its .proto file](https://github.com/per-gron/shuriken/blob/master/src/rs-grpc/test/rsgrpctest.proto).

At the moment, I am writing some real world-like services to see what it's like and to find and fix the most glaring usability issues.

ctiller:
Hey - I've been out on leave for a little bit, and I plan to get back to
this in the next few days.

hsaliak:
@ctiller  /cc @a11r  @vjpai  what's next for this RFC. Should someone else pick up the review here and make a determination on whether this can be supported?

thelinuxfoundation:

* User @per-gron isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

per-gron:
The CLA should be sorted now

per-gron:
Closing because I think this proposal maybe adds a bit too much conceptual overhead.

PR #35, opened 2017-09-08
Clarify Transparent Retries

ncteisen:
Clarification of transparent retries stemming from an offline discussion.

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -276,6 +276,8 @@ gRPC implementations must expose information about this second case in the RPC m
 
 Since retry throttling is designed to prevent server application overload, and these transparent retries do not make it to the server application layer, they do not count as failures when deciding whether to throttle retry attempts.
 
+Similarly, transparent retries do not count toward the limit of configured retry attempts (`maxRetryAttempts`).
+
 ![State Diagram](A6_graphics/transparent.png)
 
 [Link to SVG file](A6_graphics/transparent.svg)
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -268,9 +268,11 @@ RPC failures can occur in three distinct ways:
 
 The last case is handled by the configurable retry policy that is the main focus of this document. The first two cases are retried automatically by the gRPC client library, **regardless** of the retry configuration set by the service owner. We are able to do this because these request have not made it to the server application logic, and thus are always safe to retry.
 
-In the first case, in which the RPC never leaves the client, the client library will immediately retry the call once. If this immediate retry fails, then the failure will be treated just like the third bullet point. This means that it will be handled by the configured retry policy.
+In the first case, in which the RPC never leaves the client, the client library can transparently retry until a success occurs, or the RPC's deadline passes.
 
-Similarly, if the RPC reaches the server, but has never been seen by the server application logic (the second bullet), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy.
+If the RPC reaches the gRPC server library, but has never been seen by the server application logic (the second bullet), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy. This extra caution is needed because this cases involves extra load on the wire.
+
+gRPC implementations must expose information about this second case in the RPC metadata, allowing service owners to see how many RPCs are falling into this category.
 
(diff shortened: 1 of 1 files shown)


ejona86:
So honestly, this section looked fine to me. The major concerns I've raised on wait-for-ready don't apply to this pre-existing section. Saying that the wait-for-ready subsumes this section is actually a spec change and should called out as such.

ncteisen:
My view was that this section was describing an _implementation_ of the clarified/changed WFR behavior. We can discuss this today as we close the loop on WFR

ejona86:
How could it be describing wait-for-ready if it did not mention "wait for ready" or provide any conditionals that would imply wait-for-ready?

I'm calling out the _silent_ change to the spec. We should make it clear when something changes from what was previously stated and not simply claim it moved or is being clarified. There is no way WFR RPCs could subsume this spec without a spec change because this spec doesn't limit itself to WFR.

A6-client-retries.md
```diff
@@ -268,9 +268,11 @@ RPC failures can occur in three distinct ways:
...
 
-In the first case, in which the RPC never leaves the client, the client library will immediately retry the call once. If this immediate retry fails, then the failure will be treated just like the third bullet point. This means that it will be handled by the configured retry policy.
+In the first case, in which the RPC never leaves the client, the client library can transparently retry until a success occurs, or the RPC's deadline passes.
 
-Similarly, if the RPC reaches the server, but has never been seen by the server application logic (the second bullet), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy.
+If the RPC reaches the gRPC server library, but has never been seen by the server application logic (the second bullet), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy. This extra caution is needed because this cases involves extra load on the wire.
```
ejona86: You say "second bullet", but it was an ordered list (with numbers), not an unordered list (with bullets).

A6-client-retries.md
```diff
@@ -268,9 +268,11 @@ RPC failures can occur in three distinct ways:
...
+
+gRPC implementations must expose information about this second case in the RPC metadata, allowing service owners to see how many RPCs are falling into this category.
```
ejona86: How/in what format?

A6-client-retries.md
ncteisen: Done

A6-client-retries.md
ncteisen: Done

A6-client-retries.md
```diff
@@ -268,9 +268,11 @@ RPC failures can occur in three distinct ways:
...
+In the first case, in which the RPC never leaves the client, the client library can transparently retry until a success occurs, or the RPC's deadline passes.
 
-Similarly, if the RPC reaches the server, but has never been seen by the server application logic (the second bullet), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy.
+If the RPC reaches the gRPC server library, but has never been seen by the server application logic (the second case), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy. This extra caution is needed because this cases involves extra load on the wire.
+
+gRPC implementations must expose information about this second case in the RPC metadata, allowing service owners to see how many RPCs are falling into this category. The new header names for exposing this metadata will be `"grpc-transparent-retry-attempts"`. The value for this field will be an integer.
```
ejona86: s/new header names/header name/

ncteisen:
@a11r, @markdroth, @dfawley. Is this clarification in line with your view of our meetings resolution?

A6-client-retries.md
```diff
@@ -268,9 +268,11 @@ RPC failures can occur in three distinct ways:
 
 The last case is handled by the configurable retry policy that is the main focus of this document. The first two cases are retried automatically by the gRPC client library, **regardless** of the retry configuration set by the service owner. We are able to do this because these request have not made it to the server application logic, and thus are always safe to retry.
 
-In the first case, in which the RPC never leaves the client, the client library will immediately retry the call once. If this immediate retry fails, then the failure will be treated just like the third bullet point. This means that it will be handled by the configured retry policy.
+In the first case, in which the RPC never leaves the client, the client library can transparently retry until a success occurs, or the RPC's deadline passes.
```
markdroth: Might be worth noting that these retries do not count toward the limit of configured retries, if any.

ncteisen: Done

PR #37, opened 2017-09-15
Clarify maxRetryAttempts

ncteisen:
As discussed in email thread

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -83,7 +83,7 @@ Retry policies support configuring the maximum number of retries, the parameters
 
 ```
 "retryPolicy": {  
-  "maxRetryAttempts": 3,
+  "maxAttempts": 4,
   "initialBackoffMs": 100,
   "maxBackoffMs": 1000,
   "backoffMultiplier": 2,
@@ -96,9 +96,9 @@ Each of these configuration options is detailed in its own section below.
 
 ##### Maximum Number of Retries
 
-`maxRetryAttempts` specifies the maximum number of retry attempts per original RPC.
+`maxAttempts` specifies the maximum number of RPC attempts, including the original request.
 
-gRPC's call deadlines apply to the entire chain of original RPC and retry attempts. The time limit begins when the original RPC is issued to the gRPC client library. For example, if the specified deadline is 500ms and after 500ms only one retry has been attempted, the call is cancelled even if `maxRetryAttempts` is set to two or greater.
+gRPC's call deadline applies across all attempts for a given RPC. For example, if the specified deadline for an RPC is July 23 9:00:00pm PDT the operation will fail after that time regardless of how many attempts were configured or attempted.
 
 ![State Diagram](A6_graphics/too_many_attempts.png)
 
@@ -124,7 +124,7 @@ Hedged requests are configured with the following parameters:
 
 ```
 "hedgingPolicy": {  
-  "maxRequests": 3,
+  "maxAttempts": 4,
   "hedgingDelayMs": 500,
   "nonFatalStatusCodes": [  
     "UNAVAILABLE",
@@ -134,7 +134,7 @@ Hedged requests are configured with the following parameters:
 }
 ```
 
-When a method has chosen a `hedgingPolicy`, the first RPC is sent immediately, as with a standard non-hedged call. After `hedgingDelayMs` has elapsed without a successful response, the second RPC will be issued. If neither RPC has received a response after `hedgingDelayMs` has elapsed again, a third RPC is sent, and so on, up to `maxRequests`. In the above configuration, after 1ms there would be one outstanding hedged RPC, after 501ms there would be two outstanding hedged RPCs, and after 1001ms there would be three outstanding hedged RPCs. As with retries, gRPC call deadlines apply to the entire chain of hedged requests. The time limit starts when the original RPC is issued, and if the deadline is reached all outstanding hedged requests are aborted.
+When a method has chosen a `hedgingPolicy`, the original RPC is sent immediately, as with a standard non-hedged call. After `hedgingDelayMs` has elapsed without a successful response, the second RPC will be issued. If neither RPC has received a response after `hedgingDelayMs` has elapsed again, a third RPC is sent, and so on, up to `maxAttempts`. In the above configuration, after 1ms there would be one outstanding RPC (the original), after 501ms there would be two outstanding RPCs (the original and the first hedged RPC), after 1001ms there would be three outstanding RPCs, and after 1501ms there would be four. As with retries, gRPC call deadlines apply to the entire chain of hedged requests. Once a deadline has passed, the operation fails regardless of in-flight RPCS, and regardless of the hedging configuration.
 
 The implementation will ensure that the listener returned to the client application forwards its calls (such as `onNext` or `onClose`) to all outstanding hedged RPCs.
 
@@ -178,7 +178,8 @@ Neither retry attempts or hedged RPCs block when `token_count` is less than or e
 The only RPCs that are counted as failures for the throttling policy are RPCs that fail with a status code that qualifies as a [retryable](#retryable-status-codes) or [non-fatal status code](#hedging-policy), or that receive a pushback response indicating not to retry. This avoids conflating server failure with responses to malformed requests (such as the `INVALID_ARGUMENT` status code).
 
 #### Pushback
-Servers may explicitly pushback by setting metadata in their response to the client. The pushback can either tell the client to retry after a given delay or to not retry at all. If the client has already exhausted its `maxRetryAttempts`, the call will not be retried even if the server says to retry after a given delay.
+
+Servers may explicitly pushback by setting metadata in their response to the client. The pushback can either tell the client to retry after a given delay or to not retry at all. If the client has already exhausted its `maxAttempts`, the call will not be retried even if the server says to retry after a given delay.
 
 Pushback may also be received to a hedged request. If the pushback says not to retry, no further hedged requests will be sent. If the pushback says to retry after a given delay, the next hedged request (if any) will be issued after the given delay has elapsed.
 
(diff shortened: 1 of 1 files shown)


A6-client-retries.md
```diff
@@ -134,7 +136,7 @@ Hedged requests are configured with the following parameters:
...
+When a method has chosen a `hedgingPolicy`, the original RPC is sent immediately, as with a standard non-hedged call. After `hedgingDelayMs` has elapsed without a successful response, the first hedged RPC will be issued. If neither RPC has received a response after `hedgingDelayMs` has elapsed again, a second hedged RPC is sent, and so on, up to `maxHedgedRequests`. In the above configuration, after 1ms there would be one outstanding hedged RPC, after 501ms there would be two outstanding hedged RPCs, and after 1001ms there would be three outstanding hedged RPCs. As with retries, gRPC call deadlines apply to the entire chain of hedged requests. The time limit starts when the original RPC is issued, and if the deadline is reached all outstanding hedged requests are aborted.
```
dfawley: > The time limit starts when the original RPC is issued, and if the deadline is reached all outstanding hedged requests are aborted.

Maybe just a Go thing, but our timer starts when the context is created that specifies a deadline.  Even if a timeout is specified, Go converts that into a point-in-time deadline immediately (`WithTimeout(d)` shorthand for `WithDeadline(time.Now().Add(d))`).

If this is actually similar to how the other languages handle timeouts, then you may want to omit or more accurately specify when "the time limit starts".

A6-client-retries.md
```diff
@@ -339,11 +343,11 @@ This is an example of a retry policy and its associated configuration. It implem
 
 #### Hedging Policy
 
-The following example of a hedging policy configuration will issue up to three hedged requests for each RPC, spaced out at 500ms intervals, until either: one of the requests receives a valid response, all fail, or the overall call deadline is reached. Analogously to `retryableStatusCodes` for the retry policy, `nonFatalStatusCodes` determines how hedging behaves when a non-OK response is received.
+The following example of a hedging policy configuration will issue an original RPC, then up to three hedged requests for each RPC, spaced out at 500ms intervals, until either: one of the requests receives a valid response, all fail, or the overall call deadline is reached. Analogously to `retryableStatusCodes` for the retry policy, `nonFatalStatusCodes` determines how hedging behaves when a non-OK response is received.
```
dfawley: > Analogously to `retryableStatusCodes` for the retry policy, `nonFatalStatusCodes` determines how hedging behaves when a non-OK response is received.

Off-topic: maybe we should discuss renaming `retryableStatusCodes` to `nonFatalStatusCodes` for even more consistency between hedging/retries?

A6-client-retries.md
```diff
@@ -420,16 +424,16 @@ The retry policy is transmitted to the client through the service config mechani
       }
 
       "hedgingPolicy": {
-        // The hedging policy will send up to maxRequests copies of the hedged
+        // The hedging policy will send up to maxHedgedRequests copies of the hedged
         // RPC. This number is inclusive of all requests, including the original
```
dfawley: Update, please: not inclusive of the original request.

A6-client-retries.md
```diff
@@ -420,16 +424,16 @@ The retry policy is transmitted to the client through the service config mechani
...
         // RPC.
         //
         // This field is required and must be greater than 1.
```
dfawley: May this be zero now?

A6-client-retries.md
```diff
@@ -420,16 +424,16 @@ The retry policy is transmitted to the client through the service config mechani
...
-        "maxRequests": number,
+        "maxHedgedRequests": number,
 
-        // The first RPC will be sent immediately, but the maxRequests-1 subsequent
+        // The first RPC will be sent immediately, but the maxHedgedRequests-1 subsequent
```
dfawley: Remove "-1"

A6-client-retries.md
```diff
@@ -420,16 +424,16 @@ The retry policy is transmitted to the client through the service config mechani
...
         // hedged RPCs will be sent at intervals of every hedgingDelayMs. Set this
-        // to 0 to immediately send all maxRequests RPCs.
+        // to 0 to immediately send all maxHedgedRequests RPCs.
```
dfawley: maxHedgedRequests+1?

A6-client-retries.md
```diff
@@ -420,16 +424,16 @@ The retry policy is transmitted to the client through the service config mechani
       }
 
       "hedgingPolicy": {
-        // The hedging policy will send up to maxRequests copies of the hedged
+        // The hedging policy will send up to maxHedgedRequests copies of the hedged
```
dfawley: maxHedgedRequests+1?  (Is the original RPC considered a copy?)

A6-client-retries.md
ericgribkoff: I think we should keep these in the terms used by the public APIs, so I think this wording is fine as-is

A6-client-retries.md
```diff
@@ -98,7 +98,9 @@ Each of these configuration options is detailed in its own section below.
 
 `maxRetryAttempts` specifies the maximum number of retry attempts per original RPC.
 
-gRPC's call deadlines apply to the entire chain of original RPC and retry attempts. The time limit begins when the original RPC is issued to the gRPC client library. For example, if the specified deadline is 500ms and after 500ms only one retry has been attempted, the call is cancelled even if `maxRetryAttempts` is set to two or greater.
+gRPC's call deadlines apply to the entire chain of original RPC and retry attempts. The time limit begins when the original RPC is issued to the gRPC client library. For example, if the specified deadline is 500ms and after 500ms only one retry has been attempted, the call is canceled even if `maxRetryAttempts` is set to two or greater.
```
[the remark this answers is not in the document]
ericgribkoff: In the Java API, deadlines are specified as durations from the time the RPC starts.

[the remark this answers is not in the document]
ericgribkoff: Oops, I was wrong about this - the deadline can be set per-call to roughly match the semantics we talk about here ("the time limit starts when the original RPC is issued") but can also come from the context or elsewhere, so our statement about when the time limit starts is wrong.

A6-client-retries.md
ericgribkoff: As above - I was wrong here, gRPC allows setting deadlines in ways that do not start when the original RPC is issued, so Doug's right, we should remove this text.

A6-client-retries.md
ncteisen: Let's save this for a different thread if we want to make the change

ncteisen:
I addressed all comments. If we decide to move forwards with the modification to hedging semantics, I will squash commits and give a good desc before merging.

Personally, I can logic it both ways: 
* maxHedgedRequests=4 means 4 overall RPCs, because they are all hedged requests
* or maxHedgedRequests=4 means 5 overall because original+4 hedged.

It all comes down to whether we consider that original to be any different that the hedged RPCs.

@dfawley, @ericgribkoff, who feels more strongly one way or the other?

dfawley:
Semantically, the initial request isn't a hedge.  It might be "hedged by"
the subsequent attempts, but it is not in itself a hedge.  Also, I like
that there ends up being a parallel between retries and hedges, and you can
reason about the number of RPCs sent the same way.

Alternatively, we can make the settings for both hedging and retries named
"maxAttempts" (or "maxRPCs" or "maxRequests" or something similar) to avoid
any confusion.  This would be my #1 choice.

ericgribkoff:
> Alternatively, we can make the settings for both hedging and retries named
"maxAttempts" (or "maxRPCs" or "maxRequests" or something similar) to avoid
any confusion.  This would be my #1 choice.

This sounds good to me too. I think this means that we leave hedging as-is (it was already maxRequests, inclusive of the original attempt) and change (maxRetryAttempts, exclusive of the original RPC) => (maxRequests, inclusive of the original RPC).

Thus we also sidestep the unanswerable question of, given a policy with maxAttempts=2 and hedgingDelayMs=0, which RPC hedges the other :)

markdroth:
FWIW, I agree with Doug: My first choice would be to use `maxAttempts` for both retries and hedging, where the value will indicate the total number of attempts.  This would mean that we should not allow a config where `maxAttempts` is set to a value lower than 2, but I think that's less confusing than the semantic where you always have to mentally add 1 to what you see in the config to determine the total number of attempts that will occur.

ncteisen:
Updated the doc to use maxAttempts for both retry policy and hedging policy.

I will squash commits before merging

A6-client-retries.md
```diff
@@ -96,11 +96,9 @@ Each of these configuration options is detailed in its own section below.
 
 ##### Maximum Number of Retries
 
-`maxRetryAttempts` specifies the maximum number of retry attempts per original RPC.
+`maxAttempts` specifies the maximum number of RPC attempts, including the original request.
 
-gRPC's call deadlines apply to the entire chain of original RPC and retry attempts. For example, if the specified deadline for an RPC is July 23 9:00:00pm PDT the operation will fail after that time regardless of how many retries were configured or attempted.
-
-Note that if `maxRetryAttempts` is set to 3, a total of 4 RPCs could be made against the server (if the first three fail): 1 original RPC and 3 retry attempts.
+gRPC's call deadlines apply to the entire chain of original RPC and retry attempts. For example, if the specified deadline for an RPC is July 23 9:00:00pm PDT the operation will fail after that time regardless of how many attempts were configured or attempted.
```
markdroth: Suggest alternate wording for the first sentence:

gRPC's call deadline applies across all attempts for a given RPC.

A6-client-retries.md
```diff
@@ -182,15 +178,15 @@ Neither retry attempts or hedged RPCs block when `token_count` is less than or e
...
 
 #### Limits on Retries and Hedges
 
-Both `maxRetryAttempts` in `retryPolicy` and `maxHedgedRequests` in `hedgingPolicy` have, by default, a client-side maximum value of 4. This client-side maximum value can be changed by the client through the use of channel arguments. Service owners may specify a higher value for these parameters, but higher values will be treated as equal to the maximum value by the client implementation. This mitigates security concerns related to the service config being transferred to the client via DNS.
+`maxAttempts` in both `retryPolicy` and `hedgingPolicy` have, by default, a client-side maximum value of 4. This client-side maximum value can be changed by the client through the use of channel arguments. Service owners may specify a higher value for these parameters, but higher values will be treated as equal to the maximum value by the client implementation. This mitigates security concerns related to the service config being transferred to the client via DNS.
```
markdroth: The maximum value should now be 5, not 4.

A6-client-retries.md
```diff
@@ -402,10 +398,10 @@ The retry policy is transmitted to the client through the service config mechani
       // RPCs will not be retried or hedged.
 
       "retryPolicy": {
-        // The maximum number of retry attempts, exclusive of the original RPC.
+        // The maximum number of RPC attempts, including the original RPC.
         //
-        // This field is required and must be greater than zero.
-        "maxRetryAttempts": number,
+        // This field is required and must be one or greater.
```
markdroth: It should actually be 2 or greater.  Given the new semantic, if it's set to 1, it means that no retrying or hedging should be done, so there's no point in supporting that value.

A6-client-retries.md
```diff
@@ -424,16 +420,16 @@ The retry policy is transmitted to the client through the service config mechani
...
-        // The hedging policy will send up to maxHedgedRequests copies of the
-        // original RPC. This number represents the hedged RPCs, it does not
-        // count the original
+        // The hedging policy will send up to maxAttempts RPCs.
+        // This number represents the all RPC attempts, including the
+        // original and all the hedged RPCs.
         //
-        // This field is required and must be 0 or greater.
-        "maxHedgedRequests": number,
+        // This field is required and must be 1 or greater.
```
markdroth: Same here -- it should be 2 or greater.

ncteisen:
Done

PR #40, opened 2017-10-12
A14 Channelz

carl-mastrangelo:


carl-mastrangelo:
@markdroth   I made updated changes based on our discussion, PTAL.

A14-channelz.md
```diff
@@ -29,19 +29,23 @@ and connections that process currently has.  In a process, there are multiple
...
-    channel is a directed acyclic graph (DAG).  Only leaf channels may have
-    sockets.  Interior channels may only have subchannels.
-2.  A "server" represents the entry point for RPCs.  A server may have one or
-    more listening sockets, and has a collection of "services".  Unlike clients,
-    servers are not hierarchical.  A server may only have sockets.
-3.  An "endpoint" is either a channel or a server.  The local and remote
+    channel may have channels and subchannels.  A channel is a directed 
+    acyclic graph (DAG).  Only leaf channels may have
+    sockets.  Interior channels may only have channels and subchannels.
+2.  A "subchannel" represents an abstraction that is load balanced over by an
```
markdroth: Suggest saying something like this, maybe right after the first sentence here: "A subchannel will only send RPCs that are delegated to it by a parent channel or subchannel."

A14-channelz.md
```diff
@@ -71,23 +75,34 @@ The data representation of each ref:
...
+Channels and Subchannels are hierarchically organized into a DAG structure.  A
+channel may have any number of channels and subchannels.  Each subchannel can
+also have channels and subchannels.  Each channel and subchannel may also have
+any number of sockets.   However, a given channel or subchannel cannot have
+heterogeneous children.  That is, a channel or subchannel may have
+channels and subchannels, or  have sockets, but not both.
+
+A subchannel represents a channel that is load balanced over.  When a channel
+has both subchannels and channels, the channels are not delegated to by the 
+superchannel.  Channels and Subchannels will be used interchangeably unless
```
markdroth: To avoid confusion and make the spec clearer, I think it would be better to remove this last sentence.  Instead, we should look at every place in this document where we use the term "channel" and decide whether it applies to both channels and subchannels; if both, we should change it to say "channel or subchannel".

A14-channelz.md
```diff
@@ -349,6 +364,7 @@ Channelz exposes four methods to access channelz data:
 * GetTopChannels returns the root channels for channel trees
 * GetServers returns the servers in the system
 * GetChannel returns info about a specific channel
+* GetSubchannel returns info about a specific subchannel
```
markdroth: Weren't we going to change these to `GetChannels` and `GetSubchannels`, plural, so that clients could get more than one of them at once?

A14-channelz.md
```diff
@@ -630,18 +646,48 @@ import "google/protobuf/any.proto";
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
 
-// Channel is a logical grouping of subchannels and sockets.
+// Channel is a logical grouping of channels, subchannels and sockets.
```
markdroth: https://www.verbicidemagazine.com/wp-content/uploads/2011/09/Oxford-Comma.jpg

A14-channelz.md
```diff
@@ -630,18 +646,48 @@ import "google/protobuf/any.proto";
...
 message Channel {
   // The identifier for this channel.
   ChannelRef ref = 1;
   // Data specific to this channel.
   ChannelData data = 2;
-  ChannelTrace trace = 3;
-  // At most one of 'subchannel_ref' and 'socket' is set.
+  // At most one of 'channel_ref+subchannel_ref' and 'socket' is set.
```
markdroth: If you want to encode this restriction in the proto format, we could do something like this:

```
message ChannelAndSubchannelRefs {
  repeated ChannelRef channel_ref;
  repeated SubchannelRef subchannel_ref;
}
oneof {
  ChannelAndSubchannelRefs channel_and_subchannel_refs;
  repeated SocketRef socket_ref;
}
```

A14-channelz.md
```diff
@@ -684,12 +732,21 @@ message ChannelRef {
   reserved 3, 4, 5, 6;
 }
 
+message SubchannelRef {
+  // The globally unique id for this subchannel.  Must be a positive number.
+  int64 subchannel_id = 7;
+  // An optional name associated with the subchannel.
+  string name = 8;
+  // Intentionally don't use field numbers from other refs.
```
markdroth: Why does this matter?

A14-channelz.md
```diff
@@ -698,32 +755,30 @@ message ServerRef {
...
 }
 
 message Server {
   ServerRef ref = 1;
   ServerData data = 2;
-  ServerChannelTrace trace = 3;
 
-  // The list of all connected sockets refs, possibly including those in
-  // listen_socket.  There are no ordering guarantees.
-  repeated SocketRef socket = 4;
```
markdroth: Why did this go away?

A14-channelz.md
carl-mastrangelo: Done.  There are a LOT of places, so I tried to unify them back together as "descendent channels".  I think I got each place.

A14-channelz.md
carl-mastrangelo: It doesn't but I like having the refs be distinguishable by the tag numbers.

A14-channelz.md
carl-mastrangelo: Done.

A14-channelz.md
carl-mastrangelo: I thought about it, but it complicates the API.  It's no longer possible to use NOT_FOUND as a return code.

A14-channelz.md
carl-mastrangelo: Comma comma down doobie do down down.

A14-channelz.md
carl-mastrangelo: I thought about this too, but it didn't seem significantly better.  It adds another level of indirection to getting at the data you want.  The amount of restriction here is already not enforceable the proto file, so I didn't want to partially do it here and partially in the comments.  

For example, can duplicate refs exist in the list?  What about refs with the same id, but different names?   Can they exist in both channel refs and sub channel refs?  Can channel 2 be a subchannel of channel 1, and an owned channel of channel 3?  


Lastly, there is another option that someone might reasonably ask for:

```
message ChannelAndSubchannelRef {
  oneof ref {
    ChannelRef channel_ref;
    SubchannelRef subchannel_ref;
  }
}
message ChannelAndSubchannelRefs {
  repeated ChannelAndSubchannelRef ref;
}
oneof {
  ChannelAndSubchannelRefs channel_and_subchannel_refs;
  repeated SocketRef socket_ref;
}
```

None of these seem like big enough problems to modify it.

A14-channelz.md
carl-mastrangelo: @a11r raised concerns about socket pagination.   It was moved into ListServerSockets

carl-mastrangelo:
@markdroth @a11r friendly ping

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+----
+Adds channel level debug information that can be remotely queried.
+
+Background
+----
+
+gRPC needs a way to expose internal details about connections.  This
+information can be used to debug or diagnose a live program which may be
+exhibiting undesirable network behavior.  In Stubby, this need is fulfilled by
+the "rpcz" page, which exposes aggregated RPC stats as well as listing all
```
markdroth: "sub-pages of the /rpcz page"

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+
+Along with the id, a human readable name can optionally be associated with each
+the id.  The id and the name together form a "reference".  Thus, Channels,
+Servers and Sockets are identified by their respective references.  These
+references are abbreviated ChannelRefs, SubchannelRefs, ServerRefs, and 
+SocketRefs.  Note that only the id is necessary to query channelz.
+
+The data representation of each ref:
+
+![message ref structure][message ref structure]
```
markdroth: The content of this diagram is just text.  Given that, why put it in a diagram instead of inlining it directly into this document?

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+
+Each channel has a ChannelRef reference which includes the channel id and an
+optional name.  The name is included for human consumption.  There are no
+restrictions on the name but it should be limited to a reasonable length.
+
+Each subchannel has a SubchannelRef reference which includes the subchannel id
+and an optional name.  The name is included for human consumption.  There are 
+no restrictions on the name but it should be limited to a reasonable length.
+
+![channel hierarchy 1][channel hierarchy 1]
```
markdroth: Please include subchannels in this diagram.  In practice, most descendent channels will be subchannels; there will usually be no more than one descendent channel that is a channel.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+
+1.  A "channel" represents an abstraction that can start and complete RPCs.  A
+    channel may have channels and subchannels.  A channel is a directed 
+    acyclic graph (DAG).  Only leaf channels may have
+    sockets.  Interior channels may only have channels and subchannels.
+2.  A "subchannel" represents an abstraction that is load balanced over by an 
+    owning channel. A subchannel may have channels and subchannels.  A 
+    subchannel is a directed acyclic graph (DAG).  Only leaf subchannels may 
+    have sockets.  Interior subchannels may only have channels and subchannels.
+3.  A descendent channel is either a channel or subchannel that is logically
```
markdroth: Maybe put "descendent channel" in quotes, for consistency with the other definitions here?

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+SocketRefs.  Note that only the id is necessary to query channelz.
+
+The data representation of each ref:
+
+![message ref structure][message ref structure]
+
+
+## Channels and Subchannels
+
+Channels and Subchannels, or descendent channel, are hierarchically organized
```
markdroth: If the term "descendent channel" means either a channel or subchannel, then it's not clear to me why we need to mention it here, since we've already mentioned channels and subchannels.

If we do mention it here, it should probably be pluralized for consistency with the subjects at the start of the sentence.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+no restrictions on the name but it should be limited to a reasonable length.
+
+![channel hierarchy 1][channel hierarchy 1]
+
+### Channel Data
+
+Channels and Subchannels include data about themselves, notably their initial
+parameters and their call stats.  Currently the data includes:
+
+* Channel state ( See: connectivity-semantics-and-api.md )
```
markdroth: I assume this should be a link to https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+* Channel trace
+* Channel target, if applicable.
+* Number of calls starts, succeeded, and failed.  Note this is slightly
```
markdroth: s/starts/started/

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+  different than the number of streams.  See the Socket Data section.
+* Time of the last call started.
+
+In general, each piece of data included is specific to the channel itself and
+NOT of its descendent channels.  This is to say that an ancestor channel
+is not an aggregator for descendent channels.  The target of a channel or
+subchannel may be different from that of its parent, or not present at all.  The
```
markdroth: Are there cases where the target of a subchannel will be different from that of its parent?  I think this is true for descendent channels (like an LB channel), but I don't think it is for subchannels.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+
+The number of calls started, succeeded, and failed on a channel gives insight
+into the activity of the channel.  Subtracting the number of failed and
+succeeded calls from the number started indicates how many are in flight.  The
+total number of started calls indicates how active the channel has been over
+time.  The ratio between succeeded calls and failed calls gives a rough
+estimation of how healthy the channel has been.  The last call started timestamp
+indicates how close the channel is towards entering the idle state.
+
+#### Channel Trace
```
markdroth: @ncteisen, please review this section.

Are we intending that the channel trace design will be a separate doc?

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+Channels and Subchannels also have a concept of a channel trace.  This includes
+interesting events that have happened on the channel recently.  Example events
+include entering different channel states, name resolution, load balancing
+changes, unreachability, etc.  As mentioned earlier, this tracing info is
+specific to the channel, and not an aggregation of descendent channel traces.
+
+The channel traces may include ids pointing to channels and subchannels outside
+of the current DAG.  Since channel traces use the same ids as the channels 
+themselves, this allows a user to query for more information about a channel 
+given its trace.  For example, the trace may mention that name resolution
```
markdroth: I don't think this is a good example.  In general, I think traces for name resolution will only be present in the channel that does the name resolution.

I think the only cases where a trace would contain an entry that points to a node other than the current node is when a parent channel references an owned channel (like the grpclb channel) or a subchannel.  And in those cases, it would generally point to another node in the same tree.  The only edge case I can think of would be an entry that points to a subchannel that has subsequently been removed from the tree (e.g., because that address has been removed by a resolver update).

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+
+### Subchannels
+
+To keep the size of the Channel message down, only the ChannelRefs and 
+SubchannelRefs will be included in the channel or subchannel.  These refs can
+be used to walk the hierarchy.  Subchannels are not included in the top level
+channels, and thus do not need to be paginated.  A user may query for all the
+subchannels in parallel. As mentioned above, if there are any Sockets on the 
+channel, there will be no descendent channels.   This invariant exists to 
+simplify the tree structure, and more closely match the known implementations of
```
markdroth: Grammar nit: Remove the comma here.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+gRPC.
+
+The channel DAG represents a "has-a" rather than a "is-a" relationship.  In the
+common case, there will be many subchannels.  These subchannels will be used to
+balance RPCs for a given target.
+
+Alternatively, a given channel's load balancer may itself make RPCs to gather
```
markdroth: Why is this mentioned in the "subchannels" section, given that the LB channel is not a subchannel?

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+data about which subchannels to prefer.  Thus, the load balancing subchannels
+are owned by a parent, but do not share its target.
+
+![channel hierarchy 2][channel hierarchy 2]
```
markdroth: This diagram shows the LB channel as a subchannel, but it's actually a channel.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+
+Lastly, subchannels may not be shared across different channel trees.  Each
```
markdroth: I believe we agreed to remove this restriction.  In C-core, subchannels are shared between channels.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+channel has at most one parent.  If a channel tree is used between multiple
+other channels, it must reside in its own tree.  For example, a name resolver
```
markdroth: What is a "name resolver channel"?

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+
+Like channels and subchannels, only SocketRefs are included on a channel.  This 
+is done for two main reasons.  The size of the socket message may be large, and
+there may be many sockets for a given channel.  Secondly, it can be CPU 
+intensive to fill in the socket information, even though the information might 
+not be desired.  As mentioned in the section above, if there are any descendent 
+channels, there will be no sockets.  This invariant exists to simplify the
+DAG structure, and more closely match the known implementations of gRPC.
+
+![channel hierarchy 3][channel hierarchy 3]
```
markdroth: Might be a good idea to add subchannels to this diagram too.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+Note that even with no Servers present, channelz data is still collected.  If in
+the future a server is started, it can expose the channelz service which
+includes prior channel information.
+
+### Server Data
+
+Server data includes data about the server, similar to Channel Data.  As of this
+writing server data is a subset of the channel data:
+
+* Number of calls starts, succeeded, and failed.  Note this is slightly
```
markdroth: s/starts/started/

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+  message was received.
+* The amount of flow control window granted to the remote endpoint, and the
+  amount of flow control window granted by remote endpoint.
+* All known socket options set.
+
+The stream counts differ slightly than that of the number of calls.  Calls may
+be started before a socket (transport) is available, so the stream counter may
+smaller than the corresponding call counters.  Conversely, if a stream fails and
+the call is retriable, a new stream may be started.  This could make the stream
+counts higher than the call counts.  Streams are considered successful if they
```
markdroth: This implies that even if the grpc call fails, the stream will be considered successful if trailing metadata containing the grpc status is sent successfully?

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+have the HTTP/2 EoS bit set, or failed if terminated otherwise (usually an
+HTTP/2 RST_STREAM frame).
+
+The message counts represent how many gRPC messages are sent and received.
```
markdroth: It seems a little inconsistent for the socket to track streams instead of calls (which implies that it's working at the HTTP/2 layer, not the gRPC layer) but then track a gRPC-layer entity like messages.  Maybe the socket should track both streams and calls?  Or maybe it shouldn't track messages at all?

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+Calls may send and receive any number of messages, including zero.
+
+The number of keepalives sent represents how often the local endpoint wants to
+keep the socket alive.  Generally this is an HTTP/2 PING frame.  However, not
+all PING frames are keep alives.  Specifically, pings that were sent as a reply
+(ACK bit set) or were used for reasons unrelated to keep alives should not count
+towards this.  For example, if pings were used to measure network bandwidth,
+they would not be treated as keep alives.   The idea is to measure how often the
```
markdroth: s/idea/intent/

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+is also useful to know.  The last timestamps are included for messages in both
+directions.
+
+Flow control is a very tricky subject and is often the source of hard to explain
+slowdown.  The amount of window granted by each endpoint to each other is a
+constantly evolving number, which each endpoint sending chunks of window.  For
+the particular flow control windows in the socket data, they are the windows as
+defined by the underlying transport.  For example, HTTP/2 transports report the
+connection level flow control window.  Implementations that don't support flow
+control may set this to the max value, implying that there is no limitation.
```
markdroth: It might make more sense for implementations that don't support flow control to just leave this field unset.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+Note that only the connection level flow control window is reported, rather than
+the stream level.  TCP level flow and congestion control counters are exposed
+via socket options.
+
+#### Socket Options
+
+At this low level in the gRPC stack, the implementation difference begin vary
```
markdroth: This sentence is grammatically incorrect and very difficult to parse.  I suggest something like this:

At this low level in the gRPC stack, implementation differences become more significant.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+on.
+
+This means that varying amounts of socket data may be returned by the channelz
+service.  Implementations may avoid querying the options if it would be
+impractical, though they are encouraged to do so.
+
+### Addresses
+
+An important part of debugging a connection is finding "who" the socket is
+connected to.  Toanswer this question each socket exposes a remote and a local
```
markdroth: s/Toanswer/To answer/

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+address field.  The address implies what kind of socket it is (e.g. TCP, UDP,
+UDS, etc.) as well as the the identity used to establish the connection (e.g.
+IP Address, port number, filename, etc.).
+
+Not every socket has a remote address;  listening sockets may only include a
+local address.  Not every socket has a local address; UDS sockets may only
```
markdroth: s/UDS sockets/Unix domain sockets/

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+structure is extensible.  By default, only a TLS message is defined.
+
+# Querying Channelz
+
+Channelz exposes four methods to access channelz data:
+
+* GetTopChannels returns the root channels for channel trees
+* GetServers returns the servers in the system
+* GetChannel returns info about a specific channel
+* GetSubchannel returns info about a specific subchannel
```
markdroth: I think that we need some way to paginate data about subchannels.  For example, in the common case of a client with a single top-level channel and a large number of subchannels, paginating the subchannels is actually more important than paginating the top-level channels.

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+While a human user could retry the query, perhaps to see if new channels have
+been created, an automated user would know to stop here.
+
+The channelz service may return any number of entities, though the results
+should be capped to a reasonable number.  If the process is under heavy load,
+it may decide to return fewer entities.  Since there is a flag in the response
+indicating that the user is at the end, channelz may even return an empty list.
+All ids are valid start points, including non positive ids.
+
+## ListRootChannels
```
markdroth: GetTopChannels

A14-channelz.md
```diff
@@ -0,0 +1,1051 @@
...
+
+This method returns all the root channels in the process.  Users are expected
+to make this query first when investigating channels.  As mentioned in the
+previous section, channelz will respond with a list of zero or more top level
+channels, as well as a boolean flag indicating if there may be more results.
+Users who wish to get the next results should use the largest channel id from
+the previous results in the subsequent query.
+
+## ListServers
```
markdroth: GetServers

carl-mastrangelo: Done.

A14-channelz.md
carl-mastrangelo: A channel that resolves names, likely using gRPC as an underlying transport.  The NR is just an example, it can be anything (such as load balancing, or security, or etc.)

A14-channelz.md
carl-mastrangelo: Renamed the section

A14-channelz.md
carl-mastrangelo: Not so clear.  The channel that triggers name resolution may want to indicate that here.  

This is really only a stub section, and will need to be clarified when the Channel Trace design is written.   This section is informative, but not normative.

A14-channelz.md
carl-mastrangelo: Probably, especially if two channels were created to different targets, but ended up with the same addresses.  The subchannels could be shared, but the original target would be relevant.  

In any case, some one may want to do it, even if implementations today don't.

A14-channelz.md
ncteisen: This section looks good to me as a stub. The design for channel tracing won't be that long. Once I do it, I can update this doc in a subsequent pull request. No need to block this for that.

carl-mastrangelo:
@markdroth ping

A14-channelz.md
```diff
@@ -0,0 +1,1046 @@
...
+  // There are no ordering guarantees on the order of subchannel refs.
+  // There may not be cycles in the ref graph.
+  // A sub channel ref may be present in more than one channel or subchannel.
+  repeated SubchannelRef subchannel_ref = 4;
+  
+  // There are no ordering guarantees on the order of sockets.
+  repeated SocketRef socket = 5;
+}
+
+// Channel is a logical grouping of channels, subchannels, and sockets.
```
lyuxuan: Subchannel description here?

A14-channelz.md
```diff
@@ -0,0 +1,1046 @@
...
+  uint32 tcpi_reordering = 29;
+}
+
+service Channelz {
+  // Gets all root channels (e.g. channels the application has directly
+  // created). This does not include subchannels nor non-top level channels.
+  rpc GetTopChannels(GetTopChannelsRequest) returns (GetTopChannelsResponse);
+  // Gets all servers that exist in the process.
+  rpc GetServers(GetServersRequest) returns (GetServersResponse);
+  // Gets all server sockets servers that exist in the process.
```
lyuxuan: delete servers? And I think the server sockets here means the non-listener sockets, right? Also, it would be great if you can explain briefly why implement pagination for server sockets is a good idea somewhere in the doc.

A14-channelz.md
markdroth: I think it's unlikely that we would ever have an architecture where a subchannel would be shared across channels with different targets, because the authority would be different.  And even if we did, we would probably want the subchannel to not show any target rather than showing one of the two targets randomly.  So I think we can support the ability for future implementations to do this by simply saying that the target of a subchannel may not be present.

A14-channelz.md
markdroth: I cannot think of a case where we would actually want to do this.  I think this example is misleading. 
 Please either remove this example or change it to something more realistic.

A14-channelz.md
markdroth: Makes sense.  Might be a good idea to explicitly say that.

A14-channelz.md
markdroth: I'm not sure why bytes sent is hard to track.  But even if they are, I think it would make more sense for the implementation to provide an estimate of bytes sent instead of breaking the abstraction between layers this way.

I am really not comfortable with mixing the two layers this way.  If you do not think there's an alternative, then we should get @a11r to weigh in on this.

A14-channelz.md
markdroth: We can encode the difference between "unset" and "0" using the google.protobuf sub-messages.

A14-channelz.md
markdroth: Why do you think it's unlikely?  Do you doubt the existence of clients that talk to large numbers of backends?  I'd be very surprised if we have data showing that this is not a common case.

ncteisen:
I am going to revive #7 soon. What is the status of this one?

markdroth:
Carl, there are still several open comment threads here.  Can we please get these resolved soon?  Thanks!

ncteisen:
Ping. I can update this with the Channel Tracing design whenever it gets merged

A14-channelz.md
carl-mastrangelo: Bytes is hard because it means different things to different people.  For each of the following questions, there are reasonable people who would disagree about if they count:

* Http Proxy overhead
* Http/2 ping frames
* Http/2 window frames
* Http/2 padding
* Http2 Hpack / continuation frames
* Per message compression and / or whole stream compression
* SSL Framing overhead (and does this include overhead added by hardware crypto?)
* TCP framing overhead
* TCP keep alives
* TCP sidechannel
* TCP retransmissions and duplicate acks
* Quic overhead

Secondly, not all transports report all of these numbers, like Cronet, Inprocess, and possibly gRPC over gRPC.

Thirdly, we don't always know when something is "sent".   Does it get incremented at the beginning of a 1gig message or at the beginning, or update it as it sends.  

Your point about abstraction layer is valid; I don't think there is an easy way to update the upper layers of gRPC from the lower level transport.   I picked the stream because at this point it is certain which transport is servicing the message.   

Like I said, this is just a proxy for something that is easy to calculate.  It doesn't solve all the problems mentioned above, but it is a useful metric.   We can talk about adding more things based on user feedback in 1.1.

A14-channelz.md
carl-mastrangelo: Good point, done here and in the proto.

A14-channelz.md
carl-mastrangelo: more than 100 subchannels for a single channel.   I do doubt that high subchannel usages exist.    We should optimize for the common case, not for the rare one.  This means making users who have excess subchannels pay a slightly higher cost when they query the channelz service.

Also IIRC, the default number of addresses returned by the internal load balancer was 20.

A14-channelz.md
carl-mastrangelo: Done.

A14-channelz.md
carl-mastrangelo: Wow nice catch!  Done.

PR #41, opened 2017-10-17
A16: Binary logging

y-zeng:


ejona86:
The file should end in `.md`.

y-zeng:
cc @markdroth

hsaliak:
@markdroth @y-zeng could we please initiate a discussion for this in the grpc-io? So that we can get community feedback?

A15-binary-logging
```diff
@@ -0,0 +1,264 @@
...
+representation.
+
+The proto of GrpcLogEntry:
+
+```
+// Log entry we store in binary logs
+message GrpcLogEntry {
+  // Enumerates the type of logs
+  enum Type {
+    SEND_INITIAL_METADATA = 0;
```
dfawley: This enum should ideally have an `UNKNOWN=0` to make it future-proof.  Or the enum could be removed in lieu of the fact that we're using a `oneof` below.

A15-binary-logging
```diff
@@ -0,0 +1,264 @@
...
+  // This flag is currently used to indicate whether the payload is compressed,
+  // it may contain other semantics in the future. Value of 1 indicates that the
+  // binary octet sequence of Message is compressed using the mechanism declared
+  // by the Message-Encoding header. A value of 0 indicates that no encoding of
+  // Message bytes has occurred.
+  uint8 flags = 1;
+  // Length of the message. It may not be the same as the length of the
+  // data field, as the logging payload can be truncated or omitted.
+  uint32 length = 1;
+  // Binary payload, post compression (depends on the flags field), may be
```
dfawley: post-compression and pre-decompression?

In grpc-go, applications don't have access to the compressed payload, so this could be a problem for us.

A15-binary-logging
```diff
@@ -0,0 +1,264 @@
...
+// A metadata key value pair
+message MetadataEntry {
+  string key = 1;
+  bytes value = 2;
+}
+
+// Peer information
+message Peer {
+  enum PeerType {
+    PEER_IPV4 = 0;  // peer is struct sockaddr_in
```
dfawley: This should also have a `PEER_UNKNOWN = 0`.

A15-binary-logging
```diff
@@ -0,0 +1,264 @@
...
+  fixed64 low = 2;
+};
+```
+
+A call id should be enclosed in each log record, so that the analyzer can use it
+to group records generated by the same call. The generation of call id can be
+separated from the log generation code, and provided as an interface. In the
+open source version, we can generate and assign call id according to the channel
+and stream information. At the beginning of each RPC, the gRPC client generates
+a call id and sends it to the server in the client initial metadata. This call
```
dfawley: What metadata field will be used for this?

Or should this be removed?  (Have plans changed?)

zpencer:
Updates to the grfc:

- Do not mandate the use of an env variable for GRPC_BINARY_LOG_FILTER
  across languages
- Specifically call out that the log events reflect what is visible to
  an application
- Removed the section about generating a call ID on client side and
  sending it to the server. binary logs should only observe, and
  should not alter the RPC.
- Removed section about logging in c-core, since there are currently
  discussions about moving the implementation to C++ and python
  interceptors.
- pluggability: removed some unnecessary language
- updated the proto, comments inline

Further discussions:

Should we require that call id be a monotonically increasing number,
and remove the ID generation interface? The "grpc-trace-bin" header
already contains the census trace info, albeit in a custom (non
protobuf) serialization format. If some headers are special and should
never be truncated, we can expose a configuration knob for that.

Should we remove the payload compression info? In java, we don't have
access to the original compressed data, and full stream decompression
makes it difficult to determine if a message was compressed on the wire.

A15-binary-logging
zpencer: In the current state of things, we *should* log the census span as the call id, and have the option to fall back to an alternate method. I would support simplifying this further and just require that call ids be a monotonically increasing integer. If we do that, we can get rid of `Uint128`.

A15-binary-logging
zpencer: The current proto has an `UNKOWN`

A15-binary-logging
zpencer: The current proto has an `UNKOWN`

A15-binary-logging
zpencer: It's an issue for java too, so perhaps we can just remove this.

dfawley:
@zpencer could you also rename the file to end with `.md` so we can see how it renders, please?  Thanks for taking this on!

zpencer:
Done, it's now a .md file

zpencer:
@menghanl updated doc with today's discussions

zpencer:
Pasting my commit message here so it is more discoverable to reviewers:
- Since we are making a lot of changes to the original proto file, let's use a new package name so we get a fresh start and avoid the field deprecation process (change from grpc.binarylog.v1alpha to grpc.binarylog.v1).
- Remove EVENT_TYPE_END_MARKER. Add comments to EVENT_TYPE_SERVER_TRAILER and EVENT_TYPE_CANCEL stating that these events mark the end of RPC, and subsequent messages logged due to races can be safely ignored.
- Changed call id field to a int64
- Add a google.protobuf.Timestamp field
- Remove flags because we agree we don't want to log compression info. Shift up sibling field IDs.
- Make 'grpc-trace-bin' always be logged, and state that the size DOES count towards the header size limit. i.e. if the header size limit is small, then logging grpc-trace-bin will cause other things to be truncated. In the future when we design the content filtering API, there may be additional user defined headers that MUST ALWAYS be logged. Right now I'm thinking we should conut those towards the limit as well. The limit is really just a hint.
- Rename enum values to have consistent format like "<ENUM_NAME>_<VALUE>". This format helps prevent collisions in C++.
- Remove comment about "intentionally flat"

A15-binary-logging.md
```diff
@@ -0,0 +1,352 @@
...
+  google.protobuf.Duration timeout = 12;
+
+  // The entry sequence id for this call. The first GrpcLogEntry has a
+  // value of 1, to disambiguate from an unset value. The purpose of
+  // this field is to detect missing entries in environments where
+  // durability or ordering is not guaranteed.
+  uint32 sequence_id_within_call = 13;
+
+  // The timestamp of the binary log message
+  google.protobuf.Timestamp timestamp = 14;
```
dfawley: My 2c: timestamp should be the first field.  Actually, a full reordering might be worth it.  Something like this seems most logical to me:

- timestamp, call ID, sequence ID
- logger, event type
- payload (metadata / message), truncated
- header fields: method name, peer, timeout
- trailer fields: status code, message, details

Also, we may want sub-messages for header/trailer fields.

A15-binary-logging.md
```diff
@@ -0,0 +1,352 @@
...
+## Abstract
+
+Provide binary logging functionality in gRPC.
+
+## Background
+
+gRPC binary logging logs RPCs in binary format. RPC logs generated from
+production traffic will be useful in the following cases:
+
+-   Troubleshooting services, find exceptions
```
markdroth: Grammar nit: Please make verb tenses consistent.  Change "find" to "finding".  And on line 21, change "Replay" to "Replaying".

A15-binary-logging.md
```diff
@@ -0,0 +1,352 @@
...
+    // Used by EVENT_TYPE_CLIENT_MESSAGE, EVENT_TYPE_SERVER_MESSAGE
+    Message message = 5;
+  }
+
+  // Peer address information, will only be recorded in
+  // EVENT_TYPE_CLIENT_HEADER or EVENT_TYPE_SERVER_HEADER
+  Peer peer = 6;
+
+  // true if payload does not represent the full message or metadata.
+  bool truncated = 7;
```
markdroth: Suggest renaming this `payload_truncated`, so that it's clear what it refers to without looking at this comment.

A15-binary-logging.md
markdroth: +1 on the re-ordering comment.

I'm ambivalent about sub-messages for header vs. trailer fields.

A15-binary-logging.md
```diff
@@ -0,0 +1,352 @@
...
+
+  // The method name. Logged for EVENT_TYPE_CLIENT_HEADER
+  string method_name = 8;
```
markdroth: Hmm.  Do we also want to log the host?  I am thinking that might be useful for applications doing virtual hosting.

Is this specific to the HTTP/2 transport, or is it something we'd expect to support in other transports?

A15-binary-logging.md
zpencer: Done

A15-binary-logging.md
zpencer: Done

A15-binary-logging.md
zpencer: I reordered the fields and grouped header and trailer into sub messages. It ended up looking pretty reasonable so I support grouping them.

I was considering making `peer` always be logged at SERVER_HEADER because on client side CLIENT_HEADER is before name resolution / load balancing, and on server side we can easily defer the logging until SERVER_HEADER. However, in the case where the RPC gets cancelled before SERVER_HEADER, the server would not be able to log the peer info.

I'm going to leave it as a top level field that appears at the SERVER_HEADER event on client side and CLIENT_HEADER for server side.

A15-binary-logging.md
zpencer: Makes sense, `authority` is called out in the gRPC semantics doc. I'll add a `string authority`.

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+  // On server side, it is possible that the RPC is cancelled before
+  // it sends the header.
+  Peer peer = 11;
+};
+
+message ClientHeader {
+  // This contains only the metadata from the application.
+  Metadata metadata = 1;
+
+  string method_name = 2;
```
dfawley: Please document this and authority.

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+  string authority = 3;
+
+  // the RPC timeout
+  google.protobuf.Duration timeout = 4;
+
+}
+
+message ServerHeader {
+  // This contains only the metadata from the application.
+  Metadata metadata = 3;
```
dfawley: Should this be 1?

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+  uint32 status_code = 1;
+
+  // An original status message before any transport specific
+  // encoding.
+  string status_message = 2;
+
+  // The value of the 'grpc-status-details-bin' metadata key. If
+  // present, this is always an encoded 'google.rpc.Status' message.
+  bytes status_details = 3;
+}
```
dfawley: Should this have a Metadata field?

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+//
+// Implementations must always log grpc-trace-bin if it is present.
+// The pair will count towards the size limit,
+message Metadata {
+  repeated MetadataEntry entry = 1;
+}
+
+// A metadata key value pair
+message MetadataEntry {
+  bytes key = 1;
```
dfawley: Why not a string?

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+  bytes value = 2;
+}
+
+// Peer information
+message Peer {
+  enum PeerType {
```
dfawley: Nit/optional: this could be simply "Type" since it's nested inside Peer.

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+    PEER_TYPE_IPV4 = 1;
+    // address the address in canonical form (RFC5952 section 4)
+    // The scope is NOT included in the peer string.
+    PEER_TYPE_IPV6 = 2;
+    // address is UDS string
+    PEER_TYPE_UNIX = 3;
+  };
+  PeerType peer_type = 1;
+  string address = 3;
+  // only for PEER_IPV4 and PEER_IPV6
```
dfawley: PEER_TYPE_IPV4/6

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+The proposed design is adding binary logging as a built-in functionality in
+gRPC. Users can turn it on/off by setting the `GRPC_BINARY_LOG_FILTER`
+environment variable or a similarly named flag or option string.
+Turning on binary logging should be a configuration setting and require no
+code changes. The control interface should allow users to change the amount
+and contents of logs.
+
+The logged events correspond to what is visible at the application
+level, and as a result they exclude things like [hedging and
+retries](./A6-client-retries.md).
```
dfawley: Maybe we should define what it means to "exclude" retries.  Does this mean that we should only log events from the committed attempt?

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+them to the logging interface (remote logging service on diskless servers, or
+file based logging in the open source version). 
+
+### Pluggability
+
+gRPC binary logging should serve as a separate library, making it convenient to
+be changed or even replaced. Configuration knobs should be exposed directly via
+this binary logging library. The production target can remove this library
+according to its need to minimize the target size.  Other parts of gRPC code
+hould not depend on this library, so that it could be removed without any
```
dfawley: hould->should

A15-binary-logging.md
zpencer: Fixed

A15-binary-logging.md
zpencer: Fixed, this slipped in while copy pasting the sub messages around

A15-binary-logging.md
zpencer: Fixed, this slipped in while copy pasting the sub messages around

A15-binary-logging.md
```diff
@@ -0,0 +1,375 @@
...
+  // Each call may have several log entries, they will all have the same call_id.
+  // Nothing is guaranteed about their value other than they are unique across
+  // different RPCs in the same gRPC process.
+  int64 call_id = 2;
+
+  // The entry sequence id for this call. The first GrpcLogEntry has a
+  // value of 1, to disambiguate from an unset value. The purpose of
+  // this field is to detect missing entries in environments where
+  // durability or ordering is not guaranteed.
+  uint32 sequence_id_within_call = 3;
```
menghanl: Can we make this also 64 bit so the same ID generator can be reused?

A15-binary-logging.md
zpencer: String should be fine, since https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md states keys are strings only.

A15-binary-logging.md
zpencer: Done

A15-binary-logging.md
zpencer: done

A15-binary-logging.md
zpencer: done

A15-binary-logging.md
zpencer: Yes. I will add this statement to the doc.

A15-binary-logging.md
zpencer: int64 makes sense. Done.

dfawley: It's a bit of a nit, but should both of these be uint64 instead? I don't think we want to use negative numbers for either, and it's best to make things unexpressable if you don't want them.

zpencer: Agreed, a negative call_id value is probably just asking for trouble. I'll make it a uint64.

zpencer:
@dfawley @markdroth are there any remaining issues with this PR?

A16-binary-logging.md
```diff
@@ -0,0 +1,386 @@
...
+    ServerHeader server_header = 7;
+    // Used by EVENT_TYPE_CLIENT_MESSAGE, EVENT_TYPE_SERVER_MESSAGE
+    Message message = 8;
+    Trailer trailer = 9;
+  }
+
+  // true if payload does not represent the full message or metadata.
+  bool payload_truncated = 10;
+
+  // Peer address information, will only be recorded in
```
dfawley: Should we put Peer into Client/ServerHeader since they are only set with those event types?

Then the comments *might* be simpler to explain: `ClientHeader.Peer: // only set by server`.  (and vice-versa)

Should we put it in the trailer, too, in case the RPC is trailers-only?

zpencer: Having the peer in one of three places might make it harder for users to understand. The trailers-only case in particular may be something people overlook when filtering the records.

I tried to rephrase the comment. Does this make it easier to understand?

dfawley: Sounds good thanks.

A16-binary-logging.md
```diff
@@ -65,132 +74,160 @@ message GrpcLogEntry {
...
+  uint32 length = 1;
   // May be truncated or omitted.
-  bytes data = 3;
+  bytes data = 2;
 }
 
-// A list of metadata pairs, used in the payload of CLIENT_HEADER,
-// SERVER_HEADER and SERVER_TRAILER.
+// A list of metadata pairs, used in the payload of client header,
+// server header and trailer.
```
markdroth: https://i.kym-cdn.com/photos/images/original/000/946/427/5a4.jpg :)

A16-binary-logging.md
```diff
@@ -60,9 +63,9 @@ message GrpcLogEntry {
   // definition, but the same meaning is expressed here.
   enum EventType {
     EVENT_TYPE_UNKNOWN = 0;
-    // Headers sent from client to server
+    // Header sent from client to server
     EVENT_TYPE_CLIENT_HEADER = 1;
```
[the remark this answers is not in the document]
dfawley: Note that the header includes things that are not "metadata" (e.g. the method name).  Does that influence your opinion on this?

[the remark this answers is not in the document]
markdroth: That doesn't really bother me.  Those other fields are associated with these ops.  The names of the ops are IMHO the important thing here, and I think it's important for us to be consistent.

A16-binary-logging.md
```diff
@@ -0,0 +1,383 @@
+# gRPC Binary Logging
+
+*   Author(s): Yuchen Zeng, Spencer Fang
+*   Approver: a11r
+*   Status: In Review
+*   Implemented in:
+*   Last updated: 07/18/2018
+*   Discussion at:
```
carl-mastrangelo: Where is the discussion thread?

dfawley:
Why is this not A16 anymore?  The newer, colliding gRFC was changed to A18 -- this should go back to being A16 in that case.

A19-binary-logging.md
```diff
@@ -0,0 +1,388 @@
...
+    TYPE_UNKNOWN = 0;
+    // address is in 1.2.3.4 form
+    TYPE_IPV4 = 1;
+    // address is in IPv6 canonical form (RFC5952 section 4)
+    // The scope is NOT included in the peer string.
+    TYPE_IPV6 = 2;
+    // address is UDS string
+    TYPE_UNIX = 3;
+  };
+  PeerType peer_type = 1;
```
menghanl: `Type type = 1;`

A19-binary-logging.md
```diff
@@ -0,0 +1,388 @@
...
+  string address = 3;
```
menghanl: Should those be `2` and `3` instead?

A19-binary-logging.md
```diff
@@ -0,0 +1,388 @@
...
+}
+
+// A metadata key value pair
+message MetadataEntry {
+  string key = 1;
+  bytes value = 2;
+}
+
+// Peer information
+message Peer {
```
menghanl: Should this message type be called `Address` instead?

A16-binary-logging.md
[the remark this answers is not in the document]
ejona86: I don't think it is appropriate to call this initial_metadata, as it would _contain_ initial_metadata. (Calling what it contains just "metadata" is fine because there is context)

A16-binary-logging.md
```diff
@@ -1,11 +1,11 @@
...
 
-*   Author(s): Yuchen Zeng, Spencer Fang
-*   Approver: a11r
+*   Author(s): Spencer Fang, Yuchen Zeng
+*   Approver: markdroth
 *   Status: In Review
 *   Implemented in:
-*   Last updated: 07/18/2018
-*   Discussion at:
+*   Last updated: 08/28/2018
```
markdroth: Nit: Please use ISO format for dates (YYYY-MM-DD).

A16-binary-logging.md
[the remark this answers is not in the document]
markdroth: But the event is the client sending initial metadata.  I thought we had agreed to use that term in the call semantics doc?  If so, why would we use something different here?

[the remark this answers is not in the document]
zpencer: The [semantics PR as it is currently written](https://github.com/grpc/grpc/pull/15460/files#diff-449981901227883dfbe29af4b096d0daR238) and [wire protocol doc](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) use the terms "request headers", "response headers", and "trailers". Aren't these the ops we are trying to be consistent with?

A16-binary-logging.md
```diff
@@ -4,7 +4,7 @@
 *   Approver: markdroth
 *   Status: In Review
 *   Implemented in:
-*   Last updated: 08/28/2018
+*   Last updated: 2018/09/17
```
markdroth: ISO format uses dashes instead of slashes.

A16-binary-logging.md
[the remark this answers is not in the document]
markdroth: @ejona86 and @dfawley and I have discussed this and decided to try the terms "gRPC header" and "gRPC trailer" (both singular).  So it looks like this PR is using the right terms.

zpencer:
Thank you to all the reviewers for your feedback!

PR #43, opened 2017-10-26
Updates to Retry Spec

ncteisen:
Changes durations from millisecond number values, to well defined strings of the form "1.23s".

cc @carl-mastrangelo 

Also addresses two smaller clarifications of grpc-retry-attemps and grpc-retry-pushback-ms

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -84,8 +84,8 @@ Retry policies support configuring the maximum number of retries, the parameters
 ```
 "retryPolicy": {  
   "maxAttempts": 4,
-  "initialBackoffMs": 100,
-  "maxBackoffMs": 1000,
+  "initialBackoff": ".1s",
+  "maxBackoff": "1s",
   "backoffMultiplier": 2,
   "retryableStatusCodes": [  
     "UNAVAILABLE"
@@ -106,9 +106,9 @@ gRPC's call deadline applies across all attempts for a given RPC. For example, i
 
 ##### Exponential Backoff
 
-The `initialBackoffMs`, `maxBackoffMs`, and `backoffMultiplier` parameters determine the randomized delay before retry attempts.
+The `initialBackoff`, `maxBackoff`, and `backoffMultiplier` parameters determine the randomized delay before retry attempts.
 
-The initial retry attempt will occur at `random(0, initialBackoffMs)`. In general, the `n`-th attempt will occur at `random(0, min(initialBackoffMs*backoffMultiplier**(n-1), maxBackoffMs))`.
+The initial retry attempt will occur at `random(0, initialBackoff)`. In general, the `n`-th attempt will occur at `random(0, min(initialBackoff*backoffMultiplier**(n-1), maxBackoff))`.
 
 ##### Retryable Status Codes
 
@@ -125,7 +125,7 @@ Hedged requests are configured with the following parameters:
 ```
 "hedgingPolicy": {  
   "maxAttempts": 4,
-  "hedgingDelayMs": 500,
+  "hedgingDelay": ".5s",
   "nonFatalStatusCodes": [  
     "UNAVAILABLE",
     "INTERNAL",
@@ -134,7 +134,7 @@ Hedged requests are configured with the following parameters:
 }
 ```
 
-When a method has chosen a `hedgingPolicy`, the original RPC is sent immediately, as with a standard non-hedged call. After `hedgingDelayMs` has elapsed without a successful response, the second RPC will be issued. If neither RPC has received a response after `hedgingDelayMs` has elapsed again, a third RPC is sent, and so on, up to `maxAttempts`. In the above configuration, after 1ms there would be one outstanding RPC (the original), after 501ms there would be two outstanding RPCs (the original and the first hedged RPC), after 1001ms there would be three outstanding RPCs, and after 1501ms there would be four. As with retries, gRPC call deadlines apply to the entire chain of hedged requests. Once a deadline has passed, the operation fails regardless of in-flight RPCS, and regardless of the hedging configuration.
+When a method has chosen a `hedgingPolicy`, the original RPC is sent immediately, as with a standard non-hedged call. After `hedgingDelay` has elapsed without a successful response, the second RPC will be issued. If neither RPC has received a response after `hedgingDelay` has elapsed again, a third RPC is sent, and so on, up to `maxAttempts`. In the above configuration, after 1ms there would be one outstanding RPC (the original), after 501ms there would be two outstanding RPCs (the original and the first hedged RPC), after 1001ms there would be three outstanding RPCs, and after 1501ms there would be four. As with retries, gRPC call deadlines apply to the entire chain of hedged requests. Once a deadline has passed, the operation fails regardless of in-flight RPCS, and regardless of the hedging configuration.
 
 The implementation will ensure that the listener returned to the client application forwards its calls (such as `onNext` or `onClose`) to all outstanding hedged RPCs.
 
@@ -183,7 +183,7 @@ Servers may explicitly pushback by setting metadata in their response to the cli
 
 Pushback may also be received to a hedged request. If the pushback says not to retry, no further hedged requests will be sent. If the pushback says to retry after a given delay, the next hedged request (if any) will be issued after the given delay has elapsed.
 
-A new metadata key, `"grpc-retry-pushback-ms"`, will be added to support server pushback. The value is to be an integer with no unnecessary leading zeros that represents how many milliseconds to wait before sending a retry. If the value for pushback is set to -1, then it will be seen as the server asking the client not to retry at all.
+A new metadata key, `"grpc-retry-pushback-ms"`, will be added to support server pushback. The value is to be an ASCII encoded signed 32-bit integer with no unnecessary leading zeros that represents how many milliseconds to wait before sending a retry. If the value for pushback is negative or unparseble, then it will be seen as the server asking the client not to retry at all.
 
 #### Limits on Retries and Hedges
 
@@ -198,7 +198,7 @@ There are five possible types of server responses. The list below enumerates the
 
 2. Retryable/Non-Fatal Status Code
     1. Retry policy: Retry according to policy
-    2. Hedging policy: Immediately send next scheduled hedged request, if any. Subsequent hedged requests will resume at `hedgingDelayMs`
+    2. Hedging policy: Immediately send next scheduled hedged request, if any. Subsequent hedged requests will resume at `hedgingDelay`
 
 3. Fatal Status Code
     1. Retry policy: Don't retry, return failure to client application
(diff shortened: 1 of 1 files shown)


A6-client-retries.md
```diff
@@ -183,7 +183,7 @@ Servers may explicitly pushback by setting metadata in their response to the cli
 
 Pushback may also be received to a hedged request. If the pushback says not to retry, no further hedged requests will be sent. If the pushback says to retry after a given delay, the next hedged request (if any) will be issued after the given delay has elapsed.
 
-A new metadata key, `"grpc-retry-pushback-ms"`, will be added to support server pushback. The value is to be an integer with no unnecessary leading zeros that represents how many milliseconds to wait before sending a retry. If the value for pushback is set to -1, then it will be seen as the server asking the client not to retry at all.
+A new metadata key, `"grpc-retry-pushback-ms"`, will be added to support server pushback. The value is to be a signed 32-bit integer with no unnecessary leading zeros that represents how many milliseconds to wait before sending a retry. If the value for pushback is set to -1, then it will be seen as the server asking the client not to retry at all.
```
markdroth: Let's explicitly state that the value is encoded in ASCII, not binary.

markdroth: Presumably, any negative or unparseable value should be treated as the server asking the client not to retry.

A6-client-retries.md
```diff
@@ -287,7 +287,7 @@ Similarly, transparent retries do not count toward the limit of configured RPC a
 
 Both client and server application logic will have access to data about retries via gRPC metadata. Upon seeing an RPC from the client, the server will know if it was a retry, and moreover, it will know the number of previously made attempts. Likewise, the client will receive the number of retry attempts made when receiving the results of an RPC.
 
-The header name for exposing the metadata will be `"grpc-retry-attempts"` to give clients and servers access to the attempt count. The value for this field will be an integer.
+The header name for exposing the metadata will be `"grpc-retry-attempts"` to give clients and servers access to the attempt count. This value represents the number of preceding retry attempts. Thus, it will not be present on the first RPC, will be 1 for the second RPC, and so on. The value for this field will be an integer.
```
markdroth: Given the new semantic, how about changing the header name to include the word "previous"?

Also, I assume that we want to use this for both retries and hedging, but the current name implies only the former.

How about just calling it "grpc-previous-attempts"?

A6-client-retries.md
```diff
@@ -405,11 +405,11 @@ The retry policy is transmitted to the client through the service config mechani
...
 
         // Exponential backoff parameters. The initial retry attempt will occur at
-        // random(0, initialBackoffMs). In general, the nth attempt will occur at
+        // random(0, initialBackoff). In general, the nth attempt will occur at
         // random(0,
-        //   min(initialBackoffMs*backoffMultiplier**(n-1), maxBackoffMs)).
-        "initialBackoffMs": number,  // Required. Must be greater than zero.
-        "maxBackoffMs": number,  // Required. Must be greater than zero.
+        //   min(initialBackoff*backoffMultiplier**(n-1), maxBackoff)).
+        "initialBackoff": string,  // Required. Long decimal with "s" append
```
markdroth: s/append/&ed/

Same on the next line.

markdroth: Might want to explicitly note that these fields will be the JSON form of the protobuf Duration message as described in https://developers.google.com/protocol-buffers/docs/proto3#json.

ncteisen:
All comments addressed

A6-client-retries.md
```diff
@@ -429,9 +432,12 @@ The retry policy is transmitted to the client through the service config mechani
         "maxAttempts": number,
 
         // The original RPC will be sent immediately, but the maxAttempts-1 
-        // subsequent hedged RPCs will be sent at intervals of every hedgingDelayMs. 
+        // subsequent hedged RPCs will be sent at intervals of every hedgingDelay. 
         // Set this to 0 to immediately send all maxAttempts RPCs.
-        "hedgingDelayMs": number,
+        // hedgingDelay takes its form from:
+        // https://developers.google.com/protocol-buffers/docs/proto3#json
```
dfawley: https://developers.google.com/protocol-buffers/docs/proto3#json doesn't adequately describe the "Duration" format.  It just gives 2 trivial examples.

For example: is "s" the only suffix supported?  Or can one use "us, ns, ms" for sub-second precision?  What about "m", "h", "d", etc?

Is there a formal spec here, and can we link it?  If not, I'm wary of using it.

A6-client-retries.md
```diff
@@ -429,9 +432,12 @@ The retry policy is transmitted to the client through the service config mechani
         "maxAttempts": number,
 
         // The original RPC will be sent immediately, but the maxAttempts-1 
-        // subsequent hedged RPCs will be sent at intervals of every hedgingDelayMs. 
+        // subsequent hedged RPCs will be sent at intervals of every hedgingDelay. 
         // Set this to 0 to immediately send all maxAttempts RPCs.
```
dfawley: Should this say "Set this to '0s' to ..."?  Or is "0" OK?

What is the default value if it is unset?

A6-client-retries.md
markdroth: We're already using this format for the service config `timeout` field, so it's not really a new format that we're not already dealing with.

A6-client-retries.md
markdroth: Agree that this should say `"0s"` instead of `0`.

The default should be `"0s"` if unset.  Let's document that.

A6-client-retries.md
dfawley: Yes, I recently noticed it there as well, and I think this is a real problem.

If there's no spec for this format, I think we should either request one ASAP or define our own spec and use it instead (for this and `timeout`).

markdroth: We can't use something else, because we are depending on the default JSON conversion to make use of proto data internally.

While I agree that the proto documentation is not a rigorous formal spec, I think it's clear enough to understand.  The value will be encoded in seconds, with decimal digits to indicate fractional seconds, followed by a trailing 's'.  If other suffixes were supported, I think it would have said so.  So basically, when parsing the JSON, we strip off the trailing 's' from the string, and we accept any number of decimal digits up to 9.

If you want to ask the proto folks for a formal spec, I'm fine with that, but I'd be surprised if we run into any significant problems with the above.

dfawley: > We can't use something else, because we are depending on the default JSON conversion to make use of proto data internally.

But when we parse it from JSON in order to use it, we have to JSON-parse it as a string and then decode it -- you don't convert it back to a proto, right?  Right now we're using a Go library for that.  https://golang.org/pkg/time/#ParseDuration -- this _seems_ fine, but how can I know for sure if there isn't a spec?

> The value will be encoded in seconds, with decimal digits to indicate fractional seconds, followed by a trailing 's'.

That sounds perfect, let's use it.  Maybe add something about nano precision and the maximum possible value.

Otherwise our users will almost certainly try things like "50ms".  And that will work for Go clients with the parsing we're doing.  But then it might not work in C if it only supports seconds.  The failure modes at that point might be really hard to debug or even notice for the user, so I think we need to be as specific as possible.

markdroth: > > We can't use something else, because we are depending on the default JSON conversion to make use of proto data internally.
> 
> But when we parse it from JSON in order to use it, we have to JSON-parse it as a string and then decode it -- you don't convert it back to a proto, right?

Right, we don't convert it back to a proto.  What I meant was that internally, the service config is stored in proto form, and the google resolver needs to convert it from proto to JSON to feed it into gRPC core.  In order to avoid implementing custom conversion for each individual field in the google resolver (and duplicating it in each language), we want to just depend on the existing code in the protobuf library for converting the entire proto to JSON.  Thus, we need to use the format the protobuf code generates for a Duration field.  And the format for that is documented in https://developers.google.com/protocol-buffers/docs/proto3#json.

Like I said before, if you want to ask the protobuf team to give you something more formal than what's already written in that doc, that's fine with me.  But I personally don't see much ambiguity in what's already there.

> Right now we're using a Go library for that.  https://golang.org/pkg/time/#ParseDuration -- this _seems_ fine, but how can I know for sure if there isn't a spec?

It looks like that Go library will support a lot more formats that what the protobuf library will generate.  I'm not sure why you chose to use that library, but I don't think it's a very good choice, for exactly the reason you mentioned below: it will cause problems if we support certain formats in some languages but not others.  I suggest that you change grpc-go to accept only the format generated by the proto library.

> > The value will be encoded in seconds, with decimal digits to indicate fractional seconds, followed by a trailing 's'.
> 
> That sounds perfect, let's use it.

I think this is basically what's already written in the protobuf docs.  But if this wording makes you happy, I'm fine with it. :)

> Maybe add something about nano precision and the maximum possible value.

The protobuf doc already says the following:

"""
Generated output always contains 0, 3, 6, or 9 fractional digits, depending on required precision. Accepted are any fractional digits (also none) as long as they fit into nano-seconds precision.
"""

> Otherwise our users will almost certainly try things like "50ms".  And that will work for Go clients with the parsing we're doing.  But then it might not work in C if it only supports seconds.  The failure modes at that point might be really hard to debug or even notice for the user, so I think we need to be as specific as possible.

I don't know why you would think that we should support any suffix other than 's', since the protobuf doc doesn't say anything about that.  And I agree with you that it could cause problems if Go supports that and other languages don't.  So I think you should drop support for those suffixes in Go.

dfawley: > It looks like that Go library will support a lot more formats that what the protobuf library will generate. I'm not sure why you chose to use that library, but I don't think it's a very good choice

I agree, knowing what I know about the format now.  But as an example of what happens when things are under-specified, note that the official JSON->proto conversion library in Go uses `time.ParseDuration`, too:
https://github.com/golang/protobuf/blob/1643683e1b54a9e88ad26d98f81400c8c9d9f4f9/jsonpb/jsonpb.go#L760

> I think this is basically what's already written in the protobuf docs. But if this wording makes you happy, I'm fine with it. :)

Yes, this makes me happy.  Let's do it.  I can send a PR for the service config `timeout` field if you'd like.  How about this:

```
// The value is encoded in seconds, with optional decimal digits to indicate
// fractional seconds, followed by a trailing "s".  Leading and trailing zeros
// are permissible, and no zero is required before or after the decimal point,
// if present.  No more than nine digits after the decimal point are allowed.
//
// Examples: "1s", "01s", ".0120s".
```

markdroth: Works for me.

A6-client-retries.md
```diff
@@ -405,11 +405,14 @@ The retry policy is transmitted to the client through the service config mechani
...
+        // random(0, initialBackoff). In general, the nth attempt will occur at
         // random(0,
-        //   min(initialBackoffMs*backoffMultiplier**(n-1), maxBackoffMs)).
-        "initialBackoffMs": number,  // Required. Must be greater than zero.
-        "maxBackoffMs": number,  // Required. Must be greater than zero.
+        //   min(initialBackoff*backoffMultiplier**(n-1), maxBackoff)).
+        // The following two fields take their form from:
+        // https://developers.google.com/protocol-buffers/docs/proto3#json
+        // They are representations of the proto3 Duration type.
+        "initialBackoff": string,  // Required. Long decimal with "s" appended
```
markdroth: Don't we still want to say that these two fields must be greater than 0?

PR #44, opened 2017-11-10
A2 update: Specify "_grpc_config." prefix for DNS name.

markdroth:
Also a couple of other updates:
- Fix quote character used in JSON.
- Update information on C-core implementation.

### DIFF
--- a/A2-service-configs-in-dns.md
+++ b/A2-service-configs-in-dns.md
@@ -4,7 +4,7 @@ Service Config via DNS
 * Approver: a11r
 * Status: Implemented
 * Implemented in: C-core (Go and Java in progress)
-* Last updated: 2017-09-13
+* Last updated: 2017-11-10
 * Discussion at: https://groups.google.com/d/topic/grpc-io/DkweyrWEXxU/discussion
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


A2-service-configs-in-dns.md
```diff
@@ -101,16 +104,17 @@ When encoding the service config in DNS, TXT records are the "obvious"
...
-resulting service config JSON data.
+## Implementation
 
-Note that, due to platform support issues, we will initially *not*
-support the c-ares resolver under Windows or for Node.  Alternatives
-will need to be found for these environments.
+The implementation has been done in C-core as part of the c-ares DNS
+resolver.  We are currently working toward making the c-ares resolver
+the default DNS resolver for C-core.  This requires things like Windows
+and node support and adding address sorting.
```
fengli79: s/node/Node
And why node is special here? Does it use c-core?

markdroth: Done.

We are in the process of working on a native javascript implementation for Node, but the comment here refers to Node support in C-core.  There are some issues related to using c-ares with Node that we still need to work on.

PR #46, opened 2017-11-27
L12: Node.js gRPC 2.0 API changes

murgatroid99:


L12-node-2.0-changes.md
```diff
@@ -48,11 +48,15 @@ We need to increment the major version of the Node.js gRPC library in order to u
  - Update the `checkClientCertificate` argument of `ServerCredentials.createSsl` from a boolean to an enum to reflect the internal API.
  - Change `ChannelCredentials.createSsl` and `ServerCredentials.createSsl` to use objects instead of positional arguments. In particular, they should use options that are compaitible with [`tls.createSecureContext`](https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options) accepts: `key`, `cert`, and `ca`. `ServerCredentials.createSsl` will additionally accept a `checkClientCertificate` option.
 
+### Improve Server APIs
+
+ - Replace `tryShutdown` and `forceShutdown` functions with a single `shutdown` function with an additional `options` arguments that accepts a `graceful` option. With `graceful` set to `true`, the behavior will be identical to the current `tryShutdown` behavior, and with `graceful` set to `false`, the behavior will be identical to the current `forceShutdown` behavior. In both cases, it will accept an optional callback.
+
```
wenbozhu: Just to confirm: 
1) on the server side, grpc doesn't support half-close
2) shutdown(true) does a full-close, i.e. no server-generated message loss 
3) shutdown(false) does a so-called abort, which will close the underlying channel abruptly.

murgatroid99: That looks basically right. `shutdown({graceful:true})` does a C core shutdown, which makes it stop accepting requests, and calls the callback when there are no more pending requests. `shutdown({graceful:false})` does a C core shutdown followed by `cancel_all_calls`, which immediately terminates all pending requests and calls the callback immediately.

hsaliak:
@murgatroid99 if all the comments in the discussion are resolved (looks like they are) could you please merge?

murgatroid99:
I want to hold off on this. Based on some other stuff we're looking at, we may not end up making these changes now, and in that case I would prefer to close it and revisit the idea at a later date.

PR #47, opened 2017-12-07
Clarify buffering of hedged client side streaming

ncteisen:


A6-client-retries.md
```diff
@@ -251,7 +251,7 @@ Hedged RPCs with a non-zero delay will only send out the subsequent hedged reque
 
 After the RPC response has been returned to the client application layer, the RPC is removed from the buffer.
 
-For client-side streaming RPCs, if additional outgoing messages would cause the buffered size to exceed the per-RPC limit, the entire message will be removed from the buffer and thus no longer retryable.
+For client-side streaming RPCs, if additional outgoing messages would cause the buffered size to exceed the per-RPC limit, the entire message will be removed from the buffer and thus no longer retryable. In the case of hedging, this situation would cause all but one hedges RPC to stop making progress. The RPC that is furthest along in terms of outgoing messages will be the once to continue communication.
```
ejona86: s/once/one/.

markdroth: I think the wording here needs to be much more precise.  I suggest something like the following:

"""
When the application sends a message on an RPC that causes the RPC to exceed the buffer limit, the RPC becomes committed, meaning that we choose one attempt to continue and stop all others.

For retriable RPCs, when an RPC becomes committed, the client will continue with the currently in-flight attempt but will not make any subsequent attempts, even if the current attempt fails with a retriable status and there are retry attempts remaining.

For hedged RPCs, when an RPC becomes committed, the client will continue the currently in-flight attempt on which the maximum number of messages have already been sent.  All other currently in-flight attempts will be immediately cancelled, and no subsequent attempts will be started.

Once we have committed to a particular attempt, any messages that have already been sent on the that attempt can be immediately freed from the buffer, and each subsequent message that is replayed can be freed as soon as it is sent.  The new message sent by the application (which caused the RPC to exceed the buffer limit) will never be added to the buffer in the first place; it will stay pending until all previous messages have been replayed, and then it will be sent immediately without buffering.
"""

Also, I think we should delete the paragraph on line 250 above.  Its wording is very misleading, and I don't think it's actually necessary with the text above.

ncteisen: I mainly took your words verbatim, but changed the last paragraph, which I found to be confusing. 

You start with "Once we have committed to a particular attempt", but then talk about replaying messages. With my understanding, committing to an attempt means we will pick the furthest along attempt and continue to make progress. But since we will start no new hedges, we do not need to replay anything. So we can drop all of the old messages in the buffer, send the message that caused us to go over the limit, and then the attempt will continue.

markdroth: I wrote the last paragraph (which applies to both retries and hedging) the way I did because there are in fact cases where we need to replay previously sent messages even after committing.  In fact, these are exactly the cases I outlined in the email thread where we were discussing this.

For retries, consider the following:

1. We have sent 3 messages on the initial attempt, which then fails.
2. We have replayed 1 of the 3 messages on the 2nd attempt.
3. The application now starts a 4th send_message, which takes us over the buffer limit.

In this case, we need to finish replaying the 2nd and 3rd messages on the current attempt before we can send the 4th.

A similar thing can happen with hedging.  For example:

1. Application starts a hedged RPC with 3 parallel attempts, which we'll call A, B, and C.  And let's say that the hedging delay is long enough that we've sent one message on the previous attempt by the time we start the next one.
2. Application sends 3 messages, each of which is completed as soon as we send it on attempt A.  However, by the time we complete the third message on A, we've only sent two on attempt B and only one on attempt C.
3. Attempt A fails with a non-fatal status, so we're going to keep going with the other attempts.
4. The application sends a 4th message.

In this case, we will not be able to send the 4th message until we finish replaying the one remaining message on attempt B.  At that point, we'll send the 4th message on attempt B and return the completion to the application.

ncteisen: Ack. Updated to original phrasing

PR #48, opened 2017-12-12
Proposal L16 explicit C++ library initialization

yang-g:


dfawley:
For consistency, please rename this file to `L16-cpp-...` before merging.  Thanks!

PR #49, opened 2017-12-15
Create L9-Add-MaxThreads-To-C++ThreadManager

sreecha:


L9-Add-MaxThreads-To-C++ThreadManager
```diff
@@ -0,0 +1,83 @@
+Add a new option to the ThreadManager to add the max number of threads
```
ctiller: Why not integrate with ResourceQuota?

sreecha:
- I agree with both the comments by @ctiller and @vjpai and believe using ResourceQuota is a good approach. I also remember the same discussion we had offline the last time when I attempted to make this change :)..

- My concern is that a comprehensive change using ResourceQuotas will take time to implement (considering that its been a few months since we had this discussion the last time) and we currently do not have a way to prevent a thread avalanche on a sync server.  

- max_threads setting may still co-exist with a ResourceQuota based solution (where max_threads can override ResourceQuotas decision.. although admittedly, I can't think of a reason why one could want to use this other than for some artificial test scenarios).

- The other scenarios mentioned by @vjpai (like limiting the number of connections at server, number of connections per ip address etc) are important but fall a bit outside the 'ThreadManager''s scope (they fall in the scope of  'SyncServer''s implementation)

sreecha:
I had offline discussion with @ctiller and @vjpai and am summarizing the concerns:
- Our ServerBuilder API is already too bloated (and this proposal is adding to this problem)
- A ResourceQuota based solution could be very similar to what is being proposed here i.e ThreadManager is a Resource user , every thread creation requests a quota (we can implement it either based as a memory quota or a thread count quota).

I agree with this and will give the ResourceQuota based implementation a try. If that is taking too much time, I can revisit this proposal.

I am closing it for now.

PR #51, opened 2017-12-19
L17: C++ synchronous server should catch exceptions from method handlers

vjpai:
Relates to grpc/grpc#13706 and grpc/grpc#13815.
Please discuss in https://groups.google.com/forum/#!topic/grpc-io/u1XmQPAi3fA

Even though the implementation is quite simple and localized in scope, I feel that it needs a gRFC and discussion because this amounts to a permanent promise that the sync server will catch exceptions from method handlers.

Cc: @yang-g @ctiller @dgquintas

### DIFF
--- a/L17-c++-sync-server-exceptions.md
+++ b/L17-c++-sync-server-exceptions.md
@@ -29,12 +29,25 @@ N/A
 
 The C++ sync server should wrap its invocation of user method handler functions
 in a `try/catch` block.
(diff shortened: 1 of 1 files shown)


vjpai:
I have now added a commit to grpc/grpc#13815 to make sure that we test building without exceptions as part of our portability suite.

vjpai:
I will merge this on January 15th or 16th unless we get more feedback to address.

PR #52, opened 2018-01-13
P3: Core API changes (including additions) need a gRFC

vjpai:
As wrapped languages move to their own repositories and more community language bindings emerge, it is increasingly important to stabilize core surface APIs and treat them as a promise for the future rather than just an implementation mechanism.


hsaliak:
@vjpai thanks for addressing the comment I raised in the discussion. Could you please merge?

PR #54, opened 2018-01-19
L19: Ruby Client Stub close API

apolcyn:


hanmd82:
I encountered some data anomalies while benchmarking bi-directional streaming gRPC using a Ruby client stub and came across this proposal - it turns out that the client stub were not automatically closed and resources released. This proposal and API would be very helpful to improve the Ruby client 👍

dfawley:
For consistency, please rename this file to `L19-ruby-...` before merging.  Thanks!

PR #55, opened 2018-01-19
A15: Provide C-Core API to estimate memory usage of expensive objects like channels

apolcyn:
https://groups.google.com/forum/#!topic/grpc-io/FPoXprcT0d4

### DIFF
--- a/A15-core-memory-usage-api.md
+++ b/A15-core-memory-usage-api.md
@@ -1,0 +1,39 @@
+Core API to estimate the memory usage of channels and servers
+----
+* Author(s): apolcyn
+* Approver: a11r
+* Status: {In-Review}
+* Implemented in: core
+* Last updated: 1/18/2018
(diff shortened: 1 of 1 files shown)


vjpai:
I think core changes should be `L` not `A` since they have no impact on Java/Go/etc.

qingling128:
Could someone help review this proposal to see it makes sense? This has caused some excessive memory usage in Ruby. We have some short-term mitigation in place, but it's still a bit concerning in the long run.

Thanks in advance!

apolcyn:
We do have a preliminary sketch for this in the Google groups chat btw, but it is looking like it will take some more time than initially thought.

qingling128:
apolcyn@ - Cool cool. Thanks for initiating the conversation. That should be fine. We don't have an urgent need for this feature. Just wanna make sure it's still on the road map.

PR #56, opened 2018-01-25
L21: Recategorize GPR headers when possible

vjpai:


L21-gpr-review.md
```diff
@@ -0,0 +1,55 @@
...
+## Proposal
+
+Curate the contents of `include/grpc/support`
+- Keep entries that are used in wrapped language implementations or
+any public API (e.g., `time.h`, `alloc.h`, ...)
+- Move entries to `test/core/util` if they are only used to support
+  tests (e.g. `cmdline.h`, `subprocess.h`)
+- Move entries to `src/core/lib/gpr` if they are used only in the core
+  and C++ API implementations but not used for any public-facing API
+  (e.g., `avl.h`, `tls.h`)
```
nicolasnoble: I'd probably even go one step further and say that avl should in fact live in grpc, not in gpr. The goal of the name "gpr" was meant for "Google Public Runtime", meaning to hide anything that's a portability problem. But avl isn't a portability thing.

The same way we have src/core/lib/json, we could have src/core/lib/avl.

However, it makes me cringe a bit to have tls moved away. I'd be fine with having it being private as long as we keep an option to resurrect it back as being public if one of the wrapped languages suddenly needs it, which may or may not happen in the future.

vjpai: 1. I can definitely make that change for AVL. I'll make that happen in the pull request.
1. So, by that standard, I think that `useful.h` and `host_port.h` should remain gpr since the former is mostly bit-twiddling and the latter is host and port formation.
1. We can go either way on TLS: either privatize it now and make it public later or just leave it public. I'd prefer the former but am ok either way.

I'll ping back when these changes are ready.

vjpai:
Alright, this is merge day!

PR #57, opened 2018-01-26
L22 - gRPC C++ Public Header Directory Change

muxi:


L22-change-grpc++-dir-name.md
```diff
@@ -0,0 +1,43 @@
...
+## Proposal
+
+The proposal is to migrate `include/grpc++` directory to `include/grpcxx` in a backwards compatible manner. The objective is that both styles of inclusion `#include <grpc++/...>` and `#include <grpcxx/...>` can be used for all past and future gRPC C++ users.
+
+### Changes to gRPC C++ API
+
+- Make `include/grpcxx` directory the new location for all gRPC C++ public headers; move all current headers from `include/grpc++` to `include/grpcxx`; update all corresponding inclusions inside gRPC code base;
+- Make wrapper headers in `include/grpc++` for all C++ public headers, preserving the directory structure. This makes current gRPC users build with old directory name.
+
+Wrapper headers in `include/grpc++` may be generated with a script by `generate\_projects.sh` to guarantee that every header in `include/grpcxx` has a corresponding wrapper.
```
nicolasnoble: So, for API backward compatibility, I think we only want to do that once. Any future added header would simply be added into grpcxx, and not into grpc++. Meaning that such maintenance script might become not relevant.

L22-change-grpc++-dir-name.md
```diff
@@ -0,0 +1,43 @@
...
+The 3 restrictions make it impossible to make gRPC C++ library work as Apple framework. We believe this issue can happen somewhere else too in the future (e.g. [grpc#14089](https://github.com/grpc/grpc/pull/14089) is another example where this naming convention creates problem).
+
+## Proposal
+
+The proposal is to migrate `include/grpc++` directory to `include/grpcxx` in a backwards compatible manner. The objective is that both styles of inclusion `#include <grpc++/...>` and `#include <grpcxx/...>` can be used for all past and future gRPC C++ users.
+
+### Changes to gRPC C++ API
+
+- Make `include/grpcxx` directory the new location for all gRPC C++ public headers; move all current headers from `include/grpc++` to `include/grpcxx`; update all corresponding inclusions inside gRPC code base;
+- Make wrapper headers in `include/grpc++` for all C++ public headers, preserving the directory structure. This makes current gRPC users build with old directory name.
```
nicolasnoble: Maybe make sure there is some sort of deprecation warning if using these headers ? Not sure if that's a good idea or not, but, relevant, maybe.

L22-change-grpc++-dir-name.md
muxi: I am worried if there will be users who do not want to switch to `grpcxx` (for some reason) but want to use new features with new headers. But I agree that we want to discourage this and, if necessary, we can manually add wrapper headers.

L22-change-grpc++-dir-name.md
muxi: Yep. I think it is good adding deprecation warning as comment at the top of the wrapper headers.

vjpai: Can I suggest that that should be considered breaking? A lot of people mention using -Werror . Can it be a message that's not a warning? I don't know how to implement that.

muxi: `#pragma message`?

muxi: I am not very sure how portable that is though. We can give a try.

L22-change-grpc++-dir-name.md
```diff
@@ -0,0 +1,42 @@
...
+- Update build systems to expose both headers and wrappers headers to users.
+
+## Rationale
+We think this change is reasonable since, as mentioned in Background section, the old directory name may hurt again in the future. Since this is a rather big API change in C++, wrapper headers are created to make current users build. With this change in place:
+- Users who need to maintain backwards compatibility can keep using the old directory name;
+- New users should use the new directory name;
+- Users who need to use gRPC C++ as Apple framework must use the new directory name.
+
+## Implementation
+Will be implement as part of the [gRPC C++ cocoapods library](https://github.com/grpc/grpc/issues/13582) efforts.
```
nicolasnoble: *implemented :P

PR #58, opened 2018-01-30
L23 - Standalone Node.js gRPC+Protobuf.js Library API

murgatroid99:
Discussion thread: https://groups.google.com/forum/#!topic/grpc-io/t7XFE7Wevow

### DIFF
--- a/L23-node-protobufjs-library.md
+++ b/L23-node-protobufjs-library.md
@@ -2,7 +2,7 @@ Standalone Node.js gRPC+Protobuf.js Library API
 ----
 * Author(s): murgatroid99
 * Approver: wenbozhu
-* Status: Draft
+* Status: Ready for Implementation
 * Implemented in: Node.js
 * Last updated: 2018-01-30
 * Discussion at: https://groups.google.com/forum/#!topic/grpc-io/t7XFE7Wevow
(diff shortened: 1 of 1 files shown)


L23-node-protobufjs-library.md
```diff
@@ -0,0 +1,72 @@
...
+ */
+load(filename, options)
+
+/**
+ * Synchronously load a .proto file and all of its transitive imports
+ * @param {string} filename File path to the .proto file to load.
+ * @param {Object} options Options for loading the file and setting up the deserializers
+ * @return {PackageDefinition} An object containing definitions for all of the loaded services
+ */
+loadSync(filename, options)
```
wenbozhu: Is this necessary to have? Can't we remove this util there is a real use case to add this (which won't be a breaking change).

L23-node-protobufjs-library.md
```diff
@@ -0,0 +1,72 @@
...
+ - `json`: a boolean indicating that additional JSON compatibility conversions should be performed.
+ - `include`: an array of strings listing paths in which to search for imported `.proto` files.
+
+## Rationale
+
+This package provides very similar functionality to the existing `grpc.load` API, but with some differences. First, this API directly exposes all of the relevant Protobuf.js options instead of defining a subset of custom options. This package is explicitly a wrapper around Protobuf.js, so it makes more sense here to directly and consistently expose the relevant options from that library. Second, this API does not provide an equivalent to the `loadObject` function in the existing gRPC library. With the Protobuf.js options directly exposed, there is less benefit in allowing the user to separately load `.proto` files. Third, this adds an asynchronous API for loading `.proto` files. This has been requested before, and more closely matches existing Node APIs. It uses promises instead of callbacks because they are more versatile, and because Node is increasingly supporting them.
```
wenbozhu: Is it the plan that if protobuf.js introduces any breaking changes (e.g. renaming an option), this package will introduce a new release to make such an API change explicit to users?

murgatroid99: Yes, that is my intention.

L23-node-protobufjs-library.md
murgatroid99: You're right. I will remove it.

L23-node-protobufjs-library.md
wenbozhu: Maybe mention this in this section, that future breaking protobuf.js API changes will be contained in this pkg.

PR #60, opened 2018-02-26
L24: C++ Extensible api

makdharma:


vjpai:
Please retitle the actual proposal consistently with other gRFCs, as I have done in the title.

vjpai:
Please also go ahead and start a grpc-io thread.

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,45 @@
+Title
+----
+* Author(s): makdharma
+* Approver: vpai
```
vjpai: vjpai

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,45 @@
...
+* Status: Draft
+* Implemented in: https://github.com/grpc/grpc/pull/14517
+* Last updated: Feb 26, 2018
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+This proposal will enable extending functionality of GRPC C++ API.
```
vjpai: Can this abstract be more specific?

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Background
+
+All GRPC C++ API classes are currently marked "final". The rationale is that
```
vjpai: Please use backquotes rather than doublequotes to indicate language features since that allows them to be typeset to look like code.

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,45 @@
...
+safer to mark classes "final" unless there's a reason to make them extensible.
+
+### Related Proposals:
+
+## Proposal
+
+* Remove "final" keyword from all public API classes.
+* Move current private core methods and member variables to protected.
+* Mark core methods to virtual, so the functionality can be extended.
+* Each such change should go through thorough performance evaluation.
```
vjpai: Please state precisely that "and should not be accepted if there is any observed performance degradation."

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,45 @@
...
+* Do this work in stages. Begin with Server, ServerBuilder, CompletionQueue
+  classes and extend to other classes as needed.
+
+
+## Rationale
+
+The original rationale for keeping methods private and classes non-extensible
+has not held true. The performance impact of removing "final" from all public
+API classes was not measurable. Meanwhile the final classes and non-virtual
```
vjpai: I believe the statement about measurability wrt performance, but do you have any comments about this with regard to code size? Also, can you include a citation in this document indicating a lack of measurable performance? (For example, my PR measuring the impact of de-finalizing all classes would be a relevant citation for final, but maybe you should also extend that to de-virtualize all functions so that we know the behavior in the extreme limit)

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,45 @@
...
+methods preclude any experimentation with implementation. This proposal will
+allow extending and experimenting with different server and client
+implementations.
+
+## Implementation
+
```
vjpai: Please mention your plan here in more detail.

vjpai:
I commented on structural and cosmetic issues with this PR in my review so that you can fix those before targeting the grpc-io thread to actually discuss the technical merit of the idea.

L24-c++-extensible-api.md
makdharma: done

L24-c++-extensible-api.md
makdharma: done

L24-c++-extensible-api.md
makdharma: done

L24-c++-extensible-api.md
makdharma: done

L24-c++-extensible-api.md
makdharma: done

L24-c++-extensible-api.md
makdharma: done

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,61 @@
...
+* Status: Draft
+* Implemented in: https://github.com/grpc/grpc/pull/14517
+* Last updated: Feb 26, 2018
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+This proposal will enable extending functionality of GRPC C++ API by removing
+the ```final``` keyword and making some class methods ```virtual```. It will allow
+alternate implementations of the API by subclassing the API classes.
```
vjpai: by subclassing --> that subclass

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Background
+
+All GRPC C++ API classes are currently marked ```final```. The rationale is that
```
vjpai: All --> Most

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,61 @@
...
+Additionally, adding ```final``` to public API classes later on would break existing
+implementations that use classes derived from public API. Hence it is better and
+safer to mark classes ```final``` unless there's a reason to make them extensible.
+
+### Related Proposals:
+
+## Proposal
+
+* Remove ```final``` keyword from all public API classes.
+* Move current private core methods and member variables to protected.
```
vjpai: Member variables cannot be `protected` (or `public`) according to the C++ style guide. Consider only moving the methods or adding setters/getters or something like that.

vjpai: Also, what do you mean by "core" here and below?

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,61 @@
...
+* Mark core methods to ```virtual```, so the functionality can be extended.
+* Each such change should go through thorough performance evaluation and should not
+  be accepted if there is any observed performance degradation.
+* Do this work in stages. Begin with Server, ServerBuilder, CompletionQueue
```
vjpai: "and" before CompletionQueue

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,61 @@
...
+  classes and extend to other classes as needed.
```
vjpai: comma before and

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+## Rationale
+
+The original rationale for keeping methods private and classes non-extensible
+has not held true. Meanwhile the ```final``` classes and non-virtual methods
```
vjpai: Please provide a citation on that first sentence. You can cite the microbenchmark results that came from definalizing.

You should use single-backquote for all inline code content; triple-backquote is supposed to be for out-of-line code blocks.

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,61 @@
...
+preclude any experimentation with implementation. This proposal will allow
+extending and experimenting with different server and client implementations.
```
vjpai: different server, client, and support structure implementations. (since CQ is also in here)

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,61 @@
...
+
+The performance impact of removing ```final``` from all public
+API classes was not measurable. Impact of adding ```virtual``` was noticable in
+case of Inproc transport microbenchmarks and only for 256K byte messages.
+However this is a extreme case where virtual was added everywhere which is not
+the intention behind this proposal. The more realistic case is seen in
```
vjpai: Commas after However and before  before which

L24-c++-extensible-api.md
makdharma: Done

L24-c++-extensible-api.md
makdharma: Done

L24-c++-extensible-api.md
makdharma: Core is a misnomer, I changed the text. A subset of methods is what I wanted to say.

L24-c++-extensible-api.md
makdharma: Done

L24-c++-extensible-api.md
makdharma: done

L24-c++-extensible-api.md
makdharma: Done

L24-c++-extensible-api.md
makdharma: Done

L24-c++-extensible-api.md
makdharma: Done.

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,65 @@
+L24: C++ Extensible api
+-----------------------
+* Author(s): makdharma
+* Approver: vjpai
+* Status: Draft
+* Implemented in: https://github.com/grpc/grpc/pull/14517
+* Last updated: Feb 26, 2018
+* Discussion at: <google group thread> (filled after thread exists)
```
vjpai: Please fill in this Google group thread.

L24-c++-extensible-api.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Abstract
+
+This proposal will enable extending functionality of GRPC C++ API by removing
+the `final` keyword and making some class methods `virtual`. It will allow
+alternate implementations of the API by that subclass.
```
vjpai: that subclass --> allowing subclassing

PR #63, opened 2018-03-01
Serialization Mechanism is Customizable Per Call

ncteisen:
Should come after #61


L26-serialization-traits-per-call.md
```diff
@@ -0,0 +1,43 @@
...
+* Discussion at: n/a
+
+## Abstract
+
+This document proposes a change to [SerializationTraits](https://github.com/grpc/grpc/blob/master/include/grpcpp/impl/codegen/serialization_traits.h) in order to allow the serialization mechanism to be configurable per call. Most use cases will be unaffected, but this will make it much easier for user to inject custom serialization code into the gRPC pathway.
+
+## Background
+
+Currently, gRPC supports a mechanism for custom serialization. Users must implement the `grpc::SerializationTraits<Message, void>` class for their specific Message type. Forcing users to implement a class that exists in the grpc namespace is a form of internal API leakage. Additionally, this method does not give users the granularity to change the serialization system per-call.
+
```
vjpai: IMO, valuable background would be to give a comparison of how the other languages do this. I know that Go has customizable Codec per channel, not sure about Java.

L26-serialization-traits-per-call.md
```diff
@@ -0,0 +1,43 @@
...
+```C++
+template <class W, class R, 
+          class WCodec = SerializationTraits<W, void>, 
+          class RCoder = SerializationTraits<R, void>>
+class ClientAsyncReaderWriterInterface;
+```
+
+In the default case, nothing will have to change. But this will give greater control to users who would like to supply custom serialization.
+
```
vjpai: Are there implications for codegen?

L26-serialization-traits-per-call.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+We will add templates for serialization classes for both the request and the response. These templates would default to the existing serialization traits objects. This would look like:
+
+```C++
+template <class W, class R, 
+          class WCodec = SerializationTraits<W, void>, 
+          class RCoder = SerializationTraits<R, void>>
+class ClientAsyncReaderWriterInterface;
+```
```
vjpai: Can you give samples for each of the stream objects? It's hard to evaluate this without a full picture (this idea came to me when I was on a plane in a coach-class middle seat, so I'm not really sure that I was thinking through all the cases). For example, in some cases it will be the Read/Write methods that have to get the extra template parameters in addition to the object as a whole. Is this possible for the sync API as well? How would this be done, for example, in the case of a simple blocking unary call?

L26-serialization-traits-per-call.md
dfawley: grpc-go is now customizable per-call.  And we have a registry to support multiple codecs in both clients and servers.  Check out this PR that implements what support we have; you may want to do something similar:

https://github.com/grpc/grpc-go/pull/1813

Usage docs are here: https://github.com/grpc/grpc-go/blob/master/Documentation/encoding.md (this also discusses our compressor support that works the same way).

vjpai: Thanks for the info! I had seen it as a DialOption but good to know that it is now a CallOption. This is valuable background feedback.

@ncteisen : Can you also retitle the gRFC itself to indicate that it's L26 and that it's C++ specific? Likewise with references to gRPC converting to gRPC C++

L26-serialization-traits-per-call.md
ncteisen: I don't think there should be, since the existing Call family objects should just pick up the default, and keep using proto serialization traits

L26-serialization-traits-per-call.md
ncteisen: I added info on Java and Go

PR #64, opened 2018-03-07
Update Channel Tracing Proto

ncteisen:
Removes `status` and `state`, adds `is_error_condition`

### DIFF
--- a/A3-channel-tracing.md
+++ b/A3-channel-tracing.md
@@ -59,10 +59,14 @@ message SubchannelRef {}
 message ChannelTraceEvent {
   // High level description of the event.
   string description = 1;
-  // the severity of the trace event. Options are INFO, WARNING, and ERROR,
-  // which are represented by the values 1, 2, and 3, respectively. Any other
-  // value will be treated as ERROR.
-  int32 trace_level = 2;
+  // The supported severity levels of trace events.
+  enum Severity {
+    INFO = 1;
+    WARNING = 2;
+    ERROR = 3;
+  }
+  // the severity of the trace event
+  Severity severity = 2;
   // When this event occurred.
   google.protobuf.Timestamp timestamp = 3;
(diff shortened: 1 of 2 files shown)


ncteisen:
Question: do we want to make `is_error_condition` and int or enum? And then support something like DEBUG, INFO, ERROR? Or is a boolean expressive enough

A3-channel-tracing.md
```diff
@@ -53,27 +53,25 @@ The data will be accessed via the channelz service, which sends a ChannelTrace p
...
 
 // A trace event is an interesting thing that happened to a channel or
 // subchannel, such as creation, address resolution, subchannel creation, etc.
 message ChannelTraceEvent {
   // High level description of the event.
   string description = 1;
-  // Status/error associated with this event.
-  // Optional, only present is Status != OK. 
-  google.rpc.Status status = 2;
+  // the severity of the trace event. Options are INFO, WARNING, and ERROR,
```
markdroth: With regard to log levels, I agree that DEBUG is not useful, because we're not going to want to waste memory for anything that is more detailed than what we're already recording with INFO.  I also suspect that WARNING is not going to get much use, because in practice most messages are either INFO or ERROR, but I'm not necessarily opposed to including it.  So I think the two main ones we need are INFO and ERROR, which is why I suggested that this be a bool, but I'm not opposed to also supporting WARNING and making this an enum instead.

In terms of the type of this field, we should either decide to support only two types and make this a bool, or we should make it an enum.  I don't think it makes sense for the field to be an int with documentation describing enum semantics; if we really want an enum, let's use one.

ncteisen: I have made severity into an enum. I'd like to support WARNING, I think it would be useful for events that still occur in healthy channels, but need a closer look, because a lot of them might signal an underlying issue.

PR #65, opened 2018-03-07
Update ChannelTraceEvent::Severity enum

ncteisen:
This is needed for the .proto to actually compile.

Realized while working on grpc/grpc#13883


vjpai:
Since this is a bug fix on an existing approved proposal and only adds an enum to the proto, I don't think it needs the 2-week waiting period, but please get consensus on that from @a11r and/or @markdroth .

PR #68, opened 2018-03-09
A14-channelz: rename to socket_ref for consistency in Subchannel

zpencer:
The previous PR updated Channel but failed to update Subchannel.


carl-mastrangelo:
cc: @lyuxuan 

LGTM

PR #69, opened 2018-03-09
Remove grpc-transparent-retry-attempts metadata

ericgribkoff:
The number of transparent retries tracked by this metadata key can be at most 1, so it has minimal value on the server-side. 

cc @dapengzhang0


ncteisen:
Do we still want to have a _boolean_ equivalent being sent in metadata? I rememeber we added this because a server would want to know if a lot or requests were making it to the server grpc lib, but failing there.

PR #72, opened 2018-03-26
A9: Clarify MAX_CONNECTION_AGE_GRACE

ejona86:
There should not be any _required_ change to the code based on this. I use the word "ideally" to say "go probably want to do it this way" but "it's not the worst thing to do something differently." If you are doing something differently though, it may be good to discuss what and why, to make sure edge cases are covered.

CC @MakMukhi, @vjpai, @yashykt

### DIFF
--- a/A9-server-side-conn-mgt.md
+++ b/A9-server-side-conn-mgt.md
@@ -63,14 +63,20 @@ age the server should send GOAWAY with status code NO\_ERROR and ASCII debug dat
 `max_age` and then continue the [graceful connection termination][].
 Specifically, outstanding RPCs should be permitted to complete, and
 implementations are encouraged to initially set the last stream identifier to
-231-1. This option can force the client to periodically reconnect to a L4 LB
-which allows spreading load more quickly. A per-connection random jitter of
-+/-10% will be added to `MAX_CONNECTION_AGE` to spread out connection storms.
+2<sup>31</sup>-1. This option can force the client to periodically reconnect to
+a L4 LB which allows spreading load more quickly. A per-connection random jitter
+of +/-10% will be added to `MAX_CONNECTION_AGE` to spread out connection storms.
 Note that this option without jitter would not create connection storms by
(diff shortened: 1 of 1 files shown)


A9-server-side-conn-mgt.md
```diff
@@ -63,14 +63,20 @@ age the server should send GOAWAY with status code NO\_ERROR and ASCII debug dat
...
+of +/-10% will be added to `MAX_CONNECTION_AGE` to spread out connection storms.
 Note that this option without jitter would not create connection storms by
 itself, but if there happened to be a connection storm it could cause it to
-repeat at a fixed period. `MAX_CONNECTION_AGE_GRACE` is an additive period after
-`MAX_CONNECTION_AGE` after which the connection will be forcibly closed. No
-jitter is applied to the grace period.
+repeat at a fixed period.
+
+`MAX_CONNECTION_AGE_GRACE` is the maximum time the connection will be kept alive
+for outstanding RPCs to complete, ideally relative to the second GOAWAY of the
```
MakMukhi: our grace period is relative to the first goaway; it starts right after scheduling the first goaway.

ejona86: Do you have a limit on how much time exists between the first and second GOAWAY? Or were you relying on the GRACE to provide a limit?

FYI: We didn't notice this earlier because Java wasn't doing graceful shutdown.

MakMukhi: We do have a [limit](https://github.com/grpc/grpc-go/blob/bdb0727fa72e9f94e0b23df2b8c17561956811cd/transport/http2_server.go#L1100) which is the maximum amount of time that may pass after the first goaway and the second goaway. This limit is independent of the grace period.

ejona86: That seems fine. With grace times in the seconds (which is what I'd expect from most users), the difference shouldn't matter. If a user notices, it can be enhanced to what's suggested here.

PR #79, opened 2018-04-26
A10: Don't support grpclb-in-DNS and service-config for "localhost" and ip literals

apolcyn:
When a grpc client that <i>doesn't</i> support grpclb-in-DNS or service configs connects to "localhost", is can resolve "localhost" is negligible time with a quick "hosts" file lookup. But clients that do support these must also query for TXT and SRV records (of `_grpclb._tcp.localhost` and `_grpc_config.localhost`), and these queries must go over the  network and (presumably in the common case) come back with NXDOMAIN. (from what I've observed) it's not uncommon for these "TXT/SRV lookup -> NXDOMAIN" RTT's to be on the order of 10's or 100's of milliseconds, and can potentially add an order of magnitude increase to channel setup latency when resolving "localhost". This gRFC discusses this issue, and the solution described here potentially affects all languages. 

Updated with grpc-io thread for more discusssion: https://groups.google.com/forum/#!topic/grpc-io/Ln6zeUDoo3M

### DIFF
--- a/A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
+++ b/A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
@@ -1,0 +1,81 @@
+Special case localhost and ip literals for service-config and grpclb-in-DNS
+----
+* Author(s): apolcyn
+* Approver: markdroth
+* Status: Draft
+* Implemented in: none currently, suited for all languages
+* Last updated: 2018-10-29
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/Ln6zeUDoo3M
+
+## Abstract
+
+This document discusses a problem and solution to bring down channel
+setup latency for grpc clients that implement
+[grpclb-in-DNS](https://github.com/grpc/proposal/blob/master/A5-grpclb-in-dns.md)
+and
+[service-configs-in-DNS](https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md),
+in a special and degenerate case.
+
+## Background
+
+Clients that support grpclb-in-DNS and service-configs-in-DNS must
+query for TXT and SRV records of `_grpclb._tcp.<target>` and
+`_grpc_config.<target>`, along with the A/AAAA records of
+`<target>`, when creating a channel to `<target>`. In the case
+that `<target>` exists in the machines "hosts" file and
+grpclb-in-DNS/service-configs-in-DNS are not set up for `<target>` (common
+when the target is "localhost" or an ipv4/v6 literal),
+channel setup can take much longer than desired due to the need
+to make these TXT and SRV lookups over the network and then
+wait for NXDOMAIN responses. This is particularly problematic
+for automated tests, which commonly create large numbers of client
+channels having targets of "localhost".
+
+While clients which don't implement grpclb-in-DNS or
+service-configs-in-DNS potentially only need to do a "hosts" file
+lookup when the target is "localhost", clients which do implement
+them add `lookup -> NXDOMAIN` RTT to the setup of every channel.
+
+Another problem which is unrelated to latency but which also caused
+by SRV and TXT record lookups is related to the way in which
+Java treats DNS resolution failures. Internally, gRPC-Java raises a type
+of fatal error when a DNS query fails. Normally, gRPC-Java internally
+catches these types of errors and carries on, but there are certain
+automated test environments in which the Java runtime is adjusted to
+crash upon seeing these types of errors. Normally, A/AAAA queries
+always succeed in these types of test environments because it's
+unexpected for the client to target anything other than "localhost", but
+SRV and TXT record lookups change things so that clients in these
+test environments almost always see failed DNS queries, and so they
+present a problem.
+
+## Proposal
+
+This gRFC proposes that gRPC clients special case "localhost" and
+ipv4/v6 literals (e.g. "1.2.3.4" and "::1"). That is, this gRFC
+proposes that gRPC clients should not query for SRV or TXT
+records when the target is "localhost" or an ipv4/v6 literal,
+and thus that gRPC clients should not implement grpclb-in-DNS or
+service-configs-in-DNS for such targets.
+
+## Rationale
+
+The problem is likely to be noticeable
+only for "localhost" and ipv4/ipv6 literal targets, because grpc clients
+connecting to targets which are not in a machine's "hosts" file and/or parseable as
+an ip literal need to reach out to DNS servers and wait for A/AAAA lookup time anyways
+(unless the grpc client is using a resolver that does client-side caching).
+
+This could perhaps be fixed by implementing DNS caches within grpc
+clients, but that would be more complex.
+
+Also note that while special-casing "localhost" seems wrong in principle
+because its special status is really a convention rather than an
+inherent property of the DNS protocol, the convention here is so strong
+that anyone violating it is likely to have many other problems, and we
+therefore think it's reasonable to have our code treat it specially.
+
+## Implementation
+
+This should be done for all languages which implement
+grpclb-in-DNS and/or service-configs-in-DNS.
--- a/A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
+++ b/A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
@@ -60,14 +60,17 @@ service-configs-in-DNS for such targets.
 
 ## Rationale
 
-The problem is likely to be noticeable
+The latency problem is likely to be noticeable
 only for "localhost" and ipv4/ipv6 literal targets, because grpc clients
 connecting to targets which are not in a machine's "hosts" file and/or parseable as
 an ip literal need to reach out to DNS servers and wait for A/AAAA lookup time anyways
-(unless the grpc client is using a resolver that does client-side caching).
+(unless the grpc client is using a resolver that does client-side caching). This
+problem could perhaps be fixed by implementing DNS caches within grpc clients, but
+that would be more complex.
 
-This could perhaps be fixed by implementing DNS caches within grpc
-clients, but that would be more complex.
+The problem related to Java error handling in certain environments only practically
+affects gRPC-Java clients which target "localhost" or ip literals, and so this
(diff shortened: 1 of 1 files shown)


Falco20019:
In case c-ares is used with the notation dns://localhost/`servicename` this would only affect lookup for `servicename` being `localhost` and not all services just because the authority is `localhost`, correct?

Because we plan to use c-ares with a nameserver managed by our application internally for easy setup at the customer site where we usually don’t be able to setup a real nameserver.

apolcyn:
@Falco20019 

```
In case c-ares is used with the notation dns://localhost/servicename this would only affect lookup for servicename being localhost and not all services just because the authority is localhost, correct?
```

Yes, this proposal is unrelated to the DNS authority/server. This would only affect the name when `servicename` is `localhost`.

carl-mastrangelo:
Perhaps there should be a reason listed why this proposal is being made?

apolcyn:
@carl-mastrangelo  updated the description here

ejona86:
So it seems the root cause is that the results are being retrieved from mixed sources. Ideally if we resolve _any_ address via /etc/hosts we wouldn't search any for the TXT. SRV is a bit more interesting, because there's an mix of priorities. Domains earlier in the `search` should "win" over those later. But SRV "wins" over A/AAAA in our code. If those are considered independent, weird results could be produced. All that said, that edge case seems it _may_ not be a problem in practice.

My response on #15192 is of similar ilk. The wider problem looks pretty messy to solve (especially in Java). Hard-coding a fix for localhost doesn't seem that bad. Although if we do want to try to solve the wider problem, then maybe it should fix this case as well.

apolcyn:
As discussed offline, and to add a caveat, "localhost" and hosts file problems might not exactly overlap, since on some platforms (e..g windows), "localhost" can be faked up outside of the hosts file. The full fix perhaps needs to handle the "target is present in hosts file" issue <i>as well as</i> the "localhost" issue.

ejona86:
And as we discussed offline, newer Linux also doesn't include localhost in the hosts file, instead relying on nss-myhostname from systemd.

carl-mastrangelo:
@markdroth @dfawley Java plans on moving forward with an implementation of this proposal, (in order to continue with auto gRPCLB support, and not break some internal tests).

Did either of you (or @ejona86  ) have any more thoughts on this?  It would be nice if we could get a consistent story on localhost and ip address queries.

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
+Special case localhost and ip literals for service-config and grpclb-in-DNS
+----
+* Author(s): apolcyn
+* Approver: a11r
+* Status: Draft
+* Implemented in: none currently, suited for all languages
+* Last updated: 2017-12-19
```
carl-mastrangelo: Could you bump this?

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
...
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/Ln6zeUDoo3M
+
+## Abstract
+
+This document discusses a problem and solution to bring down channel
```
carl-mastrangelo: optional: The problem this doc solves has increased in scope since Java has more fatal errors.  Consider adding some verbiage.

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
...
+wait for NXDOMAIN responses.
+
+While clients which don't implement grpclb-in-DNS or service
+config only need to do a "host's" file lookup when the target
+is "localhost", clients which do implement them add
+`lookup -> NXDOMAIN` RTT to the setup of every channel.
+
+## Proposal
+
+Special case "localhost" and ipv4/v6 literals (e.g. "1.2.3.4" and "::1").
```
carl-mastrangelo: nit: not a sentence.

apolcyn: fixed

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
apolcyn: done

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
...
+Don't query for SRV or TXT records when the target is "localhost" or an ipv4/v6 literal,
+and thus don't implement grpclb-in-DNS or service-config for
+such targets.
+
+## Rationale
```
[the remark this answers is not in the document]
apolcyn: good point. copied verbatim

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
...
+Clients that support grpclb-in-DNS and service-config in DNS must
+query for TXT and SRV records of `_grpclb._tcp.<target>` and
+`_grpc_config.<target>`, along with the A/AAAA records of
+`<target>`, when creating a channel to `<target>`. In the case
+that `<target>` exists in the machines "host's" file and
+grpclb-in-DNS/service-config are not set up for `<target>` (common
+when the target is "localhost" or an ipv4/v6 literal),
+channel setup can take much longer than desired due to the need
+to make these TXT and SRV lookups over the network, and then
+wait for NXDOMAIN responses.
```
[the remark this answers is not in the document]
apolcyn: done

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
...
+
+Clients that support grpclb-in-DNS and service-config in DNS must
+query for TXT and SRV records of `_grpclb._tcp.<target>` and
+`_grpc_config.<target>`, along with the A/AAAA records of
+`<target>`, when creating a channel to `<target>`. In the case
+that `<target>` exists in the machines "host's" file and
+grpclb-in-DNS/service-config are not set up for `<target>` (common
+when the target is "localhost" or an ipv4/v6 literal),
+channel setup can take much longer than desired due to the need
+to make these TXT and SRV lookups over the network, and then
```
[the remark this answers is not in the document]
apolcyn: done

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
...
+setup latency for grpc clients that implement grpclb-in-DNS and
+service-configs in DNS, in a special and degenerate case.
+
+## Background
+
+Clients that support grpclb-in-DNS and service-config in DNS must
+query for TXT and SRV records of `_grpclb._tcp.<target>` and
+`_grpc_config.<target>`, along with the A/AAAA records of
+`<target>`, when creating a channel to `<target>`. In the case
+that `<target>` exists in the machines "host's" file and
```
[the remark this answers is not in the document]
apolcyn: done

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
+Special case localhost and ip literals for service-config and grpclb-in-DNS
+----
+* Author(s): apolcyn
+* Approver: a11r
```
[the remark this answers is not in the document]
apolcyn: done

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -0,0 +1,54 @@
...
+* Status: Draft
+* Implemented in: none currently, suited for all languages
+* Last updated: 2017-12-19
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/Ln6zeUDoo3M
+
+## Abstract
+
+This document discusses a problem and solution to bring down channel
+setup latency for grpc clients that implement grpclb-in-DNS and
```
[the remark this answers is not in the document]
apolcyn: done, linked mentions of grpclb and service config to gRFCs

apolcyn:
thanks for the reviews, I've addressed the last round of comments

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
carl-mastrangelo: yes

apolcyn: I've added background on this problem as well - can you please check to see if my description is accurate?

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -36,18 +36,18 @@ service-configs-in-DNS potentially only need to do a "hosts" file
...
-adversely to this situation, because it raises a fatal error when
-attempting to open a socket to the name server when making
-SRV/TXT queries, and it's undesireable to adjust code to catch
-these errors.
+Another problem which is unrelated to latency but which also caused
+by SRV and TXT record lookups is related to the way in which
+Java treats DNS resolution failures. Internally, gRPC-Java raises a type
+of fatal error when a DNS query fails. Normally, gRPC-Java internally
+catches these types of errors and carries on, but there are certain
+automated test environments in whic the Java runtime is adjusted to
```
carl-mastrangelo: which

A10-avoid-grpclb-and-service-config-for-localhost-and-ip-literals.md
```diff
@@ -36,18 +36,18 @@ service-configs-in-DNS potentially only need to do a "hosts" file
...
+crash upon seeing these types of errors. Normally, DNS queries
```
carl-mastrangelo: A/AAAA

PR #83, opened 2018-05-30
A15: Promote Reflection Proto out of Alpha

carl-mastrangelo:


A15-promote-reflection.md
```diff
@@ -0,0 +1,57 @@
...
+## Rationale
+
+It is unlikely that the reflection proto service will change in
+backwards-incompatible ways.  The service has been implemented in each language
+implementation of gRPC and has not changed in two years.
+
+## Implementation
+
+1.  Copy grpc/reflection/v1alpha/reflection.proto to
+    grpc/reflection/v1/reflection.proto
```
dfawley: "... and update the package name."

A15-promote-reflection.md
```diff
@@ -0,0 +1,57 @@
...
+2.  Update existing service implementations in each repo to support the new
+    location in addition to the old location.
+3.  The old reflection.proto will be marked deprecated and for removal.
+4.  Clients (such as grpc_cli) will be updated to dual request reflection
+    information, preferring the new location first.
+5.  In an upcoming release (e.g. v1.15.x), the new proto will be announced
+    and the old proto will be declared for removal.
+6.  In the subsequent release (e.g. 1.16.x), the old implementations will be
```
dfawley: Is 1 release sufficient?  What is the cost of waiting longer?

A15-promote-reflection.md
```diff
@@ -0,0 +1,57 @@
...
+supports.  Both the protocol and the [description](
+https://github.com/grpc/grpc/blob/master/doc/server-reflection.md) of the
+reflection service have not changed for a long time.
+
+## Proposal
+
+It is proposed that the package name (and corresponding directory structure)
+for [reflection.proto](
+https://github.com/grpc/grpc/blob/v1.12.x/src/proto/grpc/reflection/v1alpha/reflection.proto)
+be changed from `grpc.reflection.v1alpha` to `grpc.reflection.v1`.  Java and Go
```
dfawley: What about C and/or the wrapped languages?

carl-mastrangelo: This part is less clear to me.  There is a C# and a C++ implementation, but I couldn't find implementations for the other langs.  I can't tell if this is intentional or not, so I was only planning on looking at those ones.

A15-promote-reflection.md
dfawley: I was advocating for an even longer period of supporting both.  Binary tools like grpc_cli can go many many months without being recompiled/deployed, so keeping it around longer should be a very low cost and helpful for our users.

carl-mastrangelo: The cost of maintaining this for Java and Go is probably feasible.  I would like to get input from the C side before picking a final supported version.

A15-promote-reflection.md
yang-g: I think it is good to know whether reflection is enabled by default in all the languages (C++ servers require users to explicitly link in a library). For C++ we can leave the user to decide which library to link in (or both).

A15-promote-reflection.md
yang-g: grpc_cli is an internal too which lives under `test/cpp`, I would not guess people using that extensively. I think one release is ok, unless someone in the discussion thread raises other issues.

A15-promote-reflection.md
a11r: Can we update the status for C++ as requested by other reviewers.

carl-mastrangelo: @a11r I have audited usages of the C++ implementation and have updated the plan here.   This looks doable with minimal code change and service interruption.

PTAL

carl-mastrangelo:
Submitted, thanks all for your input.

PR #85, opened 2018-06-12
A6-client-retries: minor clarifications and define a maximum value for maxTokens

dfawley:
The maximum value for `maxTokens` is obviously open for debate.  This was just a value I chose randomly that seemed like more than most people would want to use.


ejona86:
Seems fine to me. Nothing really controversial. I didn't approve since I don't have much skin in the game.

PR #86, opened 2018-06-15
L31: PHP interceptor API Change with Deserialize Argument

ZhouyihaiDing:
Since we haven't release the PHP interceptor, I think this is a good time for us to update before the first release.

### DIFF
--- a/L31-php-intercetor-api-change.md
+++ b/L31-php-intercetor-api-change.md
@@ -1,0 +1,64 @@
+Title
+----
+* Author(s): ZhouyihaiDing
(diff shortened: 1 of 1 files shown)


dfawley:
For consistency, please rename this file to `L31-php-...` before merging.  Thanks!

ZhouyihaiDing:
Hi @mehrdada ,
Sorry about this delay. The API change is used to support the GCP extension, which I can hook the pre-process and post-process to each call.
Can you please take a look and review the PR?
Thanks!

PR #88, opened 2018-06-27
L32: Add a public Channel API to the Node library

murgatroid99:
Discussion thread: https://groups.google.com/forum/#!topic/grpc-io/VR0C7DiG7bs

### DIFF
--- a/L32-node-channel-API.md
+++ b/L32-node-channel-API.md
@@ -70,7 +70,9 @@ class Channel {
     watchConnectivityState(currentState: ConnectivityState, deadline: Date|number, callback: (error?: Error) => void);
     /**
      * Create a call object. Call is an opaque type that is used by the Client
-     * class.
+     * class. This function is called by the gRPC library when starting a
+     * request. Implementers should return an instance of Call that is returned
+     * from calling createCall on an instance of the provided Channel class.
      * @param method The full method string to request.
      * @param deadline The call deadline
      * @param host A host string override for making the request
--- a/L32-node-channel-API.md
+++ b/L32-node-channel-API.md
@@ -1,0 +1,103 @@
+Node.js Channel API
+----
+* Author(s): mlumish
+* Approver: wenbozhu
+* Status: Draft
+* Implemented in: Node.js
+* Last updated: 
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+Expose the `Channel` class in the Node API, along with APIs for overriding the channel created in the `Client` constructor.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


L32-node-channel-API.md
```diff
@@ -0,0 +1,103 @@
...
+
+Expose the `Channel` class in the Node API, along with APIs for overriding the channel created in the `Client` constructor.
+
+## Background
+
+In the past, some users have requested the ability to explicitly create channels and share them among clients. In addition, another library needs to be able to intercept some channel functionality.
+
+
+### Related Proposals: 
+* A list of proposals this proposal builds on or supersedes.
```
nicolasnoble: I think you meant N/A here instead of a placeholder text ? :) Not sure what we've done in the past, but I don't think I've recalled this placeholder text like this.

L32-node-channel-API.md
```diff
@@ -0,0 +1,103 @@
...
+     */
+    createCall(method: string, deadline: Date|number, host: string|null, parentCall: Call|null, propagateFlags: number|null): Call;
+}
+```
+
+### New `Client` Construction Options
+
+We will add the following options to the `Client` constructor's options map:
+
+ - `channelOverride`: a `Channel` instance. The `Client` will use this channel for communicating, and will ignore all other channel construction options.
```
wenbozhu: Nits: channelOverride => providedChannel ?

L32-node-channel-API.md
```diff
@@ -0,0 +1,103 @@
...
+* A list of proposals this proposal builds on or supersedes.
+
+## Proposal
+
+We will add to the API a `Channel` class and two `Client` construction options.
+
+### New `Channel` Class API
+
+```ts
+enum ConnectivityState {
```
wenbozhu: should we add closed (half and/or full) too? is this how connectivity state defined in other languages?

murgatroid99: That enum was actually based on a slightly old version of the core's channel state enum. I fixed it, and now it matches the [connectivity state documentation](https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md).

L32-node-channel-API.md
murgatroid99: I called it `channelOverride` because it overrides the `Client`'s default channel construction. But I'm not very attached to the name.

L32-node-channel-API.md
wenbozhu: I would expect to see shutdown (full-close) and abort (fatal failure) .. but whatever core says :)

PR #89, opened 2018-06-29
Add Minimal Set of Channel Trace Events

ncteisen:
@markdroth, @carl-mastrangelo, @dapengzhang0, @lyuxuan, feel free to chime in if you think this list should be longer or shorter or modified in any way

### DIFF
--- a/A3-channel-tracing.md
+++ b/A3-channel-tracing.md
@@ -43,6 +43,17 @@ _Trace events_ should only be added for events that happen relatively infrequent
 ... Entering idle mode
 ```
 
+## Minimal Set of Data
+
+Channel tracing is a language agnostic feature, and while each implementation will vary, it is useful to define a minimal set of events that must be traced in order to consider the channel tracing feature as complete.
+
+All of these events must be traced:
+- Channel creation and deletion
+- Subchannel creation and deletion
+- Address resolution
+- Channel and subchannel state changes
+
+Language specific implementations may add any additional trace that is deemed useful, as long as the trace is not expected to happen at a per-RPC frequency.
 
 ## Format of Exported Data
 
--- a/A3-channel-tracing.md
+++ b/A3-channel-tracing.md
@@ -4,7 +4,7 @@ Channel Tracing
 * Approver: markdroth
 * Status: Draft
 * Implemented in: N/A
-* Last updated: 2018-03-01
+* Last updated: 2018-07-01
 * Discussion at: https://groups.google.com/forum/#!topic/grpc-io/WFDj3KeHYTI
 
 ## Terminology
--- a/A3-channel-tracing.md
+++ b/A3-channel-tracing.md
@@ -50,10 +50,15 @@ We define a minimal set of events that must be traced, in order to consider the 
 All of these events must be traced:
 - Channel creation and deletion
 - Subchannel creation and deletion
-- Address resolution resulting from service config change
-- Address resolution that causes number of backends to go from zero to non-zero
 - Channel connectivity state changes
 - Subchannel connectivity state changes
+- Interesting address resolution events (see below for details)
+
+Address resolution is special case. We want to track resolution events, but we do not want to flood the trace buffer with them. So we define several types of "interesting" resolution events that must be traced:
+- Address resolution resulting in service config change
+- Address resolution that causes number of backends to go from zero to non-zero
+- Address resolution that causes number of backends to go from non-zero to zero
+- Address resolution that causes a new LB policy to be created
 
 Language specific implementations may add any additional trace that is deemed useful, as long as the trace is not expected to happen at a per-RPC frequency.
 
--- a/A3-channel-tracing.md
+++ b/A3-channel-tracing.md
@@ -50,8 +50,10 @@ We define a minimal set of events that must be traced, in order to consider the 
 All of these events must be traced:
 - Channel creation and deletion
 - Subchannel creation and deletion
-- Address resolution
-- Connectivity state changes
+- Address resolution resulting from service config change
+- Address resolution that causes number of backends to go from zero to non-zero
+- Channel connectivity state changes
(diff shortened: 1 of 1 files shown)


A3-channel-tracing.md
```diff
@@ -43,6 +43,17 @@ _Trace events_ should only be added for events that happen relatively infrequent
 ... Entering idle mode
 ```
 
+## Minimal Set of Data
+
+Channel tracing is a language agnostic feature, and while each implementation will vary, it is useful to define a minimal set of events that must be traced in order to consider the channel tracing feature as complete.
```
carl-mastrangelo: run on sentence.

A3-channel-tracing.md
```diff
@@ -43,6 +43,17 @@ _Trace events_ should only be added for events that happen relatively infrequent
...
+
+All of these events must be traced:
```
carl-mastrangelo: connectivity state change should be in these too.   (Ready->Idle, etc.)

A3-channel-tracing.md
ncteisen: Donne

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
```diff
@@ -43,6 +43,17 @@ _Trace events_ should only be added for events that happen relatively infrequent
...
 ```
 
+## Minimal Set of Data
+
+We define a minimal set of events that must be traced, in order to consider the channel tracing feature as complete.
+
+All of these events must be traced:
+- Channel creation and deletion
+- Subchannel creation and deletion
+- Address resolution
```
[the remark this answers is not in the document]
markdroth: BTW, as an implementation detail, it's probably a good idea to make sure that LB policies have the ability to add trace events, since there may be policy-specific events of interest.  For example, the grpclb policy may want to add an event when it establishes or loses a connection to a balancer, or when it receives a serverlist with all drop entries, or something else.

A3-channel-tracing.md
```diff
@@ -43,6 +43,19 @@ _Trace events_ should only be added for events that happen relatively infrequent
...
 ```
 
+## Minimal Set of Data
+
+We define a minimal set of events that must be traced, in order to consider the channel tracing feature as complete.
+
+All of these events must be traced:
+- Channel creation and deletion
+- Subchannel creation and deletion
+- Address resolution resulting from service config change
```
markdroth: s/from/in/

A3-channel-tracing.md
```diff
@@ -43,6 +43,19 @@ _Trace events_ should only be added for events that happen relatively infrequent
...
+- Address resolution that causes number of backends to go from zero to non-zero
```
markdroth: Or vice-versa.

markdroth: The address resolution case is complex enough that it warrants more than just a couple of bullets in this list.  Let's have a separate paragraph explaining why we need to be selective about which address resolution events get recorded, and then have a separate bulleted list identifying the cases in which we should log resolver results.

Note that we also need to record the address resolution result whenever it causes us to create a new LB policy.

A3-channel-tracing.md
ncteisen: Done

A3-channel-tracing.md
ncteisen: Done

ncteisen: Done

A3-channel-tracing.md
```diff
@@ -50,10 +50,15 @@ We define a minimal set of events that must be traced, in order to consider the
...
-- Address resolution that causes number of backends to go from zero to non-zero
 - Channel connectivity state changes
 - Subchannel connectivity state changes
+- Interesting address resolution events (see below for details)
+
+Address resolution is special case. We want to track resolution events, but we do not want to flood the trace buffer with them. So we define several types of "interesting" resolution events that must be traced:
+- Address resolution resulting in service config change
+- Address resolution that causes number of backends to go from zero to non-zero
+- Address resolution that causes number of backends to go from non-zero to zero
+- Address resolution that causes a new LB policy to be created
```
dapengzhang0: Is it possible that "Address resolution that causes a new LB policy to be created" without "resulting in service config change"? What is the case?

A3-channel-tracing.md
```diff
@@ -50,10 +50,15 @@ We define a minimal set of events that must be traced, in order to consider the
 All of these events must be traced:
 - Channel creation and deletion
 - Subchannel creation and deletion
-- Address resolution resulting from service config change
-- Address resolution that causes number of backends to go from zero to non-zero
 - Channel connectivity state changes
 - Subchannel connectivity state changes
+- Interesting address resolution events (see below for details)
+
+Address resolution is special case. We want to track resolution events, but we do not want to flood the trace buffer with them. So we define several types of "interesting" resolution events that must be traced:
```
markdroth: We should explain why flooding the trace buffer is a concern.  Users who are used to DNS-based environments where backends have fairly static addresses may not understand this.  We need to explain that there are environments where backends are more dynamically scheduled and the resolver will be delivering updates very frequently, in which case we don't want to log every result returned by the resolver, since most of them are far less interesting than the other events that would be pushed out of the trace buffer.

A3-channel-tracing.md
markdroth: Yes.  If the resolver had previously not returned any balancer addresses and now returns one, we will switch to the grpclb LB policy.  Conversely, if the resolver had previously returned a balancer address and now does not return any, then we will switch away from the grpclb LB policy.

dapengzhang0: I see.

A3-channel-tracing.md
ncteisen: Done

PR #95, opened 2018-08-22
A18: option for setting TCP_USER_TIMEOUT

yashykt:
Discussion at https://groups.google.com/forum/#!topic/grpc-io/6VZYCFZpyTI


A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,68 @@
...
+option TCP_USER_TIMEOUT for TCP sockets to a default value of 20 seconds. (The
+system defaults might be as high as 20 minutes.)
+
+gRPC will also provide a configuration option for both channels and servers to
+set the TCP_USER_TIMEOUT value.
+
+## Rationale
+
+This proposal solves two separately seen issues, but with the same root cause
+that was described above -
```
sreecha: Would help the reader if you just add a quick summary of what the issue is here..
(a short paragraph for each issue).

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,68 @@
...
+[#15871](https://github.com/grpc/grpc/issues/15871).
+
+
+## Implementation
+
+C-Core - [#16419](https://github.com/grpc/grpc/issues/16419) sets the
+TCP_USER_TIMEOUT socket option. A new channel argument,
+GRPC_ARG_TCP_USER_TIMEOUT_MS will be introduced to configure the value.
+
+JAVA -
```
hsaliak: we probably want to flesh out the story for other stacks and platforms..
what incompatibilities can happen here, if any? If we expect this to not cause incompatibility with systems that do not support this option, it would be beneficial to call that out since thats  the first question most people would have..

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,68 @@
...
+For all Linux platforms running Linux 2.6.37 and later, gRPC will set the socket
+option TCP_USER_TIMEOUT for TCP sockets to a default value of 20 seconds. (The
+system defaults might be as high as 20 minutes.)
+
+gRPC will also provide a configuration option for both channels and servers to
+set the TCP_USER_TIMEOUT value.
+
+## Rationale
+
+This proposal solves two separately seen issues, but with the same root cause
```
srini100: Solves? "Helps mitigate" would be more appropriate here. TCP_USER_TIMEOUT is not an exact replacement for RPC deadline.

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,68 @@
...
+before TCP will forcibly close the corresponding connection and return ETIMEDOUT
+to the application.  If the option value is specified as 0, TCP will to use the
+system default."
+
+At the time this proposal was written, gRPC Core faced an issue where data could
+be stuck for a very long time at the gRPC TCP layer. For cases where the network
+goes down, and the socket cannot accept more data for writing, an RPC could be
+blocked for a very long time. Even if the deadline expired on such an RPC, gRPC
+Core was incapable of shutting down the RPC, because the TCP layer would stil
+have references to user data. Setting TCP_USER_TIMEOUT to a reasonable value
```
srini100: Worth noting that fixing the reference issue would come at the cost of performance.

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,68 @@
...
+gRPC will also provide a configuration option for both channels and servers to
+set the TCP_USER_TIMEOUT value.
+
+## Rationale
+
+This proposal solves two separately seen issues, but with the same root cause
+that was described above -
+[#15889](https://github.com/grpc/grpc/issues/15889),
+[#15871](https://github.com/grpc/grpc/issues/15871).
+
```
srini100: Besides mitigating existing issues, TCP_USER_TIMEOUT by itself is a useful setting to have. It provides a fairly reliable way to detect TCP connection breakage without any additional cost.

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Implementation
+
+C-Core - [#16419](https://github.com/grpc/grpc/issues/16419) sets the
+TCP_USER_TIMEOUT socket option. A new channel argument,
+GRPC_ARG_TCP_USER_TIMEOUT_MS will be introduced to configure the value.
```
srini100: What are the allowed min/max values? What would 0 mean, gRPC default or system default?

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,68 @@
...
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/6VZYCFZpyTI
+
+## Abstract
+
+Provide an option to set the socket option, TCP_USER_TIMEOUT on TCP sockets for
+Linux kernels 2.6.37 and later.
+
+## Background
+
+TCP_USER_TIMEOUT is a socket option that has been available on Linux since Linux
```
srini100: ~Linux~ kernel version

srini100:
@ejona86 @dfawley PTAL for Java and Go.

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,55 @@
...
+Even by doing one of the two currently known solutions, the issue mentioned in #15871 would still remain, since we require sending the HTTP2 GOAWAY frame before closing the connection in certain scenarios.
+
+Other than solving the two described issues, TCP_USER_TIMEOUT would also provide a fairly reliable way to detect TCP connection breakage without any additional cost.
+
+## Implementation
+
+C-Core - [#16419](https://github.com/grpc/grpc/issues/16419) sets the TCP_USER_TIMEOUT socket option.
+A new channel argument, GRPC_ARG_TCP_USER_TIMEOUT_MS will be introduced to configure the value. The channel argument value will be passed as the socket option value. The minimum value, 0, would result in setting TCP_USER_TIMEOUT to the system default. The maximum value will be capped at 20 minutes (system default).
+
+JAVA - For JAVA, the support would be currently restricted to netty. The OkHttp library we use does not support setting such an option.
```
ejona86: Netty with NIO (which is our default) also doesn't support this. Only Netty with epoll/kqueue supports this, and when the user is configuring gRPC to use epoll/kqueue they can also configure TCP_USER_TIMEOUT via pass-through configuration in our API.

So as this proposal stands now, there no benefit to adding anything in Java.

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,55 @@
...
+* Last updated: 2018-8-22
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/6VZYCFZpyTI
+
+## Abstract
+
+Provide an option to set the socket option, TCP_USER_TIMEOUT on TCP sockets for Linux kernels 2.6.37 and later.
+
+## Background
+
+TCP_USER_TIMEOUT is a socket option that has been available on Linux since kernel version 2.6.37. As described the Linux manual page, "This option takes an unsigned int as an argument. When the value is greater than 0, it specifies the maximum amount of time in milliseconds that transmitted data may remain unacknowledged before TCP will forcibly close the corresponding connection and return ETIMEDOUT to the application.  If the option value is specified as 0, TCP will to use the system default."
```
hsaliak: given that one of the issues referenced is a python issue, has this been verified to work on manylinux?

yashykt: It will depend on the linux kernel version being used.. if it's lower than 2.6.37 then it will not work

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,55 @@
...
+* adding API to gRPC Core's TCP layer to be able to cancel writes for certain streams - This would require gRPC's TCP layer to understand streams and this breaks layers, not to mention the additional bookkeeping and overhead in doing so.
+
+Even by doing one of the two currently known solutions, the issue mentioned in #15871 would still remain, since we require sending the HTTP2 GOAWAY frame before closing the connection in certain scenarios.
+
+Other than solving the two described issues, TCP_USER_TIMEOUT would also provide a fairly reliable way to detect TCP connection breakage without any additional cost.
+
+## Implementation
+
+C-Core - [#16419](https://github.com/grpc/grpc/issues/16419) sets the TCP_USER_TIMEOUT socket option.
+A new channel argument, GRPC_ARG_TCP_USER_TIMEOUT_MS will be introduced to configure the value. The channel argument value will be passed as the socket option value. The minimum value, 0, would result in setting TCP_USER_TIMEOUT to the system default. The maximum value will be capped at 20 minutes (system default).
```
ejona86: 1) This talk about minimum values should not be in "implementation" but in "proposal". I missed this and only noticed because I glanced at a comment from @srini100
2) No. We need a greater minimum value than 0. You may define '0' to mean "system default". But we should not allow the user to set a 1ms TCP_USER_TIMEOUT. That is just asking for trouble. I'd suggest a value no less than 10s.
3) "system default" isn't really correct. '0' uses the system default. The _Linux_ default may be 20 minutes (although that doesn't seem to agree with the [tcp documentation](https://linux.die.net/man/7/tcp), and it should be based on TCP RTTs), but system administrators can change it via /proc/sys/net/ipv4/.

A16-tcp-user-timeout.md
ejona86: manylinux is older than 2.6.37. That said, you can still support the newer kernels and older kernels in the same binary. But if you're on the older machines this won't do anything.

ejona86: CentOS 5 is on 2.6.18. CentOS 6 is 2.6.32. Based on the dockerfile at https://github.com/pypa/manylinux, it seems manylinux implies CentOS/RHEL 5. I'll note that CentOS 5 is unsupported, but RHEL 5 is in "extended support".

nicolasnoble: The manylinux case isn't what you seem to talk about. Manylinux is only the building platform, not the running platform. But if the building platform doesn't have the TCP_USER_TIMEOUT symbol defined, the code to toggle it won't be compiled in, and it won't work, even in the case of the binary being run on a newer platform.

nicolasnoble: It might be relevant to link to the current pull request that's implementing this proposal :-)

https://github.com/grpc/grpc/pull/16419/files

A16-tcp-user-timeout.md
yashykt: 1. Thanks! I'll move it to proposal.
2. I agree that 1ms is a dangerous value (dangerous in the sense that the connection will be dropped), but if the kernel allows it, I don't see why we should restrict it. The application itself would be responsible for the connection being dropped due to the low setting (if the kernel honors it).
3. I'll remove the system default wording from the max value. I'm also tempted to make to same argument as the minimum value for the maximum value. If the kernel allows it, we shouldn't be restricting it.

A16-tcp-user-timeout.md
yashykt: Thanks for checking! I'll update the doc to reflect this.

A16-tcp-user-timeout.md
srini100: While 20s default is reasonable we have to look at the current limitations in Core implementation and see how users are going to deploy this feature to mitigate those issues before limiting the minimum to 10s. As in #15889, if gRPC deadline doesn't work users have no option but to rely on TCP_USER_TIMEOUT to detect connection breakage much faster than 10 seconds.

A16-tcp-user-timeout.md
yashykt: It's linked to in "implementation".

A16-tcp-user-timeout.md
ejona86: > I agree that 1ms is a dangerous value (dangerous in the sense that the connection will be dropped), but if the kernel allows it, I don't see why we should restrict it.

Because we are the experts. And we know you are only going to cause problems with a setting that low. We also know that it can appear to "work" and then behave very poorly if the network ever suffers a slight problem; that is, it makes the system brittle. The kernel may not (or may! do we know?) place any limits, but it is used directly by experts and changes slowly compared to userspace.

> While 20s default is reasonable we have to look at the current limitations in Core implementation and see how users are going to deploy this feature to mitigate those issues before limiting the minimum to 10s.

I see the Background paragraph now. I didn't realize that was part of this. If 10 seconds is not enough to solve the Core implementation problem, then that's an argument that this isn't a solution. If you only wanted to use this to be able to react to cancellations in less than "minutes" (like 20 seconds), this would be fine. But I don't think it is appropriate to abuse this option for the cancellation problem with Core if you're wanting greater than that level of precision, because they are two fundamentally different problems and they don't solve each other:

1. Timeouts can be milliseconds which is _far_ too low of a value for TCP_USER_TIMEOUT. Yes, we can tell applications to use something higher, but the fact remains that applications will want this to be much lower than is safe for the network. We must not encourage putting the network in harm's way
2. TCP_USER_TIMEOUT doesn't actually help much for less than 20 seconds: it does not provide any guarantees as to the rate of the connection. If you have a 1 byte/second connection, TCP_USER_TIMEOUT can remain happy, even though it may take hours to drain a buffer. It just requires that anything sent is ACKed, but it places no limit on the sending rate. (Relatedly, but not too important for HTTP/2, if the remote application is not reading then flow control will push back. That push-back is not impacted by TCP_USER_TIMEOUT.)
3. TCP_USER_TIMEOUT doesn't actually help: it limits the amount of time after a successful socket write before that data is acknowledged. However, C is suffering from buffering _before_ the socket write. If the socket writes successfully, then C no longer needs to retain the memory.

Instead maybe Core would need to invest in using TCP_NOTSENT_LOWAT maybe combined with SO_SNDBUF (so that write() can succeed) or figure out some other solution. Note that C would also need to combine this with how much data it is buffering _before_ getting to the socket. I believe that was hard-coded to a max of 1 MB or some such.

yashykt: TCP_USER_TIMEOUT as @ejona86 explains is not intended to guarantee the cancellation deadline. The only aim is to prevent gRPC Core from getting blocked forever (20 minutes) if the network goes down. The default setting of 20 seconds is enough to solve this and is reasonable for most applications in my opinion. 

The only argument is whether the user should be allowed to set it below 10 seconds. I  understand @srini100 's  point that users for example in https://github.com/grpc/grpc/issues/15889 might want a way to enforce a deadline of 300ms through TCP_USER_TIMEOUT. I am not entirely convinced that we should be restricting it. Also, it doesn't seem like KEEPALIVE_TIMEOUT is explicitly restricted to 10 seconds either. KEEPALIVE_TIME is stated to have a minimum of 10 seconds, but the same is not said for KEEPALIVE_TIMEOUT as per https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md

ejona86: I see, you're right, KEEPALIVE_TIMEOUT doesn't have a limit. I'd be willing to consider that an oversight and we add one.

A16-tcp-user-timeout.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Related Proposals:
+[Client-side Keepalive](https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md)
+[Server-side Connection Management](https://github.com/grpc/proposal/blob/master/A9-server-side-conn-mgt.md)
+
+## Proposal
+
+gRPC already has a mechanism to detect connection failures, called [keepalive](https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md). The KEEPALIVE_TIMEOUT configuration option controls the amount of time, gRPC waits for a HTTP/2 ping to be acknowledged. Since the effective aim of KEEPALIVE_TIMEOUT is similar to TCP_USER_TIMEOUT, the proposal is to reuse this configuration option to set the socket option TCP_USER_TIMEOUT. Since, TCP_USER_TIMEOUT is only available on Linux kernels 2.6.37 and later, this proposal is limited to those platforms.
+
```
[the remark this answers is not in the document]
yashykt: That is true, and it is also a good thing in a way since applications will not notice unwanted behavioral changes anymore due to a default TCP_USER_TIMEOUT of 20 seconds, if keepalive was disabled in the first place.

On the other hand, people who are worried about connection breakage getting detected on time would have already set keepalives, and doing it this way will allow them to continue with that.

dfawley:
Please rename the file to A18-....  (A16 was taken and A17 is taken now.)

A16-tcp-user-timeout.md
```diff
@@ -18,13 +18,18 @@ TCP_USER_TIMEOUT is a socket option that has been available on Linux since kerne
 At the time this proposal was written, gRPC Core faced an issue where data could be stuck for a very long time at the gRPC TCP layer. For cases where the network goes down, and the socket cannot accept more data for writing, an RPC could be blocked for a very long time. Even if the deadline expired on such an RPC, gRPC Core was incapable of shutting down the RPC, because the TCP layer would still have references to user data. Setting TCP_USER_TIMEOUT to a reasonable value would put a bound on the maximum time the TCP layer would wait for TCP packets to be acknowledged, and close the connection if the packets are not acknowledged in time. This would end up releasing references to user data at the gRPC TCP layer, thereby allowing the RPC to fail in a timely manner.
 
 ### Related Proposals:
-N/A
+[Client-side Keepalive](https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md)
```
ejona86: You might consider making these links relative (i.e., just "A8-client-side-keepalive.md"). The are easier to read and work properly in more environments.

A16-tcp-user-timeout.md
```diff
@@ -18,13 +18,18 @@ TCP_USER_TIMEOUT is a socket option that has been available on Linux since kerne
...
+[Server-side Connection Management](https://github.com/grpc/proposal/blob/master/A9-server-side-conn-mgt.md)
 
 ## Proposal
 
-For all Linux platforms running Linux 2.6.37 and later, gRPC will set the socket option TCP_USER_TIMEOUT for TCP sockets to a default value of 20 seconds. (The system defaults might be as high as 20 minutes.)
+gRPC already has a mechanism to detect connection failures, called [keepalive](https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md). The KEEPALIVE_TIMEOUT configuration option controls the amount of time, gRPC waits for a HTTP/2 ping to be acknowledged. Since the effective aim of KEEPALIVE_TIMEOUT is similar to TCP_USER_TIMEOUT, the proposal is to reuse this configuration option to set the socket option TCP_USER_TIMEOUT. Since, TCP_USER_TIMEOUT is only available on Linux kernels 2.6.37 and later, this proposal is limited to those platforms.
```
ejona86: "the amount of time, gRPC waits" shouldn't have a comma.

ejona86: "Since, TCP_USER_TIMEOUT" shouldn't have a comma.

A16-tcp-user-timeout.md
```diff
@@ -18,13 +18,18 @@ TCP_USER_TIMEOUT is a socket option that has been available on Linux since kerne
...
 
-gRPC will also provide a configuration option for both channels and servers to set the TCP_USER_TIMEOUT value. This configuration option will have millisecond accuracy and will have the same semantics as the socket option. Setting it to 0 will configure TCP_USER_TIMEOUT with the system default and not the gRPC default of 20 seconds. No minimum value to this option will be enforced by gRPC. It will be the responsibilty of the application to be prudent while setting this option. A very small TCP_USER_TIMEOUT value can affect TCP transmissions over paths with a high latency or packet loss. If the timeout occurs before the acknowledgement arrives, then the connection will be dropped.
+A very small TCP_USER_TIMEOUT value can affect TCP transmissions over paths with a high latency or packet loss. If the timeout occurs before the acknowledgement arrives, then the connection will be dropped. A very large value, on the other hand. will increase the time it takes gRPC to detect broken connections. Hence, it will be the responsibility of the application to be prudent while setting this option.
```
ejona86: "other hand. will" Period should be a comma.

yashykt: done!

yashykt:
All comments have been addressed, and enough time has passed since discussion started. Marking this proposal as final.

PR #98, opened 2018-09-05
L46: Create a new TLS credential API that supports SPIFFE mutual TLS

yihuazhang:
This gRFC proposes a new TLS credential API that will be used to support SPIFFE mutual TLS. 

Discussion at: https://groups.google.com/forum/#!topic/grpc-io/gMrROsizrtc


L35-grpc-core-tls-credential-API.md
```diff
@@ -0,0 +1,281 @@
+Title
+----
+* Author(s): yihuazhang
+* Approver: vjpai, markdroth
+* Status: Draft
+* Implemented in:
+* Last updated: September 3, 2018
+* Discussion at:
```
carl-mastrangelo: needs a discussion url in the grpc-io forum.

nicolasnoble:
That's not an `L`, that's a `A`. You're going to affect most languages with it, aka all of the languages wrapping the core. cc @hsaliak and @vjpai.

A19-tls-credential-API
```diff
@@ -0,0 +1,281 @@
...
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/gMrROsizrtc
+
+## Abstract
+This proposal aims to incorporate SPIFFE-based mutual TLS into gRPC C core auth stack. Towards the goal, it proposes a new credential API model that 1) allows flexible addition of new TLS-related features (e.g., credential reload) without breaking the API and 2) supports both synchronous and asynchronous implementations of those features. It then proposes SPIFFE and HTTPS-based TLS credential API's that comply with the new model.
+
+## Background
+
+The current TLS implementation in gRPC C core has the following problems when it comes to support SPIFFE-based mutual TLS.
+
+1: It is originally designed to comply with HTTPS semantics which makes it difficult to be configured to support SPIFFE-based mutual TLS. For instance, the server authorization check in the current implementation is enforced locally by leveraging whitelisted endpoint information embedded in peer certificates while in SPIFFE use case, it may be enforced via an external server authorization check service.
```
jiangtaoli2016: Please clarify what is server authorization. such as "_Server Authorization provides client-side checks that the servers clients connect to are legitimate._"

A19-tls-credential-API
```diff
@@ -0,0 +1,281 @@
...
+Regarding grpc_tls_ctx_customize_config, the config is used when a caller wants to
+configure an underlying TLS context with customized primitives. For example, there could
+be a use case in which raw private keys are not directly accessible (e.g., hardware-backed) to a caller,
+but private key methods (e.g., “Sign” functions) are, in which case the config provides a means
+to customize an TLS context with an SSL_PRIVATE_KEY_METHOD object and associated signing algorithms.
+
+## Implementation
+
+The implementation of this proposal is straightforward as all API changes will be
+made to include/grpc/grpc_security.h.
```
jiangtaoli2016: Also mention that when this proposal is implemented, we will start deprecating old SSL credentials API. User of gRPC TLS transport security shall start using this new API.

A19-tls-credential-API
```diff
@@ -0,0 +1,281 @@
...
+3: The current TLS credential API is inflexible in incorporating new TLS-related features in future as not only does it break the API, it also requires all wrapped languages to update their implementations in a lockstep manner.
+
+This proposal introduces a new credential API model that allows flexible addition of TLS-related features that can be implemented in either a synchronous or asynchronous manner. It then proposes new credential API's for both HTTPS and SPIFFE-based TLS that complies with the API model.
+
+### Related Proposals:
+N/A
+
+## Proposal
+
+The first part of proposal introduces credential options and credential create APIs that will be added to include/grpc/grpc_security.h.
```
jiangtaoli2016: Add a note that although grpc_tls_https_server_credentials_create, grpc_tls_https_credentials_create, grpc_tls_spiffe_server_credentials_create, grpc_tls_spiffe_credentials_create use the same grpc_tls_credentials_options struct as input, certain configure may not be applicable and will be ignored during credential creation. E.g., grpc_ssl_client_certificate_request_type config is only applicable to https_server_credentials creation.

yihuazhang:
Thanks Jiangtao for the review. All comments are addressed.

yihuazhang:
Friendly ping @vjpai.

euroelessar:
It looks like `.md` suffix was lost during the rename?

yihuazhang:
@euroelessar, 

The design was finalized, and we are at the implementation phase. We aim to finish it by the end of Q4.

dfawley:
Are we sure this is an "A" proposal?  It does not affect _all_ languages, just core and wrapped languages.  To me, that makes it an "L".  I.e. I get scared every time I see this PR because I think someone has approved API changes for Go and Java as part of this.

jiangtaoli2016:
@dfawley we only position this work for core and wrapped languages. It shall not affect Java/Go.
@yihuazhang please update from "A" to "L".

yihuazhang:
The proposal is changed from `L` as suggested.

dfawley:
Please name the file `L42-core-____.md`.  @vjpai, should we have a different name to indicate it affects wrapped languages as well?  Or do we assume all core changes might affect wrapped languages?

vjpai:
@dfawley : So far we've been using `L` for core changes since it's core whose API is changing and the other languages have to either react (for breakages) or choose whether to react (for extensions). It's as though the platform underneath the other languages has changed.

Now that said, this doesn't give a good answer if we want a simultaneous API extension at core and the wrapped languages. I think this is an unlikely event unless it truly is an `A`. That said, if it's just core and C++, I think `L` would still be appropriate (the callback API would be a good example of this once we decide to de-experimentalize it).

sanjaypujare:
L42 is already taken (L42-python-metadata-flags.md). Should rename this to the next available number...

sanjaypujare:
What is holding up merging of this PR? I see all comments have been addressed

L42-core-tls-credential-API.md
```diff
@@ -0,0 +1,235 @@
...
+
+/** TLS credentials options. */
+typedef struct grpc_tls_credentials_options grpc_tls_credentials_options;
+
+/** Create an empty TLS credentials options. */
+GRPCAPI grpc_tls_credentials_options* grpc_tls_credentials_options_create();
+
+/** Set grpc_ssl_client_certificate_request_type field in credentials options
+    with the provided type. */
+GRPCAPI void grpc_tls_credentials_options_set_cert_request_type(
```
markdroth: Please update this as per the latest changes in grpc/grpc#17549.

sanjaypujare:
I sort of agree with @hsaliak - this needs to include all the gRPC languages. Whether we achieve it thru making this an A or have other L proposals for Java and Go can be debated.

sanjaypujare:
@yihuazhang in that case can you comment on whether you get equivalent functionality (i.e., credential reload or server authorization) in Java/Go thru a different hook/plugin or it simply won't be available? Note that we need the functionality in all gRPC languages.

@dfawley  how does one do SPIFFE validation in Go? Are there callbacks one can implement? In that case the library itself can provide that functionality, I am thinking

dfawley:
> @dfawley how does one do SPIFFE validation in Go? Are there callbacks one can implement? In that case the library itself can provide that functionality, I am thinking

@sanjaypujare In Go you would implement a [handshaker](https://godoc.org/google.golang.org/grpc/credentials#TransportCredentials) to do the validation.

> In that case the library itself can provide that functionality, I am thinking

I'm not sure what you are saying.  Which library?

For standard TLS, we use the Go TLS library to implement the grpc-go `TransportCredentials` interface ([example implementation](https://github.com/grpc/grpc-go/blob/master/credentials/credentials.go#L228)).

sanjaypujare:
> 
> I'm not sure what you are saying. Which library?
>

The grpc-go library.
 
> For standard TLS, we use the Go TLS library to implement the grpc-go `TransportCredentials` interface ([example implementation](https://github.com/grpc/grpc-go/blob/master/credentials/credentials.go#L228)).

Yes, I was thinking of this same place (meaning same directory or parent directory) to put this additional functionality for SPIFFE certs

dfawley:
> Yes, I was thinking of this same place (meaning same directory or parent directory) to put this additional functionality for SPIFFE certs

Makes sense.  We could add SPIFFE TransportCredentials in the same location.

sanjaypujare:
> Makes sense. We could add SPIFFE TransportCredentials in the same location.

Great! So are there any plans to add this?

yihuazhang:
@sanjaypujare 

In C-core implementation, we can control where to invoke those hooks/plugins. For instance, credential reload will happen when performing the first `tsi_handshaker_next()` in `ssl_transport_security.cc`, and server authorization check happens right after handshake finishes, but before the peer information gets returned (in `ssl_security_connector.cc`). In Java and Go, I suspect that we have the same level of flexibility to control those internal details (e.g., where to invoke those plugins) since AFAIK all SSL-related details are wrapped within an object and just gets passed to the API that does the handshake. Furthermore, I also doubt that sync/async API's mentioned in this proposal can find a counterpart in Java and Go since they those API's are tightly bound to gRPC core by using the `grpc_core::exec_ctx'. 

I understand that gRPC Java and Go will still need to implement the functionality of credential reload and server authorization, but I do not think they will be implemented via plugins/hooks proposed in this doc.  I mentioned that in the gRPC proposal.

yihuazhang:
Any other comments on this gRFC? Can it be merged?

sanjaypujare:
LGTM

markdroth:
I would like to see a written design for how this functionality will be provided in Java and Go, with approvals from Eric and Doug.  I don't really care whether we add that to this gRFC and change it from `L` to `A` or whether we publish a separate `L` one for each language.  However, I do want this plan to be written and approved before I approve this gRFC, since I don't want this to go in without a plan for how we're going to do this in all languages.

yihuazhang:
@jiangtaoli2016 

@sanjaypujare Do we have a plan to support SPIFFE in gRPC Java and Go with your effort in managed gRPC? AFAIK, the security team does not have enough resources to work on it Q1 (and maybe Q2 as well) unless its priority gets bumped up.

@markdroth I understand your concern. I also believe that it will take some time to flesh out the details for SPIFFE support in gRPC Java and Go (if we decide to do). In order to unblock the current C++ effort, how about namespace the C++ API as `experimental` (as you suggested) and later come back to this gRFC once we have plans for Java and Go?

jiangtaoli2016:
@markdroth How about a high level plan (one-pager) for Java and Go for supporting SPIFFE type credentials? We have not allocated resource on SPIFFE Java/Go support this quarter. We probably don't have resource to do a detailed design on Java and Go. 

We don't want grpc c core support on SPIFFE being blocked. External customers are interested in this.

sanjaypujare:
@markdroth has a point and I agree with. My preference is to have a separate proposal for Java and Go each (high level plan @jiangtaoli2016 suggested) and unblock this proposal based on that. But I don't think we should wait for those proposals to be approved for this to be approved.

@yihuazhang to answer your question, it is not clear yet if the gRPC Java and Go contributor team have any plans to add SPIFFE (and other) support in their respective languages.

@jiangtaoli2016 can you come up with a one-pager for Java and Go as the starting point?

markdroth:
I would be willing to go forward with this with just a one-pager for each of Java and Go, leaving the detailed design work for later.

I would also be fine with deferring the gRFC and moving forward with the C-core implementation while labelling the new APIs as experimental.

My goal here is just to make sure that we're all confident that there won't be any design blockers in either of the other languages before we commit to this API in C.  Once we commit to it, we have to support it, so if we later determine that there is some constraint in some other language that changes what we plan to support here, it will be too late to fix it.

yihuazhang:
Thanks All. So my plan is to use the experimental namespace for C-core API to unblock the C-core effort then, come up with a one-pager design doc for Java and Go to get this gRFC resolved.

jiangtaoli2016:
Per offline discussion, @yihuazhang will work on one-pager design for SPIFFE credential API for Java and Go.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,236 @@
...
+/** Set grpc_ssl_client_certificate_request_type field in credentials options
+    with the provided type. options should not be NULL.
+    It returns 1 on success and 0 on failure. */
+GRPCAPI int grpc_tls_credentials_options_set_cert_request_type(
+    grpc_tls_credentials_options* options,
+    grpc_ssl_client_certificate_request_type type);
+
+/** Set grpc_tls_key_materials_config field in credentials options
+    with the provided config struct whose ownership is transferred.
+    Both parameters should not be NULL.
```
jtattermusch: clarity: `Both parameters should not be NULL.` is ambiquous.  one of them needs to be set? setting just one of them to null is illegal?  Also "should" is not the right word here I think?

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,236 @@
...
+
+The second part of proposal introduces API's related to configs of TLS-related features that will be added to include/grpc/grpc_security.h.
+
+```c++
+/** --- TLS key materials config. --- */
+
+/** Create an empty grpc_tls_key_materials_config instance. */
+GRPCAPI grpc_tls_key_materials_config* grpc_tls_key_materials_config_create();
+
+/** Set grpc_tls_key_materials_config instance with provided a TLS certificate.
```
jtattermusch: wording: `with provided a TLS certificate`

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,236 @@
...
+      API. arg is used to pinpoint an exact check request to be cancelled. The
+      operation may not have any effect if the request has already been
+      processed.
+    - destruct is a pointer to an application-provided callback used to clean up
+      any data associated with the config.
+*/
+GRPCAPI grpc_tls_server_authorization_check_config*
+grpc_tls_server_authorization_check_config_create(
+    const void* config_user_data,
+    int (*schedule)(void* config_user_data,
```
jtattermusch: note that the pattern  where C core calls a user defined callback, which in turn should call a "done" callback provided as one of the arguments is a pattern that might work well in C and C++, but it is difficult to implement in wrapper languages (it involves a callback from native code back to managed code, which can be troublesome by itself because there might be limitations on what code can run in such constrained environment and with garbage collected languages managing the lifecycle of all the involved objects correctly is hard).

We did something similar with the authentication plugin API in the past and it has resulted in quite a lot of ugly C# code, so perhaps this time we could give more thought to making sure the new API is easily wrappable in other languages than C++.

sanjaypujare: @jtattermusch good point. Are there examples of how to do it right with respect to all the wrapped languages? Or a document that describes how to do it?

yihuazhang: @markdroth
 
Thanks @jtattermusch for the feedback. When designing async SPIFFE API, we indeed followed the similar pattern as that of plugin credentials. For credential reload, we decided not to support async callback (I will update the proposal) since it turns out not giving too much benefit, but for server authorization check, async callback will be still supported. 

As @sanjaypujare suggested, it will be great if you could provide more details on what the wrapped languages expect w.r.t async callback. It is unclear to me how the current API's needs to be changed to make wrapped languages's implementation easier.

jiangtaoli2016: The SPIFFE API supports both sync and async callbacks, so that we are flexible for future changes. We can only expose sync callback to python and other wrapped languages, where sync and async callbacks are only available to C++.

bluecloudes:
Any update about SPIFFE credential API for Java and Go?

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,236 @@
...
+
+/** A callback function provided by gRPC to handle the result of credential
+    reload. It is used when schedule API is implemented asynchronously and
+    serves to bring the control back to grpc C core. */
+typedef void (*grpc_tls_on_credential_reload_done_cb)(
+    grpc_tls_credential_reload_arg* arg);
+
+/** A struct containing all information necessary to schedule/cancel
+    a credential reload request. cb and cb_user_data represent a gRPC-provided
+    callback and an argument passed to it. key_materials is an in/output
```
sanjaypujare: Param name is `key_materials_config` (change from `key_materials`)

Falco20019:
Any plans on when the gRFC is merged? C++ seems to already implement it.

markdroth:
The API is currently experimental, which is why the gRFC has not been finalized.  We will be making a few more tweaks to the API in the area of credential reloading before we finalize this.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,236 @@
...
+    with the provided config struct whose ownership is transferred.
+    Both parameters should not be NULL.
+    It returns 1 on success and 0 on failure. */
+GRPCAPI int grpc_tls_credentials_options_set_key_materials_config(
+    grpc_tls_credentials_options* options,
+    grpc_tls_key_materials_config* config);
+
+/** Set grpc_tls_credential_reload_config field in credentials options
+    with the provided config struct whose ownership is transferred.
+    Both parameters should not be NULL.
```
ak800i, line 68: should -> must

ak800i:
> This needs to be updated to include the new certificate provider/distributor API that we've designed. @yihuazhang, can you please work with @ZhenLian and @jiangtaoli2016 to get this updated? Thanks!

Until this is updated I cannot proceed with creating a proposal to migrate https://github.com/grpc/grpc/pull/22906/files into the new cert provider/distributor API.
As far as I can currently see, the provider/distributor API proposed in this iteration is the same as it previously was (with callbacks that the user provides).

yihuazhang:
@markdroth, SG. Let me work with @ZhenLian and @jiangtaoli2016 to update it.

yihuazhang:
Had an offline discussion with @ZhenLian and @jiangtaoli2016. We agreed on that the credential reload API proposed in the current gRFC is significantly different from the new TLS credential reload API proposed by @ZhenLian and @jiangtaoli2016. There is also a chance that the server authorization check API will be simplified in a way that avoids  asynchronous callback. 

@ZhenLian will continuously work on this gRFC to reflect the aforementioned changes.

markdroth:
Closing this in favor of #205.

PR #100, opened 2018-09-18
A21-Service-Config-Error-Handling

yashykt:
Discussion thread at https://groups.google.com/forum/#!topic/grpc-io/bAxyse6c2eI

### DIFF
--- a/A21-Service-Config-Error-Handling.md
+++ b/A21-Service-Config-Error-Handling.md
@@ -38,12 +38,10 @@ service config update from the resolver. On the other hand, if it is a new
 client, then the client should fail to start, i.e., not act on the resolution
 result till a valid service config is received.
 
-If a default service config is provided (as a channel argument), service config
-resolution is enabled and the received service config is invalid, the client
-should NOT fall back on the default service config. It should as before, fail to
-start if it has not received a valid service config. On the other hand, if
-service config resolution is disabled, then the client should use the default
-service config.
+If a default service config is provided (as a channel argument), and the
+received service config is invalid, the client should fall back on the default
+service config if there was no previously accepted service config. It should as
+before, fail to start if the default service config has not been set.
 
 Based on this proposal, if new load balancing policies are to be introduced and
 the rollout is not controlled, then older clients will start discarding entire
--- a/A21-Service-Config-Error-Handling.md
+++ b/A21-Service-Config-Error-Handling.md
@@ -38,6 +38,31 @@ service config update from the resolver. On the other hand, if it is a new
 client, then the client should fail to start, i.e., not act on the resolution
 result till a valid service config is received.
 
+If a default service config is provided (as a channel argument), service config
+resolution is enabled and the received service config is invalid, the client
+should NOT fall back on the default service config. It should as before, fail to
+start if it has not received a valid service config. On the other hand, if
+service config resolution is disabled, then the client should use the default
+service config.
+
+Based on this proposal, if new load balancing policies are to be introduced and
+the rollout is not controlled, then older clients will start discarding entire
+service configs due to the load balancing policy being unknown. To make this
+rollout process easier, load balancing policy could be changed into a list of
+policies. The clients would iterate through the list and use the first policy
+that it knows. If none of the policies are known, the client would treat the
+service config as invalid.
+
+For example,
+
+“loadBalancingPolicy” : [
+    “NewUnknownPolicy”,
+    “RoundRobin”
+  ]
+
+In this case, since the client does not understand the first policy, it would
+use RoundRobin.
+
 ## Rationale
 
 By discarding invalid service configs instead of ignoring fields that have
--- a/A21-Service-Config-Error-Handling.md
+++ b/A21-Service-Config-Error-Handling.md
@@ -24,7 +24,7 @@ certain parts.
 
 
 ### Related Proposals:
-* [A2-Service Config via DNS](A2-service-configs-in-dins.md)
+* [A2-Service Config via DNS](A2-service-configs-in-dns.md)
 
 ## Proposal
 
--- a/A21-Service-Config-Error-Handling.md
+++ b/A21-Service-Config-Error-Handling.md
@@ -39,8 +39,11 @@ need to take on receiving a new config.
 configs are available, clients choose the service config to use based on the
 proposal
 [A2 Service Config in DNS](A2-service-configs-in-dns.md#canarying-changes).
-2. On choosing a service config, the client validates the service config based
-on the criteria outlined in the section below.
+Only the chosen service config affects the validation of the gRPC config, i.e.,
+if the chosen service config is found invalid while the other service configs
+were invalid, the gRPC config would still be considered invalid as a whole.
+2. The client validates the choice based on the criteria outlined in the
+section below.
 3. If the service config is valid, the client uses the configuration provided.
 4. If the service config is invalid, the clients should reject the received
 config in its entirety and do different things depending on its state and
@@ -50,8 +53,8 @@ configuration.
 	config. An empty service config is valid.
 	2. If the client has not previously selected a valid service config or no
 	service config, for example a new client, then:
-		1. It uses the default service config, if configured. Note that, an
-		empty default service config is a valid service config.
+		1. It uses the default service config, if configured and valid. Note
+		that, an empty default service config is a valid service config.
 		2. If it does not have a default service config, it should treat the
 		resolution attempt as having failed (e.g., for a polling-based
 		resolver, it should retry the query after appropriate backoff). In
@@ -63,12 +66,13 @@ configuration.
 A valid service config needs to be well formatted, i.e., the JSON representation
 of the service config needs to be in a valid JSON format. If the service config
 is well formatted, the client validates each field that it recognizes while
-ignoring the unknown fields, i.e., unknown fields do not affect validation. For
-known fields, valid values are defined by the field itself. Fields can accept
-lists where some of the values may be unknown. As long as there is atleast one
-value in the list which succeeds in getting parsed, the field will be
-considered to have a valid value. If any of the known fields fails in
-validation, the entire service config is deemed invalid.
+ignoring the unknown fields, i.e., unknown fields do not affect validation.
+
+Fields in the service config must clearly define the validation criteria for
+values. Fields can accept lists where some of the values may be unknown. As
+long as there is at least one value in the list which succeeds in getting
+parsed, the field will be considered to have a valid value. If any of the known
+fields fails in validation, the entire service config is deemed invalid.
 
 ## Rationale
 
@@ -84,16 +88,16 @@ Examples of invalid service configs:
 
 * Badly Formatted service config
 ```
-“serviceConfig” : {
-  “MethodConfig” : {
-    “Service // bad format
+"serviceConfig" : {
+  "MethodConfig" : {
+    "Service // bad format
   }
 ```
 
 * Fields with invalid values are not allowed
 ```
-“serviceConfig” : {
-  “loadBalancingPolicy” : “UnknownPolicy”
+"serviceConfig" : {
+  "loadBalancingPolicy" : “UnknownPolicy”
 }
 ```
 
@@ -103,10 +107,10 @@ it supports. If it doesn’t understand any of the policies, it is treated as
 invalid, and the entire service config is rejected.
 
 ```
-“serviceConfig” : {
-  “loadBalancingConfig” : [
-    “UnknownPolicy1” : {},
-    “UnknownPolicy2” : {}
+"serviceConfig" : {
+  "loadBalancingConfig" : [
+    "UnknownPolicy1" : {},
+    "UnknownPolicy2" : {}
   ]
 }
 ```
@@ -114,9 +118,9 @@ invalid, and the entire service config is rejected.
 Examples of valid service configs :
 * Unknown fields are allowed
 ```
-“serviceConfig” : {
-  “UnknownField” : “value”, // Field is unknown and hence ignored
-  “methodConfig” : {}
+"serviceConfig" : {
+  "UnknownField" : “value”, // Field is unknown and hence ignored
+  "methodConfig" : {}
 }
 ```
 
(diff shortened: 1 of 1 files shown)


A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,73 @@
...
+invalid service config or partially accept the service config while ignoring
+certain parts.
+
+
+### Related Proposals:
+* [A2-Service Config via DNS](A2-service-configs-in-dins.md)
+
+## Proposal
+
+If a resolved service config is badly formatted or contains invalid values for
```
carl-mastrangelo: There are a ton of failure cases here.  We need to get consensus about what values are errors so that we get consistent behavior.

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,73 @@
...
+[service config](https://github.com/grpc/grpc/blob/master/doc/service_config.md)
+is a mechanism that allows service owners to publish
+parameters that clients could use. If the received service config is malformed
+or has unknown fields or has invalid values, the client could either ignore the
+invalid service config or partially accept the service config while ignoring
+certain parts.
+
+
+### Related Proposals:
+* [A2-Service Config via DNS](A2-service-configs-in-dins.md)
```
dfawley: s/dins/dns/

A21-Service-Config-Error-Handling.md
yashykt: What values are considered invalid would depend on the field itself? For example, one proposal to make it easier to roll out new loadbalancing policies is to convert loadBalancingPolicy into a list, and the client would accept the first policy it recognizes. Only if it does not recognize any of the policies in the list would the value be considered invalid.

yashykt: Also, please continue this discussion at https://groups.google.com/forum/#!topic/grpc-io/bAxyse6c2eI

ejona86:
@yashykt, this looks like a previous proposal. Please ping the PR when it is up-to-date with the latest proposal.

yashykt:
@ejona86 From the meeting, we had discussed primarily 3 changes - 
1) Service config validation should ignore unknown fields.
2) Changes to loadbalancing policy to make it a list.
3) Behavior when the two service config channel arguments are set.

2 seemed like it should be a different proposal, while 3 doesn't seem to be a change, but I guess it is better to call it explicitly. I have added the two. Thanks for looking!

yashykt:
@markdroth PTAL. I have omitted details on the implementation details and the current state of the various fields with their invalid values. Let me know if those should be added here.

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,174 @@
...
+		the channel will remain in state TRANSIENT_FAILURE until a valid service
+		config is received.
+
+### Criteria to determine the validity of a service config
+
+A valid service config needs to be well formatted, i.e., the JSON representation
+of the service config needs to be in a valid JSON format. If the service config
+is well formatted, the client validates each field that it recognizes while
+ignoring the unknown fields, i.e., unknown fields do not affect validation. For
+known fields, valid values are defined by the field itself. If any of the known
```
ejona86: This may be misleading. We should mention the "repeated field"/list pattern, where certain values could fail to parse, but as long as one of them parses the config is considered well-formed. (I know that is discussed later, but that is in a "rationale" section and not the "criteria" section)

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,174 @@
...
+fields fails in validation, the entire service config is deemed invalid.
+
+## Rationale
+
+By discarding invalid service configs instead of ignoring fields that have
+invalid values, we prevent clients from resorting to using default values which
+might be unwanted for certain systems. The clause that existing clients continue
```
ejona86: clause? I don't know what it is meaning

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,174 @@
...
+To implement the above spec -
+1. The resolver needs to be provided a parsing method(s) that it can run for a
+service config. This function will be provided to the resolver at instantiation
+time along with the default service config (if registered). This method will
+return a parsed form of the service config, or an error indicating that parsing
+failed.
+2. When a resolver gets an update, it runs the parser on this service config.
+	1. If parsing succeeds, it saves this service config and returns it to
+        the channel to apply.
+	2. If parsing fails, it provides a previously saved service config
```
ejona86: In Java (and it may be a good idea for other languages), we're planning on giving the full parse results (parsed_object, error) to the channel. That would allow the channel to use the old config. (And based on my comment above, use the old addresses+old config.)

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,174 @@
...
+[here](A2-service-configs-in-dns.md#canarying-changes).
+2. On choosing a service config, the client validates the service config based on
+the criteria outlined in the section below.
+3. If the service config is valid, the client uses the configuration provided.
+4. If the service config is invalid, the clients should reject the received
+config in its entirety and do different things depending on its state and
+configuration.
+	1. If it previously received an update with a valid service config, it should
+	continue using the configuration from that config.
+	2. If this is a new client or a client which has not yet received a valid
```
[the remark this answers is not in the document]
yashykt: Accepting the rephrasing suggestion.

I don't think failing the entire update works. In the scenario that we have a bad service config, and the addresses change later, the entire system would go down even with a default service config configured.

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,174 @@
+Service Config Error Handling
+----
+* Author: yashkt
+* Approver: markdroth
+* Status: Draft
+* Implemented in:
+* Last updated: 09-17-2018
```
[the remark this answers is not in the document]
yashykt: done

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,174 @@
...
+In brief, if a service config (JSON formatted) received is badly formatted or
+contains invalid values for known fields, the client should ignore that service
+config in its entirety. This section enumerates the exact steps that clients
+need to take on receiving a new config.
+
+### Behavior on receiving a new gRPC Config
+
+1. A gRPC config can have multiple service configs. When multiple service configs
+are available, clients choose the service config to use based on the proposal
+[here](A2-service-configs-in-dns.md#canarying-changes).
```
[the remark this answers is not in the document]
yashykt: done

A21-Service-Config-Error-Handling.md
yashykt: added

A21-Service-Config-Error-Handling.md
yashykt: removed

A21-Service-Config-Error-Handling.md
yashykt: view comment above this failing when new addresses are received with a bad service config and the entire system going down even when a default service config is provided.

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,179 @@
...
+## Proposal
+
+In brief, if a service config (JSON formatted) received is badly formatted or
+contains invalid values for known fields, the client should ignore that service
+config in its entirety. This section enumerates the exact steps that clients
+need to take on receiving a new config.
+
+### Behavior on receiving a new gRPC Config
+
+1. A gRPC config can have multiple service configs. When multiple service
```
carl-mastrangelo: We have been calling this the Service Config "choice".

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,179 @@
...
+configs are available, clients choose the service config to use based on the
+proposal
+[A2 Service Config in DNS](A2-service-configs-in-dns.md#canarying-changes).
```
carl-mastrangelo: Could you clarify explicitly what happens if there are multiple applicable service configs, and one is valid but not the other?

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,179 @@
...
+3. If the service config is valid, the client uses the configuration provided.
+4. If the service config is invalid, the clients should reject the received
+config in its entirety and do different things depending on its state and
+configuration.
+	1. If it previously had selected a valid service config or no service
+	config,	it should continue using that previous configuration from that
+	config. An empty service config is valid.
+	2. If the client has not previously selected a valid service config or no
+	service config, for example a new client, then:
+		1. It uses the default service config, if configured. Note that, an
```
carl-mastrangelo: Can this be clarified as well?  If the service config is passed as a channel arg (or via some, non-resolver based way), what should the error handling look like?  This is also the same scenario as this bullet.

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,179 @@
...
+service configs prevent an invalid service config from bringing down the
+entire system, while newer clients which do not have a previously received
+valid service config fail to start, avoiding new traffic for the duration that
+the service is set with an invalid config.
+
+Examples of invalid service configs:
+
+* Badly Formatted service config
+```
+“serviceConfig” : {
```
carl-mastrangelo: It would be clearer if you avoided using unicode quotes here, which also make this invalid.

A21-Service-Config-Error-Handling.md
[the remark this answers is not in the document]
ejona86: Sorry, I agree that we'd have the name resolver return the tuple of (list of addresses, magic cookie for parsed service config). The Channel however may decide to fail the entire update if there isn't a default service config.

A21-Service-Config-Error-Handling.md
```diff
@@ -0,0 +1,179 @@
...
+config provided as a channel argument, the client should fail to start and keep
+waiting for a valid service config. Note that the default service config
+being set to an empty service config would be a valid fallback.
+
+## Implementation
+
+To implement the above spec -
+1. The resolver needs to be provided a parsing method(s) that it can run for a
+service config. This function will be provided to the resolver at instantiation
+time along with the default service config (if registered). This method will
```
markdroth: We will no longer be passing the default service config to the resolver; only the channel needs to see that.

A21-Service-Config-Error-Handling.md
markdroth: We should change this to say that if parsing fails, the resolver just returns the parse error alongside the addresses.

A21-Service-Config-Error-Handling.md
yashykt: If the resolver does not see the default service config, then we would need a different way to validate the default service config

A21-Service-Config-Error-Handling.md
yashykt: Done!

A21-Service-Config-Error-Handling.md
yashykt: The default service config also needs to be validated if that's what you mean.

A21-Service-Config-Error-Handling.md
```diff
@@ -50,8 +53,8 @@ configuration.
 	config. An empty service config is valid.
 	2. If the client has not previously selected a valid service config or no
 	service config, for example a new client, then:
-		1. It uses the default service config, if configured. Note that, an
-		empty default service config is a valid service config.
+		1. It uses the default service config, if configured and valid. Note
+		that, an empty default service config is a valid service config.
```
markdroth: Grammar nit: The comma should be removed here.

A21-Service-Config-Error-Handling.md
yashykt: I was hoping the editor would figure that out. Done!

A21-Service-Config-Error-Handling.md
yashykt: Removed!

A21-Service-Config-Error-Handling.md
```diff
@@ -63,12 +66,13 @@ configuration.
...
 is well formatted, the client validates each field that it recognizes while
-ignoring the unknown fields, i.e., unknown fields do not affect validation. For
-known fields, valid values are defined by the field itself. Fields can accept
-lists where some of the values may be unknown. As long as there is atleast one
-value in the list which succeeds in getting parsed, the field will be
-considered to have a valid value. If any of the known fields fails in
-validation, the entire service config is deemed invalid.
+ignoring the unknown fields, i.e., unknown fields do not affect validation.
+
+Fields in the service config must clearly define the validation criteria for
```
markdroth: The wording is still imprecise here.  What does it mean for a field to define something?  I think this needs to say something like "Future gRFCs that add new service config fields must define the validation criteria for those fields."

A21-Service-Config-Error-Handling.md
```diff
@@ -63,12 +66,13 @@ configuration.
...
+values. Fields can accept lists where some of the values may be unknown. As
```
markdroth: Suggest more precise wording: "In a case where different clients may support different values for a field, a gRFC may define a field to be a list of values, where each client will use the first value in the list that it understands."

yashykt:
@markdroth I will send out PRs for the already defined fields that lack clarity on validation.

A21-Service-Config-Error-Handling.md
yashykt: Removed!

A21-Service-Config-Error-Handling.md
yashykt: Suggestion accepted!

A21-Service-Config-Error-Handling.md
yashykt: Suggestion accepted!

yashykt:
PRs #128, #129, #130 and https://github.com/grpc/grpc/pull/18225 have been  merged to update existing service config fields with the validation criteria

A21-Service-Config-Error-Handling.md
yashykt: I thought it better to be verbose here, and be clear about what needs to be done from the point after receiving the grpc config

yashykt:
Thanks for all the reviews!

MellyReg (approved):

PR #103, opened 2018-09-20
L37: Go Custom Transports

dfawley:


dfawley:
cc @ejona86 @markdroth

L37-go-custom-transports.md
```diff
@@ -0,0 +1,415 @@
...
+
+  This does not allow for a more efficient in-memory transport that is capable
+  of copying proto messages rather than serializing and deserializing.
+
+- Inject the client `TransportBuilder` directly into the `ClientConn`, rather
+  than allowing the resolver and balancer to influence it.
+
+  This would have a simpler API, however, it would require each `ClientConn` to
+  use only a single transport.  The proposed design enables a `ClientConn` with
+  different types of connections.
```
jhump: The value of this is unclear. Under what cases would a single connection want different types of connection?

The only thing I can imagine might be to provide both in-process and remote connections for a service that is exposed both locally and by the server's peers. But that would only be useful if there were a rich API for affinity -- such that some requests (based on affinity keys) might be routed locally over an in-process transport and others could be routed to a remote server. But that could also easily be implemented as a "composite channel" abstraction -- where a channel delegates to a `ClientConn` with in-process connections for some requests but to another `ClientConn` for others.

Perhaps a simpler API would be more valuable?

L37-go-custom-transports.md
```diff
@@ -0,0 +1,415 @@
...
+
+The main goal of this proposal is to allow for users to inject custom transports
+for clients and servers.  Some additional design goals of this proposal:
+
+1. Transport and stream APIs should expose only gRPC semantics as defined in the
+   [Call Semantics Specification](https://github.com/grpc/grpc/pull/15460) and
+   hide HTTP/2 semantics and grpc wire protocol semantics).
+1. Transports should be message-based, not byte-based, to facilitate an
+   in-process transport that uses `proto.Clone()` for efficiency.
+   - Note that retry support requires byte-based support from all transports,
```
jhump: I'm curious about this constraint. Could you point to what in the streaming API imposes this constraint?

dfawley: Retry for streaming RPCs necessitates the ability to replay multiple sent messages.

Spec: https://github.com/grpc/proposal/blob/master/A6-client-retries.md#memory-management-buffering
grpc-go's buffer: https://github.com/grpc/grpc-go/blob/master/stream.go#L361

In the Go streaming API, grpc-go does not acquire ownership of the messages when users pass them to `SendMsg`.  The only way we can safely reference the messages after returning from `SendMsg` is if we serialize them first.

jhump: Hmm. I'm still a little confused. If the transport copies the message (like as done in an in-process transport), then it owns its copy and does not need to resort to serialization to support retries. What am I missing?

jhump: Actually I think I get it. The issue is that the stream abstraction for the call is at a different layer than the stream abstraction for the transport. So the message will come into the call layer, and then need to be serialized in order to be cached for retries, before it gets relayed to the transport-layer (which could happen multiple times, like when the call is actually retried).

It think there must be an alternate solution to this, especially since we're talking about a brand new API (i.e. if designed right, seems like it should be able to facilitate the desired behavior: caching messages in a safe way without requiring the call/channel layer to serialize to bytes first).

dfawley: > The issue is that the stream abstraction for the call is at a different layer than the stream abstraction for the transport

Right.  The transport stream only handles a single attempt of the RPC.  Even if it were to make a copy of the message, we will kill the stream it when it fails, so the copy would be lost.  It's possible/desired to use a different transport (even if it's the same _type_ of transport) when we retry.

> It think there must be an alternate solution to this, especially since we're talking about a brand new API

The new API is only between grpc and the transport.  The problem comes about because of the existing API between the user and grpc.

Another option could be to check for and use a `Clone(interface{}) interface{}` method on the marshaller when caching messages for retry.  However, we can't require this, to preserve backward compatibility.  So I think the transport would still either need to support byte-based transmission (which seems pretty trivial to implement) or just fail the stream when a serialized message is encountered.

This has made me notice that I am missing a way to access the raw message passed to `Send`/`Recv` from inside the transport.  I believe a method on `IncomingMessage` and `OutgoingMessage` to access it (e.g. `RawMessage() interface{}`) would be the obvious way to do that.

asim:
If you're looking for inspiration around pluggable transports, clients, servers, etc you should take a look at go-micro https://github.com/micro/go-micro. We support the ability to do grpc as codec, client/server and transport.

PR #104, opened 2018-09-21
binarylog fix: conflicting rules are not supported

menghanl:


A16-binary-logging.md
```diff
@@ -270,6 +270,8 @@ order to log certain required headers.
     i.e. `-<service>/*` is not supported.
 *   Specifying a method while wildcarding the service is not supported, i.e.
     `*/<method>` is not supported.
+*   Conflicting rules are not supported, i.e. an exact method match and
+    a negation for the same method "/m{h:12;m:23},-s/m".
```
zpencer: Also conflicts:
- `/m{h:12;m:23},s/m{h}` do we log with truncation or log only header?
- `/m{h:12;m:23},s/m{h:1,m:2}` which truncation length do we use?

I think we can treat this as a fatal error, because this is can only be turned on at startup.

zpencer: Maybe it would make it easier if we disallowed duplicates entirely, then we don't need to special case duplicates that are not conflicting, e.g. `s/m{h:1,m:1},s/m{h:1,m1}`

menghanl: This sounds good to me.

I assume two negations for the same method would be find. I also added this to the spec. Let me know if you think otherwise.

menghanl: Are suggesting we allow duplicates as `s/m{h:1;m1},s/m{h:1;m1}`?

Do we want to also allow `s/m,s/m{h;m}` then?

zpencer: You're right the other equivalent way to express it adds complexity. Disallowing all duplicates including identical duplicates seems the easiest to me.

menghanl: What about identical duplicate negations?

It's easy to allow them, but maybe better to disallow them for consistence?

zpencer: I agree, disallowing them makes things easier to reason about.

PR #106, opened 2018-09-26
L39: Remove grpc_use_signal from core surface API

vjpai:
This core surface API is no longer useful and doesn't seem to be used.


vjpai:
Merging since no objections seen.

PR #109, opened 2018-10-04
Small Update to Channel Tracing Spec

ncteisen:


A3-channel-tracing.md
```diff
@@ -30,6 +30,8 @@ The data from _Channel trace_ will be exposed via the [channelz service](A14-cha
 
 Since the tracing objects may consume large amounts of space, care must be taken to prevent the _Channel trace_ from using too many resources. gRPC implementations must allow limiting the number of _Trace events_ retained to a per channel or subchannel maximum. Once the maximum is reached, adding a new _Trace event_ will cause oldest _Trace event_ to be removed. If the maximum number of trace events is set to zero, then channel tracing will be disabled.
 
+Implementations MAY choose to restrict _Channel trace_ further. For example, a memory sensitive implementation can additionally provide a configurable maximum for the amount of memory used by the _Trace events_.
```
markdroth: Why make this a separate, optional knob?  Why not simply replace the existing knob of number of trace events with a new knob that says amount of memory, and change all of our implementations to do that?

ncteisen: Tracking memory in that way is very hard for Java and Go. Also, those languages are less sensitive to memory usage, and so a simple cap of number of trace events should suffice

This comes from offline discussion with @carl-mastrangelo

carl-mastrangelo: @markdroth I was under the impression this spec is meant as the API contract for channel traces, rather than implementation guidance.   From that point of view, the relevant thing is that Channel Traces MAY drop records if they want to limit resource usage.   That makes it clear that impls may impose memory constraints, and consumers may not be getting a comprehensive trace history.  

If we start talking about bytes, it would become impl specific fast (should we count tcmalloc overhead?)  and imply excessive strictness (should we acquire a lock before tracing to avoid going over?)

markdroth: I'm not suggesting that we make this spec any more implementation-specific.

What I'm saying is that there's a discepancy between what this spec now says and what @ncteisen is implementing in C-core.  This spec says "gRPC implementations must allow limiting the number of _Trace events_ retained to a per channel or subchannel maximum" and "Implementations MAY choose to restrict _Channel trace_ further", but the implementation in C-core is doing only the latter.

I would be fine with saying something like "Implementations should provide some control for limiting the amount of memory used for channel trace events, such as a max number of trace events per node or the max amount of memory used".

ncteisen: I updated the phrasing to be more in line with this discussion. Ket me know what you think

PR #110, opened 2018-10-09
A23: RPC Representation

carl-mastrangelo:
cc: @yang-g @dfawley @ejona86


carl-mastrangelo:
The main use case is for gRPC users to diagnose their own issues easier, or failing that, provide us better information when filing bug reports.

PR #111, opened 2018-10-16
Update Implementation Status of Retries

ncteisen:
cc @wenbozhu

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -3,7 +3,7 @@ gRPC Retry Design
 * Author(s): [Noah Eisen](https://github.com/ncteisen) and [Eric Gribkoff](https://github.com/ericgribkoff)
 * Approver: a11r
 * Status: Ready for Implementation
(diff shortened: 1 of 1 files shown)


A6-client-retries.md
```diff
@@ -3,7 +3,7 @@ gRPC Retry Design
 * Author(s): [Noah Eisen](https://github.com/ncteisen) and [Eric Gribkoff](https://github.com/ericgribkoff)
 * Approver: a11r
 * Status: Ready for Implementation
-* Implemented in:
+* Implemented in: Go and Java (C Core support in progress)
```
markdroth: This is not entirely accurate.  The retry code is done in Java and Go, but hedging is not yet done, and that's part of this design.  And none of this code has really gotten any real-world testing yet.

For now, I think it's probably better to classify it as in progress in all langauges.

ncteisen: Done

PR #113, opened 2018-10-18
Rename language proposal files for consistency

dfawley:
Sorry for the churn; I do think this is useful to make it easier to scan the list of proposals in our repo and understand what you're looking at.


dfawley:
@muxi - Good point; done.

dfawley:
cc @AspirinSJL @apolcyn @ZhouyihaiDing @mehrdada, who have pending L proposals whose filenames may need to be changed if these conventions are accepted.

dfawley:
I'm fine with that.  I didn't like the "++" either, personally.  Done.

This makes L22 become "L22-cpp-change-grpcpp-dir-name.md".  Is that OK?  It *is* the *new* name of the directory, so I think it's OK - just thought I'd mention it.

carl-mastrangelo:
@dfawley I appreciate the intent, but I strongly urge you to not submit this PR.  In several places we link to the doc on github hub, and this would break all those links.  I don't think breaking links is worth the consistency of the names.

dfawley:
@carl-mastrangelo how would you feel if we left behind redirects, potentially for a limited time period?

dfawley:
I was thinking something like this:

```
This file has been moved.  The new location is [Lxx-xxx-xxxxx.md](Lxx-xxx-xxxxx.md).
```

dfawley:
I did a google search for every filename I am proposing changing.  I found only one hit for any of them*, and it was from here:

https://groups.google.com/forum/#!topic/grpc-io/UFx2vxmI1xY

* - I also found 2 forks of the grpc/proposal repo, and obviously the main repo and the PR to add the file in the main repo.

carl-mastrangelo:
ok.   If yo can show there are no links to break, I guess that's okay.  But that said, how are you going to prevent this situation from happening in the future?  We'll just end up here again and youll have to look through all the links hoping none of them are used.

dfawley:
> But that said, how are you going to prevent this situation from happening in the future?

Short of writing a bot, we can't.  The goal of this PR isn't to enforce rules, it's to formally declare the rule and fix the existing violations.  I think with the rule documented and the existing examples setting a good precendent, we should be better off moving forward.

carl-mastrangelo:
@dfawley I don't want to belabor this, bu I don't think your pattern will stick.  When creating a PR, it's not obvious that the file name doesn't match the rest of the directory.  So by the time it's merged, the damage is done.  Worse, the author won't come back to look at it and see they made a mistake.   Your goals are right with this PR, but it won't have the effect you want.

dfawley:
If we find people are unable to follow the established convention after this, we can always create a bot.  There are plenty of other problems with this repo, like discovering which number to use for your proposal (many have gotten this wrong).  Let's try to make some progress where we can, and not try to fix all the problems at once.

dfawley:
@vjpai can you PTAL please?

vjpai:
Thanks for doing this!

PR #114, opened 2018-10-24
Channelz: Add GetServer Method

ncteisen:
cc @lyuxuan @menghanl @zpencer

### DIFF
--- a/A14-channelz.md
+++ b/A14-channelz.md
@@ -424,8 +424,12 @@ the previous results in the subsequent query.
 
 This method returns all servers in the process.  The usage is identical to GetTop
 Channels, except that servers are returned.  Users are expected to make this
-query first when investigating servers.  Since servers are not hierarchical,
-there is no GetServer method.
+query first when investigating servers.
+
+## GetServer
+
+This method returns a single Server request by its id.  If the server is not
+present, a `NOT_FOUND` status is returned.
 
 ## GetChannel
 


zpencer:
Is this so we can show a single server page that can be refreshed without pulling a page of results?

ncteisen:
Yes. It also allows the Servers landing page to only show non-verbse data like the counts. then a user can click to the individual server page to see the full output with trace and lists of listen ockets

zpencer:
How about folding this into GetServerRequest?
* uint32 limit
* bool verbose

I can imagine a UI wanting to control how many max results to show in a single page.

markdroth:
Let's take a step back and consider how the RPC methods map to the UI we would want.

The top level servers page would be generated from the GetServers call.  It would contain a table with columns "name" (which would be a link to the sub-page for that individual server), "channel trace", "num calls started", "num calls succeeded", "num calls failed", and "last call started timestamp".  That seems mostly fine and doesn't seem to require a separate GetServer method.  (One minor nit is that the channel trace info might be too large to fit into a table with the other info, so we might want to move it to the server-specific sub-page, but that's something we can think about later.)

The problem comes in for the sub-page for the individual server, which will wind up showing a table of the individual sockets for the server.  With our current UI, this would be generated via the GetServerSockets method.  However, that method basically returns nothing but the name of the socket; additional information needs to be obtained via a GetSocket call for each socket.  So the current API offers two choices:

1. We can have a one-to-one mapping from UI pages to Re.PC methods.  The sub-page for the idividual server would be based on the GetServerSockets response, so it would just be a list (not a table) of sockets by name, where each socket name is a link to a sub-page for details about the socket.  That socket details page would then contain the data from the GetSocket method.

2. We can combine the GetServerSockets and GetSocket calls into a single UI page.  The page would first call GetServerSockets to get a list of sockets and then call GetSocket on each one.  The page would display a table containing all information about each socket.

The problem with approach 1 is that it's not a very useful UI; humans will want to see more information about each socket in order to decide which ones they may want to drill down into.  Conversely, the problem with approach 2 is that it's too expensive; IIRC (but please correct me if I don't), getting the socket data from the kernel is expensive, and we don't want to do that until the user has decided that they want more detailed information about a particular socket.

An alternative would be to split the socket information into two parts.  If the GetServerSockets response included the information that is cheap but not the information that is expensive, then we could have the page for the individual server display a table of the cheap information, with a link to a more detailed page.  The more detailed page for a given socket would be fed from the GetSocket call, and that would include the expensive information.

Thoughts...?

ncteisen:
The one part that you didn't mention in the GetServers page is that each server will also have it's data for listen_sockets. We don't expect a ton of those per server, but it is not bounded by the spec.

I still think we should support a GetServer method for the following reason:

SRE lands on Servers page, sees the server they care about. They want to click on to a page that has all the information about that server and only that server. GetServerSockets does not fulfill this because it only has info about the server sockets, not the server itself (ie the call counts and trace).

And the SRE might want to refresh periodically.

Separately, there is the issue you bring up about heavy vs cheap socket data. My preference would be as you suggest, to create a new message type like `FastSocketData` or something. Then GetServerSockets returns a repeated of that so we can populate stream counts and timestampes. Then the user would make a GetSocket to do the full data with sockopts and all

ncteisen:
@zpencer also brings up a point we have mentioned before of GetTopChannels and GetServers supporting a limit. I think this is a good idea from UI perspective

zpencer:
> We can have a one-to-one mapping from UI pages to RPC methods.

I would like to avoid coupling UI pages with RPCs, because we may want to tweak the UI without changing the service proto.

> We can combine the GetServerSockets and GetSocket calls into a single UI page.

When a user is drilling down to servers and server sockets, they might not care about TCP options yet. They probably do want some relatively high level info at the abstraction layer of gRPC. 

> An alternative would be to split the socket information into two parts. If the GetServerSockets response included the information that is cheap but not the information that is expensive, then we could have the page for the individual server display a table of the cheap information, with a link to a more detailed page. The more detailed page for a given socket would be fed from the GetSocket call, and that would include the expensive information.

Right, one of the design goals of the service definition is to avoid fetching expensive data that the client may not care about. We can achieve the above by moving `repeated SocketOption option` out of `SocketData` and into `GetSocketResponse`. Then `GetServerSocket` can return a cheaply created `Socket` rather than the `SocketRef`.

ncteisen:
I don't like the idea of a verbose bool though, that seems clunky. If we decide to support multiple options of verbosity, let's declare that in the messages themselves. Aka being able to query for QuickSocket or QuickServer or something

markdroth:
@ncteisen wrote:
> The one part that you didn't mention in the GetServers page is that each server will also have it's data for listen_sockets. We don't expect a ton of those per server, but it is not bounded by the spec.

That's true.  But in practice, there are unlikely to be more than a few of these per server.  I suspect this is probably in the same category as the server's channel trace data: it's something that should ideally go on the detailed page for the individual server, but it's not a big deal if it's not.

> I still think we should support a GetServer method for the following reason:
> 
> SRE lands on Servers page, sees the server they care about. They want to click on to a page that has all the information about that server and only that server. GetServerSockets does not fulfill this because it only has info about the server sockets, not the server itself (ie the call counts and trace).
> 
> And the SRE might want to refresh periodically.

I am not opposed to adding this if everyone agrees that it's useful.  I suppose if nothing else, it would proivide an easy way to move the channel trace and listen socket data to the detail page for the individual server.

> @zpencer also brings up a point we have mentioned before of GetTopChannels and GetServers supporting a limit. I think this is a good idea from UI perspective

Yes, I agree.  Pretty much any RPC that returns a repeated field should allow the client to specify how many entries to return.  If someone wants to volunteer to make that change, that would be great.


@zpencer writes:
> > We can have a one-to-one mapping from UI pages to RPC methods.
> 
> I would like to avoid coupling UI pages with RPCs, because we may want to tweak the UI without changing the service proto.

I agree that we should not tie the RPC API too closely to any particular UI.  However, I think that we should structure the RPC API in a way that it's flexible enough to allow UI development without being too cumbersome (i.e., without requiring the UI to grab a lot of data it doesn't need in order to get the data it does need, and without requiring the UI to send a large number of RPCs to get all the data it does need).

> > We can combine the GetServerSockets and GetSocket calls into a single UI page.
> 
> When a user is drilling down to servers and server sockets, they might not care about TCP options yet. They probably do want some relatively high level info at the abstraction layer of gRPC.

Agreed.

> > An alternative would be to split the socket information into two parts. If the GetServerSockets response included the information that is cheap but not the information that is expensive, then we could have the page for the individual server display a table of the cheap information, with a link to a more detailed page. The more detailed page for a given socket would be fed from the GetSocket call, and that would include the expensive information.
> 
> Right, one of the design goals of the service definition is to avoid fetching expensive data that the client may not care about. We can achieve the above by moving `repeated SocketOption option` out of `SocketData` and into `GetSocketResponse`. Then `GetServerSocket` can return a cheaply created `Socket` rather than the `SocketRef`.

Sounds like the right general idea, but I'm not sure about the specifics.  Maybe we just change `GetServerSockets` to return `Socket` instead of `SocketRef`, but make it clear that the returned data will not include anything expensive (i.e., the `SocketOption` field will not be populated), and that callers need to call `GetSocket` to get the more detailed data?  That feels a little messy, but breaking up the existing `SocketData` message seems a little more invasive.

Do we need to do this in a backward-compatible way, or can we still change the API at this point?  If we have to be backward-compatible, then this could get messy.

menghanl:
> > The one part that you didn't mention in the GetServers page is that each server will also have it's data for listen_sockets. We don't expect a ton of those per server, but it is not bounded by the spec.
> 
> That's true. But in practice, there are unlikely to be more than a few of these per server. I suspect this is probably in the same category as the server's channel trace data: it's something that should ideally go on the detailed page for the individual server, but it's not a big deal if it's not.

One reason to show listen sockets in the servers page is that, it can be used to identify the server.

We don't have a good name for each server now. So the servers page will show the server ID (just numbers), and information like calls started/succeeded/failed. If the user is interested in a particular server, it's hard to find it.

markdroth:
> One reason to show listen sockets in the servers page is that, it can be used to identify the server.
> 
> We don't have a good name for each server now. So the servers page will show the server ID (just numbers), and information like calls started/succeeded/failed. If the user is interested in a particular server, it's hard to find it.

That's a really good point.  From the user's perspective, the real identity of the server is what addresses it's listening on.

zpencer:
> Do we need to do this in a backward-compatible way, or can we still change the API at this point? If we have to be backward-compatible, then this could get messy.

I think we can still break the API, what do you think @carl-mastrangelo ?

I can add the limit to the top level pagination calls.

carl-mastrangelo:
> I think we can still break the API, what do you think @carl-mastrangelo ?

I'm not an API breaker, and I think people looking to the gRPC for best practices will see what we do and follow.

> I am not opposed to adding this if everyone agrees that it's useful. 

As a performance optimization, I don't think this is a valid issue. All calls were designed to be very cheap,  and the pagination API let's the server dynamically drop load if it wants.  We can add more methods if the ones we have prove to be slow.

From an API POV, having a method that has an error of NOT_FOUND if the requested server id is gone would be useful.  GetServer would make the API more consistent IMO, since there are getters for the other types.

markdroth:
@lyuxuan wrote:
> I think the same problem applies for the subchannel refs and nested channel refs returned by the `GetChannel` call. User will need to click on each subchannel or nested channel id to look inside and then figure out whether it is relevant. However, it may be less of a problem since there's usually a smaller number of subchannels under a channel than that of sockets under a server.

I think channels and subchannels are different from servers, because the name of the channel or subchannel (the human-readable field included in the `ChannelRef` or `SubchannelRef`) will be enough for the user to identify it.  For channels, that name should be set to the target used to create the channel.  For subchannels, it should be set to the address of the subchannel (or its name, if known).  In contrast, servers don't have any such unique name, since they're only really identified by the addresses that they're listening on.

markdroth:
Should we have a meeting to try to resolve these issues?  If so, someone should throw something on calendar.

PR #118, opened 2018-12-08
L44: New RPC Status API for Python

lidizheng:
gRFC: https://github.com/lidizheng/proposal/blob/python-rich-status/L44-python-rich-status.md
Discussion Thread: https://groups.google.com/forum/#!topic/grpc-io/8Ys6ba8gpLc

### DIFF
--- a/L44-python-rich-status.md
+++ b/L44-python-rich-status.md
@@ -9,26 +9,26 @@ New RPC Status API for Python
 
 ## Abstract
 
-Current design of status API doesn't support rich status, also the concept of status "details" doesn't comply with gRPC spec. So, after discussion with @gnossen and @ericgribkoff, we think we should add a new set of status APIs to correct them.
+The current design of status API doesn't support rich status. Also the concept of status "details" doesn't comply with the gRPC spec. So, after discussion with @gnossen and @ericgribkoff, we think we should add a new set of status APIs to correct them.
 
 
 ## Background
 
-The gRPC [Spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses) defined two trailing data to indicate the status of a RPC call `Status` and `Status-Message`. But status code and a text message themselves are insufficient to express more complicate situation like the RPC call failed because of quota check of certain policy, or the stack trace from the crashed server, or obtain retry delay from the server (see [error details](https://github.com/googleapis/api-common-protos/blob/master/google/rpc/error_details.proto)).
+The gRPC [Spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses) defined two trailing data to indicate the status of an RPC call `Status` and `Status-Message`. However, status code and a text message themselves are insufficient to express more complicated situation like the RPC call failed because of quota check of specific policy, or the stack trace from the crashed server, or obtain retry delay from the server (see [error details](https://github.com/googleapis/api-common-protos/blob/master/google/rpc/error_details.proto)).
 
-The one major use case of this feature is Google Cloud API. The Google Cloud API needs it to transport rich status from server to client, also they are the owner of those proto files.
+The one primary use case of this feature is Google Cloud API. The Google Cloud API needs it to transport rich status from the server to the client. Also, they are the owner of those proto files.
 
-So, the gRPC team introduced a new internal trailing metadata entry `grpc-status-details-bin` to serve this purpose. It takes in serialized proto `google.rpc.status.Status` message binary and transmits as a binary metadata. However, since gRPC is a ProtoBuf-agnostic framework, it can't enforce either the content of the binary data be set in the trailing metadata entry, nor the consistency of status code/status message with the additional status detail proto message. The result of this unsolvable problem is that for three major implementation of gRPC we have, each of them are slightly different about this behavior.
+So, the gRPC team introduced a new internal trailing metadata entry `grpc-status-details-bin` to serve this purpose. It takes in serialized proto `google.rpc.status.Status` message binary and transmits as binary metadata. However, since gRPC is a ProtoBuf-agnostic framework, it can't enforce either the content of the binary data be set in the trailing metadata entry, nor the consistency of status code/status message with the additional status detail proto message. The result of this unsolvable problem is that for three major implementations of gRPC we have, each of them is slightly different about this behavior.
 
 ### Behavior Different Across Implementations
 
-C++ implementation tolerates anything to be put in the `grpc-status-details-bin` entry; Java will check if the code of the proto status is matching the status code of the RPC call; Golang will enforce the both code and message of the proto status and the RPC call be the same.
+C++ implementation tolerates anything to be put in the `grpc-status-details-bin` entry; Java checks if the code of the proto status is matching the status code of the RPC call; Golang enforces both code and message of the proto status, and the RPC call be the same.
 
 So, which paradigm should Python follow?
 
 ### The Proto of Status
 
-The proto of status is well-defined and used in numerous framework. Here is the definition of `Status`, for full version see [status.proto](https://github.com/googleapis/api-common-protos/blob/master/google/rpc/status.proto).
+The proto of status is well-defined and used in many frameworks. Here is the definition of `Status`, for full version see [status.proto](https://github.com/googleapis/api-common-protos/blob/master/google/rpc/status.proto).
 ```proto
 message Status {
   int32 code = 1;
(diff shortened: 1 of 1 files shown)


L44-python-rich-status.md
```diff
@@ -0,0 +1,394 @@
...
+                detail="Can't recognize this argument",
+            )
+        )
+        rich_status = grpc_status.status_pb2.Status(
+            code=grpc_status.code_pb2.INVALID_ARGUMENT,
+            message='API call quota depleted',
+            details=[detail]
+        )
+
+        context.set_status(*rpc_status.convert(rich_status))
```
gnossen: It's a little bit odd to have to use a splat here. Any way we can eliminate it?

lidizheng: I don't have better idea than this one. If we are going to make two variable mandatory and one optional, this is my best solution. Also, I don't like to make a function accept different set of arguments.

An alternative way can satisfy both criteria is make the function signature like this:
```Python
def set_status(self, status_tuple):
```
```Python
# Usage for non-grpcio-status users

context.set_status((code, message))
```
But the additional parentheses doesn't look good.


Do you have any insight?

gnossen: My thought was to make the function accept different types of arguments and determine at runtime how to unpack them. But there are certainly arguments against that type of approach. The splat should be fine.

lidizheng:
@srini100 I have updated the gRFC. PTAL

lidizheng:
This API seems break opentracing-contrib/python-grpc, but they got it fixed by https://github.com/opentracing-contrib/python-grpc/pull/17.

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+* Author(s): lidizheng
+* Approver: ericgribkoff
+* Status: Draft
+* Implemented in: Python
+* Last updated: 12-10-2018
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/8Ys6ba8gpLc
+
+## Abstract
+
+The current design of status API doesn't support rich status. Also the concept of status "details" doesn't comply with the gRPC spec. So, after discussion with @gnossen and @ericgribkoff, we think we should add a new set of status APIs to correct them.
```
ericgribkoff: AFAIK rich status is not a term used elsewhere, and would need to be defined. I would prefer something more general here, such as:

> The current design of the gRPC Python servicer API for RPCs terminated with a non-OK status does not easily allow pairing the failure with associated additional information not expressible by the `grpc-status` and `grpc-message` fields. This proposal adds a rich `Status` class to gRPC Python that allows coupling the failure's status code, error string, and arbitrary additional information about the failure (sent as trailing metadata) in a unified interface.

(and please remove the second sentence about "after discussions with")

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+
+
+## Background
+
+The gRPC [Spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses) defined two trailing data to indicate the status of an RPC call `Status` and `Status-Message`. However, status code and a text message themselves are insufficient to express more complicated situation like the RPC call failed because of quota check of specific policy, or the stack trace from the crashed server, or obtain retry delay from the server (see [error details](https://github.com/googleapis/api-common-protos/blob/master/google/rpc/error_details.proto)).
```
ericgribkoff: `Status` -> `grpc-status`
`Status-Message` -> `grpc-message`

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+
+The one primary use case of this feature is Google Cloud API. The Google Cloud API needs it to transport rich status from the server to the client. Also, they are the owner of those proto files.
```
ericgribkoff: Delete the `Also` sentence (why does it matter who owns "those" proto files mentioned above?)

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+
+So, the gRPC team introduced a new internal trailing metadata entry `grpc-status-details-bin` to serve this purpose. It takes in serialized proto `google.rpc.status.Status` message binary and transmits as binary metadata. However, since gRPC is a ProtoBuf-agnostic framework, it can't enforce either the content of the binary data be set in the trailing metadata entry, nor the consistency of status code/status message with the additional status detail proto message. The result of this unsolvable problem is that for three major implementations of gRPC we have, each of them is slightly different about this behavior.
```
ericgribkoff: s/be set/set/

ericgribkoff: Need to say that `google.rpc.status.Status` contains its own `code` and `message` fields (probably after second sentence here, before "However,"). This lets a new reader understand the problems with consistency that are then described.

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+
+### Behavior Different Across Implementations
+
+C++ implementation tolerates anything to be put in the `grpc-status-details-bin` entry; Java checks if the code of the proto status is matching the status code of the RPC call; Golang enforces both code and message of the proto status, and the RPC call be the same.
+
+So, which paradigm should Python follow?
+
+### The Proto of Status
```
ericgribkoff: "Status Proto"

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+
+## Proposal
+
+### A New Public Interface: Status
+
+This class is used to describe the status of an RPC.
+
+The name of the class is `Status`, because it is shorter and cleaner than `RpcStatus` or `ServerRpcStatus`. In the future, we might want to utilize it at client side as well.
+
+The metadata field of `Status` is `trailing_metadata` instead of `metadata`. The `Status` should be the final status of an RPC, so it should assist developers to get information that only accessible at the end of the RPC. Also, if we want to support it at client side, we need to separate `initial_metadata` and `trailing_metadata`.
```
ericgribkoff: I don't understand the last sentence about supporting this at client side, I think it can just be deleted. On the client side the status would not included initial_metadata...but it doesn't matter for this gRFC in either case.

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+      code: A StatusCode object to be sent to the client.
+      details: An ASCII-encodable string to be sent to the client upon
+        termination of the RPC.
+      trailing_metadata: The trailing :term:`metadata` in the RPC.
+    """
+```
+
+### Server side API
+
+gRPC Python team encountered lots of disagreements about how this part should be implemented. Available options are listed in the `Rationale` section. Here is our final consensus as a team, but feel free to comment about other options.
```
ericgribkoff: I suggest removing the first sentence about internal disagreements - I don't think it adds anything for future readers...

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+        rich_status = grpc_status.status_pb2.Status(
+            code=grpc_status.code_pb2.INVALID_ARGUMENT,
+            message='API call quota depleted',
+            details=[detail]
+        )
+        context.abort_with_status(rpc_status.to_status(rich_status))
+        # The method handler will abort
+```
+
+### Optional: Client side API
```
ericgribkoff: Delete this section and leave to a future gRFC?

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+* Avoid introducing new classes
+
+#### Cons
+* Recommend using splat is not Pythonic
+* Need to educate developers about the priority of those three APIs
+
+
+### Why we need a new package `grpcio-status`?
+
+Although we can't enforce the consistency of status code, status message, and status details, there is still value for the new package to help developers validate that. It demonstrates our recommendation through the design.
```
ericgribkoff: > Although the grpcio package can't enforce the consistency of status code and status message with the information embedded in the rich `Status` proto, there is still value to provide an additional package - one which can depend on protobuf and is able to verify that these elements are consistent - to help developers use the API in its intended fashion.. It demonstrates our recommendation through the design.

L44-python-rich-status.md
```diff
@@ -0,0 +1,474 @@
...
+
+One more reason is that in the future, the `google.rpc.status.Status` may not be the only status proto we accept. This package provides a straightforward mapping between different status proto message.
+
+
+### Not Swapping the Two Concepts
+
+Ideally speaking, the **status message** and **status details** are misused in Python implementation, and this proposal is our best chance to get them right. According to spec, the correct definition should be as followed:
+
+* **Status message** should be interpreted as a text (Unicode allowed) field that indicates the status of the RPC call, and it is intended to be read by the developers.
+* **Status details** should be understood as a encoded `google.rpc.status.Status` proto message that serves as rich status error details about the RPC Call.
```
ericgribkoff: There is not actually a spec for this, is there? I would just simplify this section by saying what Python calls 'details' actually corresponds to `grpc-message`, and that this discrepancy is unfortunately deeply embedded in the gRPC Python API and unlikely to be changed.

L44-python-rich-status.md
lidizheng: Updated to your version. Thank you for making it more clear.

L44-python-rich-status.md
lidizheng: Updated.

L44-python-rich-status.md
lidizheng: Removed.

L44-python-rich-status.md
lidizheng: The first mistake is corrected.
The description for `Status` message is added:

> Notably, the `google.rpc.status.Status` message contains its own `code` and `message` fields.

L44-python-rich-status.md
lidizheng: Updated.

L44-python-rich-status.md
lidizheng: The sentence is removed. I meant there have to be two fields to support both server-side and client-side.

L44-python-rich-status.md
lidizheng: Removed.

L44-python-rich-status.md
lidizheng: Removed.

L44-python-rich-status.md
lidizheng: Updated.

lidizheng:
@ericgribkoff Thank you for the thorough review! I fixed most issue you posted in the review. The last section rewrite still in progress.

L44-python-rich-status.md
lidizheng: The specs for both "grpc-message" and "grpc-status-details-bin" fields are not directly describing their usage, and you are right we shouldn't do the definition for them in this gRFC. The last section is rewrote into:

> ### Semantic difference of `details` between gRPC Python and `Status` proto

> gRPC Python uses the field name `details` to store data that will later be transmitted in `grpc-message` header. But the `details` field in proto `Status` is a list for error details proto messages. Unfortunately, this discrepancy is deeply embedded in the gRPC Python API and unlikely to be changed.

L44-python-rich-status.md
```diff
@@ -9,24 +9,23 @@ New RPC Status API for Python
...
 
 ## Background
 
-The gRPC [Spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses) defined two trailing data to indicate the status of an RPC call `Status` and `Status-Message`. However, status code and a text message themselves are insufficient to express more complicated situation like the RPC call failed because of quota check of specific policy, or the stack trace from the crashed server, or obtain retry delay from the server (see [error details](https://github.com/googleapis/api-common-protos/blob/master/google/rpc/error_details.proto)).
+The gRPC [Spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses) defined two trailing data to indicate the status of an RPC call `grpc-status` and `grpc-Message`. However, status code and a text message themselves are insufficient to express more complicated situation like the RPC call failed because of quota check of specific policy, or the stack trace from the crashed server, or obtain retry delay from the server (see [error details](https://github.com/googleapis/api-common-protos/blob/master/google/rpc/error_details.proto)).
```
ericgribkoff: nit: grpc-message (lower case m)

lidizheng: Typo fixed.

PR #120, opened 2019-01-22
Use repeated for all instances of LoadBalancingConfig message.

markdroth:
@AspirinSJL pointed out that I missed this in grpc/grpc#17693.

I had originally intended all instances of this field to be repeated, but I somehow didn't get it right in the gRFC (or in our internal proto definition, which I'll fix if we agree on this change).  I think the rationale for this is the same in all cases: we want the ability to select an LB policy that may not be known to all clients.

Any objections?

### DIFF
--- a/A24-lb-policy-config.md
+++ b/A24-lb-policy-config.md
@@ -64,10 +64,14 @@ message XdsConfig {
   string balancer_name = 1;
   // Optional.  What LB policy to use for intra-locality routing.
   // If unset, will use whatever algorithm is specified by the balancer.
-  LoadBalancingConfig child_policy = 2;
+  // Multiple LB policies can be specified; clients will iterate through
+  // the list in order and stop at the first policy that they support.
+  repeated LoadBalancingConfig child_policy = 2;
   // Optional.  What LB policy to use in fallback mode.  If not
   // specified, defaults to round_robin.
-  LoadBalancingConfig fallback_policy = 3;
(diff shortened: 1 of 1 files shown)


AspirinSJL:
LGTM.

carl-mastrangelo:
Please only modify this if you plan on updating existing references.

markdroth:
> Please only modify this if you plan on updating existing references.

As I said in the original description, I will change the internal proto file once we agree on this change.

carl-mastrangelo:
I think you need to go through the full proto upgrade steps.  That means adding a new field number, deprecating the existing one, and then each implementation moving to the new field.

markdroth:
> I think you need to go through the full proto upgrade steps. That means adding a new field number, deprecating the existing one, and then each implementation moving to the new field.

The xds policy does not exist yet, so no one is using it yet.  I think it's safe to change this without allocating a new field number.

carl-mastrangelo:
> The xds policy does not exist yet, so no one is using it yet. I think it's safe to change this without allocating a new field number.


There is existing java code that is constructing this message today.  (internally).  Either you will need update this when you import it, or go through the normal proto update process.

zhangkun83:
There is a Java test file that constructs this proto. @markdroth please make sure to run the gRPC Java TAP and fix it (it would be a trivial fix) when you update the internal proto.

Issue #122, opened 2019-01-23
Rename L38-cpp-server-load-reporting.md to (maybe) L45


dapengzhang0:
Since L38-objc-api-upgrade.md seems merged in first.

dfawley:
It may have been merged first, but `cpp-server-load-reporting` *had* reserved L38 a few days before `objc-api-upgrade` was sent as a PR.

However, it's easier to rename this since the `objc` one has graphics.

Do NOT use L36, L37, or L44.  Those are all open PRs.  That's how this collision happened.  L45 is (currently) free:

https://github.com/grpc/proposal/pulls?utf8=%E2%9C%93&q=is%3Apr+sort%3Aupdated-desc+L45

dapengzhang0:
@dfawley, After double check, I found that actually it is `cpp-server-load-reporting` that was sent out _and_ merged first, but _**reserved**_ "L38" two days _after_ `objc-api-upgrade` was sent as a PR.

#105 L38: Objective-C API Upgrade 
PR sent out date: Sep 24, 2018, 
Merge date: Jan 7, 2019 https://github.com/grpc/proposal/pull/105#event-2058226386

#91 L38: server load reporting
PR sent out date: Jul 13, 2018,  with initial file name: L30-cpp-server-load-reporting.md
"L37" reserve date: Sep 20, 2018 https://github.com/grpc/proposal/pull/91#issuecomment-423261387
"L38" reserve date: Sep 26, 2018 https://github.com/grpc/proposal/pull/91#event-1869537366
Merge date: Oct 22, 2018 https://github.com/grpc/proposal/pull/91#event-1918690898

I agree it's easier to rename this one since the objc one has graphics.

dfawley:
You're right.  My [comment on the 20th](https://github.com/grpc/proposal/pull/91#issuecomment-423261387) *said* that I would set it to L38, but I then changed it to L37 by mistake.  It was adjusted to L38 a few days later by @AspirinSJL, after #105 was already sent.  Sorry for the mixup.  The gRFC process clearly has some rough edges.

PR #123, opened 2019-01-25
L45: rename of L38-cpp..

AspirinSJL:
Fix https://github.com/grpc/proposal/issues/122.


dapengzhang0:
@AspirinSJL After merge, it would be nice to update all the links to it, if there are any internal docs linking to it.

PR #126, opened 2019-03-01
L47: Node types package

murgatroid99:
Discussion thread: https://groups.google.com/forum/#!topic/grpc-io/30SrKiqsEis


murgatroid99:
I have determined that it is not practical to implement a library as described in this proposal so I am closing it.

PR #129, opened 2019-03-02
A17 update: Specify validation rule as per #100

yashykt:
Specify validation rule as per #100


yashykt:
Thanks for reviewing!

PR #130, opened 2019-03-02
A24 update: Add validation rules as per #100

yashykt:
Add validation rules as per #100


yashykt:
Thanks for reviewing!

PR #140, opened 2019-04-19
L50: gRPC Objective-C Interceptor

muxi:
Discussion thread: https://groups.google.com/forum/#!topic/grpc-io/7QGmsQW86Ss

Implementation: https://github.com/grpc/grpc/pull/18852


L50-objc-interceptor.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Abstract
+
+Proposes the interceptor architecture and API for gRPC Objective-C.
+
+## Background
+
+gRPC interceptor has been a popular utility for users making HTTP requests.
+gRPC Objective-C library has received a number of requests to add support of
+this feature to gRPC. The old gRPC Objective-C library API was not flexiable enough to support interceptor API. With the [new API](https://github.com/grpc/proposal/blob/master/L38-objc-api-upgrade.md), now it is possible to add in this support.
```
srini100: s/flexiable/flexible

L50-objc-interceptor.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Use cases
+The use cases for interceptor is wide; a few examples include:
+* Log RPCs
+* Collect metrics
+* Obtain/Refresh access tokens
+* Create test stub
+* ...
```
srini100: Caching should be  a popular one, especially when you already have an example on it.

L50-objc-interceptor.md
```diff
@@ -0,0 +1,263 @@
...
+
+Requirements of interceptor API are
+* Allows access to headers, messages, and trailers
+* Allows access to call destination, paths, verbs, and call configuration parameters
+* Supports Bi-di streaming
+* Allows short-circuiting calls (interceptor can terminate the call)
+* Allows interceptors to specify their own dispatch queue
+* Scope interceptors separately without direct access to each other's states
+* Allows state sharing between interceptors when initiated by the interceptor author
+
```
srini100: In the above lines, clarify if "access" means read or write. If write, you could say something like "Allows modification to...."

L50-objc-interceptor.md
```diff
@@ -0,0 +1,263 @@
...
+Some optional requirements that we may not implement at the initial stage
+* Global interceptor
+* Interception of non-serialized protobuf message
+
+### Related Proposals: 
+The interceptor API is built based on the proposals [L38: Proposal to A New gRPC Objective-C API](https://github.com/grpc/proposal/blob/master/L38-objc-api-upgrade.md) and [L49: gRPC Objective-C Flow Control](https://github.com/grpc/proposal/blob/c28ba74bbacc8c1fb5d269f48d89b06c51d21f44/L49-objc-flow-control.md).
+
```
srini100: The flow control proposal is still marked as in review. It should be changed to approved.

L50-objc-interceptor.md
```diff
@@ -0,0 +1,263 @@
...
+@end
+```
+After all the alternations in the interceptors, GRPCCall2Internal object uses the serializer to transform the data before passing it to core.
+
+Though there are some potential use cases for this feature, it can be implemented independent of the proposal above.
+
+### Interceptor Dispatch Queues
+Since each interceptor is stateful, multiple requests or responses issued to an interceptor should be processed in order. That requires the dispatch queues of an interceptor to be serial queue. This requirement currently seems reasonable for us, because multiple calls can each have their own dispatch queue so that the same class of interceptors for different calls can still run concurrently. On the other hand, if serialized process is required across all interceptors of the same class, they will be able to use the same dispatch queue provided by their interceptor factory on initialization.
+
+In case we have to support concurrent queue in the future, the following alternation may be used to achieve the goal transparently to the users and the interceptor developers:
```
sanjaypujare: alternation => alternative

L50-objc-interceptor.md
```diff
@@ -0,0 +1,263 @@
...
+@end
+
+@interface GRPCMutableCallOptions
+
+// Properties that already exist
+@property id<GRPCMarshaller> marshaller;
+
+@end
+```
+After all the alternations in the interceptors, GRPCCall2Internal object uses the serializer to transform the data before passing it to core.
```
sanjaypujare: I thought this was meant to be "alternatives" but probably you meant "alterations" as in updates? In that case "updates" itself might be a better word.

L50-objc-interceptor.md
muxi: Well I didn't really mean "alternative" too :) But I rephrased it. Thanks Sanjay.

L50-objc-interceptor.md
muxi: Yea that's correct :)

PR #142, opened 2019-05-10
A21 update: Add note on fallback to default service config if no service config is found

yashykt:
Add note on fallback to default service config if no service config is found


A21-service-config-error-handling.md
```diff
@@ -44,8 +44,11 @@ if the chosen service config is found invalid while the other service configs
 were valid, the gRPC config would still be considered invalid as a whole.
 2. The client validates the choice based on the criteria outlined in the
 section below.
-3. If the service config is valid, the client uses the configuration provided.
-4. If the service config is invalid, the clients should reject the received
+3. If no service config is found, the client falls back on the default service
+config. If there is no default service configured either, the client selects no
```
markdroth, line 48: I think if there is no default service config set for the channel, then it will use the an empty config.

markdroth:
@ejona86 and @dfawley should also review, just to make sure I haven't missed anything.

A21-service-config-error-handling.md
```diff
@@ -44,8 +44,11 @@ if the chosen service config is found invalid while the other service configs
 were valid, the gRPC config would still be considered invalid as a whole.
 2. The client validates the choice based on the criteria outlined in the
 section below.
-3. If the service config is valid, the client uses the configuration provided.
-4. If the service config is invalid, the clients should reject the received
+3. If no service config is found, the client falls back on the default service
```
dfawley, line 47: Should we clarify this a little and say "the default service config _set via the client/channel API_" (e.g.)?  I don't think we've ever documented what the "default service config" is or where it comes from.  (It is already referenced elsewhere, but was never defined AFAICT.)

dfawley (approved):

A21-service-config-error-handling.md
yashykt: Done

A21-service-config-error-handling.md
yashykt: Done

markdroth (approved):

A21-service-config-error-handling.md
ejona86, line 47: Shouldn't this be (2) and (2) be renamed to (3)?

ejona86 (approved):

A21-service-config-error-handling.md
yashykt: You mean reordering the two? Will do

ejona86: Yeah, reorder it between the current 1 and 2, since 2 is about validation.

ejona86 (approved):

PR #147, opened 2019-06-05
Add note about receiving NO SERVICE CONFIG

yashykt:
Add note about receiving NO SERVICE CONFIG.

This differentiates the NO SERVICE CONFIG case from the valid service config or invalid service config cases.

### DIFF
--- a/A21-service-config-error-handling.md
+++ b/A21-service-config-error-handling.md
@@ -60,6 +60,10 @@ configuration.
 		resolver, it should retry the query after appropriate backoff). In
 		other words,the channel will remain in state TRANSIENT_FAILURE until a
 		valid service config is received.
+5. If the resolver does not support service config resolution OR if no gRPC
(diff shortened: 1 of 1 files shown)


A21-service-config-error-handling.md
```diff
@@ -60,6 +60,10 @@ configuration.
 		resolver, it should retry the query after appropriate backoff). In
 		other words,the channel will remain in state TRANSIENT_FAILURE until a
 		valid service config is received.
+5. If the resolver does not support service config resolution OR if no gRPC
```
markdroth: Let's make it clear that the "OR if gRPC config is received but no service config is chosen" part is specific to the DNS resolver.

yashykt: Done!

yashykt:
@dfawley @lyuxuan PTAL

yashykt:
Thanks!

PR #149, opened 2019-06-12
L54: Python Server Wait API

lidizheng:
gRFC: https://github.com/lidizheng/proposal/blob/python-wait/L54-python-server-wait.md
Group Thread: https://groups.google.com/forum/#!topic/grpc-io/GhJl_JHfaHA

### DIFF
--- a/L54-python-server-wait.md
+++ b/L54-python-server-wait.md
@@ -14,7 +14,7 @@ until the server is stopped or terminated.
 
 ## Background
 
-Currently, the gRPC Python server only provides a `server.Start()` API, but
+Currently, the gRPC Python server only provides a `server.start()` API, but
 there isn't any API to block forever. gRPC Python server applications need to
 block on the main thread to keep serving explicitly. The solution the gRPC
 Python team proposed in examples is sleeping for a while then exit. See code:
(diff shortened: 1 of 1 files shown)


L54-python-server-wait.md
```diff
@@ -0,0 +1,126 @@
...
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/GhJl_JHfaHA
+
+## Abstract
+
+Add a `wait_for_termination` API for Python server that can block in any thread
+until the server is stopped or terminated.
+
+## Background
+
+Currently, the gRPC Python server only provides a `server.Start()` API, but
```
gnossen: `s/Start/start/`

L54-python-server-wait.md
```diff
@@ -0,0 +1,126 @@
...
+in Java as `blockUntilShutdown()`, in Golang as `Serve`. This gRFC is
+introducing the Python version of it.
+
+### Related Proposals: 
+
+None.
+
+## Proposal
+
+This gRFC propose to add following API to `grpc.Server` interface:
```
gnossen: `s/propose to add/proposes that we add/`

L54-python-server-wait.md
```diff
@@ -0,0 +1,126 @@
...
+        
+        1) Calling `stop` on the server in another thread;
+        2) The `__del__` of the server object is invoked.
+        """
+        raise NotImplementedError()
+```
+
+## Rationale
+
+### Alternative Design: Await For Signal
```
gnossen: I definitely think the way you chose is the way to go. The application author should be in charge of what runs on which thread. They can write their own signal handler to run on the main thread and call `server.stop()`.

lidizheng: Glad to see that we have similar view of how this API should be done :)
I added another alternative design section describing the idea proposed by @mehrdada in the implementation PR.

PR #150, opened 2019-06-17
L55 - gRPC Objective-C Global Interceptor

muxi:
Discussion thread: https://groups.google.com/forum/#!topic/grpc-io/tRKcPcfQKi4

Implementation: https://github.com/grpc/grpc/pull/19322


muxi:
@srini100 - can you take a look?

L55-objc-global-interceptor.md
```diff
@@ -0,0 +1,60 @@
...
+The proposed gRPC Objective-C library's global interceptor interface is as follows:
+
+```objectivec
+@interface GRPCCall2 (Interceptor)
+
+/**
+ * Register a global interceptor's factory in the current process. Only one interceptor can be
+ * registered in a process. If another one attempts to be registered, an exception will be raised.
+ */
++ (void)registerGlobalInterceptor:(nonnull id<GRPCInterceptorFactory>)interceptorFactory;
```
srini100: Do other langs have unregister API, in case app wants to remove/replace the global interceptor on-the-fly?

muxi: Asked Java and C++ and they do not have it, with rationale that it makes the hack (global interceptor) more complicated.

PR #152, opened 2019-06-25
Clarify specification of percentage selector in service configs in DNS.

markdroth:
It looks like our current implementations are all doing the wrong thing with the percentage selector: we are actually recomputing the client's random percentage for each choice we evaluate when parsing the service config choices, which means (a) we may not have a consistent value when there are multiple choices with different percentage values and (b) even worse, we will recompute the percentage every time the resolver re-resolves, which means we will wind up switching configs whenever (e.g.) a subchannel becomes disconnected.

I am proposing that the client compute its percentage at resolver startup and retain it for the lifetime of the resolver.  This means that if a client starts up and computes a low percentage for itself, it will wind up being an early canary, but I think that's okay.

The alternative would be to have the resolver check whether the service config has changed since the last resolution and recompute the client's percentage only if it has changed.  However, this would actually make it harder to understand what's happening when the service owner is canarying changes, because it would be much harder to identify which clients are getting the canary config.

### DIFF
--- a/A2-service-configs-in-dns.md
+++ b/A2-service-configs-in-dns.md
@@ -55,9 +55,12 @@ determine which choice will be selected by a given client:
     // "python", etc).  Each string is case insensitive.
     "clientLanguage": [string],
     // Percentage: integer from 0 to 100 indicating the percentage of
-    // clients that should use this choice.  If present, the number must
-    // match the regular expression `^0|[0-9]|[1-9][0-9]|100$`
-    // All other numbers are considered invalid.
+    // clients that should use this choice.  Each client's resolver should
+    // compute its percentage at startup and use the same percentage for the
+    // lifetime of the resolver; if the client's computed percentage is less
+    // than the value of this field, then it is considered a match.
+    // If present, the number must match the regular expression
+    // `^0|[0-9]|[1-9][0-9]|100$`.  All other values are considered invalid.
     "percentage": number,
     // Client hostname(s): a list of strings.  Each name is case 
     // sensitive and must be an exact match of the hostname according to


A2-service-configs-in-dns.md
```diff
@@ -55,9 +55,12 @@ determine which choice will be selected by a given client:
     // "python", etc).  Each string is case insensitive.
     "clientLanguage": [string],
     // Percentage: integer from 0 to 100 indicating the percentage of
-    // clients that should use this choice.  If present, the number must
-    // match the regular expression `^0|[0-9]|[1-9][0-9]|100$`
-    // All other numbers are considered invalid.
+    // clients that should use this choice.  Each client's resolver should
+    // compute its percentage at startup and use the same percentage for the
+    // lifetime of the resolver; if the client's computed percentage is less
+    // than the value of this field, then it is considered a match.
```
dfawley: This means that if two entries have the same criteria except percentage, and the second entry has a lower percentage than the first, it will never be selected.  Is this what we want?  Should we call this out here?  Should we provide a tool to help service owners vet their service configs before pushing them?

As an alternative, we could re-roll for every entry, but start with the same random seed for the first roll each time.

A2-service-configs-in-dns.md
```diff
@@ -55,9 +55,12 @@ determine which choice will be selected by a given client:
     // "python", etc).  Each string is case insensitive.
     "clientLanguage": [string],
     // Percentage: integer from 0 to 100 indicating the percentage of
-    // clients that should use this choice.  If present, the number must
-    // match the regular expression `^0|[0-9]|[1-9][0-9]|100$`
-    // All other numbers are considered invalid.
+    // clients that should use this choice.  Each client's resolver should
+    // compute its percentage at startup and use the same percentage for the
+    // lifetime of the resolver; if the client's computed percentage is less
```
carl-mastrangelo: lifetime of the process?  Does this work with python multiprocessing?

A2-service-configs-in-dns.md
markdroth: > This means that if two entries have the same criteria except percentage, and the second entry has a lower percentage than the first, it will never be selected. Is this what we want? 

Yes, I think that's the desired behavior.

> Should we call this out here?

I don't think we need to explicitly call this out; it's inherent in the first-match-wins semantic.  It's always possible to construct a list of choices such that one of them will never be used, just by having the choice be a superset of the selectors of an earlier choice.

> Should we provide a tool to help service owners vet their service configs before pushing them?

I would not be opposed to this if someone wants to write it, but I think this is out of scope of this change.

> As an alternative, we could re-roll for every entry, but start with the same random seed for the first roll each time.

I don't think it makes sense to change the computed percentage for each entry.  The percentage selector is really intended to be a way to canary changes on a slowly increasing set of clients.  And one of the biggest criticisms of the percentage selector from the original design discussion was that it was so volatile compared to the more stable selection mechanism we have internally, so I think we should minimize the churn as much as possible here.

A2-service-configs-in-dns.md
markdroth: In C-core, we don't really have anything that exists for the lifetime of the process.  Even our most global data structures only exist between when grpc_init() and grpc_shutdown() are called, and that can happen multiple times over the lifetime of a process.  I suppose we could use that scope for this, but that would add complexity, and it's not clear what the advantage of that would be.  I think lifetime of the resolver (which is basically the lifetime of the channel) is fine.

I don't think this will be a problem for python or any other wrapped language, since it's purely within the channel implementation in C-core.

carl-mastrangelo: to be clear, the number picked by the client is [0,100] ?  [0,100) ?  (0,100] ?

carl-mastrangelo: Answering my own question (and I think you should include in the text:   The random number the client picks should be [0, 100).  The config should apply if the client pct is >= svcConfig.pct.

yashykt: sgtm

markdroth: Neither 0 nor 100 seem particularly useful values for people to use in practice, although I can't think of a reason to reject either of them, since it's clear how to handle both.  What's the advantage in rejecting 100?  Why not reject 0 also?

A2-service-configs-in-dns.md
ejona86: Hmm... I'm thinking now it may make the most sense to re-roll for each selector. Because the population for each set of conditions is different, it can be hard to determine how the percentage in an earlier condition affects later conditions.

It also seems to be more dangerous to use a single roll. If I want a 1% sample for a rule, but it follows a rule that is 10%, I need to set the percentage to 11%. However, if someone removes the earlier rule that now means we must remember to re-calculate all following percentages, including determining how overlapping populations interact. Re-rolling means that you select X% of the population that matches that rule, which seems much easier to manage.

markdroth: I think that in practice, people will generally wind up using only one percentage selector, since they will generally be rolling out one change at a time.  I think the stability of doing the roll only once will be a much bigger benefit to most people than the potential to have to worry about multiple rules with percentages.

A2-service-configs-in-dns.md
dfawley: AFAICT @carl-mastrangelo was talking about the random number range used by the client to roll the die, to which I would agree [0, 100) is proper with `client pct < svcConfig.pct` as the logic.  I don't think anyone was arguing 0 or 100 should be disallowed as a setting in the config.

FWIW 0 is useful to disable a config without needing to delete it.  For 100, it's easy enough to delete the percentage field from your config list instead, but I don't see any reason to disallow it, either.

markdroth: Ah, sorry, I misunderstood.  I agree that the number picked by the client should be in the range [0, 100).

I've added a note about this.

A2-service-configs-in-dns.md
dfawley: > I think the stability of doing the roll only once will be a much bigger benefit to most people than the potential to have to worry about multiple rules with percentages.

We can have the same stability by using the same seed at the start of each processing round, too.

> I think that in practice, people will generally wind up using only one percentage selector, since they will generally be rolling out one change at a time.

This is logical, but it's dangerous to assume how people will be using a feature unless you lock it down such that it may only be used that way (e.g. "only one selector may contain a percentage; any subsequent selectors with a percentage will be ignored").

ejona86: I think Mark's argument from a usability standpoint. It's okay if someone wants to percentages, but they have to deal with the more limited usability. We're optimizing for what we expect the common case to be: a single percentage.

I don't think we have to choose one or the other though. We choose a seed when we start the resolver. We start we the same seed each time entering the selection logic. When we check a condition against a percentage we re-roll for any following percentage. That would be equally stable in the one-percentage-only case.

markdroth: I don't see how using the same seed for multiple rolls actually helps stability.  The thing that hurts stability is that we would be rolling again each time we re-resolve, and I don't see how that is avoided by using the same seed to start with.  In the common case, where there is one feature being canaried with a slowly increasing percentage of clients, we don't want individual clients constantly falling into and out of the canary set just because one of their subchannels has become disconnected.  The canary set is much more stable if a client decides its percentage when the resolver starts and doesn't change it until it is destroyed.

ejona86: Each time we look for a matching selection we will roll the same numbers.

```java
long seed;
MyResolver() { // constructor
  seed = new Random().nextLong(); // random
}

Map chooseConfig(List configs) {
  Random r = new Random(seed); // deterministic, for the life of the resolver
  for (Map config : configs) {
    // check conditions
    if (config.percentage != null && r.nextInt(100) < config.percentage) {
      return config;
    }
    ...
  }
}
```

dfawley:
> I don't see how using the same seed for multiple rolls actually helps stability. The thing that hurts stability is that we would be rolling again each time we re-resolve, and I don't see how that is avoided by using the same seed to start with.

If you pick a seed at resolver startup, and restore the seed every time you process the config list, then roll \<n\> will be the same every time.  Conceptually, instead of rolling once and saving the result, you roll N times and save them all, and apply them in the same order each time you process the list.

With this scheme, increasing the percentage of an entry in the list will never cause clients to stop selecting that entry.

A2-service-configs-in-dns.md
dfawley: > I don't see how using the same seed for multiple rolls actually helps stability. The thing that hurts stability is that we would be rolling again each time we re-resolve, and I don't see how that is avoided by using the same seed to start with.

If you pick a seed at resolver startup, and restore the seed every time you process the config list, then roll <n> will be the same every time. Conceptually, instead of rolling once and saving the result, you roll N times and save them all, and apply them in the same order each time you process the list.

With this scheme, increasing the percentage of an entry in the list will never cause clients to stop selecting that entry.

A2-service-configs-in-dns.md
```diff
@@ -55,9 +55,13 @@ determine which choice will be selected by a given client:
     // "python", etc).  Each string is case insensitive.
     "clientLanguage": [string],
     // Percentage: integer from 0 to 100 indicating the percentage of
-    // clients that should use this choice.  If present, the number must
-    // match the regular expression `^0|[0-9]|[1-9][0-9]|100$`
-    // All other numbers are considered invalid.
+    // clients that should use this choice.  Each client's resolver should
```
carl-mastrangelo: Just to be clear:   If the channel goes into idle mode, shuts down the resolver, and then comes back out of idle, it is allowed to re-roll, since it could be a new name resolver instance?

markdroth: Yes, I think that's fine.

PR #155, opened 2019-07-18
L58: Async API for gRPC Python

lidizheng:
Link to gRFC: https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md

Discussions (besides this PR):
* https://groups.google.com/forum/#!topic/grpc-io/7V7HYM_aph4
* https://github.com/lidizheng/grpc-api-examples/pull/1
* https://github.com/lidizheng/grpc-api-examples/pull/2

Fixes https://github.com/grpc/grpc/issues/19494
Fixes https://github.com/grpc/grpc/issues/19495


lidizheng:
@pfreixes @rmariano Updates and comments are welcomed.

L58-python-async-api.md
```diff
@@ -0,0 +1,453 @@
...
+* Accepts `grpc.aio.HandlerCallDetails`;
+* `continuation` is a function accepts `grpc.aio.HandlerCallDetails` and returns
+  `grpc.aio.RpcMethodHandler`.
+
+Changes in `ClientInterceptor` classes in `grpc.aio`:
+* `continuation` is a coroutine that returns `Awaitable` object.
+
+### Utility Functions
+
+Changes in `grpc.aio.channel_ready_future`:
```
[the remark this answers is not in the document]
lidizheng: What do you think for removing the future suffix, so it will look like `grpc.aio.channel_ready`?

In actual code:
```Python
channel = grpc.aio.insecure_channel("localhost:50051")
await grpc.aio.channel_ready(channel)
```

L58-python-async-api.md
```diff
@@ -0,0 +1,453 @@
...
+while 1:
+    stub.ReportLoad(ReportRequest(
+        timestamp=...,
+        metrics=[...],
+    ))  # <- RPC not sent
+    await asyncio.sleep(3)
+
+### Await on whatever function starts with "Async"
+while 1:
+    await stub.AsyncReportLoad(ReportRequest(
```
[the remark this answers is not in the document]
lidizheng: I'm also in favor of not making modification. I'm mentioning this is because people do use this naming pattern when they adopt `asyncio` (like [Instagram](https://www.slideshare.net/jimmy_lai/the-journey-of-asyncio-adoption-in-instagram)). 

Now, I'm thinking maybe we can name the service stub `async_stub` in our example code.

L58-python-async-api.md
```diff
@@ -0,0 +1,453 @@
...
+### The new usage is much verbose for same scenario
+class AsyncIter:
+    def __init__(self, items):    
+        self.items = items    
+
+    async def __aiter__(self):    
+        for item in self.items:    
+            yield item
+
+stub.SayHelloStreaming(AsyncIter([
```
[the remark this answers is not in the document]
lidizheng: The only valid use case I can think about in practice: transmitting trunks of a large file.
On the other hand, testing code usability is also something we care about. Not only gRPC Python itself is writing end2end tests, but also our users.

But doing magic for different type of input makes me hesitate a bit. I'm also in favor of doing nothing, and only accepting asynchronous generator/iterator.

L58-python-async-api.md
```diff
@@ -0,0 +1,453 @@
...
+        metrics=[...],
+    ))  # <- RPC not sent
+    await asyncio.sleep(3)
+
+### Await on whatever function starts with "Async"
+while 1:
+    await stub.AsyncReportLoad(ReportRequest(
+        timestamp=...,
+        metrics=[...],
+    ))  # <- RPC not sent
```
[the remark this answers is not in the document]
lidizheng: Good catch!

L58-python-async-api.md
```diff
@@ -0,0 +1,453 @@
...
+* Asynchronous processing that perfectly fits IO-intensive gRPC usage;
+* Resolve a long-living design flaw of thread exhaustion problem;
+* Performance is much better than multi-threaded.
+
+### Drawbacks Of `asyncio`
+
+* Non-interoperability between coroutine functions and normal functions;
+* Cognitive overhead of event loop mechanism;
+* Calls to function may cause context-switches which resulted in unexpected
+  behavior. -->
```
pfreixes: WDYM? It's hard to believe that a context switch might have as a side effect a ` unexpected behavior`

L58-python-async-api.md
```diff
@@ -0,0 +1,453 @@
...
+channel = grpc.aio.intercept_channel(...)
+```
+
+To reduce cognitive burden, the new `asyncio` API should share the same
+parameters as the current API, and keep the usage similar, except the following
+cases:
+
+* `grpc.aio.server` no longer require the application to provide a
+  `ThreadPoolExecutor`;
+* Interfaces returning `Future` object are replaced.
```
pfreixes: or removed since `asyncio` already provides a way for running a coroutine as an independent task, and implementing the `Future` contract, for example:

```python
task = asyncio.run(stub.Hi(echo_pb2.EchoRequest(message="ping")))
task.add_done_callback(my_callback)
```

L58-python-async-api.md
```diff
@@ -0,0 +1,453 @@
...
+```
+
+Client side:
+```Python
+# Python 3.6 and lower
+with grpc.aio.insecure_channel("localhost:50051") as channel:
+    stub = echo_pb2_grpc.EchoStub(channel)
+    response = await stub.Hi(echo_pb2.EchoRequest(message="ping"))
+
+# Or using asynchronous context manager for Python3.7 and up
```
pfreixes: asynchronous context manager can be implemented and used since Python 3.5

L58-python-async-api.md
```diff
@@ -0,0 +1,453 @@
...
+server.add_insecure_port(":50051")
+helloworld_pb2_grpc.add_GreeterServicer_to_server(Greeter(), server)
+server.start()
+await server.wait_for_termination()
+```
+
+Client side:
+```Python
+# Python 3.6 and lower
+with grpc.aio.insecure_channel("localhost:50051") as channel:
```
pfreixes: Give support for the none synchronous and the asynchronous one might be a bit confusing for the users, since we are not gonna be able to implement everything for both use cases. I would get stick on only implementing the asynchronous one.

L58-python-async-api.md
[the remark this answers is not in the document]
gnossen: This naming scheme tracks much better for me. `await noun_adjective()` sounds to me like a pretty straightforward way of saying "block until the `noun` is `adjective`".

L58-python-async-api.md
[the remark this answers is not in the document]
gnossen: You could always send an email to the python-dev mailing list about implementing an `aiter()` function.

L58-python-async-api.md
lidizheng: An example of unexpected behavior will be the need for synchronization. Users of `asyncio` may think that there is only one thread operating, so it is safe to get rid of `Lock`/`Event`/`Condition`.

However, it isn't. Take the following code as example, if the user is trying to operate DB or other global state, data race can still happen.

```Python
import asyncio

class Account:
    def __init__(self, name, amount):
        self.name = name
        self.amount = amount
    def __repr__(self):
        return f"<class Account name={self.name} amount={self.amount}>"

accounts = {
    1: Account(name='John', amount=70),
    2: Account(name='Smith', amount=30),
}

async def withdraw(account_id, diff):
    await asyncio.sleep(0.5)  # Mock async db read operation
    account = accounts[account_id]
    new_amount = account.amount - diff
    if new_amount < 0:
        raise ValueError('Over withdraw!')
    await asyncio.sleep(0.5)  # Mock async db write operation
    account.amount = new_amount
    print(f'Successfully withdraw [{diff}] amount of money')

async def main():
    await asyncio.gather(
        withdraw(2, 30),
        withdraw(1, 45),
        withdraw(1, 30),
        withdraw(2, 30),
    )
    print(accounts)

asyncio.run(main())
```

All withdraw operation will success, which is wrong:
```shell
Successfully withdraw [30] amount of money
Successfully withdraw [45] amount of money
Successfully withdraw [30] amount of money
Successfully withdraw [30] amount of money
{1: <class Account name=John amount=40>, 2: <class Account name=Smith amount=0>}
```

pfreixes: Agree, data races can happen and the developers need to take this into account for avoiding them. But Python, and specifically Asyncio provides the primitives as building blocks for implementing these features that would help the developer to address this kind of issues.

Calling out that the chances of having data races is a Asyncio drawbacks it's IMO a bit unfair taking into account that this can happen in other alternatives like gevent, twisted, or even in another languages/frameworks like Node and Go. 

Im wondering if there is any language and framework that providing the capacity of achieving concurrency its at the same time free of having data races, do you have any example?

lidizheng: Yes, I agree that data races is inevitable with concurrency. That's why I commented those lines. They shouldn't show with normal view. I will remove them.

Maybe my wording is not accurate here. In my very first draft, I wrote these points as a reminder for myself that introducing `asyncio` into a system might have some side effects that users might not fully aware of. So, during design phase, I can try to protect users from going into these error-prone mistakes.

L58-python-async-api.md
lidizheng: [`contextlib.asynccontextmanager`](https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager) is new in Python 3.7. I'll change it, good catch!

L58-python-async-api.md
lidizheng: Considering `async with` is supported since 3.5, I will remove the first use case.

L58-python-async-api.md
lidizheng: Yes, so for those use case, they should return a Task or our context class instead. The idea being, if we must use `Future` mechanism, we should hide the implementation details. Otherwise, we should yield the choice of `Future` implementation to our users.

https://docs.python.org/3.7/library/asyncio-future.html#asyncio.Future

lidizheng: s/asyncio.run/asyncio.ensure_future

pfreixes: So, either the user uses the Asyncio API directly by scheduling a separated task or the `aio` library does it internally and returns either an `Asyncio.Task` [1] or the context class.

Seems that I will go for the second option only if the object that needs to be returned is not the `Asyncio.Task` one. Otherwise we will be adding a superfluous wrapper IMO.

If we return a gRPC object this would need to implement all of the interfaces - like being awaitable - to be used by the set of the asyncio methods that deal with awaitable objects, for example `wait` [2] and `wait_for` [3]. Otherwise, the user would have a lot of friction using these asyncio methods when a gRPC future object is given as a parameter, for example:

```python
grpc_task = stub.Hi.future(echo_pb2.EchoRequest(message="ping"))
await asyncio.wait([grpc_task])
```

This makes me think that the simpler the better, so going for the solution where the `aio` module does not implement any ad-hoc method, and the user uses the Asyncio API directly. Going for the ad-hoc alternative would need a strong reason. 

[1] https://docs.python.org/3/library/asyncio-task.html#asyncio.Task
[2] https://docs.python.org/3/library/asyncio-task.html#asyncio.wait
[3] https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for

lidizheng: I'm not sure I fully understand your argument. What is your conclusion here? Do you support returning an `asyncio.Task` object?

L58-python-async-api.md
```diff
@@ -0,0 +1,432 @@
...
+  `stream_tream` will be `MultiCallable` objects in `grpc.aio`.
+
+Changes in `grpc.aio.Call`:
+* All methods return coroutine object.
+
+Changes for `MultiCallable` classes in `grpc.aio`:
+* `grpc.aio.UnaryUnaryMultiCallable` and `grpc.aio.StreamUnaryMultiCallable`
+  returns `Awaitable` object for `__call__`;
+* `grpc.aio.UnaryUnaryMultiCallable` and `grpc.aio.StreamUnaryMultiCallable`
+  returns `(Awaitable, grpc.aio.Call)` for `with_call`;
```
pfreixes: For the sake of simplicity and for having an orthogonal API with the synchronous one I would go for having a `with_call` with an interface like this one

```python
response, call = await stub.Hi.with_call(...)
task = asyncio.ensure_future(stub.Hi.with_call(...))
..
..
response, call = task.result()
```

So returning basically an _awaitable_ object implemented by the coroutine, and having the coroutine the ability to return a tuple with two objects when it has finished.

IMO, if I'm not missing something, this would implement the same contract as the synchronous one and will make the implementation under the hood easier since its only a matter of returning two parameters instead of one when the `with_call` is used.

Regarding the object to return for the call, we might return the same object `grpc.aio.Call`  that we could be using for the `future()` method but without yielding if the `await` expr is used, take a look to the `Future` implementation of Asyncio [1].

The following should work, I did not test it, and the user should not suffer any context switch:

```python
response, call = await stub.Hi.with_call(...)
await call
print(call.status)
```

Also a bit related, we would need to implement the cancelation of a task, which would impact the `with_call` and the `__call__` methods.

For example this should be doable:

```python
task = asyncio.ensure_future(stub.Hi.with_call(...))
task.cancel()
```

Which theoretically should be implementable by changing a bit the `call.pyx.pxi` [2] file with something like

```python
   try:
       await self._waiter_call
   except CancelledError:
       if not self._canceled:
           # Behind the scenes ends up by calling the grpc_call_cancel
           # gRPC method
           self.cancel()
    .....
```

Later on, this cancellation mechanism might be used by the `grpc.aio.Call.cancel` the method, which behind the scenes would cancel the task.

[1] https://github.com/python/cpython/blob/master/Lib/asyncio/futures.py#L260

[2] https://github.com/Skyscanner/grpc/pull/1/files#diff-d07d1d5928e4bc4defeb2e1218c57186R136

L58-python-async-api.md
pfreixes: second thought, and related with this comment [1].

Having the feeling that we would need to return a custom object where internalities will be hidden. For example for the use case of the `future()` method of the unary call, we would return a `grpc.aio.Call` object which would implement the `__await__` method, making it compatible with Asyncio API primitives that are expecting _awaitables_ like `wait`, `wait_for`, `ensure_future` or the `await` keyword, so the following should work - I haven't tried it:

```python
call = stub.Hi.future(...)
await call
```
Ofc the following snippet will have the same effect, since the object returned implement the __await__ method

```python
call = await stub.Hi.future(...)
```


Internally the `grpc.aio.Call` wold start the task in the background, having the chance of using the methods provided by the `asyncio.Task` for implementing some of the public methods of the `grpc.aio.Call` object, for example

```python
class Call:
    def __init__(coro,......):
        self._task = asyncio.create_task(coro)
        ....
    def __await__(self):
        return self._task.__await__()
   async def cancel(self):
        self._task.cancel()
        ....
```

[1] https://github.com/grpc/proposal/pull/155/files#r311288205

lidizheng: How about make the `grpc.aio.Call` class inherit from `asyncio.Task`? So we get all the subtle changes across multiple Python versions, and also get the chance to inject the logic we want?

```Python
class Call(asyncio.Task):
    def __init__(coro, ...):
        ....
```

L58-python-async-api.md
lidizheng: I hope we can merge the `__call__`, `with_call` and `futures`. It can be much simpler. By default, the invocation to the stub will return an `asyncio.Task` object. Users can control the life cycle of that RPC by invoking methods on that object. On the other hand, if users just care about the result, they can `await` on the `Task`.

```Python
# Normal call
response = await stub.Hi(HiRequest)

# In the case of cancel
call = stub.Hi(HiRequest)
if number_of_greeting > 0:
    call.cancel()
else:
    response = await call
```

pfreixes: Merging `__call__`, `with_call` would mean not having the same implementation btw the sync and async version of the API, are we ok with that?

If so, I kinda agree with your proposal, but adding the following modifications:

1) The three original methods for making a call would be implemented with just one way. So `__call__`, `with_call` and `future` implemented under the `__call__` method
leveraging on the _awaitable_ property of a `grpc.aio.Call` object, so the following expressions would be valid:

```python
response = await stub.Hi(...)
stub.Hi(...).cancel()
asyncio.wait_for(stub.Hi(...), timeout=10)
```

2) I would move everything that is not related to the call managing from the `grpc.aio.Call` object under a new `grpc.aio.Response` object. In the synchronous version,
the `_Rendezvous` [1] implements both features, managing the call and returning properties of the response. If we do not change this, with the proposal of unifying the `with_call` method with the `__call__` the user will be forced to repeat the following pattern:

```python
call = stub.Hi(...)
response = await call
print(response)
print(call.status)
```

The alternative with a dedicated response object would be:

```python
response = await stub.Hi(...)
print(response.content)
print(response.code)
```


[1] https://github.com/Skyscanner/grpc/blob/7b0576787898ee670c7a56b3fac34acfb1a62df0/src/python/grpcio/grpc/_channel.py#L251

L58-python-async-api.md
pfreixes: Inheriting directly from the `asyncio.Task` would put our API at merci of the Asyncio maintainers. If a new method is added for the `asyncio.Task` object, will we have 100% confidence that is fully supported in our use case? I don't think so. While not exposing the task gives us the chance and freedom of using the methods that we want and in the way that we want.

But this discussion might be an implementation detail.

L58-python-async-api.md
lidizheng: Glad you agree with modification 1.  Your modification 2 is a good catch that I overlooked the situation when users want to check `status-code` and `status-message`. It is a valid option; many frameworks are using it.

A slight drawback is that if we are going to add a `grpc.aio.Response`, we might want to also add `grpc.aio.Request`, `grpc.aio.ResponseIterator` etc.. To start a new pattern, we should be cautious and discuss them thoroughly.

To solve this problem, this is my proposal. In current design, if a call finishes abnormally (status not OK), an exception will be raised. We can use the same pattern in the new API.

```Python
try:
    response  = await stub.Hi(...)
catch grpc.RpcError as rpc_error:
    print(rpc_error.code())
    print(rpc_error.details())
```

L58-python-async-api.md
lidizheng: I think both direction is fine for me. My original intent to propose inherit from `Task` is to reduce our maintenance burden. It can cut down the boilerplates we have write in `grpc.aio.Call` class. The person who implements it can decide which way to go.

Also, there is another upside that users can use `isinstance(call, asyncio.Task)` logic.

pfreixes: I'm in favor of resolving this during the implementation :)

L58-python-async-api.md
pfreixes: Is there any other attribute that the user might need when there is no error? What about initial and trailing metadata [1]?

So just for summarizing the output of the discussion:

1) Won't be any more `__call__`, `with_call` and `future` methods, everything will be handled using the unique `__call__` available.
2) The `__call__` method will return a `grpc.aio.Call` object which would be _awaitable_, returning the body once the task is done.

am I right? If the metadata and trailing metadata are not an issue, I'm completely Ok with that proposal. 


[1] https://github.com/Skyscanner/grpc/blob/7b0576787898ee670c7a56b3fac34acfb1a62df0/src/python/grpcio/grpc/_channel.py#L387

lidizheng: Metadata does matter.

As you suggested, the metadata can be accessed with slightly different way. It separate the receiving of metadata and the entire response message. E.g. `with_call` returns only after the RPC is finished. It prevents users from reacting to `initial_metadata` if received early.

```Python
call = stub.Hi(...)
if validate(call.initial_metadata()):
    response = await call
else:
    call.cancel()
```

IMO this proposal should facilitate the usage to `Call` objects since the logic suits `asyncio` environment better.

pfreixes: umm, you mean:

```python
initial_metadata = await call.initial_metadata()
```

?

lidizheng: Good catch!

lidizheng:
@gnossen @pfreixes @argaen @rmariano Please take another look.
I have updated the design doc to include recent discussion details and some suggestions from the thread in Google Group [grpc-io](https://groups.google.com/forum/#!topic/grpc-io/7V7HYM_aph4). Following is the mapping between topics and new sections:

| No. | Topic                                           | Section                                                                                                                                                               | In this Doc?  |
| --- | ----------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------- |
| 1   | Merge `__call__`, `with_call`, `futures`        | [Unified Stub Call](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#unified-stub-call)                                       | Yes           |
| 2   | Should we eliminate interfaces?                 | [Concrete Class Instead of Interfaces](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#concrete-class-instead-of-interfaces) | Yes           |
| 3   | C-Core constraint about read/write concurrency. | [Flow Control Enforcement](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#flow-control-enforcement)                         | Yes           |
| 4   | Should we support run_in_executor?              | [Support Thread And Process Executors](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#support-thread-and-process-executors) | Yes           |
| 5   | Story for grpcio-* package under async API?     | [Other Official Packages](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#other-official-packages)                           | Yes           |
| 6   | Should we allow mixing async & sync handlers?   |                                                                                                                                                                       | Merged with 4 |
| 7   | Streaming API without iterator.                 |                                                                                                                                                                       | No yet            |
| 8   | Redesign connectivity API.                      |                                                                                                                                                                       | No yet            |
| 9   | Ease usage of channel arguments.                |                                                                                                                                                                       | No yet            |

pfreixes:
If there is enough agreement about the Unified Stub call [1], could you please create an issue in the gRPC Easy board? We can start working on this ASAP.

[1] https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#unified-stub-call

lidizheng:
@pfreixes Filed as #20001

L58-python-async-api.md
```diff
@@ -1660,6 +1660,11 @@ class StreamStreamClientInterceptor:
 
 ### Utility Functions
 
+```Python
+# grpc.aio.EOF is a unique object per process that evaluates to False
+(grpc.aio.EOF or False) == False
```
gnossen: I think it would be clearer to simply say `bool(grpc.aio.EOF) == False` or to simply list its implementation:

```python
class EOF:
    def __bool__(self):
        return False
```

lidizheng: I hope there should only be one instance of `grpc.aio.EOF`, so the code you provided should be the actual implementation class named `_EOF`. And then assign to EOF: `EOF = _EOF()`. Or we can do some crazy meta class magic here.

lidizheng:
@euroelessar @pfreixes @gnossen  PTAL. I have updated the proposal, and here are the four major sections I added in this batch. Comments, suggestions are welcomed.

| No. | Issue                                   | Section                                                                                                                                                                         |
| --- | --------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Streaming API Design                    | [New Streaming API](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#new-streaming-api)                                                 |
| 2   | API Interfaces with Type Annotations              | [API Interfaces](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#api-interfaces)                                                       |
| 3   | Context Propagation Through contextvars | [Context Propagation Through contextvars](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#implicit-context-propagation-by-contextvars) |
| 4   | Typing in Generated Code                | [Introduce Typing To Generated Code](https://github.com/lidizheng/proposal/blob/grpc-python-async-api/L58-python-async-api.md#introduce-typing-to-generated-code)               |

L58-python-async-api.md
```diff
@@ -0,0 +1,1710 @@
...
+Acceptence critiria of this feature:
+* The implimentation of such feature should supported by official package, and
+users are not expected to directly access those metadata;
+* Application logic has higher priority than the implicit propagation (e.g.
+  setting timeout explicitly will override the propagated value);
+* Application has the choice to diverse coroutine local variables;
+* The exception error string should be informative (e.g. pointing out the
+  timeout is due to upstream deadline).
+
+Further discussion around this topic, please see related sections under [Rationale].
```
pfreixes: I was expecting a first proposal about how this could be implemented :)

lidizheng: @gnossen Can you take a look?

pfreixes:
Open question, should be the interceptors functions asynchronous? This for example what is being done by `aiohttp` for the middlewares [1] - like server interceptor - and for the client tracing - like the channel interceptors [2].

Running an asynchronous application and having callables that are not asynchronous would add friction, at least in the following examples:

- If interceptors need to communicate with external services, they most likely would provide asynchronous functions. For example, for sending traces or logs.
- If interceptors need to offload the work to another task most likely it will use asynchronous interfaces, for example for adding items in a queue [3]

[1] https://docs.aiohttp.org/en/stable/web_advanced.html#example
[2] https://docs.aiohttp.org/en/stable/client_advanced.html#aiohttp-client-tracing
[3] https://docs.python.org/3/library/asyncio-queue.html

L58-python-async-api.md
gnossen: @pfreixes I have an in-progress example of an implicit context API as a branch off of https://github.com/lidizheng/grpc-api-examples. I'll prioritize getting that up. Do you think that is an appropriate place to continue this discussion. Also, based on your comment above, is it implementation that you're wondering about, or is this still an API-driven discussion?

pfreixes: Not in a rush, let's have the discussion in your PR for the `grpc-api-examples` project.

lidizheng:
@pfreixes You are correct. I should mark them as "async def". I was focused on the typing of `Callable`. I have tested that coroutine functions can be typed as `Callable`, so I put Callable here.

Thank you for catching this.

L58-python-async-api.md
```diff
@@ -0,0 +1,1708 @@
...
+from inbound metadata to outbound metadata. The tracing information will be used
+to monitor the life cycle of an request, and the length it has stayed in each
+services.
+
+For deadline propagation, the deadline of upstream server will be implicitly
+pass down to downstream server, so downstream services can react to that
+information and save computation resources or perform flow control.
+
+Acceptence critiria of this feature:
+* The implimentation of such feature should supported by official package, and
```
gnossen: s/implimentation/implementation/g

L58-python-async-api.md
```diff
@@ -0,0 +1,1708 @@
...
+
+`contextvars` is a Python 3.7 feature that allows applications to set coroutine
+local variables (just like thread local variables). In the past, gRPC Python
+uses thread local to achieve two goals:
+
+* Support distributed tracing library that work across languages;
+* Support deadline propagation along the chain of RPCs.
+
+For distributed tracing library (like OpenCensus), the tracing information is
+preserved as one of the metadata, and it is expected to be implicitly promoted
```
gnossen: Tracing should *not* be propagated by default because gRPC shouldn't have an opinion on what tracing instrumentation looks like. This needs to be pluggable. Instead, there needs to be an API by which individual tracing library authors may write their own middleware interceptors that determine how to interpret metadata into tracing data and how to convert tracing data to metadata. I will follow up with an example in grpc-api-examples on this.

L58-python-async-api.md
```diff
@@ -0,0 +1,1708 @@
...
+1) It would be quite confusing to mix functions that return `asyncio` coroutine
+   and normal Python functions;
+2) We don't want to imply that switching to use the new stack requires zero code
+   changes.
+3) Also, the current contract of API constraint ourselves from changing their
+   behavior after GA for two years.
+
+### Explicit Context vs. Implicit Context
+
+TBD
```
gnossen: TBD? Or just TODO?

L58-python-async-api.md
```diff
@@ -0,0 +1,1708 @@
...
+    response_serializer: Optional[Callable[[Any], bytes]]
+    unary_unary: Optional[Callable[[Any, grpc.aio.ServicerContext], Any]]
+    unary_stream: Optional[Callable[[Any, grpc.aio.ServicerContext], Optional[AsyncIterable[Any]]]]
+    stream_unary: Optional[Callable[[AsyncIterable[Any], grpc.aio.ServicerContext], Any]]
+    stream_stream: Optional[Callable[[AsyncIterable[Any], grpc.aio.ServicerContext], Optional[AsyncIterable[Any]]]]
+```
+
+```Python
+# grpc.aio.ServicerContext
+class ServicerContext(grpc.RpcContext):
```
gnossen: This is going to need an equivalent of [`add_callback()`](https://grpc.github.io/grpc/python/grpc.html#grpc.RpcContext.add_callback).

L58-python-async-api.md
```diff
@@ -0,0 +1,1708 @@
...
+        """
+
+    def unsubscribe(self,
+                    callback: Callable[[ChannelConnectivity], None]) -> None:
+        """Unsubscribes a subscribed callback from this Channel's connectivity.
+
+        Args:
+          callback: A callable previously registered with this Channel from
+          having been passed to its "subscribe" method.
+        """
```
euroelessar: Can it be replaced by the following API, please?
```python
def check_connectivity_state(self, try_to_connect: bool=False) -> ChannelConnectivity:
    pass

async def watch_connectivity_state(self,
                                   last_observed_state: ChannelConnectivity,
                                   timeout_seconds: float) -> ChannelConnectivity:
    pass
```

It aligns better with grpc-core API, and `subscribe`/`unsubscribe` API can be easily built on top of this primitives outside of base grpc package.

lidizheng: The API you proposed is much neat than existing version.

The only concern I have is that `subscribe`/`unsubscribe` are not marked experimental. I think we need keep it. But another problem will be introduced, since the subscribe and the new API doesn't work well together. The underlying C-Core propagation of connectivity state can only be observed by one watch. If users is using both API, their application will have undefined behavior.

As a framework, we can document that don't use both API at the same time, and potentially perform check for users.

WDYT?

lidizheng: Also, in the `watch_connectivity_state`, we need to return something when timeout, so I make it return `Optional[grpc.ChannelConnectivity]`.

euroelessar: > The only concern I have is that subscribe/unsubscribe are not marked experimental. I think we need keep it.

We're designing **the new API in entirely different namespace** here (`grpc.aio.Channel`), so promises about the old API should not be a concern.

It's not breaking any users of the old API (as they have to make explicit code changes to move to async API), and it covers them by providing the same functionality via different means. It means that **old methods can simply be removed**.

_Of course the points would be different if this proposal discusses modifications of `grpc.Channel` API, but I have an impression that it's not the case or goal._

> The underlying C-Core propagation of connectivity state can only be observed by one watch.

Based on the documentation and implementation of `grpc_channel_watch_connectivity_state` it looks like it's safe to call it by multiple users. What makes you believe that it's not the case?

> Also, in the watch_connectivity_state, we need to return something when timeout, so I make it return Optional[grpc.ChannelConnectivity].

Yeah, I agree.

lidizheng: The C-Core `grpc_channel_watch_connectivity_state` does allow multiple watcher. I was thinking about the implementation in Python wrapper, which only have one completion queue for the connectivity state propagation.

It shouldn't matter in Async IO ecosystem, since it will propagate through callback instead of `cq_next`.

I have written some application using the connectivity API, I would say in some cases the "watch" is more effective, and in some cases e.g. if you want some sort of global handler for certain connectivity state, the "subscribe" semantic is more handy.

For the sake of simplicity, probably we should remove the old API.

@gnossen WDTY about the old connectivity API?

L58-python-async-api.md
```diff
@@ -0,0 +1,1739 @@
...
+    """Affords invoking an async unary RPC from client-side."""
+
+    def __call__(self,
+                 request: Any,
+                 timeout: Optional[int]=None,
+                 metadata: Optional[Sequence[Tuple[Text, Text]]]=None,
+                 credentials: Optional[grpc.CallCredentials]=None,
+                 wait_for_ready: Optional[bool]=None,
+                 compression: Optional[grpc.Compression]=None
+    ) -> grpc.aio.Call[Any]:
```
euroelessar: 1. Please avoid `Any` in all type definitions.
Signature of this class could be this instead:
```python
class UnaryUnaryMultiCallable(Generic[Request, Response]):

    def __call__(self, request: Request, ...) -> grpc.aio.Call[Response]:
        pass
```

This allows to expose more runtime errors at compile time.

2. `metadata` should be `Sequence[Tuple[str, Union[bytes, str]]]` as grpc allows binary metadata (if suffixed by `-bin`)
3. `Text` is an alias for `str` in python3-only code, it's only useful for migrations from py2 to py3.

L58-python-async-api.md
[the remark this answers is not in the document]
euroelessar: Something like this looks easier to read:
```python
stream = stub.SayHelloStreaming()
for chunk in ["Golden", "Retriever", "Pan", "Cake"]:
  await stream.write(HelloRequest(name=chunk))
```

Why is there a need to use async iterators?

L58-python-async-api.md
```diff
@@ -0,0 +1,1739 @@
...
+        Raises:
+          An RpcError exception if the write failed.
+        """
+```
+
+### Server-Side
+
+```Python
+# grpc.aio.Server
+class Server:
```
gnossen: It would be nice if there were some way to get an [`asyncio.Event`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event) signalling termination of the server so that I can run another long-running process concurrently with the server in the event loop.

L58-python-async-api.md
gnossen: I think @euroelessar is right. The old API inverted core's pull API into a push API. This was particularly bad for the sync API because it spun up a thread outside of the [application author's control](https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_channel.py#L1011). You're always going to need another concurrent process (used in the pure concurrent computing sense, e.g. thread, fiber, coroutine, or OS process) to invert that model. This is perhaps not quite as bad in the asyncio case, since we're not preempting the application to run this poller, but it's still not ideal.

I'm leaning towards providing a pull model mirroring C-core's and letting application author's invert the model if they so choose. We can add an example of this in the examples directory.

As @euroelessar pointed out, we have no obligation to maintain "backwards compatibility" with the new set of APIs, so this change seems relatively innocuous to me.

In regards to the timeout, I'll also throw in that we could raise an exception. Either option is fine so long as all of your timeouts behave the same way across the asyncio API surface.

L58-python-async-api.md
gnossen: Discussed offline. We decided to have `server.wait_for_termination()` return a `Future` instead.

L58-python-async-api.md
lidizheng: As discussed offline, the old API will be removed, and we shall write an example to demonstrate how to implement the "subscribe" / "unsubscribe".

L58-python-async-api.md
lidizheng: As discussed offline, the functionality is similar to `asyncio.Future`, I think the goal you want to achieve can be fulfilled by calling `asyncio.ensure_future` on `server.wait_for_termination`.

L58-python-async-api.md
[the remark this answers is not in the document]
lidizheng: In the client-side, the async iterator helps existing user to adopt `asyncio` if they are using iterators today. On the other hand, the `yield` and `async for` semantic is Pythonic and useful for some use cases if the request and response are not tightly coupled locally. E.g. there is a robot application that requires uploading its sensor readings, while executes the command sent by its brain server:

```Python
async def environment_report_generator(sensors: List[RoboticSensor]):
    report = EnvironmentReport()
    for sensor in sensors:
        report[sensor.name()] = await sensor.read()
    yield report

response_iterator = stub.request_for_command(environment_report_generator(sensors))
async def command in response_iterator:
    await actuators[command.actuator_name].perform(command.actuator_command)
```

L58-python-async-api.md
lidizheng: Good point! Changed to `typing.AnyStr`.

euroelessar: I'm not sure `typing.AnyStr` is correct, e.g. this input is valid but will be rejected by typing system:
```python
metadata=[
  ("foo", "text"),
  ("boo-bin", b"binary"),
],
```

L58-python-async-api.md
```diff
@@ -0,0 +1,1739 @@
...
+
+To keep the function signature stable, the new streaming API only requires
+enhancement to the context object (`grpc.aio.Call` on client-side, and
+`grpc.aio.ServicerContext` on the server-side).
+
+Notably, both new and current streaming API allows read and write messages
+simultaneously in different thread or coroutine. However, **the application should**
+**not mixing new and current streaming API at the same time**. Reading or writing to
+the iterator and context object might have synchronization issue that the
+framework does not guaranteed to protect against.
```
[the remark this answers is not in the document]
euroelessar: Thinking more about it, the following snippet provides the behavior which is very close to the current stream-stream api:
```python
def async_stream_stream(stream: AsyncStream[Request, Response],
                        iter: AsyncIterator[Request]) -> AsyncIterator[Response]: 
  asyncio.ensure_future(async_iterator_to_stream(stream, iter))
  return stream_to_async_iterator(stream)

### Usage
with stub.StreamingHi() as streaming_call:
  request_iterator: AsyncIterator[echo_pb2.EchoRequest] = ...
  async for response in async_stream_stream(stream, request_iterator):
    process(response)
```

Cancellation semantics are unclear, but one could argue that they are not very well defined in old iterator-based api as well.

L58-python-async-api.md
```diff
@@ -0,0 +1,1739 @@
...
+```Python
+# grpc.aio.UnaryUnaryClientInterceptor
+class UnaryUnaryClientInterceptor:
+    """Affords intercepting unary-unary invocations."""
+
+    async def intercept_unary_unary(self,
+                              continuation: Callable[[grpc.ClientCallDetails, Any], grpc.aio.Call[Any]],
+                              client_call_details: grpc.ClientCallDetails,
+                              request: Any
+        ) -> grpc.aio.Call[Any]:
```
pfreixes: Returning an `awaitable` object from a coroutine could be problematic. Since we agreed on returning the `Call` object the  for the `UnaryUnaryMultilCallable.__call__` method, it has needed to be changed, becoming a normal function, like:

```python
    def __call__(self,
                 request,
                 timeout=None,
                 metadata=None,
                 credentials=None,
                 wait_for_ready=None,
                 compression=None):

        if timeout:
            raise NotImplementedError("TODO: timeout not implemented yet")

        if metadata:
            raise NotImplementedError("TODO: metadata not implemented yet")

        if credentials:
            raise NotImplementedError("TODO: credentials not implemented yet")

        if wait_for_ready:
            raise NotImplementedError(
                "TODO: wait_for_ready not implemented yet")

        if compression:
            raise NotImplementedError("TODO: compression not implemented yet")

        call = asyncio.ensure_future(
            self._channel.unary_unary(self._method, _common.serialize(request, self._request_serializer)))

        return Call(call, self._response_deserializer)
```

Using a coroutine, instead of a function, does not work as we would like since the awaitable object is never awaited. For example, the following line would await the coroutine but never the awaitable object

```python
response = await stub.hi(...)
```

Because the interceptors are implemented with coroutines, they would suffer from that problem where the client will be receiving as a response of the previous statement a `Call` object and never performing the proper and expected _await_..

So we would need to find a solution, either by changing how the `Call` object is being gathered or addressing the issue, which could not be fixable.

L58-python-async-api.md
```diff
@@ -0,0 +1,1741 @@
...
+#### Snippet For New Streaming API
+
+```Python
+### Client side
+with stub.StreamingHi() as streaming_call:
+  request = echo_pb2.EchoRequest(message="ping")
+  await streaming_call.write(request)
+  response = await streaming_call.read()
+  while response:  # or response is not grpc.aio.EOF
+      process(response)
```
jtattermusch: the pattern to read a streaming response seems pretty dirty (2 copies of await streaming_call.read() seems like something that should be avoided).
Does python have a concept of "asynchronous iterator" (you should use it if so).

for comparison, C# uses this interface https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/IAsyncStreamReader.cs (and it the new versions of C#, there's even an async enumerable which allows you to do an asychronous foreach over the collection of responses).
Also, python seems to be exposing the read() and write() operations directly on the call object, while C# defines an "request stream" and "response stream" objects.

L58-python-async-api.md
```diff
@@ -0,0 +1,1741 @@
...
+So, this gRFC introduces a new pattern of streaming API that reads/writes
+message to peer with explicit call.
+
+#### Snippet For New Streaming API
+
+```Python
+### Client side
+with stub.StreamingHi() as streaming_call:
+  request = echo_pb2.EchoRequest(message="ping")
+  await streaming_call.write(request)
```
jtattermusch: is this a unary only request or a streaming request?
If streaming, I don't see a way of expressing that client is done sending requests.

See https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/IClientStreamWriter.cs and https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/IAsyncStreamWriter.cs for comparison.

L58-python-async-api.md
```diff
@@ -0,0 +1,1741 @@
...
+### Server side
+class AsyncGreeter(helloworld_pb2_grpc.GreeterServicer):
+
+    async def StreamingHi(self,
+                          unused_request_iterator,
+                          context: grpc.aio.ServicerContext
+            ) -> None:
+        bootstrap_request = await context.read()
+        initialize_environment(bootstrap_request)
+        while has_response():
```
jtattermusch: why on serverside, when reading a stream, we invoke `had_response()` and on the client side (example above) we do `while response:` and just evaluate the `response` itself? The way you read a stream of messages should be consistent IMHO.

L58-python-async-api.md
```diff
@@ -0,0 +1,1741 @@
...
+
+### Server side
+class AsyncGreeter(helloworld_pb2_grpc.GreeterServicer):
+
+    async def StreamingHi(self,
+                          unused_request_iterator,
+                          context: grpc.aio.ServicerContext
+            ) -> None:
+        bootstrap_request = await context.read()
+        initialize_environment(bootstrap_request)
```
jtattermusch: qq: what is the initialize_environment() good for?

lidizheng: `initialize_environment` is a pseudo code for application logic, it is not part of the API surface.

L58-python-async-api.md
lidizheng: `has_response` is a pseudo code for application to decide whether it want to write another response or not. For checking incoming message, let's discussion in above thread.

L58-python-async-api.md
lidizheng: > 2 copies of await streaming_call.read() seems like something that should be avoided

This is the issue for the snippet I provide. For usage of new streaming API, it should not be bound with simple loop, but more complex use case.

> Does python have a concept of "asynchronous iterator"?

Yes, in example 2, I provided the complex version of async iterator. For simpler case, it looks like this:

```Python
# Client-side
response_iterator = stub.StreamingHi(async_request_iterator)

async for response in response_iterator:
    process(response)
```
```Python
# Server-side
class Greeter(helloworld_pb2_grpc.GreeterServicer):
    async def StreamingHi(self, request_iterator, context):
        async for request in request_iterator:
            yield response
```

lidizheng: Split the call object is a good idea. Let me put more thought about it.

L58-python-async-api.md
lidizheng: @pfreixes I think this can be done by schedule the call to event loop directly, then wrap the returned future object as our `grpc.aio.Call` object. The implementation will be similar to:
```Python
class UnaryUnaryMultiCallable(...):

    def __call__(self, ...):
        task = asyncio.event_loop.create_task(_perform_actual_unary_unary_call())
        return grpc.aio.Call(task)
```

My biggest concern is about the performance impact, since the "create_task" may introduce more AsyncIO overhead to our framework. But turned out, the impact is not as significant as I thought.

```
lidiz-macbookpro:test-python-asyncio lidiz$ python3.7 coro_overhead.py
Spent 0.10s for 10000 runs of directly_await
10170.10 ns
Spent 0.10s for 10000 runs of schedule_task
10017.37 ns
Spent 14.91s for 1000000 runs of directly_await
14913.95 ns
Spent 15.51s for 1000000 runs of schedule_task
15512.92 ns
```

The snippet I used to run the benchmark:

```Python
import asyncio
import timeit
import time

async def work():
    return "I am not working"

async def directly_await():
    await work()

def schedule_task():
    return asyncio.create_task(work())

async def main():
    number = 10000
    start_time = time.time()
    await asyncio.gather(*(directly_await() for i in range(number)))
    elapsed_time = time.time() - start_time
    print("Spent %.2fs for %d runs of %s" % (elapsed_time, number, 'directly_await'))
    print("%.2f ns" % (elapsed_time / number * 1e9))
    
    start_time = time.time()
    await asyncio.gather(*(schedule_task() for i in range(number)))
    elapsed_time = time.time() - start_time
    print("Spent %.2fs for %d runs of %s" % (elapsed_time, number, 'schedule_task'))
    print("%.2f ns" % (elapsed_time / number * 1e9))


if __name__ == "__main__":
    asyncio.run(main())
```

L58-python-async-api.md
lidizheng: Can I ask for what went wrong about the `typing.AnyStr`? I wrote a snippet, and `pytype` is happy with it.

```Python
from typing import Text, AnyStr, Sequence, Tuple

metadata=[
  ("foo", "text"),
  ("boo-bin", b"binary"),
]

def process(metadata: Sequence[Tuple[Text, AnyStr]]):
    for key, value in metadata:
        print(f"I'm okay with ({type(key)} {type(value)})")

process(metadata)
```

L58-python-async-api.md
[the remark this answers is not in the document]
lidizheng: @euroelessar I agree with you that the two streaming API is mutual convertible. It is easy to convert stream-object-based API to async-iterator-based API, vice versa.

In the implementation, we can focus on one of them, and another API is just a wrapper around it. Natively, the stream-object-based API is more similar to what we are doing in C-Core, so I think the async-iterator-based API will be a wrapper.

> It increases API surface, which will have to be maintained ≈forever.

Iterator, generator is one of the core feature for Python language, I think to some extend, we should support them. Also, it makes simple use case looks really nice (Pythonic). 

> It introduces two different ways to perform the same logic and documentation has to ask user nicely to not mix them.

The same logic applies to try to create async iterator using stream-object-based API (user have to stick to one of them). For underlying implementation, it is okay to mix them, as long as user provides proper synchronization around it.

> It's easy to re-construct old-style API on top of stream-based primitives in a generic way without codegen.

I agree. One problem I haven't figure out is how do we deal with the function signature of existing code. If we remove the async-iterator, then users method handlers accepts 2 arguments, and we only provide 1. Their code will break in this case.

---

Appendix: Convert async-iterator-based API to stream-object-based API.

```Python
class Reader(Genric[Message]):
    def __init__(self, async_iterator: AsyncIterator[Message]):
        self._async_iterator = async_iterator

    async def reader(self) -> Message:
        return await self._async_iterator.__anext__()

class Writer(Genric[Message]):
    def __init__(self):
        self._queue = asyncio.queue()

    async def write(self, message: Message):
        await self._queue.put(message)

    async def __anext__(self) -> Message:
        return await self._queue.get()

    def __aiter__(self) -> AsyncIterator[Message]:
        return self
```

L58-python-async-api.md
pfreixes: Yeps that's the plan, either by calling `ensure_future` and wrapping the output within our own `Call` object or by directly calling the `create_task`. IMO they are basically two ways of doing the same which both have little diferents.

The concern is more about where the interceptor code path should live once we have the unified call implemented, the whole problem and a likely solution has been described here [1]. TL;DR of my proposal would be having the interceptors code path within the task.

Yes, maybe its an issue for the future but seeing this as a likely blocker for the future  it's worth it to have everybody aligned since now.

[1] https://github.com/grpc/grpc/issues/20001#issuecomment-539712798

L58-python-async-api.md
jtattermusch: Yes, the client can tell the server that it's done sending messages. The server will receive a notification (a pending/new read on the server side finishes with "no message"/end of stream), but server can still continue sending responses.  `done_writing()` is ok as a name (I can't speak on how idiomatic that name is in python).  `half_close()` is a bad name because half-close is a HTTP/2 terminology and it's a technical detail that should not be exposed to the end user.

More remarks:
- server of course cannot half close, it can only finish the RPC with a status (which terminates both request and response streams). That also means that the interface for sending  on the client side and server side are going to be different (see the difference in c#) and only client is going to provide `done_writing()`.

Btw, in C# it is customary to add "Async" suffix to names of all async methods, to make it obvious they should be awaited - python doesn't have a style that requires that?

L58-python-async-api.md
```diff
@@ -0,0 +1,1666 @@
...
+  stop the execution of current coroutine, but `call.write` can be scheduled in
+  event loop);
+* Enable application to have both pending read and pending write in one
+  coroutine, and react on the first finisher (with iterator syntax this will be
+  quite hard to get it right).
+
+So, this gRFC introduces a new pattern of streaming API that reads/writes
+message to peer with explicit call.
+
+#### Snippet For New Streaming API
```
jtattermusch: maybe I'm missing something, but the snippet for "new streaming API" looks  way more confusing and less fluent than the "current streaming API", so I'm a bit lost.
Btw, you're talking about "current streaming API" but you're talking about an API that AFAIK doesn't exist yet and it's being proposed in the proposal - hence my confusion.

L58-python-async-api.md
jtattermusch: I just think that that initialize_environment() and has_response() are very confusing - can you make the example clearer to make it obvious what they symbolize?  (or at least add a comment what the operations mean?)

L58-python-async-api.md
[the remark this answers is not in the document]
euroelessar: @lidizheng 

> In the implementation, we can focus on one of them, and another API is just a wrapper around it. Natively, the stream-object-based API is more similar to what we are doing in C-Core, so I think the async-iterator-based API will be a wrapper.

Great, it sounds we have an agreement?

> I agree. One problem I haven't figure out is how do we deal with the function signature of existing code. If we remove the async-iterator, then users method handlers accepts 2 arguments, and we only provide 1. Their code will break in this case.

Users have to already mark their code as `async def` so there is a change in the signature.

Can the issue be slightly mitigated by providing a decorator for implementing handler iterator-based api? For example, allow marking endpoints with `@grpc_playground.asyncio_iterator_handler`. Decorator can construct request iterator from the stream object & convert response iterator into `.write` calls.

L58-python-async-api.md
lidizheng: `done_writing()` added to the client-side context object.

#### About the "Async" suffix
Python doesn't have that hard requirement, and official modules do not follow that pattern. I have read articles about enforcing the requirement, and I think it may get messier once we introduce that pattern through out our API due to:

* How to name async getters?
* How to name functions that return an `Awaitable` object (like future)?

The function name might get much longer.

L58-python-async-api.md
lidizheng: I replaced them with `...` to avoid future confusion.

L58-python-async-api.md
lidizheng: I renamed it into `Reader / Writer` API for the new one, and `Async Iterator` for the existing one. Hope this will reduce the confusion.

jtattermusch: Much better! Thanks.

L58-python-async-api.md
jtattermusch: It's still not crystal clear IMHO, but it's an improvement.

Looking at the two different APIs, I'm not sure if there's actually any significant difference
in the way they read:
- first example uses context.read() and doesn't use the `unused_request_iterator` at all
- the second example uses the async request iterator

instead of context.read(),   I could just use `anext(request_iterator)` to read the messages one by one and the result will be exactly the same, right? If so, does it make sense to expose the context.read() method at all?

I agree that for writing (on the server side), just relying on async generator (`yield`) would be too clumsy for many applications, so providing 2 ways of writing makes sense. QQ: how does the python server knows if that you've chosen API1(context.write() or API2 (yield) for given method?

jtattermusch: btw, IIRC the "bootstrap request" in example 1 is there to demonstrate that the first API is there to support more complex scenarios (like the first request in the stream being of a different kind ~ perhaps in the 'bootstrap' sense), but it's not obvious from the example that this is the point. perhaps the example can contain a comment with that point `# Reader/Writer API allows to handle more complex semantics, such as handling the first request as "bootstrap" request and then processing other requests according to the configuration`. Btw that particular scenario is relatively easy to handle with `yield` as well so perhaps this might not be the best example to use.

```
# async generator with bootstrap request (pretty easy to implement)
def async_generator:
    async for req in request_iterator:
         if is_first_request:
             bootstrap(req)
            is_first_request = False
         else:
            process_request()
```

I think the `yield` model might be difficult to use when there's not a 1 to 1 correspondence between requests and responses or when there's more logic involved (interaction with other internal components or calling into other services) - just mentioning that if we want to demonstrate what each of the APIs is useful for, we should use examples that really reflect that.

L58-python-async-api.md
```diff
@@ -0,0 +1,1673 @@
...
+
+#### Snippet For Reader / Writer Streaming API
+
+```Python
+### Client side
+with stub.StreamingHi() as streaming_call:
+  request = echo_pb2.EchoRequest(message="ping")
+  await streaming_call.write(request)
+  response = await streaming_call.read()
+  while response:  # or response is not grpc.aio.EOF
```
jtattermusch: nit: EOF = end of file, perhaps EOS is more fitting? (I have zero context here, so feel free to ignore me).

L58-python-async-api.md
```diff
@@ -0,0 +1,1673 @@
...
+      response = await streaming_call.read()
+
+### Server side
+class AsyncGreeter(helloworld_pb2_grpc.GreeterServicer):
+
+    async def StreamingHi(self,
+                          unused_request_iterator,
+                          context: grpc.aio.ServicerContext
+            ) -> None:
+        bootstrap_request = await context.read()
```
jtattermusch: thought: having access to context.read() might not be necessary as (IMHO) anext(request_iterator) does exactly the same thing, but it it might make sense to have servicer_context.read() for consistency as on the client side, one can read using call.read() (and there's no iterator access available on the client side?)

L58-python-async-api.md
```diff
@@ -0,0 +1,1673 @@
...
+iterator-based API needs to stay. So, both new and current streaming API needs
+to be included in the surface API.
+
+To keep the function signature stable, the new streaming API only requires
+enhancement to the context object (`grpc.aio.Call` on client-side, and
+`grpc.aio.ServicerContext` on the server-side).
+
+Notably, both new and current streaming API allows read and write messages
+simultaneously in different thread or coroutine. However, **the application should**
+**not mixing new and current streaming API at the same time**. Reading or writing to
```
jtattermusch: Even though mixing the two APIs is not supported, it might make sense to implement in a way that this is possible where the implementation would be simple enough (e.g.   servicer.read() could just call anext(request_iterator) and that's it?).

L58-python-async-api.md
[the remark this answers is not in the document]
gnossen: Let me add my two cents. I think having to use a decorator on your handler is heavier a burden than I'd like users to have to take on. They are demonstrating their intent to use an iterator by simply using the iterator. Ideally, they shouldn't have to do anything more than that.

How about always providing two arguments (the request iterator and the context), each with their own fixed, uncomplicated type, and simply raising an exception if a handler ever tries to use both at once? I think this strikes a happy medium between good, unsurprising typing and aesthetics/usability.

L58-python-async-api.md
```diff
@@ -0,0 +1,1673 @@
...
+2. Generating an additional file for the typed code.
+
+## Rationale
+
+### Compatibility With Other Asynchronous Libraries
+
+By other asynchronous libraries, they mean libraries that provide their own
+Future, Coroutine, Event-loops, like `gevent`/`Twisted`/`Tornado`. In general,
+it is challenging to support multiple async mechanisms inter-operate together,
+and it has to be discussed case by case.
```
danielgtaylor: I would also consider the newer async libraries like [Curio](https://github.com/dabeaz/curio) and [Trio](https://github.com/python-trio/trio). Each has thousands of stars and is an alternative to `asyncio`.

L58-python-async-api.md
```diff
@@ -0,0 +1,1673 @@
...
+
+Typing makes a difference. This gRFC intended to introduce typing to gRPC Python
+as much as we can. To change generated code, we need to update the gRPC protoc
+plugin to generate Python 3 stubs and servicers. To date, **this gRFC intended to**
+**make the new async API compatible with existing generated code**. So, the Python 3
+generated code will have two options:
+
+1. Add a third generation mode to protoc plugin that generates stubs / servicers
+   with typing;
+2. Generating an additional file for the typed code.
```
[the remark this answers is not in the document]
lidizheng: Before we dive into details, let me point to https://github.com/lidizheng/grpc-api-examples/pull/1 where we are having discussion about how typing should be done in AsyncIO.

I found `grpclib` is using `grpc_tools` which is the `protoc` plugin maintained by grpc/grpc repo. Current code generation doesn't support typing, so it allows application to pass in any type of messages. **In near future, we are planning to support stubs from existing generated code**, so applications should have the choice to between old stubs and new stubs.

@gnossen who might be interested in the usability improvement of `betterproto`.

`dataclass` and `Protocol` are great 3.8 features, maybe this effort can include them as well, some day.

L58-python-async-api.md
lidizheng: +@euroelessar who is the advocator of reader / writer API.

Sorry I didn't response timely, I think GitHub is not ideal for long threads like this. Critique gets one point for complex reviews.

> instead of context.read(), I could just use anext(request_iterator) to read the messages one by one and the result will be exactly the same, right? If so, does it make sense to expose the context.read() method at all?

One month ago, Eric, Richard and I discussed this specific design. Eventually, the consistency prevails the redundancy. For new users who want to use stream API alone, this design makes them only considering the stream object itself, instead of switching the mental model between async generator and reader / writer.

> QQ: how does the python server knows if that you've chosen API1(context.write() or API2 (yield) for given method?

Yes, it does. If the function uses `yield`, it will return a generator (async generator for `async def`). We might need some reflection here, but I would say it is feasible.

> A better example

I'm trying to come up with an example that is extremely helpful when using reader / writer model, but haven't found a simple & valid one  yet. The example proposed in above comment is using the async generator API, which is not the focus point for the first example. The first example is meant to promote the usage of reader / writer API.

> I think the yield model might be difficult to use when there's not a 1 to 1 correspondence between requests and responses or when there's more logic involved

Yes, that is the main motivation of introducing reader / writer API.

> if we want to demonstrate what each of the APIs is useful for, we should use examples that really reflect that.

I agree. I always welcome suggestions 👍

lidizheng: Second thought, about the `unused_request_iterator`, it can be removed if we are generating new servicer code. Currently, we don't have a detailed designed yet.

L58-python-async-api.md
```diff
@@ -0,0 +1,1752 @@
...
+    async def SayHello(self,
+                       request: helloworld_pb2_grpc.HelloRequest,
+                       context: grpc.aio.ServicerContext
+                ) -> helloworld_pb2_grpc.HelloReply:
+        await asyncio.sleep(1)
+        return helloworld_pb2.HelloReply(message="Hello, %s!" % request.name)
+
+server = grpc.aio.server()
+server.add_insecure_port(":50051")
+helloworld_pb2_grpc.add_GreeterServicer_to_server(Greeter(), server)
```
argaen: Shouldn't it be `AsyncGreeter()`?

lidizheng: Good catch!

L58-python-async-api.md
```diff
@@ -0,0 +1,1752 @@
...
+            status.
+          client_call_details: A ClientCallDetails object describing the
+            outgoing RPC.
+          request: The request value for the RPC.
+
+        Returns:
+            An object that is both a Call for the RPC and an iterator of
+            response values. Drawing response values from the returned
+            Call-iterator may raise RpcError indicating termination of
+
```
rmariano: Should we also mention the presence of new exception classes, like ``AioRpcError`` ?

L58-python-async-api.md
```diff
@@ -0,0 +1,1752 @@
...
+    async def StreamingHi(self, request_iterator, context):
+        async for request in request_iterator:
+            yield response
+```
+
+#### Co-existence Of New And Current Streaming API
+
+Existing API still has advantage over simpler use cases of gRPC, and iterator
+syntax is Pythonic to use. Also, for backward compatibility concern,
+iterator-based API needs to stay. So, both new and current streaming API needs
```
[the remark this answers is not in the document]
lidizheng: Fixed. Good catch!

L58-python-async-api.md
```diff
@@ -0,0 +1,1752 @@
...
+stub.SayHelloStreaming(iter([
+    HelloRequest(name='Golden'),
+    HelloRequest(name='Retriever'),
+    HelloRequest(name='Pan'),
+    HelloRequest(name='Cake'),
+]))
+
+### The new usage is much verbose for same scenario
+async def request_iterator():
+    for request from [
```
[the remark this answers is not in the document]
lidizheng: Good catch!

L58-python-async-api.md
lidizheng: Good idea. I adding a section description for exceptions that on our road map: `grpc.BaseError`, `grpc.AbortError`, and `grpc.UsageError`.

As for `grpc.RpcError`, existing design of the `grpc.RpcError` class is empty (no methods / attributes). It generates confusion for many new gRPC users that it is unclear why the caught exception type doesn't contain any information. I hope at some point, we should promote the design of `AioRpcError` to `grpc.RpcError`.

ntmartin:
Hi, any update on when this might be merged?

wintonzheng:
@lidizheng 

Hey we see this [helloworld example](https://github.com/grpc/grpc/blob/d6298ed3dc12641c48f53cbeabe09e130e00a01a/examples/python/helloworld/async_greeter_server.py#L26-L27) that does the unary_unary rpc. is there a working example for unary_stream, stream_unary or stream_stream?

we also saw [this example](https://github.com/lidizheng/grpc-api-examples/blob/master/Examples.md#stream-stream-handler) but it's also pseudo code. We could really benefit from some real working code if there is any.

Thanks

lidizheng:
@wintonzheng We are still working on examples. For now, you can check our unit tests for API usage, e.g., https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests_aio/unit/call_test.py.

PR #156, opened 2019-08-19
L59 - Allow C++ standard library in gRPC Core Library

veblush:
This is a proposal to allow C++ standard library in gRPC Core Library and gRPC Core will be full C++11  compliant with this.

[Discussion Thread](https://groups.google.com/forum/#!topic/grpc-io/umQyyXmNr2w)

### DIFF
--- a/L59-core-allow-cppstdlib.md
+++ b/L59-core-allow-cppstdlib.md
@@ -1,0 +1,97 @@
+Allow C++ standard library in gRPC Core Library
+----
+* Author(s): veblush
+* Approver: a11r
+* Status: Draft
+* Implemented in: n/a
+* Last updated: August 19, 2019
+* Discussion at: <TBD>
+
+## Abstract
+
+Allow C++ Standard Library to be used in the gRPC Core library.
+
+## Background
+
+Since C++ was allowed to write gRPC, more and more code has been written in C++.
+But most features from C++ standard libraries were not allowed to use because
+we didn't want to introduce new dependency against C++ standard library.
+This gRFC proposes allowing C++ standard library to be used within the core
+library, but continues to restrict the public interface to C89.
+
+### Related Proposals:
+
+* [Allow C++ in gRPC Core Library](L6-core-allow-cpp.md)
+
+## Proposal
+
+Allow C++11 standard library usage in gRPC Core. The previous C++ usage rule
+will be valid except the restriction on C++ features and header-only library.
+
+- All C++11 features are allowed to use. Followings were not possible to use
+  but it becomes availabie with this.
+  - `new` and `delete`
+    - Even though `new` and `delete` now become possible to use, it doesn't
+      necessarilly mean that all code should use.
+      For the place where `gpr_malloc` and `gpr_free` should be used
+      intentionally, `grpc_core::New` and `grpc_core::Delete` need to be
+      used.
+  - Pure virtual functions.
+    Regular form `= 0` can be used instead of `GRPC_ABSTRACT`
+
+- All features from C++11 standard library are allowed to use.
+  Previosuly only header-only features were allowed to use.
+
+There are caveats since gRPC wrapped library are being distributed as a binary
+form and still there are many old but active linux distributions which don't
+have full C++11 standard library.
+
+- Only features from C++ standard library available on
+  [manylinux1](https://www.python.org/dev/peps/pep-0513)
+  can be allowed to use. It sounds a bit disappointing but most of essential
+  features in C++ standard library are available in `manylinux1`.
+  Following is a partial list of missing features because of `manylinux1`.
+  - std::chrono.
+  - std::numeric_limits for long long and some methods.
+  - string and stream support for wchar_t.
+  - new implementation of string and list compliant with C++11.
+    (still old implementation of string and list are available)
+
+## Rationale
+
+Using C++ standard library gives us several advantages:
+- gRPC Core can use all useful and essential classes in C++ standard library.
+  This allows us not to reinvent everything which are already available.
+- gRPC Core can use other libraries which requires C++ standard library.
+  Previously linking to abseil or protobuf was impossible because it requires C++
+  standard library. With this, we can consider it.
+
+All previous restrictions not to link the library are arbitrary and it can
+change anytime because it's not part of standard or specification but
+implementation details. Having gRPC built by `gcc` and `libstdc++` doesn't
+guarantee that it can achieve the same goal with `clang` and `libc++`.
+
+## Implementation
+
+1. Change build configuration for all wrapped language to add the dependency
+   to C++ standard library.
+2. Make changes to use C++ standard library to prove it's working.
+   At the same time, it should be small enough so that they are easily
+   rolled back just in case
+   - Replace `grpc_core::map` with `std::map`
+   - Replace `GRPC_ABSTRACT` with `= 0`
+3. Add new build tests to make sure that the version of library gRPC uses
+   is old enough for most platforms to have it.
+   - For Python and Ruby, manylinux1 is a good candidate to target.
+
+## Schedule
+
+1. New dependency against the C++ standard library is added with
+   some code depending it. This has to be easily rolled back just in case.
+   -> Target gRPC 1.25 (schedulled on October 22, 2019)
+2. Once this version looks fine, all gRPC Core now is allowed to use the library.
+   -> From gRPC 1.26-dev
+
+## Open issues (if applicable)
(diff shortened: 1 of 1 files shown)


lidizheng:
As @veblush wrote, depending on C++ std works for `manylinux1`. In the past, we have once encountered a distribution issue related to standard libraries is for Python 2 build on Windows. We tried to use `stdint.h`, but it doesn't exist in VC++ 2008. It is a fairly old version, but the only working Python 2.7 C extension compiler in Windows.

If we want to keep releasing Python 2.7 wheels for Windows, we might need to prevent such breakage. Existing [prod:grpc/core/master/windows/grpc_distribtests](https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:grpc%2Fcore%2Fmaster%2Fwindows%2Fgrpc_distribtests?e=-TargetLevelInsightsEnabled&jsmode=du) fusion tests may help.

chwarr:
Can you please add a discussion link? I have some comments that aren't typographic.

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,97 @@
...
+   - Replace `GRPC_ABSTRACT` with `= 0`
+3. Add new build tests to make sure that the version of library gRPC uses
+   is old enough for most platforms to have it.
+   - For Python and Ruby, manylinux1 is a good candidate to target.
+
+## Schedule
+
+1. New dependency against the C++ standard library is added with
+   some code depending it. This has to be easily rolled back just in case.
+   -> Target gRPC 1.25 (schedulled on October 22, 2019)
```
jtattermusch: nit: s/schedulled/scheduled/

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,97 @@
...
+
+* [Allow C++ in gRPC Core Library](L6-core-allow-cpp.md)
+
+## Proposal
+
+Allow C++11 standard library usage in gRPC Core. The previous C++ usage rule
+will be valid except the restriction on C++ features and header-only library.
+
+- All C++11 features are allowed to use. Followings were not possible to use
+  but it becomes availabie with this.
```
jtattermusch: s/availabie/available/

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,97 @@
+Allow C++ standard library in gRPC Core Library
```
jtattermusch: nit: I saw several typos in this doc, doing a quick pass on language would be useful.

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,97 @@
...
+* Last updated: August 19, 2019
+* Discussion at: https://groups.google.com/forum/#!topic/grpc-io/umQyyXmNr2w
+
+## Abstract
+
+Allow C++ Standard Library to be used in the gRPC Core library.
+
+## Background
+
+Since C++ was allowed to write gRPC, more and more code has been written in C++.
```
jtattermusch: Can we use language that focuses more on what the motivation and rationale (dev productivity etc.) rather than describing dry facts of what happened/is going to happen?

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,97 @@
...
+      For the place where `gpr_malloc` and `gpr_free` should be used
+      intentionally, `grpc_core::New` and `grpc_core::Delete` need to be
+      used.
+  - Pure virtual functions.
+    Regular form `= 0` can be used instead of `GRPC_ABSTRACT`
+
+- All features from C++11 standard library are allowed to use.
+  Previosuly only header-only features were allowed to use.
+
+There are caveats since gRPC wrapped library are being distributed as a binary
```
jtattermusch: we should describe the full story behind the caveat.
- wrapped languages' packages contain binary pre-compiled version of core
- packages need to be installable on a wide range of platforms without the user needing to install any extra packages (e.g. apt-get install some libraries).
- we must only use a libstdc++ version that's available on all of the platforms we supporting (which happens to be manylinux1?  which libstdc++ version is that?).
-

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,122 @@
...
+
+Allow C++ Standard Library to be used in the gRPC Core library.
+
+## Background
+
+Since C++ was allowed to write gRPC, more and more code has been written in C++.
+But most features from C++ standard libraries were not allowed to use because
+we didn't want to introduce new dependency against C++ standard library.
+
+The decision to use C++ has been helping us to be more productive but it's not
```
jtattermusch: Rephrase the two paragraph with this (has the same meaning but it more condensed)?

In the past, we allowed the use of C++ code in the gRPC Core library (link to the proposal), but we stayed conservative and avoided introducing a dependency on C++ standard library. Being able to use C++ has lead to increased productivity, but there are still significant limitations, because many useful C++ features depend on the standard C++ library.

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,122 @@
...
+Core library. It may free us from reinventing wheels which already exists
+in the standard library and help us become more productive.
+
+### Related Proposals:
+
+* [Allow C++ in gRPC Core Library](L6-core-allow-cpp.md)
+
+## Proposal
+
+Allow C++11 standard library usage in gRPC Core. The previous C++ usage rule
```
jtattermusch: typo: s/rule/rules/

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,122 @@
...
+will be valid except the restriction on C++ features and header-only library.
+
+- All C++11 features are allowed to use. Followings were not possible to use
```
jtattermusch: rephrase as:
"Following C++11 features will become available with this proposal"?

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,122 @@
...
+    users don't need to worry about this.
+
+Simply speaking, we can use C++11 library features which are available
+in `manylinux1`. For details on `manylinux`, you can see the 
+[doc](https://www.python.org/dev/peps/pep-0513).
+Here is what matters to C++.
+
+  - GLIBCXX <= 3.4.9 (GCC 4.2.0)
+
+Fortunately, most of C++ standrad library availble on this restriction.
```
jtattermusch: s/availble/available/

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,122 @@
...
+* [Allow C++ in gRPC Core Library](L6-core-allow-cpp.md)
+
+## Proposal
+
+Allow C++11 standard library usage in gRPC Core. The previous C++ usage rule
+will be valid except the restriction on C++ features and header-only library.
+
+- All C++11 features are allowed to use. Followings were not possible to use
+  but it becomes available with this.
+  - `new` and `delete`
```
jtattermusch: it's not immediately clear what's the value of being able  `new` and `delete`. a brief explanation would be useful.

L59-core-allow-cppstdlib.md
```diff
@@ -0,0 +1,122 @@
+Allow C++ standard library in gRPC Core Library
```
jtattermusch: I did another pass on this document and I was able to get the info I needed (I think), but overall I felt the readability of the doc could be improved. Since the gRFCs are basically a part of official gRPC documentation, I'd recommend getting a language review before merging the final version of the doc.

L59-core-allow-cppstdlib.md
veblush: Thank you for the better writing :) I'll update the doc with this.

L59-core-allow-cppstdlib.md
veblush: Done

L59-core-allow-cppstdlib.md
veblush: Done.

L59-core-allow-cppstdlib.md
veblush: Done.

L59-core-allow-cppstdlib.md
veblush: I added the benefit.

L59-core-allow-cppstdlib.md
veblush: I really appreciate your thorough review! I'll reach out to the language reviewer.

PR #159, opened 2019-09-15
delete a "," ,because json format without a ,

abserari:
in Integration with Service Config

          "method": string,   
-> 
          "method": string


yhyddr:
@dfawley you are right, I do it 

and could you please help my these codes about retry example https://github.com/yhyddr/grpc-go/blob/master/examples/features/retry/client/main.go#L93

I'm confused about these set serviceConfig now because which I set seems not be available.

abserari:
could it merge or close?

PR #162, opened 2019-10-08
L60: Remove custom allocation function overrides from core surface API

veblush:
I'll add a link to the discussion once initial review is done.

### DIFF
--- a/L60-core-remove-custom-allocator.md
+++ b/L60-core-remove-custom-allocator.md
@@ -2,7 +2,7 @@ Remove custom allocation function overrides from core surface API
 ----
 * Author(s): veblush
 * Approver: mark
-* Status: Draft
+* Status: Approved
 * Implemented in: https://github.com/grpc/grpc/pull/20462
 * Last updated: 2019-10-09
 * Discussion at: https://groups.google.com/forum/#!topic/grpc-io/kDswzg-cEZU
--- a/L60-core-remove-custom-allocator.md
+++ b/L60-core-remove-custom-allocator.md
@@ -1,0 +1,91 @@
+Remove grpc_use_signal from core surface API
+----
+* Author(s): veblush
+* Approver: mark
+* Status: Draft
+* Implemented in: https://github.com/grpc/grpc/pull/20462
+* Last updated: 2019-10-08
+* Discussion at: TBD
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


L60-core-remove-custom-allocator.md
```diff
@@ -0,0 +1,91 @@
+Remove grpc_use_signal from core surface API
```
markdroth: Looks like the title needs to be updated.

veblush: Oops. Done!

L60-core-remove-custom-allocator.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Proposal
+
+Following functions managing custom memory allocator will be removed.
+
+- `gpr_get_allocation_functions`
+- `gpr_set_allocation_functions`
+
+gRPC memory allocation functions such as `gpr_malloc` and `gpr_free` will
+remain with this change because they have been used when data allocated in
```
markdroth: If this is the only remaining use-case for these functions, does that mean that we don't need to use these functions inside of core?  In other words, external code that creates metadata to pass it into us will need to use these, but once inside of core, we can use `free()` directly instead of `gpr_free()`.  Is that right?

veblush: That's right. Since they're identical inside, it's up to us deciding which gRPC core will use.

L60-core-remove-custom-allocator.md
```diff
@@ -0,0 +1,92 @@
...
+
+In addition to this, some of C++ libraries cannot use custom allocators.
+For example, `std::function` doesn't support an allocator. Moreover,
+none of gRPC wrapped libraries including gRPC C++ doesn't support this.
+As a result, memory allocation can be done either in a built-in allocator
+or a custom allocator. This can be misleading to developers.
+
+## Implementation
+
+Core: https://github.com/grpc/grpc/pull/16706
```
chwarr: PR https://github.com/grpc/grpc/pull/16706/files is about removing `grpc_use_signal`. Is that what you intended to link to here? If so, can you expand on how this PR is related to this proposal?

veblush: Oops. That's right. Thanks!

PR #164, opened 2019-10-22
A26: gRPCLB Selection

dfawley:
#### Abstract

Modify the method for enabling gRPCLB to not require any special handling in core, allowing for stabilization of the gRPCLB protocol and load balancing policy.

@ejona86 @markdroth - Please also provide implementation steps for C/Java if appropriate.  Thanks!

### DIFF
--- a/A26-grpclb-selection.md
+++ b/A26-grpclb-selection.md
@@ -3,9 +3,9 @@ A26: gRPCLB Selection
 * Author(s): Doug Fawley
 * Approver: a11r
 * Status: Draft
-* Implemented in: None
+* Implemented in: C
 * Last updated: 2019-10-22
-* Discussion at: <google group thread> (filled after thread exists)
+* Discussion at: https://groups.google.com/d/msg/grpc-io/pB3vWC6MtlQ/vo681HkiAwAJ
 
 ## Abstract
 
@@ -25,8 +25,8 @@ can be declared stable, yet not create long-lasting maintenance problems.
 
 ## Proposal
 
-C core will promote its existing channel arg that configures the DNS resolver to
-fetch SRV records to "stable".
+C core will continue to provide its channel arg that configures the DNS resolver
+to fetch SRV records.
 
 In Go and Java, the default DNS resolver will not fetch SRV records. However,
 when gRPCLB is imported, it will override the default DNS resolver with a
@@ -132,4 +132,4 @@ The new approach will be implemented by:
 
 ### C
 
-TBD
+Implementation: https://github.com/grpc/grpc/pull/20803
--- a/A26-grpclb-selection.md
+++ b/A26-grpclb-selection.md
@@ -113,7 +113,22 @@ method.  This will be implemented by:
 
 ### Java
 
-TBD
+Java will remove the legacy method of gRPCLB selection (via balancer addresses
+produced by the resolver via the Address list). This was internal-only
+functionality already, as it required using
+`io.grpc.internal.GrpcAttributes.ATTR_LB_ADDR_AUTHORITY`.
+
+The new approach will be implemented by:
+
+1. Move `io.grpc.internal.GrpcAttributes.ATTR_LB_ADDR_AUTHORITY` to
+   `io.grpc.grpclb.GrpclbConstants.ATTR_LB_ADDR_AUTHORITY`.
+2. Add `io.grpc.grpclb.GrpclbConstants.ATTR_LB_ADDRS` with type
+   `Attributes.Key<List<EquivalentAddressGroup>>`. Every
+   `EquivalentAddressGroup` in the list must have a `ATTR_LB_ADDR_AUTHORITY`.
+   This key would pass addresses via the Attributes arguments of
+   `NameResolver.Listener.onAddresses(List<EAG>, Attributes)`
+3. Change grpclb policy to look for LB addresses via ATTR_LB_ADDRS instead of
+   the normal list of EAG.
 
 ### C
 
--- a/A26-grpclb-selection.md
+++ b/A26-grpclb-selection.md
@@ -2,7 +2,7 @@ A26: gRPCLB Selection
 ----
 * Author(s): Doug Fawley
 * Approver: a11r
-* Status: Draft
+* Status: Final
 * Implemented in: C
 * Last updated: 2019-10-22
 * Discussion at: https://groups.google.com/d/msg/grpc-io/pB3vWC6MtlQ/vo681HkiAwAJ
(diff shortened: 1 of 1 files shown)


ejona86:
CC @zhangkun83 . You may want to tweak Attribute.Keys names in the Java section at the end.

A26-grpclb-selection.md
```diff
@@ -0,0 +1,135 @@
...
+the gRPCLB LB policy, balancer addresses have no value. Users that do not wish
+to use gRPCLB will not take the performance penalty caused by attempting to
+resolve the balancer addresses through SRV records.
+
+Additionally, in all languages, to remove special handling of gRPCLB from the
+channel:
+- The loadBalancingConfig field in service config will be the only way to
+  specify the "grpclb" policy.
+- Balancer addresses will now only be provided via side-channel data passed from
+  the resolver to the balancer (similar to the xDS client instance).
```
zhangkun83: It's not called out -- are the balancer addresses now included in the loadBalancingConfig?

ejona86: No. They are through a side-channel API that we don't have a name for. In Java it is via the arbitrary Attributes passed from NameResolver (not in EAGs).

A26-grpclb-selection.md
```diff
@@ -0,0 +1,135 @@
...
+   `Attributes.Key<List<EquivalentAddressGroup>>`. Every
+   `EquivalentAddressGroup` in the list must have a `ATTR_LB_ADDR_AUTHORITY`.
+   This key would pass addresses via the Attributes arguments of
+   `NameResolver.Listener.onAddresses(List<EAG>, Attributes)`
+3. Change grpclb policy to look for LB addresses via ATTR_LB_ADDRS instead of
+   the normal list of EAG.
+
+### C
+
+TBD
```
markdroth: I've started putting together a PR for this.  I'll keep you posted.

A26-grpclb-selection.md
```diff
@@ -0,0 +1,135 @@
...
+experimental. In addition, gRPCLB itself is deprecated, to be replaced by xDS as
+the official advanced load balancing protocol for gRPC ([announcement
+email](https://groups.google.com/d/msg/grpc-io/0yGihF-EFQo/A4QKdXffBwAJ)). We
+would like to enable production uses of gRPCLB, but cannot do so if gRPCLB
+support is not stable. Therefore, we need a mechanism for selecting gRPCLB that
+can be declared stable, yet not create long-lasting maintenance problems.
+
+## Proposal
+
+C core will promote its existing channel arg that configures the DNS resolver to
```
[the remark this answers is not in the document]
ejona86: We _must_ have a stable API for client libraries to use it. I can agree that is separate from this proposal, but it was more a prerequisite.

I did not realize the "official word" is there is no stable API for users to control grpc.max_receive_message_length, grpc.max_connection_idle_ms, grpc.max_connection_age_ms, and similar in wrapped languages. I'm really confused why channel args were ever proposed as something Java should add to handle enabling SRV.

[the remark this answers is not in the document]
markdroth: All we really need for client libraries is to coordinate any changes with them ahead of time, which I'm fine with doing.  And while we've never promised stability for channel args strings, we also don't generally change them, so I don't expect this to cause any problem in practice.  I just don't want this gRFC to imply something to the world at large that we don't currently intend to imply.

As I said earlier, I do think that this is something that we should look more closely at in the future, but I think that's a broader topic, and I don't want to hold up this gRFC on that issue.  So, in the interest of moving forward with this gRFC, let's change the wording here to avoid the issue for now.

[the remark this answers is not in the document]
dfawley: > All we really need for client libraries is to coordinate any changes with them ahead of time

Are client libraries and gRPC expected to be kept in lockstep by users?  If not, CL should only be using stable gRPC APIs, or else users will run into problems when they attempt to use differing versions of the two libraries.

A26-grpclb-selection.md
markdroth: https://github.com/grpc/grpc/pull/20803

A26-grpclb-selection.md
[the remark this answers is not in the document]
dfawley: Per their [documentation](https://github.com/googleapis/google-cloud-cpp#requirements), client libraries only specifies a *minimum* version of gRPC.  Any backward-incompatible changes to any gRPC APIs they use will cause breakages of their older versions.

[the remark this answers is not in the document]
markdroth: Oh, interesting.  I was under the impression that they built with a particular version linked directly into the client library.  But now that I think about it, I guess that would not work in many languages anyway.

In any case, I don't think this changes anything for this gRFC.  I believe (but @apolcyn can confirm) that the client libraries are already depending on this particular channel arg, and I'm comfortable with them doing so.  It's not a problem for us to ensure that we don't break that.  But I still don't want the language here to imply a broader promise.

A26-grpclb-selection.md
```diff
@@ -0,0 +1,135 @@
...
+In Go and Java, the default DNS resolver will not fetch SRV records. However,
+when gRPCLB is imported, it will override the default DNS resolver with a
+resolver that does fetch SRV records. Justification for this approach: without
+the gRPCLB LB policy, balancer addresses have no value. Users that do not wish
+to use gRPCLB will not take the performance penalty caused by attempting to
+resolve the balancer addresses through SRV records.
+
+Additionally, in all languages, to remove special handling of gRPCLB from the
+channel:
+- The loadBalancingConfig field in service config will be the only way to
```
apolcyn: If I'm correct, this is an intentionally breaking change, that will cause clients that are currently enabling grpclb policy via SRV only (and not resolving or statically configuring <i>any</i> service config), to no longer use grpclb policy.

If that's the case, then don't see issues with this in itself, but at least a handful of experimental users would be broken by this. I.e., AFAICS we need to plan a migration and then help migrate those users, as a part of doing this.

markdroth: It should already be possible to enable grpclb via the service config even if there are no balancer addresses, so I think there's a transition plan.  We change the experimental users' code first and then do a new gRPC release with this change.

dfawley: In Go, since our gRPCLB support is not experimental, we will keep the existing selection logic, probably throughout v1.

dfawley:
@markdroth PTAL

dfawley:
This PR will wait until November 8 for community feedback, in accordance with the documented gRFC process.

PR #166, opened 2019-11-15
update retry proposal

jadekler:
Currently, the retry proposal specifies sending "grpc-previous-rpc-attempts"
metadata on each RPC beyond the first RPC.

However, this makes it hard to use: the absence of the metadata could either be
RPC attempt #0, or an RPC (of any attempt #) using a version of the gRPC library
before this logic is implemented. (or, the version of the client library that
was updated to use this field)

That is: we could never tell, for example, %QPS that are related to retries
because we don't know what our sample size is. By specifying first RPC must send
grpc-previous-rpc-attempts=0, we can do #rpcs-whose-attempts>0 / #rpcs-with-any-attempts.

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -312,7 +312,7 @@ Similarly, transparent retries do not count toward the limit of configured RPC a
 
 Both client and server application logic will have access to data about retries via gRPC metadata. Upon seeing an RPC from the client, the server will know if it was a retry, and moreover, it will know the number of previously made attempts. Likewise, the client will receive the number of retry attempts made when receiving the results of an RPC.
 
-The header name for exposing the metadata will be `"grpc-previous-rpc-attempts"` to give clients and servers access to the attempt count. This value represents the number of preceding retry attempts. Thus, it will not be present on the first RPC, will be 1 for the second RPC, and so on. The value for this field will be an integer.
+The header name for exposing the metadata will be `"grpc-previous-rpc-attempts"` to give clients and servers access to the attempt count. This value represents the number of preceding retry attempts. On the first RPC, will be 0; on the second RPC, it will be 1, and so on. The value for this field will be an integer.
 
 #### Disabling Retries
 


jadekler:
cc @dfawley @ejona @markdroth

markdroth:
It's not clear to me how this would be useful.  A previous retry implementation not conforming to this spec will not specify this metadata even on subsequent attempts, so there is no way to tell what's a retry and what's not, and the change proposed here does not help with that.  And for a retry implementation that does conform to this spec, it's not clear to me why it's useful to differentiate between an RPC that will not be retried and the first attempt of an RPC that will be retried.

Can you give a concrete example of server-side logic that would make use of this metadata on the first attempt?  What would it do differently than it would if this metadata were not present?

jadekler:
Can you give a concrete example of server-side logic that would make use of this metadata on the first attempt? What would it do differently than it would if this metadata were not present?

Imagine a query like the following, where `max_retry_attempt` is nillable:

```
SELECT a.retries / (a.retries+b.no_retries) AS retry_percentage
FROM
    (
        SELECT COUNT(*) as retries
        FROM rpcs
        WHERE max_retry_attempt > 0
    ) a,
    (
        SELECT COUNT(*) as no_retries
        FROM rpcs
        WHERE max_retry_attempt == 0
    ) b
```

With the update to this PR, this would be accurate. Without this PR, there would be no zeroes: we could only use nil, which would mean both "I was not a retry request" and "I might have been a retry request but I'm on an older version of gRPC, and didn't send the metadata".

> A previous retry implementation not conforming to this spec will not specify this metadata even on subsequent attempts, so there is no way to tell what's a retry and what's not, and the change proposed here does not help with that.

It's intention is to help by distinguishing "latest enough version of gRPC that 0 is meaningful" and "not latest enough version of gRPC to count".

dfawley:
> Its intention is to help by distinguishing "latest enough version of gRPC that 0 is meaningful" and "not latest enough version of gRPC to count".

I'm not understanding.  Any version of gRPC that implements retry should implement this header (for 1-N values).  Otherwise, it did not properly support retry.  Unless I'm mistaken, retry support is still only experimental and is disabled by default in all three languages (this is true in Go for sure).  Making new versions of gRPC set zero here would only make that worse: older versions of gRPC that do support retry but don't yet set this header to zero will skew the numbers, since they will have non-zero values but no zeroes.

markdroth:
I am also concerned that this will add overhead.  Even when retries are enabled, the common case is expected to be that the first attempt will succeed, and one of the goals of this design was to minimize the additional overhead in that case.  Adding a header to every single request seems like it would add a non-trivial amount of overhead, given the number of requests that happen per unit time.

I am not in favor of this change.  The cost seems non-trivial, and the benefit seems very marginal.

jadekler:
> I'm not understanding. Any version of gRPC that implements retry should implement this header (for 1-N values).

@dfawley As the design currently stands, any version of gRPC that implements retry _may_ send the header depending on whether it's the first request or not. I'm proposing to update the design such that any newer version of gRPC that implements retry _will always_ send the header.

> I am also concerned that this will add overhead. Even when retries are enabled, the common case is expected to be that the first attempt will succeed, and one of the goals of this design was to minimize the additional overhead in that case. Adding a header to every single request seems like it would add a non-trivial amount of overhead, given the number of requests that happen per unit time.

I agree the overhead is a concern. However, what is the value of the header without being able to determine % QPS retry? The best I can think of is that we would see spikes in retries for whatever percentage of clients are sending this. That has some value, but it seems very shallow in comparison to the sophisticated planning and design we could do when we could understand retry numbers in percentage context.

markdroth:
The purpose of this header is primarily for servers to decide to shed retry load when they're overloaded.  We do not need the header on the initial request for that purpose.

Retries are configured on a per-method basis.  If you need to know the percentage of requests that were retries, you can figure that out based on which methods are configured to be retried.  We do not need the header on the initial request for that purpose.

markdroth:
There wasn't a compelling argument for this change, so I'm going to close this.

PR #173, opened 2020-03-26
A27 update: localities may contain 0 endpoints

markdroth:
As per internal discussion, the wording here was too strict.


markdroth:
We will do the same thing as Envoy does, so I don't think we need to explicitly call that out here.

Envoy's behavior is that it basically just ignores localities with 0 endpoints.  If it gets two P0 localities and one of them has no endpoints, it sends all traffic to the other one.  If it gets a P0 locality with no endpoints and a P1 locality with endpoints, it will fail over and send all traffic to the P1 locality.

markdroth:
I added a note about the behavior.

PR #175, opened 2020-04-09
L64: Python Runtime Proto Parsing

gnossen:


L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+## Background
+
+gRPC and protocol buffers evolved in a Google ecosystem supported by a [monorepo](https://www.wired.com/2015/09/google-2-billion-lines-codeand-one-place/) and [hermetic build system](https://arxiv.org/pdf/1702.01715.pdf). When engineers check a `.proto` file into source control using that toolchain, the resultant code generated by the protocol compiler does not end up source control. Instead, the code is generated on demand during builds and cached in a distributed build artifact store. While such sophisticated machinery has started to become available in the open source community (e.g. [Bazel](https://bazel.build/)), it has not yet gained much traction. Instead, small Git repos and non-hermetic language-specific build tools are the norm. As a result, code generated by the protocol compiler is often checked into source contcontrol alongside the code which uses it.
```
srini100, line 17: typo: does not end up **in the** source control.

srini100, line 17: typo: ~~cont~~control

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+### Dependency Considerations
+
+gRPC makes a point not to incur a direct dependency on protocol buffers. It is not the intent of this feature to change that. Instead, the implementations of these new functions will live in the `grpcio-tools` package, which necessarily *already* has a hard dependency on `protobuf`. If the `grpcio` package finds that `grpc_tools` is importable, it will do so and use the implementations found there to back the `protos` and `services` functions. Otherwise, it will raise a `NotImplementedError`.
+
+In order to take advantage of newer language features, support will only be added for Python versions 3.6+. All other versions will result in a `NotImplementedError`.
+
+## Implementation
+
+This proposal has been implemented [here](https://github.com/grpc/grpc/pull/21458). This implementation uses the C++ protobuf runtime already built into the `grpcio-tools` C extension to parse the protocol buffers and generate textual Python code in memory. This code is then used to [instantiate the modules](https://www.python.org/dev/peps/pep-0302/) to be provided to the calling application.
+
```
srini100, line 96: What version of protobuf runtime is supported in grpcio-tools and when is it upgraded?

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+The wrapper function around the import servers several purposes here. First, it puts the user in control of naming the module (in a manner similar to Javascript), meaning the user never has to concern themself with the confusing `_pb2` suffix. Second, the function provides a wrapping layer through which the layer can provide guidance in the case of failed imports.
```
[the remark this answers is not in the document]
gnossen: I somehow can't type "serve" without typing "server". :facepalm:

L64-python-runtime-proto-parsing.md
gnossen: It's determined by the git submodule in the core repo. In other words, the same version as C++ at the same commit. Updating the submodule is now part of the gRPC Core release process.

gnossen:
@srini100 Thank you for the review! PTALA.

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+## Background
+
+gRPC and protocol buffers evolved in a Google ecosystem supported by a [monorepo](https://www.wired.com/2015/09/google-2-billion-lines-codeand-one-place/) and [hermetic build system](https://arxiv.org/pdf/1702.01715.pdf). When engineers check a `.proto` file into source control using that toolchain, the resultant code generated by the protocol compiler does not end up in source control. Instead, the code is generated on demand during builds and cached in a distributed build artifact store. While such sophisticated machinery has started to become available in the open source community (e.g. [Bazel](https://bazel.build/)), it has not yet gained much traction. Instead, small Git repos and non-hermetic language-specific build tools are the norm. As a result, code generated by the protocol compiler is often checked into source control alongside the code which uses it.
```
lidizheng, line 17: Should we mention that community are building tools to ease this process like https://github.com/uber/prototool.

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+
+The protocol compiler paired with Google's build system also means that the average engineer never has to manually invoke `protoc`. Instead, when an update is made to a `.proto` file and some code file references it, the code for the protocol buffer is regenerated without any manual intervention on the part of the engineer. Compare that to today's workflow for gRPC users:
+
+ 1. Update .proto file.
+ 2. Manually regenerate code. (remembering how to use all of the CLI flags)
+ 3. Make necessary corresponding updates to code using the protocol buffer.
+ 4. Rerun.
+
+It's easy for several of those steps to slip one's mind while developing. Moreover, figuring out how to invoke `protoc` in a way that meshes with your imports can be quite difficult. Python developers in particular are unused to build-time steps such as these; it is much more common to perform them at runtime.
```
lidizheng, line 30: Should we mention it's also beneficial to environments that only have access to ephemeral file systems? Like serverless.

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+They will now also have the option of completely skipping the manual protoc invocation and instead writing
+
+```python
+protos = grpc.protos('foo.proto')
+services = grpc.services('foo.proto')
+```
+
+These two new functions return the same module objects as the `import foo_pb2` and `import foo_pb2_grpc` statements. In order to maintain interoperability with any proto-backed modules loaded in the same process, after these functions are invoked, `import foo_pb2` and `import foo_pb2_grpc` will be no-ops. That is, a side-effect of calling `grpc.protos` and `grpc.services` is insertion of the returned modules into the per-process cache. This ensures that regardless of whether the application calls `grpc.protos('foo.proto')` or `import foo_pb2` or in which order, only a single version of the module will ever be loaded into the process. This avoids situations in which interoperability breaks due to two modules expecting the same protobuf-level message type, but expecting two *different* Python-level `Message` classes, one backed by a `_pb2.py` file and one backed by a `.proto` file.
```
lidizheng, line 53: Want to check if I read it correct. In order to avoid duplication, the `grpc.protos('foo.proto')` registers a global module with name "foo_pb2"? So, the following `import foo_pb2` will import the generated module instead of trying to load from disk?

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+
+The final function, `protos_and_services` is a simple convenience function allowing the user to import protos and services in a single function call.
+
+The change will be entirely backward compatible. Users manually invoking `protoc` today will not be required to change their build process.
+
+### Import Paths
+
+These functions will behave like normal `import` statements. `sys.path` will be used to search for `.proto` files. The path under which each particular `.proto` file was found will be passed to the protobuf parser as the root of the tree (equivalent to the `-I` flag of `protoc`). This means that a file located at `${SYS_PATH_ENTRY}/foo/bar/baz.proto` will result in the instantation of a module with the fully qualified name `foo.bar.baz_pb2`.  Users are expected to have a directory structure mirroring their desired import structure.
+
+Users have reported that getting Protobuf paths to match Python import paths is quite tricky today using `protoc`. In the case of a failure to import, the library will print troubleshooting information alongside the error.
```
lidizheng, line 81: In the implementation, I recall you try to make the import path identical to the package name. Is it still true, should we mention it?

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Arbitrary Protos
+
+It should be stated that, in practice, it is possible to use these new functions to load totally arbitrary protos. Suppose you wrote a server that took `.proto` files as inputs from clients, instatiated modules from them, and returned some data about the file. For example, the number of message types contained within the supplied file. This could become problematic as new syntax features are added to the Protobuf specification. In the worst case, this would require a redeploy of the server with a sufficiently up-to-date version of `grpcio-tools`. But, regardless, we will claim no support for this use case. The intent of these functions is to enable import of *fixed* `.proto` files, known at build time.
```
lidizheng, line 85: Should we support runtime reloading? So, the application can invoke `protos_and_services` multiple times during development.

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
+gRPC Python Runtime Proto Parsing
+----
+* Author(s): gnossen
+* Approver: lidizheng
+* Status: Draft
+* Implemented in: Python
+* Last updated: April 9, 2020
+* Discussion at: **TODO(gnossen)**
```
lidizheng, line 8: We might still have the 10 days open discussion restriction before merging the PR?

L64-python-runtime-proto-parsing.md
gnossen: Almost. The canonical name of the module imported by `grpc.protos('foo.proto')` will be `foo_pb2`, so the subsequent `import foo_pb2` operation won't result in an "import" at all. It will just make `foo_pb2` point to the module loaded from the file "foo.proto". That's a very nitpicky difference from how you stated it though.

L64-python-runtime-proto-parsing.md
gnossen: I'm not sure what you're referring to here. Unfortunately, both Python and Protobuf use the terms "package" and "import", so there are actually four things to worry about for every module. As written right now, the Python package, Python import path, and Protobuf import path all match up, while the Protobuf package is entirely independent.

So, taking the `grpc/grpc` repo as an example, supposing the repo root were on your `PYTHONPATH` (which would be the case if the interpreter were started from the repo root), [test.proto](https://github.com/grpc/grpc/blob/master/src/proto/grpc/testing/test.proto) would be importable with `grpc.protos('src/proto/grpc/testing/test.proto')` or, equivalently, `src.proto.grpc.testing.test_pb2`. Notice that, within the proto file itself, it's importing other files with the "src/proto" prefix. This is the Protobuf-level import path. This means that the repo root *must* be on the protocol buffer parser's search path, otherwise, this file will fail to import its dependencies.

The Protobuf package name for this file is "grpc.testing", which is much more palatable, since it really is just a namespacing mechanism and doesn't let the repo organization directories leak through. This has no bearing on how the module is imported from within Python. It *is* very common, however, to have the package at least be a suffix of the Protobuf import path, as is the case here. In that case, (in theory -- I haven't tested this), you should be able to add both the repo root and "src/proto" to your `PYTHONPATH` and be able to import you modules with `grpc.protos('grpc/testing/test.proto')` or `import grpc.testing.test_pb2`.

In general, I would recommend that people not include anything like the "src/proto" prefix in their ".proto" files at all. There should be some directory within their repo that they expect to be the root of their proto source tree and all Protobuf-level `import` statements should be made relative to *that*. That is, [this line](https://github.com/grpc/grpc/blob/e677f3d166ebcac000e2657263b3505c5c2f701f/src/proto/grpc/testing/test.proto#L21) should be changed to `import "grpc/testing/empty.proto"`. This would get all four attributes to line up and would ensure that the user only ever has to add that unique directory to their `PYTHONPATH`. Or, equivalently, to start the interpreter from that directory.

I'll add a section to the document explaining this.

L64-python-runtime-proto-parsing.md
gnossen: I don't think this is a good idea. Suppose you instantiate some messages before reloading and save them off. Then reload the module. Create some more messages. The messages before the reload cannot interact at all with the messages after the reload. You cannot fill one's containers with instances of the others. You cannot equality compare them.

Furthermore, this is already possible in the current design by doing

```python
bar_protos = grpc.protos("foo/bar.proto")
reload(bar_protos)
```

In any case, it's much safer to simply restart the process.

L64-python-runtime-proto-parsing.md
gnossen: Yes. I wanted to some eyeballs on it as a sanity check before creating the thread. Added!

L64-python-runtime-proto-parsing.md
gnossen: I'm not sure how this benefits environments with ephemeral filesystems? Anything that has the ability to host the application's main ".py" file also has the ability to host a pre-generated "_pb2.py" file. Even if running from a REPL, you'll still need a filesystem that's at least permanent enough to host the ".proto" file in order for `grpc.protos("foo.proto")` to work.

L64-python-runtime-proto-parsing.md
gnossen: We could. But it's not exactly apropos of this proposal. It's certainly good information though. Perhaps this sort of thing would make more sense as part of a post on the blog?

L64-python-runtime-proto-parsing.md
lidizheng: Make the import path look like the Protobuf package name is much better than the arbitrary path created by users. The explanation is very informative, people who read this design doc will appreciate this rationale.

L64-python-runtime-proto-parsing.md
lidizheng: Good to know that is possible. As you mentioned, this pattern is not recommended. Then we should not recommend user to restart instead of reload?

L64-python-runtime-proto-parsing.md
gnossen: I see. This is dangerously getting close to the "arbitrary protos" use case that I explicitly claim no support for in this document because it presents problems when new syntax is introduced to the protobuf specification. It also looks like IPython allows [inclusion of arbitrary resource files](https://stackoverflow.com/a/10628360/10598399) (like images). In that case, you could include a ".proto" file in the notebook and use `grpc.protos` to import it, but there isn't anything specific to the IPython environment there.

As for serverless, I don't actually know of any product that is so restrictive that it actually limits you to a single textbox. [Not appengine](https://cloud.google.com/appengine/docs/flexible/python/configuration-files). [Not Cloud Functions](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/functions/helloworld). [Not Cloud Run](https://github.com/grpc-ecosystem/grpc-cloud-run-example/tree/master/python). [Not AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/code-editor.html). It's not a usage that I want to support or advertise.

L64-python-runtime-proto-parsing.md
gnossen: I'll add a section clarifying that these functions are idempotent and will not cause a reload from disk.

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,99 @@
...
+
+```python
+import foo_pb2
+import foo_pb2_grpc
+```
+
+They will now also have the option of completely skipping the manual protoc invocation and instead writing
+
+```python
+protos = grpc.protos('foo.proto')
```
pfreixes, line 49: Would be any chance on having a more explicit API for returning explicitly a specific module, for example:

```python
foo_pb2 = grpc.protos('foo.proto',  modules=["foo_pb2"])
```
This will make the API more explicit, which would help users and developers to not have the cognitive burden of remembering that modules were magically imported behind the scenes.

Another concern of not having an explicit creation of the module would be with some complaining that syntax checkers might give you, which they would be false positive. Have you checked what would happen with pylint or IDEs like PyCharm?

gnossen: @pfreixes Thanks for the suggestion!

The problem is that whenever someone specified a module name besides the default (`foo_pb2` in your example), they'd introduce the possibility of the unintentionally loading a logically equivalent, but nevertheless incompatible module into the process. For example:

```python
foo = grpc.protos('foo.proto', modules=["foo"]) # load from .proto file
import foo_pb2 # load from pre-generated code by accident
```

Types from `foo` and `foo_pb2` look the same. They have the same names, constructors, and methods. But they are completely incompatible. Even if they hold the same values, they will never evaluate as equal. Containers from `foo` cannot be filled with elements from `foo_pb2` and vice versa.

The usage of the `foo_pb2` format as the canonical name is meant be a detail that can just be forgotten about, even in the case that the process mixes pre-generated code and runtime parsing.

You make an interesting point about syntax checkers. But if they do have a problem, it will exist for anything that uses `importlib`, including modules loaded from zip files, modules loaded from shared object libraries, and modules loaded from `.pyc` files with no corresponding `.py` file.

pfreixes: Regarding the example that you gave, is there a typo? Should be: 

```python
foo = grpc.protos('foo.proto', modules=["foo_pb2"]) # load from .proto file
import foo_pb2 # load from pre-generated code by accident
```
?

If so, why the user might end up mixing both implementations? just because different modules, or even packages, might be loading objects in a different way? for example, the caller which might be the main module using the new `grpc.protos` and the dependency might be still using the `import foo_pb2`?

Why the implementations would be different? I thought - maybe I'm totally wrong - that the `protos` field was indeed generating in memory the same python file that would be generated in disk. If so, why they would need to be considered incompatible?

L64-python-runtime-proto-parsing.md
lidizheng: Sorry, I wasn't express it clear enough. A single textarea is one setup option, but it is not the only way. At least AWS Lambda allows user to supply code from its UI. I'm trying to think more scenarios that benefited by this work.

L64-python-runtime-proto-parsing.md
gnossen: There was no typo in my example. I explicitly chose a module name besides what would be the default for such an option. The example that you included above is equivalent to the following in the current proposal.

```python
foo = grpc.protos('foo.proto') # Import from a ".proto" file
import foo_pb2 # Does not trigger a re-import.
assert foo is foo_pb2
```

In my example, however,
```python
foo = grpc.protos('foo.proto', modules=["foo"])
import foo_pb2
foo.FooMessage(a=2) == foo_pb2.FooMessage(a=2)  # This will always be false.
foo.FooMessage is foo_pb2.FooMessage  # Because this is always false.
```
> If so, why the user might end up mixing both implementations? just because different modules, or even packages, might be loading objects in a different way? for example, the caller which might be the main module using the new grpc.protos and the dependency might be still using the import foo_pb2?
> Why the implementations would be different? I thought - maybe I'm totally wrong - that the protos function was indeed generating in memory the same python file that would be generated in disk. If so, why they would need to be considered incompatible?

The problem isn't that there are two modules loaded in different ways. The problem is that there are two modules *at all*. This comes down to two things:

- [How imports in a proto file are handled](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/python/python_generator.cc#L430) (If you didn't put it under its canonical name, the import mechanism within the generated source code won't be able to find it)
- [The protobuf C extension requires pointer equality of the descriptors backing message types for them to work.](https://github.com/protocolbuffers/protobuf/blob/3be9322e28dfcacdb669f646f76fe833fa274107/python/google/protobuf/pyext/message.cc#L2176). This effectively kills interoperability through duck typing.

The only way to guarantee compatibility is to ensure that there is only ever one module loaded in the interpreter for each Protobuf file, regardless of whether the code was generated at build time or at runtime.

pfreixes: Oks got it, it seems that the main issue for having compatibility comes from the fact that the comparison would not work since pointers are being compared. But in any case, I would kinda agree that having two different modules loaded and trying to make them compatible could become a nightmare.

Sorry for the usage of the `modules` parameter, in somehow I wanted to add something for narrow down the modules and I ended up mixing two different things: modules and messages.

Oks let me just add a last little detail, should we prescript the user for doing something like this:

```python
foo_pb2 = grpc.protos('foo.proto')
```

Again trying to be as much explicit as we can, providing this recipe  the user would make something like:

```python
foo_pb2 = grpc.protos('foo.proto')

def create_message(a: int) -> foo_pb2.FooMessage:
    return foo_pb2.FooMessage(a=a)
```

gnossen: So, I explicitly chose not to include the `_pb2` suffix because it's confusing. Newcomers to gRPC have no idea what it is. When I write a file called "foo.py", I expect to be able to `import foo`, not `import foo_py3`.

You follow up with an example about typing. I don't think that the example you included would actually work. I'm not super familiar with how Python type checkers work, but I'm fairly certain they don't execute the full source code of the file being checked. In general though, I do see good integration with type checkers as being valuable. I'll do some investigation into how the feature as implemented interacts with them.

gnossen: @pfreixes I dug into the type checking a bit more. It turns out that this is just [not supported by the current Python type checkers](https://github.com/python/mypy/issues/5059). From the issue,
 
> `importlib.import_module` returns type `types.ModuleType`. We don't know anything else about the attributes of that module, beyond the attributes of that type. Therefore, the error message boils down to #1424.

> Perhaps we should warn when someone calls __import__, importlib.import_module or the like, because it almost always requires a type ignore.

At least for the moment, I would recommend that people wanting support for type checking continue to statically generate code and pull that in.

pfreixes: Thanks for having a look, just for clarifying would this be fixed by doing 
```python
protos = grpc.protos('foo.proto')
import foo_pb2
```
So being foo_pb2 a noop but, maybe, allowing the type checker to know that there is behind?

Out of curiosity, would this issue happen also with `pytype` ?

gnossen: I gave the `import foo_pb2` approach a shot as well, but alas, it doesn't work either.

`pytype` behaves slightly differently, but also doesn't satisfactorily work. Instead of throwing an error about an unknown module, it seems to simply ignore types dependent on the unknown module. For instance, you could change your example above to the following without an error:

```python
foo_pb2 = grpc.protos('foo.proto')

def create_message(a: int) -> foo_pb2.NonexistentType:
    return foo_pb2.FooMessage(a=a)
```

pfreixes: sad panda.

We should add a disclaimer since typing is nowadays well spread and used for many and many projects. And the side effect If I'm not wrong when mypy is used is an annoying exception.

L64-python-runtime-proto-parsing.md
```diff
@@ -0,0 +1,136 @@
...
+
+
+## Background
+
+gRPC and protocol buffers evolved in a Google ecosystem supported by a [monorepo](https://www.wired.com/2015/09/google-2-billion-lines-codeand-one-place/) and [hermetic build system](https://arxiv.org/pdf/1702.01715.pdf). When engineers check a `.proto` file into source control using that toolchain, the resultant code generated by the protocol compiler does not end up in source control. Instead, the code is generated on demand during builds and cached in a distributed build artifact store. While such sophisticated machinery has started to become available in the open source community (e.g. [Bazel](https://bazel.build/)), it has not yet gained much traction. Instead, small Git repos and non-hermetic language-specific build tools are the norm. As a result, code generated by the protocol compiler is often checked into source control alongside the code which uses it.
```
catern, line 17: >Instead, the code is generated on demand during builds and cached in a distributed build artifact store. While such sophisticated machinery has started to become available in the open source community (e.g. [Bazel](https://bazel.build/)), it has not yet gained much traction. Instead, small Git repos and non-hermetic language-specific build tools are the norm.

I think you need to support this point more. It sounds like you're implying here that the issue is missing technology, but generating protocol files doesn't require Bazel or a monorepo, it just requires a basic build system, no sophisticated machinery; even a basic Makefile can do it; you should more explicit point out why the build systems that open source projects have today cannot support generating protocol files.

gnossen: Other build systems can and (some) do support it ([setuptools](https://github.com/grpc/grpc/blob/e63fa5fdf0fc24bd041ba99d2c58d2974595bf5d/tools/distrib/python/grpcio_tools/grpc_tools/command.py#L51), [Bazel](https://github.com/grpc/grpc/blob/master/bazel/python_rules.bzl#L209)). It's simply a question of effort to implement (in the case that no tooling exists for your particular build system) or effort to integrate (in the case that it does). The weakness that all build systems share is that they live *outside the language*. The strongest use case for this is "running as a script". That is, running a `.py` file invoking a gRPC method without any build system at all (.e.g `python3 grpc_client.py`). In my experience, the vast majority of Python usage falls into that category.  These users *could* integrate with a build system, but, for a simple script, given the choice, even I wouldn't.

I'll clarify in the proposal.

PR #177, opened 2020-05-13
L65: Additional PyPI Packages for gRPC Python

lidizheng:
> gRPC Python is uploaded as `grpcio` on PyPI, but there is another package named
`grpc`. Hence some users are confused. This document proposes gRPC Python to
upload additional packages named `grpc` and `grpc-*` as proxies to our official
packages.

### DIFF
--- a/L65-python-package-name.md
+++ b/L65-python-package-name.md
@@ -1,0 +1,132 @@
+Additional PyPI Packages for gRPC Python
+----
+* Author(s): lidiz
+* Approver: gnossen
+* Status: Draft
+* Implemented in: python
+* Last updated: 05/13/2020
+* Discussion at: **TBD**
+
+## Abstract
+
+gRPC Python is uploaded as `grpcio` on PyPI, but there is another package named
+`grpc`. Hence some users are confused. This document proposes gRPC Python to
+upload additional packages named `grpc` and `grpc-*` as proxies to our official
+packages.
+
+## Background
+
+### Package `grpc`
+
+Package `grpc` is an alternative version of gRPC Python only supports Python 2,
+and the first version is uploaded in 2013. We have seen complaints about the
+installation failure or unable to execute our example code. The tricky package
+name is a usability issue for years.
+
+### Package `grpcio`
+
+Package `grpcio` is first uploaded in 2015, given the fact that the name `grpc`
+already being taken, this name is likely a workaround.
+
+## Proposal
+
+### List of Package Name
+
+Here is a list of new proxy packages:
+* `grpc`
+* `grpc-status`
+* `grpc-channelz`
+* `grpc-tools`
+* `grpc-reflection`
+* `grpc-testing`
+* `grpc-health-checking`
+
+### Proxy Mechanism
+
+Each proxy package will raise a `RuntimeError` exception upon installation,
+asking users to install with the correct package name. For example, for proxy
+package `grpc`, the exception will look like:
+
+```
+RuntimeError: Please install the official package with: pip install grpcio
+```
+
+### Maintenance Cost
+
+The maintenance cost for this solution is minimum since these packages are
+agnostic to our releases.
+
+### Package Version
+
+All new proxy packages will be versioned as `1.0`.
+
+## Rationale
+
+### Why not switch the name to `grpc` and `grpc-*`?
+
+#### Hyrum's Law
+
+Although the package name is not ideal, it is part of our contract with our
+users. If we change our name from `grpcio` to `grpc`, it is not backward
+compatible. For users who perform strict dependency management, their
+environment might not automatically pick-up this change. It could result in
+crashes since all gRPC logic was moved to another package.
+
+There are two valid reasons for Python users to use strict dependency
+management:
+
+1. Freezing the dependency for shipping;
+2. Version conflict in the dependency tree.
+
+For the second case, some package may pin to an old version of gRPC, but some
+requires a newer version. It's impossible to make both dependency happy, so the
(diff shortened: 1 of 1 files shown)


L65-python-package-name.md
```diff
@@ -0,0 +1,132 @@
...
+### Package `grpcio`
+
+Package `grpcio` is first uploaded in 2015, given the fact that the name `grpc`
+already being taken, this name is likely a workaround.
+
+## Proposal
+
+### List of Package Name
+
+Here is a list of new proxy packages:
```
gnossen, line 35: I"d recommend you move the section defining "proxy package" to before this first usage of the term.

L65-python-package-name.md
```diff
@@ -0,0 +1,132 @@
...
+* `grpc`
+* `grpc-status`
+* `grpc-channelz`
+* `grpc-tools`
+* `grpc-reflection`
+* `grpc-testing`
+* `grpc-health-checking`
+
+### Proxy Mechanism
```
gnossen, line 44: "Proxy" isn't the right name for what we're doing here. What we're really doing is nerfing the `grpc` package with a message suggesting that they install `grpcio` instead. This mechanism is the sign in this picture: 

Whereas a proxy would actually be a cop.

L65-python-package-name.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+### Maintenance Cost
+
+The maintenance cost for this solution is minimum since these packages are
+agnostic to our releases.
+
+### Package Version
+
+All new proxy packages will be versioned as `1.0`.
```
gnossen, line 61: I didn't fully appreciate what this meant the first time I read through. I would reword to something more along the lines of "Rather than releasing new versions of the `grpc-*` packages in lockstep with the canonical `grpcio-*` packages for each release, we will instead upload only one new artifact for each package name, versioned as '1.0'.". You might also want to make an argument for this strategy. If we go with this strategy as opposed to uploading a new copy of the same `RunTimeError`-producing package, then people running `pip install grpc==1.29.0` will get a "package not found" error instead of a more informative "I think you have the wrong number"-style message.

L65-python-package-name.md
```diff
@@ -0,0 +1,132 @@
...
+gRPC Python is uploaded as `grpcio` on PyPI, but there is another package named
+`grpc`. Hence some users are confused. This document proposes gRPC Python to
+upload additional packages named `grpc` and `grpc-*` as proxies to our official
+packages.
+
+## Background
+
+### Package `grpc`
+
+Package `grpc` is an alternative version of gRPC Python only supports Python 2,
```
gnossen, line 21: We probably want to add a bit of history as well on how we obtained ownership of the package name.

lidizheng: Add a link to https://github.com/pypa/pypi-support/issues/3.

L65-python-package-name.md
lidizheng: Good idea. Moved.

L65-python-package-name.md
lidizheng: How should we call it? "Cop packages" / "Guard packages" / "Watchdog packages" / "Toxic packages"? What do you recommend?

L65-python-package-name.md
lidizheng: Thanks for the suggestion. Updated to:
```
Rather than releasing new versions of the `grpc` and `grpc-*` packages in
lockstep with the canonical `grpc` and `grpcio-*` packages for each release, we
will instead upload only one new artifact for each package name, versioned as
`v1.0`. A stable version number suggests this package's behavior won't change
actively.
```

I reword this section, so it is less confusing. For people who want to pin package version, they are using gRPC Python in a more serious scenario. I think they would need to double check the package name.

I'm also okay to upload all future versions with a script. But they would taken our ability to change its logic forever. Instrumenting our infrastructure to produce more artifacts for each release is time-costly, we might not want to make existing release procedure more complex.

Also, separated version suggests the nature of release agnostic for the new packages.

L65-python-package-name.md
gnossen: No. I didn't mean things quite so literally with the meme. "Archived package" or "retired package" is probably best.

L65-python-package-name.md
gnossen: > lockstep with the canonical `grpc` and `grpcio-*`

The current wording implies that both `grpc` and `grpcio` prefixes are canonical. How about `grpcio*`? That includes `grpcio`, `grpcio-channelz`, etc.

> I'm also okay to upload all future versions with a script. But they would taken our ability to change its logic forever

I don't think that's true. We can always decide that the next release of `grpc` will contain something other than a `RunTimeError`.

> Instrumenting our infrastructure to produce more artifacts for each release is time-costly, we might not want to make existing release procedure more complex.

Totally agree.

L65-python-package-name.md
lidizheng: It's sad that those new packages never shined, but went straight to retirement. Perhaps that's been the story of life. Could it be better to reference it as just "package", and avoid using the word "proxy"?

gnossen: They did have a chance to shine. As "gevent RPC". *That's* the package that I'm referring to as archived, not a hypothetical "gRPC" `grpc` package.

I'm fine referring to them as just "the `grpc`-prefixed packages" or something like that. So long as it's clear in each instance whether we're referring to `grpcio`, `grpcio-reflection`, etc. *or* `grpc`, `grpc-testing`, etc.

PR #178, opened 2020-05-20
A28: gRPC xDS traffic splitting and routing

menghanl:


menghanl:
cc @srini100 @markdroth @ejona86 @dfawley

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+The weighted target LB policy will have a group of child policies and their
+corresponding weights, and will do weighted pick among the child policies.
+
+This LB policy will have the following configuration
+
+```proto
+message WeightedTargetConfig {
+  message Target {
+    uint32 weight = 1;
+    repeated LoadBalancingConfig child_policy = 2;
```
sanjaypujare, line 67: A link to the definition of `LoadBalancingConfig`  will be useful.

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+      string regex = 3;
+    }
+    repeated HeaderMatcher headers = 3;
+
+    // Every time the route is considered for a match, it
+    // must also fall under the percentage of matches
+    // indicated by this field. For some fraction N/D, a
+    // random number in the range [0,D) is selected. If the
+    // number is <= the value of the numerator N, or if the
+    // key is not present, the default value, the router
```
sanjaypujare, line 196: Not sure if `key` and `default value` belong to this sentence and are related to the `match_fraction` field since the meaning is not clear.

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+    // continues to evaluate the remaining match criteria.
+    Fraction match_fraction = 4;
+
+    string action = 5; // The action name
+  }
+
+  repeated Route route = 1;
+  map<string, Action> action = 2;
```
sanjaypujare, line 204: A comment to explain that this map maps an action name to the actual action will be helpful.

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+
+### Balancer config
+
+The `xds_routing_experimental` LB policy will have the following
+configuration:
+
+```proto
+/*
+
+Route List                         Action list
```
sanjaypujare, line 132: Action list => Action map?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+so that one action can be shared by multiple routes.
+
+The config must fulfill the following:
+
+*   Routes
+    *   There must be exactly one `path_matcher` rule.
+        *   `path_matcher` is an oneof so this is always true in proto. But json
+            doesn’t have the oneof concept, so a route may have multiple path
+            matchers, and it will be considered invalid.
+    *   The `action` name must be present in the action list.
```
sanjaypujare, line 218: action list => action map (?)

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+        *   Because the first match wins, the second matcher will never be used.
+*   Actions
+    *   The actions in list must be referred by the routes
+    *   There must be at least one valid config in `child_policy`
+
+Violating any of these will make the balancer config invalid, and be rejected.
+
+For each action, the config in `child_policy` to be used to create a child LB
+policy will be picked using the same algorithm as [service
+config](https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto#L297).
```
sanjaypujare, line 231: Line 297 points to some field inside `LrsLoadBalancingPolicyConfig` . The actual line number should be 377?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+tricky.
+
+The pick will always be delegated to the picker from the matched child
+policy, even if it’s not in connectivity state READY (as opposed to
+weighted\_target will only pick the READY child policies).
+
+There will also be cases where for an RPC, a match is found, but the child
+policy hasn’t generated the picker (for example, the child policy is still
+initializing, and no picker has been created). When this happens, the RPCs
+will be always queued until a picker is updated.
```
sanjaypujare, line 271: just to clarify the meaning: "the RPCs will always be queued at the matched child policy until a picker is created". Is that right?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+
+If no match can be found, the RPC will fail with code Unavailable, and
```
sanjaypujare, line 273: code Unavailable => status UNAVAILABLE

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+state is the state of the child policy that will be picked), it represents
+the overall state of the channel.
+
+## xDS resolver and xDS client
+
+The xDS resolver’s overall behavior will be similar as before. It will call
+the xDS client, and will send a service config generated from
+RouteConfiguration (the RDS response) to the parent gRPC client channel.
+
+The xDS client will make an LDS request, with resource\_names set to the
```
sanjaypujare, line 298: Do we need to repeat stuff that is not new or has not changed (the one at https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md#xds-resolver)

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+
+*   Must have
+    [path\_specifier](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L360)
+    *   Options supported:
+        [prefix](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L365),
+        [path](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L369)
+        or
+        [safe\_regex](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L405)
+    *   NOT regex
+        *   Justification: unsafe and deprecated -- replaced by safe\_regex
```
sanjaypujare, line 340: Does Envoy also reject (NACK) regex? If not then we will have a problem when migrating from proxyful to proxyless.

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+`xds_routing_experimental` as the LB policy. As an optimization, when there’s
+only one route, the LB policy picked will be `cds_experimental` or
+`weighted_target_experimental` depending on the action, so we can skip the
+`xds_routing` policy that has only one child.
+
+The `xds_routing_experimental` balancer config will contain the validated
+`{matchers, action}` pairs. The actions will be in a separate list, and will
+be referred to by the routes by their names.
+
+#### Action list
```
sanjaypujare, line 415: Should this be called "Action Map"?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+
+All the actions from the validated `{matchers, action}` pairs will be
+de-duplicated and stored in a cache. All actions in the cache will be
+assigned a unique name.
```
sanjaypujare, line 419: Is that a global cache or cache per RouteConfiguration? Is the unique name internally generated or does it map to anything in the xDS response?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+*   If the action is weighted\_clusters, the child\_policy will be
+    `weighted_target_experimental`. The balancer config will contain a list of
+    `cds_experimental` LB configs, and their corresponding weights.
+*   If this action is already in the cache, it will be reused, and nothing needs
+    to be done
+*   If the action is new, it will be be added to the list and assigned a name
+    *   The name should be a human readable hash of the config, to be logged for
+        debugging purposes.
+    *   For example, the names can be `cds:cluster1` and
+        `weighted:cluster1_cluster2_<index_number>`.
```
sanjaypujare, line 434: The example names don't look like hash of a config (as mentioned on line 431)

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+
+For each `{matchers, action}`, a route will be generated:
+
+*   With matchers in json representation.
+*   The action will be the name of this action from the action list.
+
+The processed response will be converted to a [JSON service
+config](#JSON-service-config-generated-by-xDS-resolver).
+
+### Multiple watches for the same resource
```
sanjaypujare, line 464: IIRC, all of the logic in this section is existing logic and not a new proposal. If true, can we reference existing doc/section?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+        *   The cache should only contain the responses for the in-process
+            watches. If we keep the item in cache after the watch is cancelled,
+            it will quickly become outdated.
+
+#### Handle responses
+
+When a response is received, the xDS client will
+
+*   Parse and notify all the watchers for this type
+*   Store the updated in cache, to prepare for future watches for the same
```
sanjaypujare, line 510: s/Store the updated in cache/Store the updated resource in the cache/

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+
+The problem we want to solve immediately is the duplicate TCP connections. We
+can do
+
+*   Share at the top level `xds_routing_experimental` LB policy
+*   Share all cds policy instances
+*   Share subchannels
+
+The first solution is chosen because it’s easy to implement, and it solves
+most of the subchannel duplication problem. Even wIthout subchannel sharing,
```
sanjaypujare, line 553: s/wIthout/without/

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+  "URL_MAP/4",
+  prefix: "/service_2",
+  weighted_clusters: [
+    {name: cluster_1, weight: 75},
+    {name: cluster_2, weight: 25}
+  ]
+},
+{
+  "URL_MAP/5",
+  regex: "^/service_2/method_3$",
```
sanjaypujare, line 593: shouldn't it be safe_regex based on above?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+      // Specifies the denominator. If the denominator specified is
+      // less than the numerator, the final fractional percentage
+      // is capped at 1 (100%).
+      uint32 denominator = 2;
+    }
+
+    oneof path_matcher {
+      string path = 1;
+      string prefix = 2;
+      string regex = 3;
```
sanjaypujare, line 187: Should we call it `safe_regex` to be consistent with the field in RDS we will use?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+When assigning names to new actions, we will try to reuse the existing names.
+This will help the xds\_routing policy to not always close/create child
+policy. But we will limit the name reuse only between weighted cluster
+actions that share the same set of clusters (the only update is weights).
+This covers the typical use case where the user keeps updating the cluster
+weights during a canary rollout.
+
+To do this, for each new RDS response, the client will
+
+1.  Calculate the list of new actions to add, and the list of old actions to remove.
```
markdroth, line 445: I think this can wind up being a simple two-pass algorithm that generates a new map based on the new list of WeightedClusters actions and the old map:

1. Pass 1: For each new WeightedClusters action, if there is an exactly matching action in the old map, move it from the old map to the new map.
2. Pass 2: For each new WeightedClusters action to add that did not match in pass 1, if there is an action in the old map that has the exact same list of clusters but different weights, move it from the old map to the new map.  Otherwise, add a new action name to the new map.

Note that the above needs to be done for WeightedClusters actions only.  Actions that directly specify the cluster do not need to go through this algorithm at all.

This is what we implemented in C-core: https://github.com/grpc/grpc/blob/832da8bd26ce4617adf667837f30d664ed191187/src/core/ext/filters/client_channel/xds/xds_client.cc#L2141

markdroth:
CC @htuch

A28-xds-traffic-splitting-and-routing.md
menghanl: Reworded this paragraph.

A28-xds-traffic-splitting-and-routing.md
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+```
+
+The route refers to the action name instead of embedding the action directly,
+so that one action can be shared by multiple routes.
+
+The config must fulfill the following:
+
+*   Routes
+    *   There must be exactly one `path_matcher` rule.
+        *   `path_matcher` is an oneof so this is always true in proto. But json
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+            doesn’t have the oneof concept, so a route may have multiple path
+            matchers, and it will be considered invalid.
+    *   The `action` name must be present in the action list.
+    *   Duplicate matchers are allowed.
+        *   Duplicate matcher means that more than one routes have the same
+            matcher, no matter whether the actions are the same or not.
+        *   Because the first match wins, the second matcher will never be used.
+*   Actions
+    *   The actions in list must be referred by the routes
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
menghanl: Updated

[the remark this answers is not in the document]
menghanl: Updated

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+        next entry.
+
+Action names are used as ID for the child policies. If an action name exists
+in the previous service config, the action’s balancer config will be sent as
+an update to the corresponding child policy (instead of creating a new child
+policy for it).
+
+NOTE: this LB policy is named `xds_routing` (instead of just routing) because
+we plan to make it deeply integrated with the xds resolver as part of
+RouteAction design.
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+```
+
+`ChildPolicy-1` will be picked for `/MyService/MyMethod`, even though the
+second route is an exact match. It is this way because Envoy always picks the
+first matched route. And with header matching, finding the exact match can be
+tricky.
+
+The pick will always be delegated to the picker from the matched child
+policy, even if it’s not in connectivity state READY (as opposed to
+weighted\_target will only pick the READY child policies).
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
menghanl: Not always waiting for the child picker. It can also be a new parent picker (meaning a new RDS response), which will select a different action.

A28-xds-traffic-splitting-and-routing.md
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+The xDS resolver’s overall behavior will be similar as before. It will call
+the xDS client, and will send a service config generated from
+RouteConfiguration (the RDS response) to the parent gRPC client channel.
+
+The xDS client will make an LDS request, with resource\_names set to the
+target URI with “xds:” scheme removed. The client will get
+HttpConnectionManager configuration from the listener response, and make an
+RDS request to get RouteConfiguration.
+
+All new features should be flag-protected until interop testing. Balancers
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
menghanl: Removed

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+field **_best matches_** the server name from URI (with scheme stripped off).
+If no matching VirtualHost is found, the xDS client will report that to the
+xDS resolver, and the xDS resolver will return an error to the client
+channel.
+
+In the matched VirtualHost, the xDS client will iterate over the list of
+[routes](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L76),
+and validate match field and action field:
+
+The route’s [match
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+field](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L179)
+
+*   Must have
+    [path\_specifier](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L360)
+    *   Options supported:
+        [prefix](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L365),
+        [path](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L369)
+        or
+        [safe\_regex](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L405)
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+        *   The presence of this field will cause a NACK
+    *   Not having path\_specifier will cause a NACK
+*   Can have
+    [headers](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L435),
+    with
+    [header\_match\_specifier](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L1382)
+    set to
+    *   Options supported:
+        [exact\_match](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L1384),
+        [safe\_regex\_match](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L1409),
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+will never be used.
+
+Violating any of the validation causes the response to be rejected (NACK’ed).
+The fields not covered by this section will all be ignored.
+
+### Service config generating
+
+The service config generated by the xDS resolver will pick
+`xds_routing_experimental` as the LB policy. As an optimization, when there’s
+only one route, the LB policy picked will be `cds_experimental` or
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
menghanl: It's not a map here yet (it's more of an action set)

A28-xds-traffic-splitting-and-routing.md
menghanl: Global, between RDS responses.

A28-xds-traffic-splitting-and-routing.md
menghanl: Updated

A28-xds-traffic-splitting-and-routing.md
menghanl: We don't have another doc with this behavior. This is the first time we will actually have multiple watchers for the same resource.

A28-xds-traffic-splitting-and-routing.md
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
menghanl: Updated

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,1095 @@
...
+      bool invert_match = 8;
+    }
+    message Fraction {
+      // Specifies the numerator. Defaults to 0.
+      uint32 numerator = 1;
+
+      // Specifies the denominator. If the denominator specified is
+      // less than the numerator, the final fractional percentage
+      // is capped at 1 (100%).
+      uint32 denominator = 2;
```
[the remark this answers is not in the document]
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -173,12 +158,11 @@ message XdsRouting {
     }
     message Fraction {
```
markdroth, line 159: This message isn't needed.  We can just use `google.protobuf.UInt32Value` type for the `match_fraction` field.

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -369,7 +363,18 @@ field](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/r
     [query\_parameters](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L442)
     *   Note: the entire route containing this will be ignored, regardless of
         the other matchers, this is the same as this always evaluates to false,
-        because gRPC does not and will not support query parameters)
+        because gRPC does not and will not support query parameters
+*   Can have
+    [grpc](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L447)
+    *   This matcher will be ignored (but the other matchers in the route will
+        be evaluated). The `content-type` can be used instead to filter gRPC
```
markdroth, line 370: I don't think this last sentence is necessary, because there is no case in which users would need to do this.  In gRPC, all requests will have the same content type, so there's no need for any filtering here.

A28-xds-traffic-splitting-and-routing.md
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
menghanl: Removed.

A28-xds-traffic-splitting-and-routing.md
menghanl: I think is in sync with what the design is proposing. Pass 1 produces 3 lists: `new-actions-to-add`, `actions-that-didn't-change`, and `actions-to-remove`. Pass 2 tries to reuse `actions-to-remove` for `new-actions-to-add`.

I removed the pseudocode because it has unnecessary details, and is just causing confusion.

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -175,10 +167,15 @@ message XdsRouting {
     // match_fraction specifies the percentage this route should be considered.
     // For fraction `N/D`, route is considered if `random(0,D) <= N`.
     //
+    // The number in this field specifies the numerator. Defaults to 0.  The
```
markdroth, line 170: It doesn't default to 0; it defaults to unset (or 1M, if you want to think of it that way).

Actually, I think you can just remove "Defaults to 0" here, since the default is described correctly in the next paragraph.

menghanl: Right. Removed. Thanks!

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,668 @@
...
+# Overview
+
+To support traffic splitting and routing, the xDS resolver will be extended
+to handle the routing fields in RDS responses, and convert them to service
+config.
+
+A new top level LB policy (`xds_routing_experimental`) will be added to do
+RPC routing. It will pick the action to perform for the matched RPCs.
+
+The action can be either to route to a cluster, or to do a weighted cluster
```
htuch, line 39: Do you want to cover what will happen if an unsupported action is specified in the `RouteConfiguration`?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,668 @@
...
+  message Action {
+    repeated LoadBalancingConfig child_policy = 1;
+  }
+  // Route represents a routing rule to send an RPC. It contains a matcher and
+  // the name of the action to take. The action with the name can be found in
+  // the action list.
+  //
+  // Route contains the action name instead of embedding the Action, so Action
+  // can be shared and reused by multiple routes.
+  message Route {
```
htuch, line 142: Why have you redefined Envoy's internal protos here rather than referenced them?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,668 @@
...
+```
+
+`ChildPolicy-1` will be picked for `/MyService/MyMethod`, even though the
+second route is an exact match. It is this way because Envoy always picks the
+first matched route. And with header matching, finding the exact match can be
+tricky.
+
+The pick will always be delegated to the picker from the matched child
+policy, even if it’s not in connectivity state READY (as opposed to
+weighted\_target, which will only pick the READY child policies).
```
htuch, line 251: Why does this distinction exist in gRPC? I don't think this happens in Envoy.

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,668 @@
...
+The weighted target LB policy will have a group of child policies and their
+corresponding weights, and will do weighted pick among the child policies.
+
+This LB policy will have the following
+[configuration](https://github.com/grpc/grpc-proto/blob/dd2dca318eb197b96b60f22297871fb1ed862800/grpc/service_config/service_config.proto#L317)
+
+```proto
+message WeightedTargetConfig {
+  message Target {
+    uint32 weight = 1;
```
htuch, line 67: @markdroth do you think that skipping `request_headers_to_remove` here might have problematic implications? Similar to matchers, this is the type of field that if silently ignored could have security implications.

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,668 @@
...
+    // For fraction `N/D`, route is considered if `random(0,D) <= N`.
+    //
+    // The number in this field specifies the numerator.  The implicit
+    // denominator is 1M (1,000,000). If the numerator specified is bigger than
+    // the denominator, the final fractional percentage is capped at 1 (100%).
+    //
+    // The default value (when this field is not set) is 100%, which means
+    // always consider this route (continue to evaluate the remaining match
+    // criteria).
+    google.protobuf.UInt32Value match_fraction = 4;
```
htuch, line 177: Which field in xDS `RouteConfiguration` does this correspond with?

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,668 @@
...
+GRPC_XDS_EXPERIMENTAL_ROUTING = (true|false)
+```
+
+### Response validation
+
+After LDS/RDS, when the xDS client gets a
+[RouteConfiguration](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route.proto#L24),
+it will look at the list of VirtualHosts and find the element whose
+[domains](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L72)
+field **_best matches_** the server name from URI (with scheme stripped off).
```
htuch, line 304: I would say "most specifically matches".

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,668 @@
...
+*   Can be
+    [Weighted\_clusters](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L719)
+    *   The sum of
+        [weights](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L278)
+        must add up to the
+        [total\_weight](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L335).
+*   Can be
+    [cluster\_header](https://github.com/envoyproxy/envoy/blob/v1.13.1/api/envoy/api/v2/route/route_components.proto#L712)
+    *   Note: the route containing this action will be ignored. This feature can
+        be accomplished by header matching.
```
htuch, line 395: @mattklein123 @envoyproxy/api-shepherds @envoyproxy/udpa-wg FYI the above section in this document provides an interesting example of how xDS `RouteConfiguration` is in practice interpreted by another client (in this case gRPC).

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -0,0 +1,668 @@
...
+
+Duplicate matches are allowed, but because the first match wins, the second
+will never be used.
+
+Violating any of the validation causes the response to be rejected (NACK’ed).
```
htuch, line 400: Do you plan on setting error details to explain why this NACK occurs?

A28-xds-traffic-splitting-and-routing.md
markdroth: In general, we've decided to stick with the "ignore unsupported fields" approach.  This is what happens when new fields are added to a proto API, so it's something users need to be aware of anyway.

We've had quite a bit of debate over this, and I'm not entirely happy with the outcome, because I do think that there are cases where this can cause surprising behavior.  But we've tried to mitigate this by including as many of the matchers as we can in this design, since ignoring a matcher can cause much more surprising behavior than ignoring an action.  Ignoring an action generally just means that a feature is not yet available -- e.g., if retries are configured and we don't retry, it's probably not the end of the world.

A28-xds-traffic-splitting-and-routing.md
markdroth: Is that field used frequently?  What is the typical use-case for it?

If this is something that is used frequently, we can considering pulling it into this design.

@ejona86 @yihuazhang If we do wind up wanting to support this, we'll have to make sure that there are no security implications.

A28-xds-traffic-splitting-and-routing.md
markdroth: We internally pass this config around in JSON form, so having a more compact representation is valuable.

A28-xds-traffic-splitting-and-routing.md
markdroth: It corresponds to the `runtime_fraction` field:

https://github.com/envoyproxy/envoy/blob/eeaebc1bea6d5b52368df18378797b1dcc163141/api/envoy/api/v2/route/route_components.proto#L459

As mentioned below, we will support only the `default_value` field, not the `runtime_key` field, since we don't have the concept of runtime.

A28-xds-traffic-splitting-and-routing.md
markdroth: In gRPC, it's normal for an LB policy to monitor the connectivity state of its children and not send traffic to children that are not reporting READY.  I understand that in the WeightedClusters case, Envoy does not do this, but I think that behavior would be surprising to gRPC users.  And it seems like the Envoy behavior is kind of an accident based on where the WeightedClusters functionality was implemented, rather than something that users would actually want; if we instead moved that functionality to an LB policy in Envoy (as per discussion in https://github.com/envoyproxy/envoy/issues/5598), then it probably would have the same behavior as gRPC.

A28-xds-traffic-splitting-and-routing.md
markdroth: Yes, we do that whenever we NACK a config.

A28-xds-traffic-splitting-and-routing.md
yihuazhang: My concern is with `weight`, it may result in an easily observable pattern that can be used as a side-channel. In HeaderMatcher, can we ignore the rules that want to match the security headers (e.g., "authorization", "x-goog-iam-authorization-token", and "x-goog-iam-authority-selector"). I think it's fair to assume rational end-users will not configure the control plane with those types of rules. Furthermore, I am not sure if it's an over-kill but since users may add a custom metadata that may also contain sensitive information, how about adding a boolean field in metadata struct that dictates whether or not participating the header matching and allowing the user to configure it?

markdroth: @yihuazhang I think we already discussed all of those concerns internally and decided that they were not enough of a problem to worry about.  Has something changed since that discussion?

The question in this thread is not about the `weight` field but rather about supporting the `request_headers_to_remove` field.

yihuazhang: Opps, sorry my bad. Let me take a look again. 

Regarding the discussion on side-channel attack, IIRC although it may not be an issue to prod (since the token is bound to a certificate/authority), it could be a problem to cloud where the token could be a bearer token and there is also a privacy issue as well. An action item is to follow up with Envoy security on this issue (as Julien suggested).

markdroth: Regarding the side-channel attack, yes, that was the follow-up item.  Please keep us in the loop on that, and we can coordinate with whatever changes need to happen on the Envoy side.

yihuazhang: Will do! 

Regarding `request_headers_to_remove`, one common use-case could be as follows: A service frontend received a user's RPC request containing an oauth token. The frontend 1) wants to send the request to its backend and also 2) wants to log the request without the security header (similar to binary logging). It is easy to configure it with the help of `request_headers_to_remove` (i.e., taking the action of `request_headers_to_remove` when the URI matches that of log service).

markdroth: Yes, I understand that use-case.

What I was asking was whether there are any security implications to having the client support `request_headers_to_remove`.  For example, if someone compromises the xDS server, they could presumably cause a DoS attack by causing clients to drop oauth headers.

I suspect this is not really a concern, since someone that wants to do a DoS attack and has gained access to the xDS server can simply shut down the xDS server.  But please consider whether there are any other issues here.

A28-xds-traffic-splitting-and-routing.md
menghanl: Done

A28-xds-traffic-splitting-and-routing.md
htuch: Fair, but it might be reasonable to state this explicitly.

A28-xds-traffic-splitting-and-routing.md
htuch: Fair, but I think it would help to provide some wording on the context here.

A28-xds-traffic-splitting-and-routing.md
htuch: The value would be in allowing a header to be available to the gRPC client equivalent of Envoy's filters, e.g. RBAC, but then not exposing it to the upstream. So, going with the auth token, the client shares an auth token with its internal RBAC but doesn't want to put this on the wire.

A28-xds-traffic-splitting-and-routing.md
htuch: I see, maybe explain this explicitly in the comment or writeup, I don't see many mentions of this field when searching.

A28-xds-traffic-splitting-and-routing.md
markdroth: Good point.  Menghan, please add a note somewhere indicating that this is our general approach to unsupported fields in the xDS API.

A28-xds-traffic-splitting-and-routing.md
markdroth: We don't currently support RBAC, nor do we have a mechanism for filters that would see the headers at this level of the stack.  We do have interceptors, but they live higher in the stack than the xDS code, so there's not currently any way to drive them from xDS.

We will need to add RBAC support and an extensibility mechanism here at some point, but since we don't have this yet, it's not clear that there's currently any use-case for `request_headers_to_remove`.  My inclination is to defer implementing this until we add support for something that would require it.

A28-xds-traffic-splitting-and-routing.md
markdroth: Good point.  Menghan, please document this.

A28-xds-traffic-splitting-and-routing.md
menghanl: In the "Response validation" section, [line 401](https://github.com/grpc/proposal/pull/178/files#diff-86f80616477d2453f4ea068d913c9b0aR401), there's
> The fields not covered by this section will all be ignored.

It seems to be the right section for this, because it's where we talk about each field in the response.

Let me know if you think there's a better place for it.

A28-xds-traffic-splitting-and-routing.md
menghanl: Added one line in section "Service config generating">"Route list", line 456.

A28-xds-traffic-splitting-and-routing.md
menghanl: What kind of explanation do you have in mind?

The weighted_target behavior is the normal gRPC balancing policy behavior, the routing policy is the special one, so there's this one sentence explaining that.

Do you want me to talk about the difference between Envoy and gRPC here?
Or one more sentence on why routing is different from weighted_target?

A28-xds-traffic-splitting-and-routing.md
markdroth: LGTM, thanks!

A28-xds-traffic-splitting-and-routing.md
markdroth: I think it will be useful to point out both (a) differences from Envoy behavior for those who have experience with Envoy but not gRPC and (b) differences from normal gRPC behavior for those that have experience with gRPC and are now starting to use xDS.

Specifically:
- In the xds_routing section, we should mention that, unlike most gRPC LB policies, the xds_routing policy will not ignore children that are not reporting READY.
- In the weighted_target section, we should mention that, unlike Envoy, gRPC will not send any traffic to clusters in the WeightedClusters list that are not reporting READY.

menghanl: Done. Added explanation for weighted_target and routing policy

A28-xds-traffic-splitting-and-routing.md
```diff
@@ -246,9 +250,11 @@ second route is an exact match. It is this way because Envoy always picks the
 first matched route. And with header matching, finding the exact match can be
 tricky.
 
-The pick will always be delegated to the picker from the matched child
-policy, even if it’s not in connectivity state READY (as opposed to
-weighted\_target, which will only pick the READY child policies).
+The pick will consider the child policies that are not reporting READY (unlike
+in most gRPC LB policies, e.g. weighted_target, only the READY child policies
```
markdroth, line 254: s/, only/, where only/

A28-xds-traffic-splitting-and-routing.md
markdroth: Looks great (modulo one minor wording nit, which I pointed out below).

A28-xds-traffic-splitting-and-routing.md
menghanl: Done.

Thanks for the review!

A28-xds-traffic-splitting-and-routing.md
ejona86: > What I was asking was whether there are any security implications to having the client support `request_headers_to_remove`

I agree it would need some thought.

One attack could be to remove user creds so that a request uses the service's transport credentials instead of the user's. The attacker could then send the request it wishes to a backend. But that depends on specifics of how the system determines the "effective" creds.

Removing :authority could impact virtual hosting. We were already uncertain how severe attacks like this would be.

Removing X-Forwarded-For could defeat IP checks in poor implementations that _expect_ the proxy to provide the header and disable the checks if the header isn't present. That seems broken though.

markdroth: For now, it doesn't seem like we need to support this feature, so I think we can defer the security debate until we do wind up adding this.

PR #180, opened 2020-05-29
L67: C++ callback-based asynchronous API

vjpai:
The callback API allows for easy use of asynchronous RPCs in gRPC C++.


L67-cpp-callback-api.md
```diff
@@ -0,0 +1,105 @@
...
+* No explicit polling required for notification of completion of RPC actions.
+   - In practice, these requirements mean that there must be a library-controlled poller for monitoring such actions. This _EventManager_ is described in another gRFC.
+* As in the synchronous API, server RPCs have an application-defined method handler function as part of their service definition. The library invokes this method handler when a new server RPC starts.
+* LIke the synchronnous API and unlike the completion-queue-based asynchronous API, there is no need for the application to "request" new server RPCs. Server RPC context structures will be allocated and have their resources allocated as and when RPCs arrive at the server.
+
+### Reactor model
+
+The most general form of the callback API is built around a _reactor_ model. Each type of RPC has a reactor base class provided by the library. These types are:
+
+* `ClientUnaryReactor` for unary calls
```
markdroth, line 48: This could be a table, if you're feeling ambitious. :)

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,105 @@
...
+   - `OnReadDone`, `OnWriteDone`, `OnWritesDoneDone`: methods invoked by the library at the completion of an explicitly-specified streaming RPC operation. Provides a `bool ok` argument to indicate whether the completion was successful.
+   - `OnReadInitialMetadataDone`: notifying client RPCs that the server has sent initial metadata. It has a bool `ok` argument to indicate whether the completion was successful. This is automatically invoked by the library for all RPCs; it is uncommonly used however
+   - `OnCancel`: for server RPCs only, called if an RPC is canceled before it has a chance to successfully send status to the client side
+   - `OnDone`: called when all outstanding RPC operations are completed. Provides status on the client side. 
+1. RPC completion prevention methods. These are methods provided by the class and are not virtual.
+   - `AddHold`, `RemoveHold`: for client RPCs, prevents the RPC from being considered complete (ready for `OnDone`) until each AddHold is matched to a corresponding RemoveHold. These are used to perform _extra-reaction flows_, which refers to  streaming operations initiated from outside a reaction method or method handler.
```
markdroth, line 71: This probably deserves a more detailed explanation, probably with an example.  I don't think it's going to be clear to readers why this is needed or what it's useful for.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,105 @@
...
+* `ServerReadReactor` for client-streaming 
+* `ServerWriteReactor` for server-streaming 
+
+Client RPC invocations from a stub provide a reactor pointer as one of their arguments, and the method handler of a server RPC must return a reactor pointer.
+
+These base classes provide three types of methods:
+
+1. Operation-initiation methods: start an asynchronous activity in the RPC. These are methods provided by the class and are not virtual.
+   - `StartRead`, `StartWrite`, `StartWritesDone`: streaming operations as appropriate to the RPC variety
+   - `StartSendInitialMetadata`: for server responses where the application wants to send the metadata separately from the first response message (uncommonly used)
```
markdroth, line 63: Couldn't this be used on the client also, for bidi streaming?

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,105 @@
...
+* `ClientReadReactor` for server-streaming 
+* `ClientWriteReactor` for client-streaming 
+* `ServerUnaryReactor` for unary calls
+* `ServerBidiReactor` for bidi-streaming 
+* `ServerReadReactor` for client-streaming 
+* `ServerWriteReactor` for server-streaming 
+
+Client RPC invocations from a stub provide a reactor pointer as one of their arguments, and the method handler of a server RPC must return a reactor pointer.
+
+These base classes provide three types of methods:
```
markdroth, line 59: I would like to see a more detailed description of each individual method here.  Also, please explicitly indicate in a consistent way whether each method is client-only, server-only, or for both client and server.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Unary RPC shortcuts
+
+As a shortcut, client-side unary RPCs _may_ bypass the reactor model by directly providing a `std::function` for the library to call at completion rather than a reactor object pointer. This is semantically equivalent to a reactor in which the `OnDone` function simply invokes the specified function (but can be implemented in a slightly faster way since such an RPC will definitely not wait separately for initial metadata from the server). In practice, this is the common and recommended model for client-side unary RPCs, unless they have a specific need to wait for initial metadata before getting their full response message.
+
+Server-side unary RPCs have the option of returning a library-provided default reactor when there method handler is invoked. This default reactor provides a Finish method, but does not provide a user callback for `OnCancel` and `OnDone`. In practice, this is the common and recommended model for most server-side unary RPCs unless they specifically need to react to an `OnCancel` callback or do cleanup work after the RPC fully completes.
```
markdroth, line 77: s/there/their/

markdroth, line 77: s/Finish/`Finish`/

L67-cpp-callback-api.md
chwarr: Yes, example, please!

L67-cpp-callback-api.md
vjpai: person being put into a shame cube

vjpai:
Gentlemen, I appreciate your enthusiasm for the callback API, but this is still a draft PR. I'll click "ready to review" soon, promise!

L67-cpp-callback-api.md
chwarr, line 77: Where are the method handlers methods run? Can a slow handler method prevent the rest of the system from making forward progress? Are there any mitigations for this?

My experience implementing such callback APIs has been that user callbacks, by default, need to run on separate threads from the network/IO threads. This helps mitigate deadlocks when handlers inadvertently do blocking work. An "I solemnly swear I won't block" escape hatch can be provided for apps that need the performance gained by one fewer context switch. Another approach I've seen is to periodically inject another worker thread when they're all block to see if that frees things up.

chainmasterx10:
@vjpai What is the status of this proposal now? I already saw corresponding changes in the code base. It would be great make the plan of record clear. So does the related proposal L69.

L67-cpp-callback-api.md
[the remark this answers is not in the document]
vjpai: We actually had an alternative proposal for such but didn't pursue it under the idea that our initial users had some experience with AddHold/RemoveHold because of some internal tooling and since that could be added either later or in a wrapper.

L67-cpp-callback-api.md
vjpai: Absolutely correct. I added it in the comments and will be adding it to the proposal.

vjpai:
@jingairpi Now that #25169 is done, we can push forward with this proposal. My plan is to add some examples, address reviewer comments on the proposal, and then mark this ready for review.

L67-cpp-callback-api.md
vjpai: SendInitialMetadata on the client is the RPC initiation; it is done when the RPC is called from the stub.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Proposal
+
+The callback API is designed to have the performance and thread scalability of an asynchronous API without the burdensome programming model of the completion-queue-based model. In particular, the following are fundamental guiding principles of the API:
+
+* Library directly calls user-specified code at the completion of RPC actions.
+* No explicit polling required for notification of completion of RPC actions.
+   - In practice, these requirements mean that there must be a library-controlled poller for monitoring such actions. This _EventManager_ is described in another gRFC.
+* As in the synchronous API, server RPCs have an application-defined method handler function as part of their service definition. The library invokes this method handler when a new server RPC starts.
+* LIke the synchronnous API and unlike the completion-queue-based asynchronous API, there is no need for the application to "request" new server RPCs. Server RPC context structures will be allocated and have their resources allocated as and when RPCs arrive at the server.
```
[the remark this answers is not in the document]
vjpai: I think my previous keyboard had some sticky keys, I guess I should get a mechanical one like you.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,105 @@
...
+   - Application associates each RPC action that it initiates with a tag
+   - The library performs each RPC action
+   - The library posts the tag of a completed action onto a completion queue
+   - The application must poll the completion queue to determine which asynchronously-initiated actions have completed
+
+The goal of the synchronous version is to be easy to program. However, this comes at the cost of high thread-switching overhead and high thread storage for systems with many concurrent RPCs. On the other hand, the asynchronous API allows the application full control over its threading and thus can scale further. It does, however, present challenges in deciding how many completion queues to use and how many threads to use for polling them, as one can either optimize for reducing thread hops, avoiding stranding, reducing CQ contention, or improving locality. These goals are often in conflict and require substantial tuning.
+
+### Related proposals
+
+* The [EventManager](https://www.github.com/grpc/proposal/XXX) is critical to the operation of the callback API.
```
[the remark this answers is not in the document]
vjpai: Since the EM project is abandoned and not used in the callback API implementation, I'm removing this line.

L67-cpp-callback-api.md
vjpai: I think the API is the limit of my ambition, formatting not so much....

L67-cpp-callback-api.md
vjpai: I've written up full example code for the greeter and route_guide and linked to them. In particular, the route_guide uses a hold to account for delayed writes.

vjpai:
This is being revived. I will complete the remaining sections early next week.

pyssling:
I'm wondering if maybe you could state something about thread safety of the API?

vjpai:
@pyssling How does the latest look?

pyssling:
@vjpai Good, thanks!

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,126 @@
...
+   - `OnDone`: called when all outstanding RPC operations are completed for both client and server RPCs. For the client-side, it additionally provides the status that the server sent with its `Finish` call. Should be used for any application-level RPC-specific cleanup. *Client and server.*
+   - _Thread safety_: the above calls may take place concurrently, except that `OnDone` will always take place after all other reactions. No further RPC operations should be issued after `OnDone` is invoked.
+   - **IMPORTANT USAGE NOTE** : code in any reaction must not block for an arbitrary amount of time since reactions are executed on a finite-sized library-controlled threadpool. If any long-term blocking operations (like sleeps, file I/O, synchronous RPCs, or waiting on a condition variable) must be invoked as part of the application logic, then it is important to push that outside the reaction so that the reaction can complete in a timely fashion. One way of doing this is to push that code to a separate application-controlled thread.
```
chwarr, line 72:
```suggestion
   - **IMPORTANT USAGE NOTE** : code in any reaction must not block for an arbitrary amount of time since reactions are executed on a finite-sized library-controlled threadpool. If any long-term blocking operations (like sleeps, file I/O, synchronous RPCs, or waiting on a condition variable) must be invoked as part of the application logic, then it is important to push that outside the reaction so that the reaction can complete in a timely fashion. One way of doing this is to push that code to a separate application-controlled work queue.
```

Consider using "work queue" instead of "thread" to minimize the change for misinterpretation that the recommendation is that start a new thread for each blocking call. A "work queue" may do that, but it may also be able to handle multiple blocking calls with one thread.

chwarr (approved):
I don't remember what this looked like two years ago, when I initially reviewed it. :-) This looks good, at a high level.

L67-cpp-callback-api.md
chwarr, line 72: When you were prototyping this inside of Google, did you ever emit a log message when a reaction blocked for a long time? What that useful? Did it end up in the gRPC version of the implementation?

I'm unsure if such a log message would be useful or if it would be noise. Is there a "good enough" static threshold?

vjpai:
@chwarr Thanks for the review! I'm integrating and/or addressing your comments.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,126 @@
...
+1. RPC completion prevention methods. These are methods provided by the class and are not virtual.
+   - `AddHold`, `RemoveHold`: for client RPCs, prevents the RPC from being considered complete (ready for `OnDone`) until each `AddHold` is matched to a corresponding `RemoveHold`. These are used to perform _extra-reaction flows_, which refers to  streaming operations initiated from outside a reaction method or method handler. *Client only*. (These are not needed at the server because even if there are extra-reaction flows, the end of the application's involvement with a server RPC is very clearly documented by the invocation of the `Finish` call.)
```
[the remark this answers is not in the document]
vjpai: You're correct about that. That said, how about we push that to a future gRFC? I think the API is useful without it and has been used plenty in the current form.

L67-cpp-callback-api.md
vjpai: Let me think about this one; I think I want something stronger than work queue even if thread isn't the right term since work queue isn't really a well-defined term whereas thread is.

vjpai: I should clarify that the Google version wasn't a separate prototype; the code that we've been running in Google is literally the callback API that has been developed out here at Github in bits and pieces since 2019. It was just usable internally because we had an additional  pre-existing piece that is hinted at in the implementation: an iomgr variant that satisifes `grpc_iomgr_run_in_background`. We had hoped to make an OSS version of that in #182 but had to abandon that project because of dependences, though it might be revived in yet another form down the road. That was why this ended up taking longer than planned and how we worked around that with grpc/grpc#25169 .

But getting back to your question: we never tried such a log but it seems like it would be a good idea. We could, for example, start a timer when a callback is initiated for say 100 ms in the future and then either cancel it at callback completion or let it fire and log. I suppose that this could be a useful debugging tool.

L67-cpp-callback-api.md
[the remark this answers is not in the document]
chwarr: SGTM.

L67-cpp-callback-api.md
chwarr: > We could, for example, start a timer when a callback is initiated for say 100 ms in the future and then either cancel it at callback completion or let it fire and log. I suppose that this could be a useful debugging tool.

That would work. It may be even cheaper to just time how long the callback takes and log after it returns, even if it takes 12 minutes. That would reduce the overhead of having a timer and another callback.

```scheme
(let [(timer (start-timer))
      (callback-result (callback ...))]
  (begin
    (if (> (timer-duration timer) (ms 100)) (log-too-long ...))
    callback-result))
```

> We had hoped to make an OSS version of that in #182

I remember that proposal. The "handle" type in there really needs to be at least pointer sized to work on non-Linux OSes. The native implementations that one would likely use to implement an I/O manager on both Windows (I/O completion queues) and macOS (libdispatch) use pointer-sized handles, not `int`. In general, wherever Linux has `int fd`, it seems like gRPC will need need `void* handle` to work cross-platform.

vjpai: Are you taunting me with Lisp as a fellow Rice alum? Seems like a reasonable thing to do. We should add some plumbing to pass the callback's `__FILE__` and `__LINE__` into the library so that the log would be human-readable and useful.

#182 failed for reasons beside handlesize iirc

chwarr: > Are you taunting me with Lisp as a fellow Rice alum?

Yup! And, come now, it was Scheme. It had square brackets. 😜

vjpai: It's been 30 years since I took COMP 210 so you'll have to forgive my lapse. The book was called The Little LISPer back then even though it was Scheme.

vjpai:
@markdroth Since you've now approved grpc/grpc#26249, I think there are no remaining blockers for this gRFC. Do you want any additional things addressed here, or can you approve this? Thanks in advance.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+* `ServerUnaryReactor` for unary calls
+* `ServerBidiReactor` for bidi-streaming 
+* `ServerReadReactor` for client-streaming 
+* `ServerWriteReactor` for server-streaming 
+
+Client RPC invocations from a stub provide a reactor pointer as one of their arguments, and the method handler of a server RPC must return a reactor pointer.
+
+These base classes provide three types of methods:
+
+1. Operation-initiation methods: start an asynchronous activity in the RPC. These are methods provided by the class and are not virtual.
```
markdroth, line 60: I suggest reorganizing this such that each method is described in its own bullet.  For example:

- `StartCall()`: (Client only.)  Initiates a new call with a specific set of initial metadata.  Must be called exactly once.  No reads or writes will actually be started until this is called (i.e., any previous calls to `StartRead()` or `StartWrite()` will be queued until `StartCall()` is invoked).
- `StartSendInitialMetadata()`: (Server only.) Sends initial metadata.  To be used in cases where initial metadata should be sent without sending a message.  Optional; if not called, initial metadata will be sent when `StartWrite()` or `Finish()` is called.  May not be invoked more than once.
- `StartRead()`: Starts a read of a message.  `OnReadDone()` will be invoked when the read is complete.  Only one read may be outstanding at any given time.  On the client, if invoked before `StartCall()`, 
- `StartWrite()`: Starts a write of a message.  `OnWriteDone()` will be invoked when the write is complete.  Only one write may be outstanding at any given time.
- `StartWritesDone()`: Similar to `StartWrite()`, but indicates that this will be the last message sent on the call.  When the operation is complete, `OnWritesDoneDone()` will be invoked.  May be used at most once per call.
- `Finish()`:  (Server only.) Sends completion status to the client.  Must be invoked exactly once, even for calls that have already been cancelled.

Similar thing for the completion methods and the completion-prevention methods.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+   - `StartRead`, `StartWrite`, `StartWritesDone`: streaming operations as appropriate to the RPC variety. *Client and server*.
```
markdroth, line 61: Is `StartWritesDone()` useful on the server?  If so, what case does it cover?  When would an application want to use that instead of `StartWrites()`?

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+
+Examples are provided in [the PR to de-experimentalize the callback API](https://github.com/grpc/grpc/pull/25728).
+
+### Unary RPC shortcuts
+
+As a shortcut, client-side unary RPCs _may_ bypass the reactor model by directly providing a `std::function` for the library to call at completion rather than a reactor object pointer. This is semantically equivalent to a reactor in which the `OnDone` function simply invokes the specified function (but can be implemented in a slightly faster way since such an RPC will definitely not wait separately for initial metadata from the server). In practice, this is the common and recommended model for client-side unary RPCs, unless they have a specific need to wait for initial metadata before getting their full response message. As in the reactor model, the function provided as a callback may not include operations that block for an arbitrary amount of time.
```
markdroth, line 80: What API is this callback passed to?

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+
+Server-side unary RPCs have the option of returning a library-provided default reactor when their method handler is invoked. This default reactor provides a `Finish` method, but does not provide a user callback for `OnCancel` and `OnDone`. In practice, this is the common and recommended model for most server-side unary RPCs unless they specifically need to react to an `OnCancel` callback or do cleanup work after the RPC fully completes.
```
markdroth, line 82: How is this option selected by the application?

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+
+* `DefaultReactor` may be used by a method handler to return a default reactor from a unary RPC.
+* `GetRpcAllocatorState`: see advanced topics section
+
+Additionally, the `AsyncNotifyWhenDone` function is not present in the `CallbackServerContext`. All method handler functions for the callback API take a `CallbackServerContext*` as their first argument. `ServerContext` (used for the sync and CQ-based async APIs) and `CallbackServerContext` (used for the callback API) actually use the same underlying structure and thus their object pointers are meaningfully convertible to each other via a `static_cast` to `ServerContextBase*`. We recommend that any helper functions that need to work across API variants should use a `ServerContextBase` pointer or reference as their argument rather than a `ServerContext` or `CallbackServerContext` pointer or reference. For example, `ClientContext::FromServerContext` now uses a `ServerContextBase*` as its argument; this is not a breaking API change since the argument is now a parent class of the previous argument's class.
```
markdroth, line 91: Suggest starting a new paragraph after the first sentence here, since the first sentence is really a separate topic from the rest of the paragraph.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+
+The goal of the synchronous version is to be easy to program. However, this comes at the cost of high thread-switching overhead and high thread storage for systems with many concurrent RPCs. On the other hand, the asynchronous API allows the application full control over its threading and thus can scale further. The biggest problem with the asynchronous API is that it is just difficult to use. Server RPCs must be explicitly requested, RPC polling must be explicitly controlled by the application, lifetime management is complicated, etc. These have proved sufficiently difficult that the full features of the asynchronous API are basically never used by applications. Even if one can use the async API correctly, it also presents challenges in deciding how many completion queues to use and how many threads to use for polling them, as one can either optimize for reducing thread hops, avoiding stranding, reducing CQ contention, or improving locality. These goals are often in conflict and require substantial tuning.
+
+### Related proposals
+
+* The C++ callback API is built on top of a new [callback completion queue in core](https://github.com/grpc/proposal/pull/181)
```
markdroth, line 30: Is it actually still built on that?  Or did your callback alternative approach wind up using a different type of CQ instead?

markdroth (commented):
Thanks for doing this!

L67-cpp-callback-api.md
vjpai: Oops! That wasn't meant to be on the server.

L67-cpp-callback-api.md
vjpai: Sounds good, will do.

L67-cpp-callback-api.md
vjpai: To the stub call itself. I guess I should give the example more clearly. It's in the examples code in the de-experimentalization PR linked.

L67-cpp-callback-api.md
vjpai: The method handler is passed in a CallbackServerContext and you just have to call DefaultReactor() on that. That's mentioned below but it should be made more clear here.

L67-cpp-callback-api.md
vjpai: Good observation. It's built on 2 possible implementations, so I should list both. Will get back to this on Monday.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Abstract
+
+Provide an asynchronous gRPC API for C++ in which the completion of RPC actions in the library will result in callbacks to user code,
+
+## Background
+
+Since its initial release, gRPC has provided two C++ APIs:
+
+* Synchronous API
```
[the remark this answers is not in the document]
vjpai: Added just below this part.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+* The C++ callback API is built on top of a new [callback completion queue in core](https://github.com/grpc/proposal/pull/181)
+* The API structure has substantial similarities to the gRPC-Node and gRPC-Java APIs
+
+## Proposal
+
+The callback API is designed to have the performance and thread scalability of an asynchronous API without the burdensome programming model of the completion-queue-based model. In particular, the following are fundamental guiding principles of the API:
+
+* Library directly calls user-specified code at the completion of RPC actions.
+* No explicit polling required for notification of completion of RPC actions.
+   - In practice, these requirements mean that there must be a library-controlled poller for monitoring such actions.
```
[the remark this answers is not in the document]
vjpai: Explained in the Implementation section now. Linked from here.

L67-cpp-callback-api.md
```diff
@@ -0,0 +1,125 @@
...
+* `ServerWriteReactor` for server-streaming 
+
+Client RPC invocations from a stub provide a reactor pointer as one of their arguments, and the method handler of a server RPC must return a reactor pointer.
+
+These base classes provide three types of methods:
+
+1. Operation-initiation methods: start an asynchronous activity in the RPC. These are methods provided by the class and are not virtual.
+   - `StartRead`, `StartWrite`, `StartWritesDone`: streaming operations as appropriate to the RPC variety. *Client and server*.
+   - `StartCall` : actually causes operations to be initiated. Before this point, any `StartRead`/`StartWrite`/etc. are just backlogged. This exists to allow precise control over the initiation of streaming operations. *Client only*. (This is not needed at the server side since streaming operations at the server are released from backlog automatically by the library as soon as the application returns a reactor from the method handler.)
+   - `StartSendInitialMetadata`: for server responses where the application wants to send the metadata separately from the first response message (uncommonly used). *Server only*.
```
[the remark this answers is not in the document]
vjpai: Done.

L67-cpp-callback-api.md
vjpai: Thanks for the great suggestion. Done.

vjpai:
@markdroth this is ready for review again. Thanks in advance!

yayapichfficial (approved):

markdroth (commented):
This is looking pretty good overall!  Please let me know if you have any questions.

L67-cpp-callback-api.md
```diff
@@ -17,69 +17,73 @@ Since its initial release, gRPC has provided two C++ APIs:
...
+   - `AddHold`: (*Client only*) This prevents the RPC from being considered complete (ready for `OnDone`) until each `AddHold` on an RPC's reactor is matched to a corresponding `RemoveHold`. An application uses this operation before it performs any  _extra-reaction flows_, which refers to streaming operations initiated from outside a reaction method. Note that in general, an RPC cannot complete
+before `StartCall`, so this is not needed for any extra-reaction flows that take place before `StartCall`. As long as there are any holds present on an RPC, though, it may not have `OnDone` called on it, even if it has already received server status and has no other operations outstanding. May be called 0 or more times on any client RPC.
+   - `RemoveHold`: (*Client only*) Removes a hold reference on this client RPC. Must be called exactly as many times as `AddHold` was called on the RPC, and may not be called more times than `AddHold` has been called so far for any RPC. Once all holds have been removed, the server has provided status, and all other operations have completed for an RPC, the library may invoke `OnDone` for that RPC.
```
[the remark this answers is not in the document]
vjpai: I meant "may" as in "is allowed to" but I see your point.

L67-cpp-callback-api.md
```diff
@@ -118,8 +124,10 @@ Besides the content described in the background section, the rationale also incl
...
+
+The other implementation is internal to Google; although it uses most of the same code as the open-source project, it additionally incorporates a library that provides threads for efficient event polling and callback execution. This implementation uses the open-source [core callback API](https://github.com/grpc/proposal/pull/181) that is also used by the Python async implementation, but it cannot currently be open-sourced because the library described above is tightly integrated with other Google internals and has a non-portable implementation. It is possible that this approach will be open-sourced at some point.
```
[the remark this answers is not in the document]
vjpai: Done.

L67-cpp-callback-api.md
```diff
@@ -17,69 +17,73 @@ Since its initial release, gRPC has provided two C++ APIs:
...
+1. Operation-completion reaction methods: notification of completion of asynchronous RPC activity. These are all virtual methods that default to an empty function (i.e., `{}`) but may be overridden by the application's reactor definition. These are invoked by the library.
+   - `OnReadInitialMetadataDone`: (*Client only*) Invoked by the library to notify that the server has sent an initial metadata response to a client RPC. It has a bool `ok` argument. If `ok` is true, then the RPC received initial metadata normally. If it is false, there is no initial metadata either because the call has failed or because the call received a trailers-only response (which means that there was no actual message and that any information normally sent in initial metadata has been dispatched instead to trailing metadata, which is allowed in the gRPC HTTP/2 transport format). This reaction is automatically invoked by the library for RPCs of all varieties; it is uncommonly used as an application-defined reaction however.
```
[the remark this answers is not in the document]
vjpai: Done.

L67-cpp-callback-api.md
```diff
@@ -17,69 +17,73 @@ Since its initial release, gRPC has provided two C++ APIs:
...
+   - `OnWritesDoneDone`: (*Client only*) Invoked by the library in response to a `StartWritesDone` operation. Has a bool `ok` argument that indicates whether the writes-done message was successfully sent; a false value indicates an RPC failure.
+   - `OnCancel`: (*Server only*) Invoked by the library if an RPC is canceled before it has a chance to successfully send status to the client side. The reaction may be used for any cleanup associated with cancellation or to guide the behavior of other parts of the system (e.g., by setting a flag in the service logic associated with this RPC to stop further processing since the RPC won't be able to send outbound data anyway). Note that even cancelled RPCs _must_ get a `Finish` call to cleanup all their library state.
+   - `OnDone`: Invoked by the library when all outstanding RPC operations are completed for a given RPC. For the client-side, it additionally provides the status that the server sent with its `Finish` call. Should be used for any application-level RPC-specific cleanup.
```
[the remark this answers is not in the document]
vjpai: Done.

L67-cpp-callback-api.md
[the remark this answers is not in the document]
vjpai: Fixed this.

L67-cpp-callback-api.md
```diff
@@ -17,69 +17,73 @@ Since its initial release, gRPC has provided two C++ APIs:
...
-   - `OnDone`: called when all outstanding RPC operations are completed for both client and server RPCs. For the client-side, it additionally provides the status that the server sent with its `Finish` call. Should be used for any application-level RPC-specific cleanup. *Client and server.*
+1. Operation-initiation methods: start an asynchronous activity in the RPC. These are methods provided by the class and are not virtual. These are invoked by the application logic.
+   - `StartCall`: (*Client only*) Initiates the operations of a call from the client, including sending any client-side initial metadata associated with the RPC. Must be called exactly once. No reads or writes will actually be started until this is called (i.e., any previous calls to `StartRead`, `StartWrite`, or `StartWritesDone` will be queued until `StartCall` is invoked). This operation is not needed at the server side since streaming operations at the server are released from backlog automatically by the library as soon as the application returns a reactor from the method handler, and because there is a separate method just for sending initial metadata.
```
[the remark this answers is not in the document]
vjpai: I've made it like other gRFCs - provide the full method signature but write about methods in text format rather than as comments.

vjpai:
@markdroth Thanks for the detailed reviews so far. This is ready to review again.

L67-cpp-callback-api.md
```diff
@@ -124,10 +124,10 @@ Besides the content described in the background section, the rationale also incl
...
+There is more than once mechanism available for implementing the background polling required by the C++ callback API. One has been implemented [on top of the C++ completion queue API](https://github.com/grpc/grpc/pull/25169). In this approach, the callback API uses a number of library-owned threads to call `Next` on an async CQ that is owned by the internal implementation. Currently, the thread count is automatically selected by the library with no user input and is set to half the system's core count, but no less than 2 and no more than 16. This selection is subject to change in the future based on our team's ongoing performance analysis and tuning efforts. Despite being built on the CQ-based async API, the developer using the callback API does not need to consider any of the CQ details (e.g., shutdown, polling, or even the existence of a CQ).
```
markdroth, line 127: s/more than once/more than one/

L67-cpp-callback-api.md
```diff
@@ -124,10 +124,10 @@ Besides the content described in the background section, the rationale also incl
...
+It is the gRPC team's intention that that implementation is only a temporary solution. A new structure called an `EventEngine` is being developed to provide the background threads needed for polling, and this sytem is also intended to provide a direct API for application use. This event engine would also allow the direct use of the [core callback API](https://github.com/grpc/proposal/pull/181) that is currently only used by the Python async implementation. If this solution is adopted, there will be a new gRFC for it. This new implementation will not change the callback API at all but rather will only affect its performance. The C++ code for the callback API already has `if` branches in place to support the use of a poller that directly supplies the background threads, so the callback API will naturally layer on top of the `EventEngine` without further development effort.
```
markdroth, line 129: Please link to https://github.com/grpc/proposal/blob/master/L68-core-callback-api.md instead of to the PR that created it.

markdroth (approved):
This looks great!  Feel free to merge after addressing the few remaining nits.

Kleclerc18 (approved):

PR #181, opened 2020-06-01
L68: Core callback completion queue

vjpai:


L68-core-callback-api.md
```diff
@@ -0,0 +1,106 @@
...
+tag
+
+The completion queue accessor functions may block until a specified
+deadline, and during that time the user's thread is available for the
+library to use in performing RPC operations or polling the operating
+system for the completion of operations.
+
+### Related proposals
+
+* The [EventManager](https://www.github.com/grpc/proposal/XXX) is
```
markdroth, line 47: This can now point to #182.

L68-core-callback-api.md
```diff
@@ -0,0 +1,106 @@
...
+a C89 int) specifying whether or not the operation completed
+normally (corresponding to the `ok` result of
+`grpc_completion_queue_next` on the `GRPC_CQ_NEXT` queue).
+* `int inlineable`: specifies whether this functor can be run inline
+in the same thread as the one that detected it rather than sending
+it to an executor thread. This should only be
+used very carefully for trivial internally-generated callbacks that
+do not use locks.
+
+## Rationale
```
markdroth, line 80: It may be worth noting here that this is essentially an internal hack that is being used to help us transition from the current CQ-based polling model to the new EventManager polling model.  Once everything has moved to EventManager and we have eliminated support for the current thread-borrowing polling model, we will look at completely eliminating `grpc_completion_queue` from the C-core API.

L68-core-callback-api.md
```diff
@@ -0,0 +1,106 @@
...
+`GRPC_CQ_CALLBACK`. Although this is called a completion queue because
+it supports the semantic interface of `grpc_completion_queue`,
+operations in this will never be queued in this structure. Instead, a
+completion will cause a user-specified function to be called.
+
+The tags injected into this completion queue
+are still of type `void*` so as not to affect the API of RPC operation
+initiation functions, but these will be interpreted by the
+library. These tags *must* point to a struct of type
+`grpc_completion_queue_functor` which has two user-provided members:
```
markdroth, line 65: Is this type exposed in the C-core API?  Are wrapped languages expected to use this?

More generally, if the callback completion queue is required to use EventManager, then what work is needed to get wrapped languages using EventManager?  @muxi @guantaol

L68-core-callback-api.md
```diff
@@ -0,0 +1,106 @@
...
+completion queue type will not provide any accessor function to drive
+the progress of the library
+* The callback completion queue is the implementation mechanism for the [C++ callback API](https://www.github.com/grpc/proposal/pull/180)
+
+## Proposal
+
+gRPC core will now support a new completion queue type called
+`GRPC_CQ_CALLBACK`. Although this is called a completion queue because
+it supports the semantic interface of `grpc_completion_queue`,
+operations in this will never be queued in this structure. Instead, a
```
chwarr, line 58: Minor: I'm not sure what "operations in this will never be queue in this structure" is trying to say. Reword?

L68-core-callback-api.md
```diff
@@ -0,0 +1,106 @@
...
+completion will cause a user-specified function to be called.
+
+The tags injected into this completion queue
+are still of type `void*` so as not to affect the API of RPC operation
+initiation functions, but these will be interpreted by the
+library. These tags *must* point to a struct of type
+`grpc_completion_queue_functor` which has two user-provided members:
+
+* `void (*functor_run)(struct grpc_completion_queue_functor*, int)` : a pointer to a
```
chwarr, line 67: Did you consider having a `void*` argument/field in `grpc_completion_queue_functor` that can be used to pass arbitrary user state? It looks like the functor gets passed back to itself, so that can be used to track state, but that can be a bit awkward to work with.

vjpai: Core is still basically an implementation detail to support the wrapped languages (including C++). So I would expact that the standard way to do this is to actually do what we do in the C++ API implementation:

* the core surface user defines a class that derives from `grpc_completion_queue_functor`
   - `class MyTag : public grpc_completion_queue_functor {`
* the `functor_run` is just a pointer to a static class method
   -  `MyTag() { functor_run = &MyTag::StaticRun; inlineable = 0; ... }`
   - `static void StaticRun(grpc_completion_queue_functor* tag, int ok) {`
* that static method casts the argument to the actual class and invokes the "real" run function on it
   -    `static_cast<MyTag*>(tag)->ActualRun(ok);`
* the state that you're talking about lives in the class which is derived from the core class rather than being pointed at by the core class.

If you actually wanted to do this directly in C, I guess you would follow the same concept but manually extend the struct by making a bigger struct whose first member is the core surface struct and whose remaining members hold the needed state.

L68-core-callback-api.md
vjpai: I should remove `in this` and then I think it makes sense.

L68-core-callback-api.md
lidizheng: Is it the other way around that callback completion queue depends on EventManager? If an application enables event manager, it can use callback completion queue or other completion queues. But if the application disables event manager, it can only use next/pluck completion queues.

Python used the callback completion queue directly. The exposed API in this doc is sufficient, with one of the asynchronous mechanism to manage I/O (asyncio, event manager, etc.).

L68-core-callback-api.md
chwarr: Yeah, I was asking primarily from a C perspective. I find `void*` easier to work with in these cases. Looking at something like the C# support library, it's a [C library that wraps Core](https://github.com/grpc/grpc/blob/d075d49a1b45bd15dcb88588286d18c262be5735/src/csharp/ext/grpc_csharp_ext.c) providing slightly higher level abstractions over Core that are easy to invoke from C#.

(C is the de facto universal FFI and all that jazz.)

L68-core-callback-api.md
vjpai: TIL that Python uses the callback CQ. Cool.

vjpai: @markdroth - `grpc_completion_queue_functor` is core surface API and will be used by wrapped languages as well. It is a C89 struct.

L68-core-callback-api.md
vjpai: Right, even if someone wants to use this in C, I would strongly recommend that their library always have:

```
struct MyFunctor {
  grpc_completion_queue_functor functor;
  int my_other_field1;
  char* my_other_field2;
  ...
};

MyFunctor functor = {.functor = {.functor_run = &RunFunction, .inlineable = false}, .my_other_field1 = 18, .my_other_field2 = "hello world"};

grpc_call_start_batch(..... , &functor);

void RunFunction(grpc_completion_queue_functor *functor_arg, int* ok) {
  MyFunctor* functor = (MyFunctor*)functor_arg;
  ...
}

```

We're aiming to provide basic mechanism in core without dictating policy.

L68-core-callback-api.md
vjpai: Thanks for the suggestion; I did it.

vjpai:
Thanks for the early review comments! I've marked this ready for review and will be moving on to #180 now....

L68-core-callback-api.md
```diff
@@ -73,16 +76,31 @@ normally (corresponding to the `ok` result of
 `grpc_completion_queue_next` on the `GRPC_CQ_NEXT` queue).
 * `int inlineable`: specifies whether this functor can be run inline
 in the same thread as the one that detected it rather than sending
-it to an executor thread. This should only be
-used very carefully for trivial internally-generated callbacks that
-do not use locks.
+it to an executor thread. This must be
+used very carefully and only for trivial internally-generated callbacks that
```
chwarr, line 80: What does "internally-generated" mean here?

L68-core-callback-api.md
chwarr: Yeah, that's the awkwardness I was talking about. I guess it just like this bikeshed colored `void*`. 😉

vjpai: Yeah, I appreciate your comments, but I think we're going to keep it without a user-specified `void*` in the mix. We have the Python AIO binding using this struct via composition (https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_cython/_cygrpc/aio/callback_common.pxd.pxi#L24) and the C++ API implementation using this struct via inheritance (e.g., https://github.com/grpc/grpc/blob/master/include/grpcpp/impl/codegen/callback_common.h#L136-L137) so I think I'd prefer to leave it alone for now. That said, I suppose I could provide a usage example in this gRFC.

L68-core-callback-api.md
vjpai: Not end-user code, but rather code that is owned by the language-specific library.

chwarr: Maybe "known safe" or "library code" instead of "internally-generated", so that this doesn't imply that only callbacks from a code generator can be used here?

L68-core-callback-api.md
vjpai: Added the usage example.

vjpai:
Thanks for all the reviews and comments! I didn't start a proper discussion thread for this gRFC until today, so I'll hold off on merging until June 23 at the earliest.

vjpai:
Merging this as a complete gRFC although its implementation will not be completed for some time because it is blocked on the implementation of #182 .

PR #182, opened 2020-06-01
L69: EventManger API

muxi:
Discussion thread: https://groups.google.com/forum/#!topic/grpc-io/7EDteJbI2nc

### DIFF
--- a/L69-cpp-core-event-manager.md
+++ b/L69-cpp-core-event-manager.md
@@ -186,7 +186,7 @@ class EpollEventManagerInterface: public BaseEventManagerInterface {
 };
 ```
 
-Note that there will be no native implementation provided by gRPC for the epoll based interface in the first step. Instead, gRPC will provide a default event manager implementation based on libuv. The libuv EM implementation will be applied to all the gRPC languages based on gRPC core, and the users may choose to override it with their custom implementation of the EM interface. gRPC may release more native event manager implementations down the road map.
+Note that in the first step gRPC will provide a default EventManager implementation based on libuv. The libuv EM implementation will be used by all the gRPC languages based on gRPC core, and the users may choose to override it with their custom implementation of the EM interface. gRPC may release more native event manager implementations down the road map.
 ```cpp
 // Represents a file descriptor when using libuv.
 class LibuvDescriptorInterface {
--- a/L69-cpp-core-event-manager.md
+++ b/L69-cpp-core-event-manager.md
@@ -48,10 +48,10 @@ class BaseEventManagerInterface {
   // closure with true. The deadline here is an absolute time. The AlarmKey pointer 
   // is an output parameter which can be used to cancel the alarm before it fires. 
   // Note that the closure should not be blocking or long-running, as it would 
-  // block the worker thread in the EventManager from polling the file 
-  // descriptors. In case that the alarm is cancelled successfully in 
-  // ConsistentCancel, the closure will be run with false in the calling thread of 
-  // ConsistentCancel. Callers of ConsistentCancel need to be cautious to avoid 
+  // block the worker thread in the EventManager from polling the file
+  // descriptors. In case that the alarm is cancelled successfully in
+  // CancelAndRun, the closure will be run with false in the calling thread of
+  // CancelAndRun. Callers of CancelAndRun need to be cautious to avoid
   // deadlock.
   virtual void ScheduleAlarmAt(gpr_timespec deadline, std::function<void(bool)> c, Priority priority, AlarmKey* key = nullptr) = 0;
 
@@ -88,11 +88,11 @@ class BaseEventManagerInterface {
   // ScheduleAlarmAt(). This cancel operation is "consistent" in that it will
   // still run the alarm closure with false as parameter when successfully
   // cancelled the alarm.
-  virtual void ConsistentCancel(const AlarmKey& key) = 0;
+  virtual void CancelAndRun(const AlarmKey& key) = 0;
 
-  // Cancels an alarm as ConsistentCancel(), but wait if the task is currently 
+  // Cancels an alarm as CancelAndRun(), but wait if the task is currently 
   // running.
-  virtual void BlockingConsistentCancel(const AlarmKey& key) = 0;
+  virtual void BlockingCancelAndRun(const AlarmKey& key) = 0;
 
   // These two methods adjust a refcount of the EventManager, preventing it from 
   // shutdown if the refcount is positive. ShutdownRef() should be called before 
--- a/L69-cpp-core-event-manager.md
+++ b/L69-cpp-core-event-manager.md
@@ -1,0 +1,281 @@
+EventManager API
+----
+* Author(s): @muxi
+* Approver: @markdroth
+* Status: Draft
+* Implemented in: gRPC Core
+* Last updated: June 1, 2020
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+This proposal describes a plan to upgrade gRPC's event engine to a newer set of interfaces called EventManager API.The gRPC team is making a fundamental change in the gRPC core's polling model. Instead of borrowing threads from the application and giving the application control over which threads poll which fds, gRPC will migrate to a new model based on EventManager, which provides dedicated polling threads with an internal thread pool. User callbacks and timers will also be handled by the EventManager. The EventManager interface will be a public API, which allows users to integrate gRPC into the event loops of their applications if necessary.
+Background
+The gRPC has been using a custom event handling framework called iomgr API. The original iomgr API had only two implementations, one for posix (Linux and MacOS) and one for Windows. As part of an effort to improve performance, we changed the posix iomgr implementation to support pluggable polling engines, selectable at run-time.  We then wrote [a whole bunch of different polling engines](https://github.com/grpc/grpc/blob/master/doc/core/grpc-polling-engines.md) to experiment on and determine which one(s) worked best.  Some of them didn't work out and have been deleted, but we still support three of them (see below).
+
+The EventManager interface aims at simplifying the usage and maintenance of gRPC event framework. It allows users of gRPC to write their own event management implementation to be used by gRPC for specific needs. The project also paves the way for supporting features such as the C++ callback API in gRPC (See proposal [gRFC L67: C++ callback-based asynchronous API](https://github.com/grpc/proposal/pull/180)).
+
+## Proposal
+The following event manager interfaces will be available in the gRPC core.
+
+```cpp
+// An external handle to an EventManager task, which can be used to cancel the 
+// task. The EventManager task is added via RunFunction.
+struct TaskKey;
+
+// An external handle to an EventManager alarm, which can be used to cancel the
+// alarm. The EventManager alarm is added via AddAlarmIn.
+struct AlarmKey;
+
+// Scheduling priority for closures and alarms added into the EventManager.
+enum Priority {
+  PRIORITY_MIN = 0,
+  PRIORITY_LOW = 0,
+  PRIORITY_MED = 1,
+  PRIORITY_HIGH = 2,
+  PRIORITY_IMMEDIATE = 3,
+  PRIORITY_SYSTEM = 4,
+  PRIORITY_MAX = 4,
+  NUM_PRIORITIES = 5
+};
+
+// A top-level base class of EventManagerInterface.
+class BaseEventManagerInterface {
+ public:
+  virtual ~BaseEventManagerInterface();
+
+  // Adds a new one-shot alarm, which will be scheduled at the deadline to run the 
+  // closure with true. The deadline here is an absolute time. The AlarmKey pointer 
+  // is an output parameter which can be used to cancel the alarm before it fires. 
+  // Note that the closure should not be blocking or long-running, as it would 
+  // block the worker thread in the EventManager from polling the file 
+  // descriptors. In case that the alarm is cancelled successfully in 
+  // ConsistentCancel, the closure will be run with false in the calling thread of 
+  // ConsistentCancel. Callers of ConsistentCancel need to be cautious to avoid 
+  // deadlock.
+  virtual void ScheduleAlarmIn(gpr_timespec deadline, std::function<void(bool)> c, Priority priority, AlarmKey* key = nullptr) = 0;
+
+  // Adds a closure to the EventManager, which will be run at the earliest possible 
+  // time. The TaskKey pointer is an output parameter which can be used to cancel 
+  // the closure at a later time (currently not supported in 
+  // BaseEventManagerInterface). Note that the closure should not be blocking or 
+  // long-running, as it would block the worker thread in the EventManager from 
+  // polling the file descriptors.
+  virtual void RunFunction(std::function<void()> handler, Priority priority, TaskKey* key = nullptr) = 0;
+
+  struct Task {
+    std::function<void()> handler;
+    Priority priority;
+    TaskKey* key;
+    int internal_added_queue;
+  };
+  // Adds multiple closures to the event management system all at once.
+  virtual void RunFunctions(Task* tasks, int length) = 0;
+
+  // Cancels an alarm which has been added to the EventManager via 
+  // ScheduleAlarmIn(). This function will run the closure with false in the 
+  // calling thread if it successfully cancels the alarm.
+  virtual void ConsistentCancel(const AlarmKey& key) = 0;
+
+  // Cancels an alarm as ConsistentCancel(), but wait if the task is currently 
+  // running.
+  virtual void BlockingConsistentCancel(const AlarmKey& key) = 0;
+
+  // These two methods adjust a refcount of the EventManager, preventing it from 
+  // shutdown if the refcount is positive. ShutdownRef() should be called before 
+  // adding the first closure to the EventManager, and ShutdownUnref() can be 
+  // called after removing the last closure from the EventManager. Also, if you 
+  // need to keep the EventManager serving before receiving an RPC response, you 
+  // need to call ShutdownRef() before sending the RPC, and invoke ShutdownUnref() 
+  // when the response arrives.
+  virtual void ShutdownRef() = 0;
+  virtual void ShutdownUnref() = 0;
+};
+```
+
+The BaseEventManagerInterface is platform agnostic in order to accommodate the different polling models on different platforms (e.g. POSIX vs Windows).
+
+A specific interface EpollEventManagerInterface is derived from BaseEventManagerInterface for all the epoll-based polling models.
+
+```cpp
+// Represents a file descriptor on epoll-capable platforms.
+class EpollDescriptorInterface {
+ public:
+  virtual ~EpollDescriptorInterface();
+
+  // Returns the underlying file descriptor.
+  virtual int fd() const = 0;
+
+  // Returns the EventManager object where this Descriptor is registered.
+  virtual EpollEventManagerInterface* event_manager() const = 0;
+
+  // Executes all the registered notification closures, and closes the underlying 
+  // file descriptor.
+  virtual void Close() = 0;
+
+  // Returns whether the underlying file descriptor is closed or not.
+  virtual bool IsClosed() const = 0;
+
+  // Dissociates the underlying file descriptor from the Descriptor object, and 
+  // stop monitoring the file descriptor. Executes all the notification closures 
+  // registered for this file descriptor, and returns the file descriptor.
+  virtual int ReleaseFD() = 0;
+
+  // Sets various priorities for this file descriptor. The priorities are ignored 
+  // if the Descriptor does not support Priority.
+  virtual void SetReadPriority(Priority priority) = 0;
+  virtual void SetWritePriority(Priority priority) = 0;
+  virtual void SetErrorPriority(Priority priority) = 0;
+
+  // Registers a closure to be run immediately in a worker thread of the 
+  // EventManager, when the file descriptor becomes readable/writable/has error.
+  virtual void NotifyWhenReadable(std::function<void()> upcall) = 0;
+  virtual void NotifyWhenWritable(std::function<void()> upcall) = 0;
+  virtual void NotifyWhenHasError(std::function<void()> upcall) = 0;
+
+  enum NotificationOutcome {
+    NOTIFY_FD_BECAME_READY,
+    NOTIFY_TIMED_OUT
+  };
+
+  // Similar to NotifyWhen*, but with a deadline. If the file descriptor state 
+  // changes before the deadline, then upcall will be called with 
+  // NOTIFY_FD_BECAME_READY. If the deadline expires before the descriptor state 
+  // changes, then upcall will be called with NOTIFY_TIMED_OUT.
+  virtual void NotifyWhenReadableOrTimeout(std::function<void(NotificationOutcome)> upcall, gpr_timespec deadline) = 0;
+  virtual void NotifyWhenWritableOrTimeout(std::function<void(NotificationOutcome)> upcall, gpr_timespec deadline) = 0;
+
+  // Clears the descriptor's status. This allows the NotifyWhen* calls to 
+  // distinguish between old and new events.
+  virtual void ClearReadable() = 0;
+  virtual void ClearWritable() = 0;
+  virtual void ClearHasError() = 0;
+
+  // Forces the descriptor’s status to change, and runs the closures if registered. 
+  // This is especially useful for tearing down closures associated with this 
+  // descriptor by forcing them to run without waiting for the event on the 
+  // underlying socket.
+  virtual void ForceReadable() = 0;
+  virtual void ForceWritable() = 0;
+  virtual void ForceHasError() = 0;
+
+  // Returns a human-readable description of the file descriptor state.
+  virtual string ToString() const = 0;
+};
+
+class EpollEventManagerInterface: public BaseEventManagerInterface {
+ public:
+  virtual ~EpollEventManagerInterface();
+
+  virtual EpollDescriptorInterface* RegisterFileDescriptor(int fd) = 0;
+
+  virtual void DeleteDescriptor(EpollDescriptorInterface* d) = 0;
+
+  // Adds a new one-shot alarm with a slack, which will be scheduled at the given 
+  // deadline plus the timespan specified in slack. Reducing the precision of 
+  // alarms through a timer slack allows the EventManager to perform 
+  // timer-coalescing, which minimizes thread wake-up and has the potential of 
+  // substantially reducing CPU utilization.
+  virtual void ScheduleAlarmInWithSlack(gpr_timespec deadline, std::function<void(bool)> c, Priority priority, TaskKey* key, gpr_timespec slack) = 0;
+};
+```
+
+Note that there will be no native implementation provided by gRPC for the epoll based interface in the first step. Instead, gRPC will provide a default event manager implementation based on libuv. The libuv EM implementation will be applied to all the gRPC languages based on gRPC core, and the users may choose to override it with their custom implementation of the EM interface. gRPC may release more native event manager implementations down the road map.
+```cpp
+// Represents a file descriptor when using libuv.
+class LibuvDescriptorInterface {
+ public:
+  virtual ~LibuvDescriptorInterface();
+
+  // Returns the underlying TCP or UDP handle. Valid until Close() is called
+  // (i.e. not thread-safe with Close()).
+  virtual uv_handle_t fd() const = 0;
+
+  // Retruns the EventManager object where this descriptor is registered.
+  virtual LibuvEventManagerInterface* event_manager() const = 0;
+
+  // Executes all the registered notification closures, and closes the
+  // underlying TCP or UDP handle.
+  //
+  // Note that we cannot release the underlying TCP or UDP handle, since it is
+  // initialized with the libuv event loop of LibuvEventManager, and it is not
+  // thread-safe to access the handle externally.
+  virtual void Close() = 0;
+  virtual bool IsClosed() const = 0;
+
+  // Registers a closure \a upcall to be run immediately in a worker thread of
+  // the LibuvEventManager, after libuv reads from the socket. \a alloc_cb will
+  // be passed to libuv to allocate the read buffer. Note that this will start
+  // a read asynchronously.
+  //
+  // It is an error to call NotifyAfterRead() again before the previous closure
+  // completes.
+  //
+  // If there is more to read on the socket, re-arms a closure after upcall
+  // finishes.
+  virtuar void NotifyAfterRead(std::function<void(LibuvReadStatus)> upcall,
+                       uv_alloc_cb alloc_cb) = 0;
+
+  // Registers a closure \a upcall to be run immediately in a worker thread of
+  // the LibuvEventManager, after libuv writes to the socket. \a info contains
+  // the write message, and target address in cases of UDP. Note that this will
+  // start a write asynchronously.
+  //
+  // It is an error to call NotifyAfterWrite() again before the previous closure
+  // completes.
+  //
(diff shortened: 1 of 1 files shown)


L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,281 @@
...
+* Author(s): @muxi
+* Approver: @markdroth
+* Status: Draft
+* Implemented in: gRPC Core
+* Last updated: June 1, 2020
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+This proposal describes a plan to upgrade gRPC's event engine to a newer set of interfaces called EventManager API.The gRPC team is making a fundamental change in the gRPC core's polling model. Instead of borrowing threads from the application and giving the application control over which threads poll which fds, gRPC will migrate to a new model based on EventManager, which provides dedicated polling threads with an internal thread pool. User callbacks and timers will also be handled by the EventManager. The EventManager interface will be a public API, which allows users to integrate gRPC into the event loops of their applications if necessary.
+Background
```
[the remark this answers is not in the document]
muxi: Nice catch Mark. Fixed

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  // ConsistentCancel. Callers of ConsistentCancel need to be cautious to avoid 
+  // deadlock.
+  virtual void ScheduleAlarmIn(gpr_timespec deadline, std::function<void(bool)> c, Priority priority, AlarmKey* key = nullptr) = 0;
+
+  // Adds a closure to the EventManager, which will be run at the earliest possible 
+  // time. The TaskKey pointer is an output parameter which can be used to cancel 
+  // the closure at a later time (currently not supported in 
+  // BaseEventManagerInterface). Note that the closure should not be blocking or 
+  // long-running, as it would block the worker thread in the EventManager from 
+  // polling the file descriptors.
```
chwarr, line 63: Minor: "file descriptor" in this comment seems out of place. What if the thing being polled isn't a file descriptor?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  virtual void RunFunction(std::function<void()> handler, Priority priority, TaskKey* key = nullptr) = 0;
+
+  struct Task {
+    std::function<void()> handler;
+    Priority priority;
+    TaskKey* key;
+    int internal_added_queue;
+  };
+  // Adds multiple closures to the event management system all at once.
+  virtual void RunFunctions(Task* tasks, int length) = 0;
```
chwarr, line 73: `size_t` for length?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+
+  // Cancels an alarm which has been added to the EventManager via 
+  // ScheduleAlarmIn(). This function will run the closure with false in the 
+  // calling thread if it successfully cancels the alarm.
+  virtual void ConsistentCancel(const AlarmKey& key) = 0;
```
chwarr, line 78: Can you explain the use of the word "Consistent" in this member function name?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  virtual void BlockingConsistentCancel(const AlarmKey& key) = 0;
+
+  // These two methods adjust a refcount of the EventManager, preventing it from 
+  // shutdown if the refcount is positive. ShutdownRef() should be called before 
+  // adding the first closure to the EventManager, and ShutdownUnref() can be 
+  // called after removing the last closure from the EventManager. Also, if you 
+  // need to keep the EventManager serving before receiving an RPC response, you 
+  // need to call ShutdownRef() before sending the RPC, and invoke ShutdownUnref() 
+  // when the response arrives.
+  virtual void ShutdownRef() = 0;
```
chwarr, line 91: Has there been any consideration of modeling this lifetime behavior with RAII types? It _is_ a C++ interface, after all. 😉 E.g., some mixture of `std::shared_ptr` and factory functions for creation?

L69-cpp-core-event-manager.md
chwarr, line 73: Was a begin/end iterator pair considered?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  // BaseEventManagerInterface). Note that the closure should not be blocking or 
+  // long-running, as it would block the worker thread in the EventManager from 
+  // polling the file descriptors.
+  virtual void RunFunction(std::function<void()> handler, Priority priority, TaskKey* key = nullptr) = 0;
+
+  struct Task {
+    std::function<void()> handler;
+    Priority priority;
+    TaskKey* key;
+    int internal_added_queue;
```
chwarr, line 70: What is the `internal_added_queue` member for?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  virtual EpollDescriptorInterface* RegisterFileDescriptor(int fd) = 0;
+
+  virtual void DeleteDescriptor(EpollDescriptorInterface* d) = 0;
+
+  // Adds a new one-shot alarm with a slack, which will be scheduled at the given 
+  // deadline plus the timespan specified in slack. Reducing the precision of 
+  // alarms through a timer slack allows the EventManager to perform 
+  // timer-coalescing, which minimizes thread wake-up and has the potential of 
+  // substantially reducing CPU utilization.
+  virtual void ScheduleAlarmInWithSlack(gpr_timespec deadline, std::function<void(bool)> c, Priority priority, TaskKey* key, gpr_timespec slack) = 0;
```
chwarr, line 179: `ScheduleAlarmInWithSlack` doesn't look like it's specific to epoll. Many other platforms support timer coalescing.

Should it belong in the base interface? Should it belong in its own interface that can be queried for from the base interface?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  // start a write asynchronously.
+  //
+  // It is an error to call NotifyAfterWrite() again before the previous closure
+  // completes.
+  //
+  // If there is more to write, re-arms a closure after upcall finishes.
+  virtual void NotifyAfterWrite(std::function<void(int)> upcall, LibuvWriteInfo info) = 0;
+
+  // Accessor and mutator of arbitrary user-defined data. NOT thread-safe.
+  virtual void* data() = 0;
```
chwarr, line 231: What are these member functions enabling? Could, instead, any user-defined data live in the concrete implementation of the interface?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  // Retruns the EventManager object where this descriptor is registered.
+  virtual LibuvEventManagerInterface* event_manager() const = 0;
+
+  // Executes all the registered notification closures, and closes the
+  // underlying TCP or UDP handle.
+  //
+  // Note that we cannot release the underlying TCP or UDP handle, since it is
+  // initialized with the libuv event loop of LibuvEventManager, and it is not
+  // thread-safe to access the handle externally.
+  virtual void Close() = 0;
```
chwarr, line 203: Can either of these member functions be const?

L69-cpp-core-event-manager.md
chwarr, line 231: Can `data()` be const?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+
+  // Returns the EventManager object where this Descriptor is registered.
+  virtual EpollEventManagerInterface* event_manager() const = 0;
+
+  // Executes all the registered notification closures, and closes the underlying 
+  // file descriptor.
+  virtual void Close() = 0;
+
+  // Returns whether the underlying file descriptor is closed or not.
+  virtual bool IsClosed() const = 0;
```
chwarr, line 117: const member function?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+
+The EventManager interface aims at simplifying the usage and maintenance of gRPC event framework. It allows users of gRPC to write their own event management implementation to be used by gRPC for specific needs. The project also paves the way for supporting features such as the C++ callback API in gRPC (See proposal [gRFC L67: C++ callback-based asynchronous API](https://github.com/grpc/proposal/pull/180)).
+
+## Proposal
+The following event manager interfaces will be available in the gRPC core.
+
+```cpp
+// An external handle to an EventManager task, which can be used to cancel the 
+// task. The EventManager task is added via RunFunction.
+struct TaskKey;
```
chwarr, line 24: I'm a little unclear on how `TaskKey` works. It's an empty struct, but it's also an output parameter of `RunFunction`. Does that mean that `RunFunction` has recorded the address of the `TaskKey` that I passed in? If that's the case, then should `TaskKey` be non-copyable and non-movable? (Perhaps also non-assignable).

I have the same question for `AlarmKey`.

Was there any consideration of having `RunFunction` and company return `TaskKey` instances as the handles?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+User implementations of EventManager can be registered with gRPC Core by the following API:
+```cpp
+// Register a user-provided EventManager, which will be used globally in
+// gRPC. gRPC won't take the ownership of this EventManager; the
+//  EventManager must outlive gRPC.
+//
+// The function must be called before gRPC is initialized (i.e. grpc_init is first called).
+// It can only be called at most once; otherwise an exception will be thrown. If the function 
+// is not called at all, gRPC will use its internal default EventManager implementation.
+void grpc_register_global_event_manager(BaseEventManagerInterface *em);
```
lidizheng, line 254: As the definition of `LibuvEventManagerInterface` and `EpollDescriptorInterface` showed, a working event manager requires not only implement the task managing logic, but also I/O handling. They are tightly coupled with the [grpc_endpoint](https://github.com/grpc/grpc/blob/master/src/core/lib/iomgr/endpoint.h) logic. So, there might be several solutions:

1. Let user provide grpc_endpoint implementation (vtable);
2. Alter the definition of event manager to include the logic currently in grpc_endpoint;
3. Instead of allow user-provided event manager, we say only the two implementation provided by us is supported.

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+Note that there will be no native implementation provided by gRPC for the epoll based interface in the first step. Instead, gRPC will provide a default event manager implementation based on libuv. The libuv EM implementation will be applied to all the gRPC languages based on gRPC core, and the users may choose to override it with their custom implementation of the EM interface. gRPC may release more native event manager implementations down the road map.
+```cpp
+// Represents a file descriptor when using libuv.
+class LibuvDescriptorInterface {
+ public:
+  virtual ~LibuvDescriptorInterface();
+
+  // Returns the underlying TCP or UDP handle. Valid until Close() is called
+  // (i.e. not thread-safe with Close()).
+  virtual uv_handle_t fd() const = 0;
```
lidizheng, line 192: File descriptors are used directly in Core's code. E.g., [local credentials](https://github.com/grpc/grpc/blob/master/src/core/lib/security/security_connector/local/local_security_connector.cc#L73). If we return anything else than system fd, some features might be impacted.

muxi:
@chwarr @lidizheng  Thanks for the comments. I'm in something else right now and will get back to update the gRFC next week.

L69-cpp-core-event-manager.md
muxi: I am not familiar with local credentials, but according to @guantaol local credentials need to move away from directly using fd's. @markdroth this might be a blocking issue for EM?

L69-cpp-core-event-manager.md
muxi: We'll require one implementation for each eventmanager implementation.

muxi: @markdroth - does that not mean we'll have to expose grpc_endpoint interface too?

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  // descriptors. In case that the alarm is cancelled successfully in 
+  // ConsistentCancel, the closure will be run with false in the calling thread of 
+  // ConsistentCancel. Callers of ConsistentCancel need to be cautious to avoid 
+  // deadlock.
+  virtual void ScheduleAlarmIn(gpr_timespec deadline, std::function<void(bool)> c, Priority priority, AlarmKey* key = nullptr) = 0;
+
+  // Adds a closure to the EventManager, which will be run at the earliest possible 
+  // time. The TaskKey pointer is an output parameter which can be used to cancel 
+  // the closure at a later time (currently not supported in 
+  // BaseEventManagerInterface). Note that the closure should not be blocking or
```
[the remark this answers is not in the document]
muxi: According to @guantaol the interface of cancelling a closure is not implemented because gRPC does not need it.

L69-cpp-core-event-manager.md
muxi: Each implementation will define their own TaskKey and AlarmKey struct as they can be different. The caller of `RunFunction` provides the struct and `RunFunction` fills TaskKey contents in. I did not really understand the "record the address" part of your question.

L69-cpp-core-event-manager.md
muxi: Not sure if I missed anything. It's already const, is it not?

L69-cpp-core-event-manager.md
muxi: I think so. Thanks for pointing out.

L69-cpp-core-event-manager.md
muxi: Can you elaborate? Close() cannot be const because it *closes* the thing, right?

L69-cpp-core-event-manager.md
muxi: It's for program specific contexts associated with the file descriptors.

L69-cpp-core-event-manager.md
muxi: It's fine to be in the base interface. Thanks for the suggestion.

L69-cpp-core-event-manager.md
muxi: I've added some comment to clarify.

L69-cpp-core-event-manager.md
muxi: It's possible but we don't see much benefit in doing so. This interface should work and the conversion with iterators is kind of trivial.

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+};
+```
+
+The BaseEventManagerInterface is platform agnostic in order to accommodate the different polling models on different platforms (e.g. POSIX vs Windows).
+
+A specific interface EpollEventManagerInterface is derived from BaseEventManagerInterface for all the epoll-based polling models.
+
+```cpp
+// Represents a file descriptor on epoll-capable platforms.
+class EpollDescriptorInterface {
```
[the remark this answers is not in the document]
muxi: I don't think an IO completion queue interface is within the scope of this design. Completion queues may need to use the eventmanager API.

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+  // Adds a new one-shot alarm, which will be scheduled at the deadline to run the 
+  // closure with true. The deadline here is an absolute time. The AlarmKey pointer 
+  // is an output parameter which can be used to cancel the alarm before it fires. 
+  // Note that the closure should not be blocking or long-running, as it would 
+  // block the worker thread in the EventManager from polling the file 
+  // descriptors. In case that the alarm is cancelled successfully in 
+  // ConsistentCancel, the closure will be run with false in the calling thread of 
+  // ConsistentCancel. Callers of ConsistentCancel need to be cautious to avoid 
+  // deadlock.
+  virtual void ScheduleAlarmIn(gpr_timespec deadline, std::function<void(bool)> c, Priority priority, AlarmKey* key = nullptr) = 0;
```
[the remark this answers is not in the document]
muxi: The interface is designed to be compatible with a Google EventManager interface, so a lot of the names were copied. But for this particular function, there's a slight change in the definition of the parameters. I think in this case `ScheduleAlarmAt` makes a bit more sense. Will change the text.

L69-cpp-core-event-manager.md
muxi: It is to differentiate from Google's implementation of `Cancel` in EM where canceling an alarm will not trigger the closure to run (in other words, `ConsistentCancel` will consistently invoke the alarm closure despite that the alarm is cancelled). I'll update the comments.

L69-cpp-core-event-manager.md
muxi: I don't think we'll do that for compatibility with the Google EM interface.

L69-cpp-core-event-manager.md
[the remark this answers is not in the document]
chwarr: If an IOCP is out of scope for this design, why is an epoll one in scope? Aren't both just ways to implement the event manager API? 😉

L69-cpp-core-event-manager.md
chwarr: Yeah. Not sure what I meant myself.

L69-cpp-core-event-manager.md
chwarr: Hmmm... Not sure anymore myself. What I see now wrt. `const` looks fine here.

L69-cpp-core-event-manager.md
chwarr: Gotcha. Should this be a `void*` so that the implementation can use whatever "handle" type is most natural for it? E.g., for an IOCP based event manager, that could be a `HANDLE`, which won't fit in an `int`.

L69-cpp-core-event-manager.md
```diff
@@ -0,0 +1,282 @@
...
+// An external handle to an EventManager task, which can be used to cancel the 
+// task. The EventManager task is added via RunFunction.
+struct TaskKey;
+
+// An external handle to an EventManager alarm, which can be used to cancel the
+// alarm. The EventManager alarm is added via AddAlarmIn.
+struct AlarmKey;
+
+// Scheduling priority for closures and alarms added into the EventManager.
+enum Priority {
```
chwarr, line 31: Can this be an enum class?

L69-cpp-core-event-manager.md
chwarr: Are `TaskKey` and `AlarmKey` base types for opaque handles? If so, should the parameter to `RunFunction` and `ScheduleAlarmAt` be `TaskKey**` and `AlarmKey**` respectively? Cancellation would need to be changed to take `TaskKey*` and `AlarmKey*` parameters too.

Assuming that the empty struct is intentional, what would help me is an example showing how cancellation will work. Right now, I don't see any way to cancel a function.

```lang-cpp
std::shared_ptr<BaseEventManagerInterface> em = GetEventManager();

TaskKey tk;
em->RunFunction(some_handler, PRIORITY_LOW, &tk);

// Now what?
// No member functions on TaskKey...
// I don't see a CancelTask member function...
```

Alarms looks more usable:

```lang-cpp
void SomeFun() {
    std::shared_ptr<BaseEventManagerInterface> em = GetEventManager();

    AlarmKey ak;
    em->ScheduleAlarmAt(in_the_future, some_handler, PRIORITY_LOW, &ak, some_slack);

    DoSomethingElse();

    em->ConsistentCancel(ak);
}
```

However, `AlarmKey` is an empty struct, so there's nowhere for the EM implementation to store any state inside of `AlarmKey`. The only thing the EM can use to identify the alarm is the address of the `AlarmKey` instance that was passed in to `ScheduleAlarmAt`. (Unless I'm missing something...) That _can_ work, but then the following will break in a subtle way:

```lang-cpp
void SomeFunc() {
    std::shared_ptr<BaseEventManagerInterface> em = GetEventManager();
    
    AlarmKey ak;
    em->ScheduleAlarmAt(in_the_future, some_handler, PRIORITY_LOW, &ak, some_slack);
    
    // Need to capture em and ak by value as we're returning from
    // SomeFunc() after scheduling the callback.
    // AlarmKey has a copy ctor, so I assume it's OK to copy it.
    ScheduleCallbackForLaterUsingSomeOtherCallbackSystem([ak, em] {

        // 💥 AlarmKey doesn't have value semantics... Making a copy wasn't OK...
        // The event manager was actually using the address of AlarmKey and this ak
        // has a different address.
        em->ConsistentCancel(ak);
    });
}
```

If the intention is to use the address of the TaskKey/AlaramKey as the identity for the event manager, then can copy, move, and assignment be disabled?

```lang-cpp
struct AlarmKey {
    AlarmKey(const AlarmKey &) = delete;
    AlarmKey& operator=(const AlarmKey &) = delete;

    // IIRC, these aren't needed but some people add them to signal that
    // non-movable was intentional.
    AlarmKey(AlarmKey &&) = delete;
    AlarmKey& operator=(AlarmKey &&) = delete;
};
```

That will force the caller to keep the address stable:

```lang-cpp
void SomeFunc() {
    std::shared_ptr<BaseEventManagerInterface> em = GetEventManager();
    
    auto ak = std::make_unique<AlarmKey>();
    em->ScheduleAlarmAt(in_the_future, some_handler, PRIORITY_LOW, ak.get(), some_slack);
    
    // This uses C++14 capture. For C++11, could .release() into a pointer, capture that,
    // then create a unique_ptr inside the body of the lambda. Or use a shared_ptr.
    ScheduleCallbackForLaterUsingSomeOtherCallbackSystem([ak = std::move(ak), em] {
        em->ConsistentCancel(*ak);
    });
}
```

L69-cpp-core-event-manager.md
chwarr: Comment updates look good. Thanks.

As a non-Google user who's never seen the Google EM implementation, the word "Consistent" in the name still comes off as a bit odd. For me, just the comments and "Cancel" would be enough. I encourage you to balance how much internal Google bits "leak", if you will, into this design.

(If this interface had both behaviors, that would be a different story, but it only has one behavior.)

L69-cpp-core-event-manager.md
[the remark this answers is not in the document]
chwarr: Is the plan to use libuv to implement this interface on Windows? If so, let's capture that. If not, I'd like to see at least a sketch of how this could be implemented on Windows. I assume such an implementation would build atop IO completions ports.

(There was an earlier comment here that I've since deleted. I hadn't fully digested the libuv parts yet.)

L69-cpp-core-event-manager.md
chwarr: Can you expand on what compatibility with the Google EM interface means? I'm assuming that there would be (is?) an adapter of some sort between the Google EM and this EM interface if for no other reason than to allow the two to change decoupled from each other. Such an adapter could, for example, turn RAII type construction/destruction into the needed Ref/Unref calls.

It's incredibly useful to have an implementation of an event manager to prove that the architecture works. Thank you for doing that: running code speaks volumes. It just looks like there's a lot of room here to have the type system help with ownership while also being compatible with the Google EM interface.

L69-cpp-core-event-manager.md
markdroth: > As the definition of `LibuvEventManagerInterface` and `EpollDescriptorInterface` showed, a working event manager requires not only implement the task managing logic, but also I/O handling. They are tightly coupled with the [grpc_endpoint](https://github.com/grpc/grpc/blob/master/src/core/lib/iomgr/endpoint.h) logic.

This is a good point.  I think we do need some way to register both the EventManager impl and the corresponding descriptor implementation.  We'll probably need to define some sort of common interface for descriptors to make this work.  But it seems like we'll also need some sort of factory interface to create new descriptors, and the factory is the thing that will probably need to be injected here along with the EM instance.

@guantaol, how are you making this work for the libuv EM?

> So, there might be several solutions:
> 
> 1. Let user provide grpc_endpoint implementation (vtable);

The issue here is descriptors, not endpoints.  Once we figure out how to have a base descriptor interface, we can figure out how to wire that into endpoints, but endpoints should not be exposed as part of this API.

Also, note that the current `grpc_endpoint` API is part of iomgr; ultimately, it will be removed along with the rest of the iomgr API.

> 2. Alter the definition of event manager to include the logic currently in grpc_endpoint;

We talked about this extensively, and it's not feasible due to the need to be compatible with the existing google3 EM API.

> 3. Instead of allow user-provided event manager, we say only the two implementation provided by us is supported.

Part of the goal of this API is to allow users to integrate gRPC into their own third-party event loops.  That won't work if users can't provide their own EM implementation.

L69-cpp-core-event-manager.md
markdroth: I think we need to do one of two things here:

1. Provide a common inteface for descriptors that exposes source and peer addresses, and expose that from the endpoint.

2. Change the endpoint API to stop exposing the underlying descriptor and instead expose methods to return the source and peer addresses.

My inclination would be to prefer option 1, since we are probably going to need to have a common descriptor interface anyway, as per discussion below.  And option 2 would still require doing that work anyway.

Note that, as mentioned below, the current `grpc_endpoint` API is part of iomgr and will eventually be going away.  Ideally, it should be replaced with a C++ API that is better designed to work in the EM world.  But it might be that the best transition path is to make changes to the existing API and then convert it to C++ later.  We can figure out those details later, though.

L69-cpp-core-event-manager.md
muxi: Agree that the name is odd and agree that we will try to minimize 'internal leak' in this design. However in this case a leak is inevitable and we can't simply name the method as `Cancel`. I can rename it to something more sensible like `CancelAndTrigger`. Suggestion welcomed.

chwarr: Maybe `CancelAndRun` for parallelism with `RunFunction`? ("Trigger" doesn't show up in the API names or comments yet. "Run" does.)

Is conflict trying to be avoided a member function name clash, a conceptual/semantic clash, or something else? I'm still a bit fuzzy here.

muxi: `CancelAndRun` is good.

The conflict is a semantics clash as Google's EM is going to inherit from this interface, so if we name it `Cancel` the semantics is different between the interface and Google's implementation.

L69-cpp-core-event-manager.md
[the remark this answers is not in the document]
muxi: Libuv will be implementing `LibuvEventManagerInterface` below. I'm not that familiar with libuv but I don't think libuv should implement this interface.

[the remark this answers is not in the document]
chwarr: "this interface" wasn't clear. I meant the `BaseEventManagerInterface`. Comment updated above.

It looks like the plan is to use libuv for all the platforms and add native implementations as needed. If that's the case, I suggest the following revisions to the paragraph before the libuv interface (additions in **bold**, removals <strike>struck through</strike>):

> Note that there will be no native EM implementation provided by gRPC for <strike>the epoll based interface</strike> **any platform** in the first step. Instead, gRPC will provide a default event manager implementation based on libuv. The libuv EM implementation will be <strike>applied to</strike> **used by** all the gRPC languages based on gRPC core, and the users may choose to override it with their custom implementation of the EM interface. gRPC may release more native event manager implementations down the road map.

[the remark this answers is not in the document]
muxi: Thanks for the paragraph. I don't really see why we should state that no native EM implementation is provided for any platform. The libuv EM implementation will be a native one provided by gRPC, and it will be provided for most of the platforms. Maybe I misunderstood your question.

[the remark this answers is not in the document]
muxi: Oh I see what you mean. When you say "native" you probably mean native for the target platform and I was more talking about "native" for gRPC.

I think I will rephrase it a little bit. Probably just say "Note that in the first step gRPC will provide a default EventManager implementation based on libuv. The libuv EM implementation will be...".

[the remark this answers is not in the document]
chwarr: That phrasing looks good to me.

L69-cpp-core-event-manager.md
muxi: The design is derived from a Google internal EM implementation, and the plan is to have Google EM as one of the implementations of the BaseEventManagerInterface. There are some semantics difference between the two so yes there will be some sort of adapting methods in the Google EM implementation, but as the team just discussed again, though it's a somewhat unideal situation, one of the design consideration is to minimize the differences (hence minimize the number of adapting methods needed) between the API in BaseEventManagerInterface and the existing GoogleEM implementation. So we will to keep the interface similar to the Google EM's interface except for some rare cases (e.g. to enable a big use cases that the current design is not capable of).

As mentioned in the doc, the libuv implementation will be provided and will be using this architecture. It's an work in progress by the team.

L69-cpp-core-event-manager.md
muxi: Based on the same compatibility argument above, we will probably not make it a `void*`.

chwarr: That's really quite unfortunate. Is there really no way to make this pointer sized?

On Windows, the handles/descriptors for the OS primitives that would likely be used in an EM implementation, like an IO Completion Port `HANDLE`, a `RIO_CQ`, a WinSock `SOCKET`, and a file's `HANDLE` are 64-bits wide when targeting amd64. If this is only an `int` (32-bits wide when targeting amd64), then such implementations will have to have a mapping table so they can expose an `int` key.

libdispatch/Grand Central Dispatch similarly has pointer-sized handles/descriptors, like `dispatch_queue_t` which is typedef for a pointer to an internal struct.

(I'm assuming that the Google EM uses this field to store a file descriptor that is associated with the task queue. If not, knowing how it's used could allay my concerns.)

L69-cpp-core-event-manager.md
guantaol: One thing I would like to call out here is that this function won't run the closure if the alarm has completed already (i.e. we fail to cancel the alarm).

On the other hand, the new name `CancelAndRun` looks good to me.

L69-cpp-core-event-manager.md
guantaol: For libuv EM, it must work together with its associate endpoint implementation. Then for grpc runtime, it only needs to read from/write to the endpoints, but do not need to operate on the fd/socket directly.

Note that not only we need to consider the descriptor implementation with the event loop, but also take the polling model into consideration (NotifyAfterRead on Windows vs. NotifyWhenReadable on Unix). Thus, EM implementation is not platform-agnostic and we need an unified endpoint interface & implementation to wrap all the platform-specific details.

markdroth: That's a good point.  So how are you doing this for the libuv EM implementation?

For the API, it sounds like we have more work to do here.  We need to define a unified API that wraps both the EM and the endpoint implementation, so that users can override both of those things.  And once we do that, we might as well make that unified API the public API and not bother exposing the EM API at all.  This would certainly address the problem of leaking details of our existing internal EM API into this OSS API.

Guantao, I think this essentially reopens the discussion about inheritance vs. composition.  This unified interface is essentially exactly the composition approach that I was originally arguing for.  If we're going to need this anyway, then we're already going to need it to wrap the EventManager, in which case we are not actually getting the efficiency win that was the main benefit of the inheritance approach.  Remind me, were there other benefits to the inheritence approach?

chainmasterx10:
@muxi What is the status of this proposal? Thanks!

drfloob:
This work has been superseded by https://github.com/grpc/proposal/pull/245. It's a similar effort with a bit of a different approach. I've reviewed the feedback here and used it to inform the new design. Please review and comment!

PR #183, opened 2020-06-01
L70: @grpc/proto-loader TypeScript Type Generator CLI Tool

murgatroid99:
Discussion thread: https://groups.google.com/g/grpc-io/c/FwLprMC-5Qc


L70-node-proto-loader-type-generator.md
```diff
@@ -0,0 +1,65 @@
...
+## Proposal
+
+In the `@grpc/proto-loader` library, provide a command line tool that will generate the types of the objects that will be created by passing the output of `load` or `loadSync` to `grpc.loadPackageDefinition` when loading a specific set of `.proto` files with a specific set of options. This will allow users to write code like this to use the type information in the rest of the code:
+
+```ts
+import * as grpc from 'grpc';
+// OR 
+import * as grpc from '@grpc/grpc-js';
+
+import ProtoGrpcType from 'proto-file-name.d.ts'; // File generated by the tool
```
alexander-fenster, line 31: I think this will cause `error TS2691: An import path cannot end with a '.d.ts' extension. Consider importing './proto-file-name' instead.`.

L70-node-proto-loader-type-generator.md
```diff
@@ -0,0 +1,67 @@
...
+ - `--longs=String|Number`: Specify the type that should be used to output 64 bit integer values
+ - `--enums=String`: Specify that enum values should be output as strings
+ - `--bytes=Array|String`: Specify the type that should be used to output `bytes` fields
+ - `--defaults`: Indicates that default values should be output for missing fields
+ - `--arrays`: Indicates that empty arrays should be output for missing repeated fields even if `--defaults` is unset
+ - `--objects`: Indicates that empty objects should be output for missing message fields even if `--defaults` is unset
+ - `--oneofs`: Indicates that virtual "oneof" fields will be set to the present field's name in the output
+ - `--includeDirs=<directories...>`, `-I`: A list of directories to search for included `.proto` files
+
+ - `--outDir`, `-O`: The path in which to output the `.d.ts` file
```
alexander-fenster, line 56: if the output is a single `.d.ts` file, would it be better to have a regular `-o` option for an output file name, dumping it to `stdout` by default if the path is not given?

alexander-fenster: if it's not a single `.d.ts` but multiple `.d.ts` (e.g. one per service), it's worth mentioning that too.

murgatroid99: I've clarified what files will be generated and in what directory structure.

DBosley:
Will this support the `commonjs_strict` option for the protoc JS generator? The exports when you set the generator to use strict mode are slightly different than when you use `commonjs`?

murgatroid99:
This has no relationship to protoc. This is a generator for `@grpc/proto-loader`, which is a separate library.

DBosley:
Ah! I found this via threads in related to ts-protoc-gen talking about how it doesn't support @grpc/grpc-js so I thought it was to support protoc generated code.

If this ts support is added, @grpc/proto-loader might be the solution my team needs. Thank you!

mjpitz:
I've gone down this route a couple of times. I've been contemplating writing a custom plugin to do some of it. Some of the rough edges I've hit is around the context generation part. When code generating protos for proto-loader, the provided context (i.e. -I) plays a semi-important role. I needed to overlay the typescript generation with some bash scripts to handle the overall context generation.

murgatroid99:
@mjpitz Can you elaborate on the context issues? Is that something that you think isn't well addressed in this proposal, or would benefit from more detail?

I have a mostly complete implementation of this proposal at grpc/grpc-node#1474, and it seems to be working pretty well for my usage, which has multiple separate include directories.

mjpitz:
@murgatroid99: Of course! So I think the proposal scopes this out by requiring clients to specify the options block. This might be mitigated by how clients end up using the tool to produce libraries. I know when I've done multi-language projects, managing contexts across each language can be tricky to do.

Based on cursory view of this doc, it's not clear to me why we export 3 different types (grpctype, servicehandlers, messages). I would expect this to generate something a little closer to the way other languages generate messages and structures. Last I went down this route, I was trying to base it on some of the manual ones I had [written by hand](https://github.com/depscloud/api/blob/main/packages/depscloud-api-nodejs/v1alpha/tracker/index.d.ts#L66-L80). This was nice for many reasons as it produced similar [call signatures to other languages](https://github.com/depscloud/extractor/blob/main/src/main.ts#L60).

I'm sure this proposal works but may end up being toilsome for library owners / consumers. For example, I need to unroll the packages twice (one for the messages, one for the handlers, and even once again for registration). From a consumer's perspective, I would expect this to be somewhat captured by the import rather than duplicated in the package. For example, this feels really dry:

```typescript
import { messages } from "./path/to/package/proto";
messages.path.to.package.proto.XXXX
```

vs

```typescript
import { MyMessage } from "./path/to/package/proto";
```

Again, I feel like a lot of this could be mitigated if owers wrap this in a client library, but it seems like extra work for something a lot of other languages get for free.

Does this make sense?

L70-node-proto-loader-type-generator.md
```diff
@@ -0,0 +1,237 @@
...
+
+```ts
+import * as grpc from 'grpc';
+// OR 
+import * as grpc from '@grpc/grpc-js';
+
+import { ProtoGrpcType, ServiceHandlers, messages } from 'path/to/proto-file-name_proto'; // File generated by the tool is path/to/proto-file-name_proto.d.ts
+import * as protoLoader from '@grpc/proto-loader';
+
+const packageDefinition = protoLoader.loadSync('other/path/to/proto-file-name.proto', options);
```
mjpitz, line 35: It's quite common for folks to want to load more than one proto file. It would be good to switch this over to an array instead of a simple string to show how multiple types may be loaded.

L70-node-proto-loader-type-generator.md
```diff
@@ -0,0 +1,237 @@
...
+    // Implementation
+  }
+}
+
+const server = new grpc.Server();
+server.addService(loadedPackageDefinition.fully.qualified.package.ServiceName.service, serviceHandler);
+
+```
+
+The tool will be named `proto-loader-gen-types` and will have this usage:
```
mjpitz, line 55: Will this work as a protoc plugin? Or will it stand on it's own.

murgatroid99:
First, I want to clarify that the runtime exports from the `@grpc/proto-loader` library already exist, and this proposal would not make any changes to them. This is only about the tool for generating the typescript type definitions to describe the output of those functions. So that API you point to that accepts the `options` object already exists and won't change.

The 4 exports in the example could probably be consolidated into 3, but fewer would be tricky because of name collisions. To be fair, the `messages` namespace essentially exports the same information as the individual message files, and the `ClientInterfaces` and `ServiceHandlers` namespaces could probably be split out into similar individual files. I think it would probably be cleaner to separate them into separate files.

But the `ProtoGrpcType` export is different, and needs to live in a root-level file. That's because it is not a namespace aggregating several type definitions, but rather a single type definition that describes the type returned by `grpc.loadPackageDefinition`.

As for the imports and namespaces, my proposal is a little different from what you describe there. With this proposal, you would be able to write

```ts
import { messages } from './generated/protoFilename';
messages.path.to.package.proto.XXXX
```

but you would actually also be able to write

```ts
import { MyMessage } from './generated/path/to/package/proto/MyMessage'
```

L70-node-proto-loader-type-generator.md
murgatroid99: This API does accept an array of strings as an alternative to a single string. When loading multiple files you can get the type of the output from `grpc.loadPackageDefinition` by loading all of the corresponding `ProtoGrpcType` interfaces from the corresponding generating files and combining them with `&`.

Also, maybe this isn't clear, but the proposed CLI tool can also accept multiple filenames to generate types for.

L70-node-proto-loader-type-generator.md
murgatroid99: This is a standalone tool, distributed in the `@grpc/proto-loader` package. It uses Protobuf.js rather than protoc for parsing.

murgatroid99:
I made some changes to the proposal in response to the recent comments:

 - The `ServiceNameClient` and `ServiceNameHandler` interfaces are now exported from `generated/fully/qualified/package/ServiceName.ts`, similar to the existing files generated for messages and enums.

 - The root files now only export the `ProtoGrpcType` interface. Everything else can be found in other files.

L70-node-proto-loader-type-generator.md
```diff
@@ -0,0 +1,216 @@
...
+ - **Generate files corresponding to the input files, all directly in the output directory.** This greatly increases the risk of filename conflicts.
+ - **Generate a single file with all of the types.** This creates a large, unwieldy file, making it hard for a human to evaluate changes. This can also greatly increase code duplication if multiple files need to be generated.
+
+The gRPC implementation is specified at build time because the types depend on types from that library and the user should know at that point which implementation they are using. An alternative is to use generics to insert the types in a different way, but that increases the complexity of both the generated code and the code that uses it for relatively little gain.
+
+Two types are generated for each message definition to represent that at runtime the library is permissive with what it accepts, but stricter with what it outputs. Having those stricter output types allows the user to get tigher type guarantees where possible.
```
wenbozhu, line 212: This sentence is  hard to read. Do you mean the following: 

"Two types are generated for each message definition. To represent the message at runtime, the library (tool?) is .... "

L70-node-proto-loader-type-generator.md
```diff
@@ -0,0 +1,216 @@
...
+
+ - `--outDir=<directory>`, `-O <directory>`: The directory in which to output files
+ - `--grpcLib=grpc|@grpc/grpc-js`: The gRPC implementation library that these types will be used with
+ - `--verbose`, `-v`: Enable various logging output
+ - `--includeComments`: Include comments from the `.proto` files in the generated files
+
+The output will be one file for each `message` and `enum` loaded, with file paths based on the package and type names, plus a master file per input file that combines all of those to produce the type that the user will load, as described above. `ProtoGrpcType` types from different files can be intersected to get the type that results from loading those files together at runtime. Messages will have an additional type generated, suffixed with `__Output`, that describes the type of objects that will be output by gRPC, i.e. response messages on the client, and request messages on the server. These "output" message types will be subtypes of the main message type, restricted based on the options that are set.
```
wenbozhu, line 81: Could you expand a bit on the need for stricter types, here or under the rationales section?  

Also, I wonder if "_Strict" will work better. Conceivably, I may want to use the strict types for both input and output messages, e.g. to improve the overall type conformity of my codebase.

L70-node-proto-loader-type-generator.md
murgatroid99: What I'm trying to say is that the library is permissive about what types it accepts as input, but we can be more specific about what types it outputs and the two different types reflect that.

L70-node-proto-loader-type-generator.md
murgatroid99: I have reworded the part of the rationales section about the stricter types.

L70-node-proto-loader-type-generator.md
```diff
@@ -0,0 +1,218 @@
...
+
+The gRPC implementation is specified at build time because the types depend on types from that library and the user should know at that point which implementation they are using. An alternative is to use generics to insert the types in a different way, but that increases the complexity of both the generated code and the code that uses it for relatively little gain.
+
+The goal of generating two separate interfaces for each message type is to describe two separate things as narrowly as possible: the objects that users can pass to the library as input, and the objects that the library will output. We have more control over what the library outputs, so we can be more specific in that type. This simplifies handling of messages output by the library, while still allowing the same flexibility when providing input messages that users get with the JavaScript interface.
+
+For example, all Protobuf 3 fields are optional, so the input type allows the user to omit fields, but with the `defaults` code generation option, we know that the library will always output the default value for omitted fields, so the output type can guarantee that every field will have a value.
```
wenbozhu, line 214: LG. Thanks for the update.

PR #184, opened 2020-06-03
A29: xDS-Based Security for gRPC Clients and Servers

sanjaypujare:
Draft PR for xDS transport security


A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+or [combined_validation_context.validation_context_certificate_provider](https://github.com/envoyproxy/envoy/blob/35702fed462f63a0a237cfbfdf26184272207c11/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L145). For mTLS, it will also include local identity certificate
+information in [tls_certificate_certificate_provider](https://github.com/envoyproxy/envoy/blob/35702fed462f63a0a237cfbfdf26184272207c11/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L169).
+
+If a channel is using `XdsCredentials`, whenever its CDS balancer receives a
+CDS response containing UpstreamTlsContext, it will ask the
+[Certificate Provider Store](#certificate-provider-store) to find or create a
+[Certificate Provider](#certificate-provider) for each of the two `certificate_provider`
+values received. If these two values turn out to be the same, the same Certificate
+Provider will be used in both the cases. The [Distributor](#distributor)s
+associated with the Certificate Providers are passed via channel args to all the descendants
```
markdroth, line 291: The discussion of "channel args" seems specific to C-core.  Can we phrase this in a more implementation-agnostic way?

A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+of the CDS policy which enables a subchannel created from any of these descendant policies
+to receive Certificate updates and provide them to the associated security connector.
+
+![gRPC xDS Security Client Flow Diagram](A29_graphics/grpc_xds_security_client_flow.png)
```
markdroth, line 295: There's actually another layer of indirection here that is not captured in either the description or the diagram.  The distributor passed down into the subchannels is *not* the same one that provides data from the CertificateProvider to the CDS policy.  This layer of indirection is important for two reasons:

1. If we are using a different CertificateProvider for the root cert than for the identity cert, the CDS policy needs to merge the data from two different distributors before passing it down to the subchannels.
2. We don't want to have to recreate all subchannels whenever we get a CDS update that changes the CertificateProvider.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+
+#### gRPC Server (Listener) Side Changes
```
markdroth, line 297: As mentioned above, let's remove server-side discussion from this document for now.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+![gRPC xDS Security Server Flow Diagram](A29_graphics/grpc_xds_security_server_flow.png)
+
+## Supporting SDS (for interop with Envoy)
+
+Envoy currently uses SDS to acquire certificates from an SDS server (which eventually
+gets them from a Certification Authority). Current Envoy deployments receive
+SDS configuration from the xDS control plane. For gRPC to seamlessly interoperate with
+Envoy, it needs to be able to consume the same SDS config. It is possible to support
+SDS within this plugin architecture where an SDS client can be implemented as a
+Certificate Provider plugin which talks to the SDS server (a Node Agent) over SDS and
```
markdroth, line 332: s/a Node Agent/typically a local daemon, like Istio's Node Agent/

A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+
+![gRPC xDS Security Server Flow Diagram](A29_graphics/grpc_xds_security_server_flow.png)
+
+## Supporting SDS (for interop with Envoy)
+
+Envoy currently uses SDS to acquire certificates from an SDS server (which eventually
+gets them from a Certification Authority). Current Envoy deployments receive
+SDS configuration from the xDS control plane. For gRPC to seamlessly interoperate with
+Envoy, it needs to be able to consume the same SDS config. It is possible to support
+SDS within this plugin architecture where an SDS client can be implemented as a
```
markdroth, line 331: Let's not be proscriptive here.  We haven't yet agreed on whether we would implement this as a CertificateProvider plugin or as a separate code path.  And in the context of this document, it doesn't really matter which of those routes we choose.  I think the important thing to say here is that we do intend to eventually add SDS support as well.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+
+A gRPC developer explicitly signals the use of this feature by using a new type of
+credentials on the client side and server side.
+
+#### Client Side Usage
+
+A prerequisite is to use xDS on the client side as described
+[here](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md#grpc-client-architecture).
+The gRPC developer also needs to use a new type of (client-side) credentials called
+`XdsCredentials` to use the feature on the client side. An example of this in C++ is:
```
howardjohn, line 56: This seems to drift from how Envoy behaves in unexpected ways. This is all from the perspective of a user, and not someone familiar with gRPC internals, but this seems un-intuitive to me.

Today, when writing an XDS control plane, the control plane is the source of truth and declares *all* behavior for the client. Aside from mismatched versions (sending new fields to old Envoys), if I send the same config to two Envoys I would expect them to both start using that configuration.

However, by requiring `XdsCredentials`, I believe that is no longer the case. The control plane may send the gRPC client some config to use mTLS, and its dropped on the floor. I think at the very least this could use some indication (perhaps client_features?) to the server that its going to ignore the TLS settings. As a control plane developer, its already confusing enough which fields gRPC supports - if its selectively supporting them it sounds like a nightmare to develop around, and certainly a major challenge to just drop an existing XDS server in place (rather than making a custom gRPC XDS server).

One example of where this could blow up - Istio's XDS server automatically determines whether to tell clients/servers to use mTLS based on its global view of the world. If well tell all connected clients to use mTLS, but half of them decide to ignore it, we are going to have problems.

I also feel like this has a dangerous precedent. Why is TLS the only setting that can override the XDS response? Will we soon need add a `grpc::XdsRetryOptions`, a `grpc::XdsLoadBalancingPolicy`, etc that overrides the XDS response and applies some local rules instead? If so, we may suffer the same issues mentioned above. If not, it feels inconsistent to me.

Finally, by adding this requirement, there are artificial barriers to adoption, as a user cannot just drop in the xds:/// scheme, but rather needs more code changes. This is discussed extensively in https://github.com/grpc/grpc-go/issues/4123 though, so it may make sense to continue the discussion there instead of splitting the conversation.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+server credentials to be used when the xDS-provided TLS-configuration cannot be used
+as in the following cases:
+
+- when xDS is not in use such as when the `xds:` scheme is not used on the client side or
+the xDS-server builder-option is not used on the server side.
+
+- xDS is in use but the xDS server didn't provide any TLS-configuration.
+
+In these cases the user can provide a fallback credentials for the channel or server to use.
+Only if the fallback credentials itself is of plaintext (or insecure) type, fallback to
```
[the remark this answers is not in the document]
howardjohn: I think looking at this as "fallback" is misleading, for Envoy. Its not really a fallback - the control plane is explicitly telling the proxy to use plaintext.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+```
+
+### Architecture
+
+#### Certificate Provider Plugin Architecture
+
+We take advantage of the new
+[Certificate Provider protos](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto#extensions-transport-sockets-tls-v3-commontlscontext-certificateprovider)
+added to xDS API v3. This makes our transport security solution agentless as well
+as proxyless since an "SDS-server" in the form of a Node Agent is not needed with this
```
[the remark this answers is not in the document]
howardjohn: FWIW Istio also hasn't used a "node agent" for a year or so. we just run a local sds server in the same pod, in the agent

A29-xds-tls-security.md
markdroth: Unfortunately, we can't do that with breaking the security guarantees of our existing API.  The gRPC API requires the application to explicitly indicate the type of credentials to use, and they have an expectation that we will use the type of credentials that they told us to use.  We cannot simply decide to ignore what the user asked us to do and instead use what xDS tells us to do, because (e.g.) if they told us to use TLS and xDS told us to use plaintext, we would wind up sending their data unencrypted, which would be completely surprising to them and could cause a significant security hole.

The only safe way for us to do this without causing security problems is for the user to explicitly tell us that it's okay to trust xDS to tell us what credentials to use.  If they don't do that, we can't just assume it's okay to do so.  I understand that this does mean that applications need to make an additional change to use the xDS mTLS config, but we can't avoid it in this case.

I do not expect this same thing to come up for other features (e.g., retry options or LB policy).  It's a fairly security-specific constraint.

howardjohn: @markdroth understood, thanks for the explanation. I will say if you flipped xds <-> grpc, it would all still make sense. Ultimately, we have two different sources of configuration for the same field, which is pretty tricky.

I can imagine a use case where I just want XDS for load balancing, and now I get a big security hole since the XDS server didn't send TLS configs. However, I can also see the inverse where someone wants essentially an Envoy replacement and is happy with the xDS being the complete source of truth. Have we explored exposing such information in the bootstrap configuration as well? One benefit of bootstrap vs code is the latter can be more easily changed I would imagine, and likely invisible to the user.

markdroth: I assume that the inverse case you're talking about is where a user was originally sending gRPC traffic through Envoy, so their client and server are both configured to use InsecureCreds (i.e., plaintext), and now they want to convert to proxyless.

In that situation, the first thing they would do is change the server side to use the new xDS-enabled server APIs that we're working on.  As part of that change, they would need to have their server listen on an external IP instead of localhost, and they would need to specify creds at the same time, so it should be fairly obvious that InsecureCreds is not the right thing anymore (although we should make sure to clearly document this).  And if they get the server side right, then if they change the client to use an "xds:" URI without using XdsCreds, then they'll immediately see the problem, because the client will be unable to connect to the server.

I don't think the bootstrap file solves this problem.  The fundamental issue here is that the user has told us what creds to use via the gRPC API, and I think we would severely violate their expectations if we do something different than what they told us due to some external source of data, regardless of whether that external source is the bootstrap file or the xDS server.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,340 @@
...
+When transport security (i.e. TLS/mTLS) is enabled, the response will include
+[UpstreamTlsContext](https://github.com/envoyproxy/envoy/blob/35702fed462f63a0a237cfbfdf26184272207c11/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L27)
+containing the TLS configuration needed by all the sub-channels in that cluster.
+The root-certificate information is contained in the validation context fields - either
+[validation_context_certificate_provider](https://github.com/envoyproxy/envoy/blob/35702fed462f63a0a237cfbfdf26184272207c11/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L190)
+or [combined_validation_context.validation_context_certificate_provider](https://github.com/envoyproxy/envoy/blob/35702fed462f63a0a237cfbfdf26184272207c11/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L145). For mTLS, it will also include local identity certificate
+information in [tls_certificate_certificate_provider](https://github.com/envoyproxy/envoy/blob/35702fed462f63a0a237cfbfdf26184272207c11/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L169).
+
+If a channel is using `XdsCredentials`, whenever its CDS balancer receives a
```
ejona86, line 285: I may be missing it, but I don't see any mention of how clients will do server authorization/hostname verification.

sanjaypujare: This PR needs more work and your point will be addressed as part of that rework

sanjaypujare:
I am going to close this PR and open a brand new one after incorporating comments and other changes in the proposal. Let me know if that's going to be a problem (e.g. links to this PR might break etc)

sanjaypujare:
Closing this in favor of another one.

PR #189, opened 2020-06-23
A30: xDS v3 Support

markdroth:


A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+}
+```
+
+For each new version of xDS, a server feature will be defined.  The
+presence of that feature in the `server_features` list will indicate
+that the server supports that version of xDS.
+
+Note that because each release of gRPC will support only two versions of
+xDS and will always assume that servers support the lower of the two
+versions, a server feature added to support version N+1 will no longer
```
ejona86, line 117: "will no longer be used" can easily be misinterpreted. It would be used, but by older clients, and the bootstrap generator doesn't necessarily know the client version. This paragraph seems to muddy the waters instead of making things clearer.

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+
+This means that the logic for determining the transport protocol version
+is completely independent of the logic for handling the resource version.
+gRPC will always request resources of the same version as the transport
+protocol it is using.  However, it will accept any supported version of
+resources in responses from the server, regardless of what version of
+the transport protocol we are using.
+
+In other words, if we're using the v2 transport protocol, we will
+request v2 resources, but it's fine for the server to send us either v2
```
ejona86, line 214: Except that the server doesn't know that the client supports v3.

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+or v3 resources.  Similarly, if we're using the v3 transport protocol,
+we will request v3 resources, but it's fine for the server to send us
+either v2 or v3 resources.
```
ejona86, line 217: Except that the server doesn't know that the client supports v2. When we add support for v3.1, we would drop support for these v2 resources and newer clients would break.

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+
+The purpose of this proposal is to describe how gRPC will handle the
+transition from v2 to v3 of the xDS API and to document how gRPC will handle
+xDS versioning on an ongoing basis.
+
+## Background
+
+gRPC has recently added support for obtaining configuration via the
+[xDS API](https://github.com/envoyproxy/data-plane-api).  Our current
+implementation uses uses v2 of the xDS API, but we are now adding
```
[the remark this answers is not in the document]
markdroth: Done.

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+
+This policy will give users 1-2 years to upgrade their xDS servers in
+order to continue using the latest release of gRPC.  Older releases of
+gRPC will of course continue to function using the older versions of the
+xDS API.
+
+### Deciding What Version to Use
+
+Any given gRPC release will support two versions of xDS, versions N-1 and N.
+It will assume that all servers speak version N-1, and there will be a
```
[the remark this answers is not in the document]
markdroth: You were the one who suggested this mechanism. :)  The main benefit of expressing this as a sever capability instead of just configuring what version to use is that in the future, it can allow the same bootstrap file to work for both a client that supports versions [N-1, N] and a client that supports versions [N, N+1].

I think we can support a good error message even with this design.  We know what versions we think the server supports, and we know what version we decided to try, so we can put both of those things in the error message if we get an UNIMPLEMENTED status.

A30-xds-v3.md
markdroth: I've clarified the wording here.

A30-xds-v3.md
markdroth: Sure, but because of the make-before-break approach, we can still interpret the protos the same way, so it doesn't matter.  I'm pretty sure Envoy is doing the same thing here, but @htuch can confirm.

A30-xds-v3.md
markdroth: Again, because of make-before-break, I think this is okay.  The client will still deserialize the resources as whatever version it wants, regardless of which version the server used when serializing.

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+and N+1.
+
+This policy will give users 1-2 years to upgrade their xDS servers in
+order to continue using the latest release of gRPC.  Older releases of
+gRPC will of course continue to function using the older versions of the
+xDS API.
+
+### Deciding What Version to Use
+
+Any given gRPC release will support two versions of xDS, versions N-1 and N.
```
htuch, line 55: Do you think it's worth distinguishing between transport and resource version here?

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+It will assume that all servers speak version N-1, and there will be a
+server feature defined in the bootstrap file (see below) to tell it that
+the server supports version N.  When speaking to the server, gRPC will use
+the highest version that is common to both itself and the server.
```
htuch, line 59: The minor versioning proposal features version negotiation, should this be the preferred mechanism all things being equal?

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+  example, the type for a Listener resource in v2 is `envoy.api.v2.Listener`,
+  whereas in v3 the type is `envoy.config.listener.v3.Listener`.  We
+  call this the *resource version*.
+
+When gRPC uses v3 to talk to a server, the primary thing that it is
+choosing is the transport protocol version, not the resource version.
+
+Luckily, we can be more flexible with the resource version, because
+there's essentially only one case where we use a field in the v2 protos
+that was removed in the v3 protos.  (That case is the node `build_version`
```
htuch, line 200: Doesn't `build_version` only appear in the transport protocol? I.e. as a part of `DiscoveryRequest`? So for all actual resources, it seems there is no exception (maybe `Node` appears somewhere else?`).

A30-xds-v3.md
htuch: Yes, because the v2/v3 wire formats are compatible (excepting the newer fields in v3 and deprecated fields in v2), this is basically as simple as either deserializing in the right way.

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+For now, we will use the same server feature in the bootstrap file to
+indicate both the xDS and the LRS transport protocol version to use.
+
+Note that this means that there will not be a way to configure the client
+to use different transport versions for the two protocols.  In the future,
+if this becomes a problem, we can make use of `transport_api_version`
+field being added in https://github.com/envoyproxy/envoy/pull/11754 to
+specify the transport version of LRS.  However, since this field will exist
+only in xDS v3, this would still not provide a clean way to specify use
+of LRS v3 while using xDS v2.
```
htuch, line 230: I think maybe we could add this to v2. The v2 freeze is there to encourage folks to move to v3, if we're adding a field to explicitly make this easier, then why not.

A30-xds-v3.md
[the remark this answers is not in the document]
ejona86: I'm not arguing against the mechanism. The server capability works. I'm arguing against the "assume."

> We know what versions we think the server supports

No we don't. N-1 doesn't need to be included at all. I want N-1 to be included and us not to assume it. I'm fine with assuming v2, for backward compatibility, but once we support v3.1 we should only care about explicitly specified versions.

> so we can put both of those things in the error message if we get an UNIMPLEMENTED status

In later minor versions, we won't get UNIMPLEMENTED status, because the package didn't change.

A30-xds-v3.md
ejona86: I agree that it will "work." But it won't actually _work_, because when the client drops v2 it will stop functioning even though the server claims to support v3.

I think it would be fine to say, it "works," but it should also be noted that the server can't actually do it safely.

A30-xds-v3.md
ejona86: There is a mapping from v2 resource names to v3 resource names. You don't just look at the message name, "oh, this ends with `Listener`, let's assume it is an `envoy.config.listener.v3.Listener`". The version is part of the `Any` and `type_url`. When the client drops v2 support, are you saying it will keep the v2→v3 message name mapping and continue supporting v2 protos? That is not at all obvious to me.

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+* Status: In Review
+* Implemented in: 
+* Last updated: 2020-06-25
+* Discussion at: https://groups.google.com/g/grpc-io/c/YRVqRMUQwhM
+
+## Abstract
+
+The purpose of this proposal is to describe how gRPC will handle the
+transition from v2 to v3 of the xDS API and to document how gRPC will handle
+xDS versioning on an ongoing basis.
```
sanjaypujare, line 14: In that case would it be better to rename the gRFC (and the filename) to something like "xDS versioning in gRPC" since that will be the long term purpose?

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+For each new version of xDS, a server feature will be defined.  The
+presence of that feature in the `server_features` list will indicate
+that the server supports that version of xDS.
+
+Note that because each release of gRPC will support only two versions
+of xDS and will always assume that servers support the lower of the
+two versions, the server feature indicating support for version N+1
+will no longer be used by a release of gRPC that supports version N+2.
+However, it will not cause any problems to be left in the bootstrap file.
+So at any given time, it useful to have a bootstrap file indicating the
```
sanjaypujare, line 120: s/it useful/it is useful/

A30-xds-v3.md
```diff
@@ -0,0 +1,255 @@
...
+of which version of the transport protocol is being used.
+
+### Environment Variable
+
+Initially, support for xDS v3 in gRPC will be considered experimental,
+since we are planning to release this support in the client before we
+have been able to perform interop testing with any real-world v3 xDS
+server.  As a result, we are taking the added safety precaution that the
+new server feature in the bootstrap file will only be used if the
+`GRPC_XDS_EXPERIMENTAL_V3_SUPPORT` environment variable is set to
```
sanjaypujare, line 242: This is a good idea and it will be nice to standardize the mapping from the server_feature value to the env variable name. So for server_feature of "xds_v3" the env var name to check is "GRPC_EXPERIMENTAL_SUPPORT_XDS_V3"  i.e. you append capitalized "XDS_V3" to "GRPC_EXPERIMENTAL_SUPPORT_"

A30-xds-v3.md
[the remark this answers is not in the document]
markdroth: I see, so you'd prefer an approach where, when we add support for v3.1, gRPC always explicitly looks for a server feature for both v3 and v3.1, and doesn't automatically assume v3?

I'm not actually sure this makes much practical difference, but I guess it is a little easier to explain.  I've updated the doc accordingly.  PTAL.

A30-xds-v3.md
markdroth: I think the current title is probably going to seem more relevant to existing and near-term prospective users.

A30-xds-v3.md
markdroth: I think this distinction is described in more detail below, but I'm concerned that trying to distinguish between the two this early in the document will make it more confusing.

In general, I do think that we need a clearer separation between transport version and resource version.  But I think we'll need to see how the ecosystem evolves before we know how much we can really pull them apart in practice.

A30-xds-v3.md
markdroth: The minor versioning proposal says that there will be some mechanism for version negotiation, but it does not actually define what that mechanism will be.  The current text of the proposal suggests using context params, but in offline conversation, you and I and Matt agreed that that won't actually work due to the exact-match semantics of context_params.  I believe that where we left this was that when someone actually gets around to implementing the minor version proposal, they will need to come up with a concrete proposal for version negotiation.  So I think that how this works is an open question, and I don't want to say anything about that here until we have a concrete proposal to evaluate.

It is also worth noting that the way our implementation is going to work, the version configured in the bootstrap file is effectively the transport version, not the resource version, whereas the version negotiation mechanism is by necessity primarily about resource version.  As an example, the version negotiation mechanism probably won't be usable to control things like the "xds transport next steps" design that you're currently working on implementing, because the client needs to know whether to use it before it sends the initial request containing its supported versions.  This is another reason why I believe it would be beneficial to consider the transport version and the resource version independently: in terms of mechanism, transport changes are generally going to be configured via client-side configuration rather than version negotiation.

A30-xds-v3.md
markdroth: Sounds good.  I've sent you envoyproxy/envoy#11824.

A30-xds-v3.md
htuch: Yeah, the "accept v2 resources even though we are really asking for v3" is only acceptable for use during the 1 year transition period between deprecation and removal. It's a rollout strategy, not a production one.

A30-xds-v3.md
htuch: As above. The idea is that mixed version support only exists during the 1 year transition period and is only there for supporting rollouts where they may be mixed client/server features as an ephemeral process.

A30-xds-v3.md
```diff
@@ -0,0 +1,267 @@
...
+  whereas for v3 it is
+  `/envoy.service.discovery.v3.AggregatedDiscoveryService/StreamAggregatedResources`.
+  We call this the *transport protocol version*.
+- The type of resources that we request via the transport protocol.  For
+  example, the type for a Listener resource in v2 is `envoy.api.v2.Listener`,
+  whereas in v3 the type is `envoy.config.listener.v3.Listener`.  We
+  call this the *resource version*.
+
+When gRPC uses v3 to talk to a server, the primary thing that it is
+choosing is the transport protocol version, not the resource version.
```
htuch, line 200: One consideration worth noting is that the transport version only can currently change based on major version. There is no minor/patch in the gRPC endpoint.

A30-xds-v3.md
```diff
@@ -0,0 +1,267 @@
...
+the highest version that is common to both the client and the server is
+N-1, so the client will speak version N-1.  However, if the bootstrap file
+contains the server features for both versions N-1 and N, the highest
+version that is common to both the client and server is N, so the client
+will speak version N.
+
+As a special case for backward compatibility, there will be no server
+feature defined for xDS v2 support.  Existing releases of gRPC
+automatically assume that servers support v2, and we will continue to
+assume that when we add support for v2.  When we add support for the
```
ejona86, line 72: s/v2/v3/

markdroth: Done.

A30-xds-v3.md
markdroth: Okay.  I've added a paragraph clarifying that this is just intended for use as a rollout strategy, not something that can be assumed in the general case.

A30-xds-v3.md
markdroth: I've added a paragraph clarifying that this is just intended for use as a rollout strategy, not something that can be assumed in the general case.

A30-xds-v3.md
markdroth: Yeah, I realize that.  But this section of the doc is specific to the v2->v3 transition, and in this case, the major version number is the only relevant part.

A30-xds-v3.md
```diff
@@ -0,0 +1,275 @@
...
+
+In other words, if we're using the v2 transport protocol, we will
+request v2 resources, but it's fine for the server to send us either v2
+or v3 resources.  Similarly, if we're using the v3 transport protocol,
+we will request v3 resources, but it's fine for the server to send us
+either v2 or v3 resources.
+
+Note that the ability to accept resource versions that are different
+than what was requested is only intended for use as a rollout strategy
+during transitions where the management server has some a prior
```
ejona86, line 234: "a priori" or "prior"

markdroth: Done.

A30-xds-v3.md
```diff
@@ -0,0 +1,275 @@
...
+
+### Related Proposals: 
+
+[A27: xDS-Based Global Load Balancing](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)
+
+## Proposal
+
+### Version Support Policy
+
+gRPC's policy will be to support two versions of xDS at any given time.
```
dfawley, line 42: Is this "at most two versions"?  Or "exactly two versions"?  If v4 isn't released for 5 years, will we support v2 for that long?

markdroth: The current plan is to have one new version per year, so I think this distinction isn't important.  If the version release schedule changes, then we can change our policy at that point.

dfawley: That sounds fair.  It does seem likely there will be an xDS version bump every time there is an opportunity to do so.

PR #190, opened 2020-07-06
L72: Allow Call Credentials to be Specified in `grpc_google_default_credentials_create`

gnossen:


L72-core-google_default_credentials-extension.md
```diff
@@ -0,0 +1,62 @@
...
+* Status: Draft
+* Implemented in: Core
+* Last updated: July 6th, 2020
+* Discussion at: (TODO)
+
+## Abstract
+
+This document proposes that `grpc_google_default_credentials_create` be
+amended to allow the user to specify their desired call credential, while
+also adding a `void* reserved` argument to lessen the burden of any such
```
markdroth, line 14: I don't see the value in adding a `void* reserved` argument.  If we were to actually use that for anything, we'd want to change the type from `void*` to something more specific anyway, so the change would still require a gRFC.

L72-core-google_default_credentials-extension.md
```diff
@@ -0,0 +1,62 @@
...
+changes in the future.
+
+
+## Background
+
+The so-called "Google default credentials" created by the
```
markdroth, line 20: The term "so-called" seems misplaced here.  We're not pretending; these really are Google default credentials. :)

L72-core-google_default_credentials-extension.md
```diff
@@ -0,0 +1,62 @@
...
+`grpc_google_default_credentials_create` function in Core enables connection to
+GCP via a combination of ALTS and SSL credentials, along with a special oauth2
```
markdroth, line 22: It doesn't enable connection to GCP; it enables connection to OnePlatform services.

L72-core-google_default_credentials-extension.md
```diff
@@ -0,0 +1,62 @@
...
+token that must assert the same identity as the channel-level ALTS credential.
+
+This oauth2 token attached by `grpc_google_default_credentials_create` is based
+on the default VM service account, but the auth libraries for wrapped languages
```
markdroth, line 26: Does this mean that the wrapped languages have a mechanism for changing the service account used in the channel-level ALTS credential?  How does that work?

gnossen: So I'm told. I don't have the context myself. @apolcyn should be able to provide it.

L72-core-google_default_credentials-extension.md
gnossen: Switching to "Google services" since OnePlatform doesn't seem to be a public name.

L72-core-google_default_credentials-extension.md
gnossen: Fair enough. Is there a reason that this pattern features so heavily in the core APIs then?

gnossen: Removed.

markdroth: I suspect that this pattern predated the decision that we are allowed to make breaking changes in the C-core API.

L72-core-google_default_credentials-extension.md
apolcyn: Wrapped languages don't have a mechanism for changing the service account used in the ALTS credentials, but they don't really need to (as long as they're OK with authenticating as the default service account).

ALTS credentials by default authenticate the VM's default service account.  So long as the call credentials also authenticate the VM's default service account, then we're ok. And note that any call credentials which ultimately fetches the token from `http://metadata.google.internal/computeMetadata/v1/project/service-accounts/default/token` will authenticate the VM's default service account, creating a guarantee that we don't have a call-creds/ALTS identity mismatch.

Whether or not the wrapped language layers and above do the right thing, i.e. being sure that per-RPC tokens authenticate the VM's default service account, when potentially using ALTS, will be up to them though.

L72-core-google_default_credentials-extension.md
```diff
@@ -0,0 +1,58 @@
...
+Google services via a combination of ALTS and SSL credentials, along with a special oauth2
+token that must assert the same identity as the channel-level ALTS credential.
+
+This oauth2 token attached by `grpc_google_default_credentials_create` is based
+on the default VM service account, but the auth libraries for wrapped languages
+allow the creation of channels with non-default service accounts. In this case,
+a mismatch in identity occurs that may cause an unexpected persistent failure of
+RPCs at runtime.
+
+While the default service account is the correct option is most cases, there
```
apolcyn, line 28: I think we might want to change wording here. This is not so much about allowing higher layers to authenticate alternative identities, rather than it is allowing client libraries to continue to use their own mechanisms to get per-RPC tokens, when potentially using ALTS. I.e. this API described here allows  client libraries to continue uses the auth metadata plugin APIs to get per-RPC metadata, when using ALTS.

L72-core-google_default_credentials-extension.md
```diff
@@ -0,0 +1,58 @@
...
+I propose that the signature of `grpc_google_default_credentials_create` be
+amended to the following:
+
+```C
+GRPCAPI grpc_channel_credentials*
+grpc_google_default_credentials_create(grpc_call_credentials* call_credentials);
+```
+
+Supplying `nullptr` for `call_credentials` will result in the current behavior
+of the function -- oauth2 call credentials based on the default service account
```
apolcyn, line 43: we should change wording here -- the current behavior of how google default credentials chooses call credentials is actually based on the "Application Default Credentials" spec (see https://cloud.google.com/docs/authentication/production?_ga=2.68587985.1354052904.1594166352-2074181900.1593114348#finding_credentials_automatically). This  <i>by default</i> authenticates the VM's default service account but isn't necessarily guaranteed to, e.g. if the `GOOGLE_APPLICATION_CREDENTIALS` environment variable is set and points a JSON config file for another identity.

gnossen:
@apolcyn Thank you for the clarifications! Very helpful. I've tried to incorporate them into the proposal. PTALA. Leaving it to you to resolve the comments if you feel they've been addressed properly.

L72-core-google_default_credentials-extension.md
```diff
@@ -0,0 +1,65 @@
...
+
+This document proposes that `grpc_google_default_credentials_create` be
+amended to allow the user to specify their desired call credentials.
+
+## Background
+
+The Google default credentials created by the
+`grpc_google_default_credentials_create` function in Core enable connection to
+Google services via a combination of ALTS and SSL credentials, along with a special oauth2
+token that must assert the same identity as the channel-level ALTS credential.
```
apolcyn, line 20: nit: related to below comment, for accuracy lets replace "must assert" with "by default asserts"

apolcyn: Maybe describe how the token that google default credentials uses is decided per the Application Default Credentials spec

apolcyn, line 20: lets note here that the identity we're discussing here is specifically the one authenticated by the token available on the metadata server's `http://metadata.google.internal/computeMetadata/v1/project/service-accounts/default/token` endpoint.

L72-core-google_default_credentials-extension.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Abstract
+
+This document proposes that `grpc_google_default_credentials_create` be
+amended to allow the user to specify their desired call credentials.
+
+## Background
+
+The Google default credentials created by the
+`grpc_google_default_credentials_create` function in Core enable connection to
```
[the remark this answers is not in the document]
gnossen: "The Google default credentials ... enables"? "Enable" agrees with the plural subject, unless we've atomized "credentials" to be singular.

gnossen:
@apolcyn Thanks for the suggestions! PTALA.

L72-core-google_default_credentials-extension.md
apolcyn: these comments are still relevant, to improve accuracy of the wording here

apolcyn: sorry, to be clear I meant to say that we should clarify that the ALTS channel-level identity is that authenticated by the token served from this metadata server endpoint.

Google default credentials picks it's per-RPC metadata token via the ADC spec, though (which by default is, but in general isn't necessarily going to result in a per-RPC token from that same endpoint).

gnossen: You're right. The current wording is ambiguous. Updated to clarify that it's the ALTS identity that's taken from that HTTP endpoint.

gnossen: I missed the `s/must/by default/`. Sorry about that. I added a numbered list describing how ADC works in detail.

PR #191, opened 2020-07-10
L73:  BinderChannel for native cross-process communication on Android.

markb74:


L73-binderchannel/L73-binderchannel.md
```diff
@@ -0,0 +1,317 @@
...
+pass platform-define parcelable objects (e.g. `PendingIntent`), some of which
+can't simply be serialized into a byte stream.
+
+## Proposal
+
+We propose the creation of client and server transport that supports
+communication via [Binder] transactions with an Android [bound service], with
+corresponding channel and server builders. This transport will support the
+inclusion of Android Parcelable objects in Metadata, to allow Parcelables when
+necessary, while discouraging Parcelables as a general message format.
```
sanjaypujare, line 45: Not using Parcelable for serialization/deserialization is understandable because of versioning issues described above, but I am not sure there is enough justification for why Parcelables should be Metadata. Metadata is sent as headers/trailers and so has different semantics. Can one convert Parcelable to protobuf (and reverse) and include as part of request/response or is that also discouraged?

ejona86: It's because "some of which can't simply be serialized into a byte stream." It is not possible to form a protobuf encoding; they are handled specially by the OS. They have features similar to fd-passing for unix-domain sockets.

sanjaypujare: Okay the UDS FD example and special handling by OS make sense. Otherwise I was thinking Metadata is "metadata" about an RPC (such as tracing/telemetry or routing info) but not the content of the RPC itself. So I got the impression that the limitations of serializing/encoding is why we "promoted" Parcelable to be metadata.

L73-binderchannel/L73-binderchannel.md
```diff
@@ -0,0 +1,317 @@
...
+burden on application developers vs. lower-level APIs (e.g. posix sockets), we
+expect the same to hold for cross-process communication on Android.
+
+Protocol buffers are also more resilient to version skew than traditional
+Android [Parcelable] instances, an important consideration when calls are
+cross-application.
+
+Despite this, the use of [bound services] is necessary for the Android platform
+to be aware of inter-process dependencies, and there's occasionally a need to
+pass platform-define parcelable objects (e.g. `PendingIntent`), some of which
```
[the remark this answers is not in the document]
markb74: Done.

L73-binderchannel/L73-binderchannel.md
markb74: FWIW, the transport _does_ actually support Parcelable messages internally (it's also documented in the wire format), so someone could create a marshaller to send Parcelable messages if they really wanted.

That gets quite ugly though, and you don't get to use generated stubs, so we want to discourage it.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,173 @@
...
+transactions relating to that stream (in both directions).
+
+IDs are allocated sequentially from FIRST_CALL_ID to LAST_CALL_ID, wrapping
+around at the end. Since there are roughly 2^24 available IDs, there’s no chance
+of a collision except in pathological cases. If we _do_ detect a collision
+however, we immediately shut down the transport with an INTERNAL error code.
+
+Each normally-completed stream transmits (in both directions) a prefix, 0 or
+more blocks of message data, and a suffix. The contents of prefix and suffix
+data are direction dependenet, and all three may can be sent in a single
```
ejona86, line 73: s/may// ?

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,173 @@
...
+A single endpoint binder exists for each BinderServer or OnDeviceServerEndpoint.
+A client binder / server binder pair is created for each transport pair.
+
+## Transaction Types
+
+There are two types of transactions, control and stream.
+
+### Transport-level Control Transactions
+
+For general transport lifecycle, using one of 1000 reserved transaction codes.
```
ejona86, line 25: We should define the behavior for when an unknown transaction code is received.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,173 @@
...
+### Stream Transactions
+
+Aside from the reserved 1000 transaction codes, all others are used for streams.
+When the ClientTransport creates a new Stream, it assigns an ID to use for all
+transactions relating to that stream (in both directions).
+
+IDs are allocated sequentially from FIRST_CALL_ID to LAST_CALL_ID, wrapping
+around at the end. Since there are roughly 2^24 available IDs, there’s no chance
+of a collision except in pathological cases. If we _do_ detect a collision
+however, we immediately shut down the transport with an INTERNAL error code.
```
ejona86, line 69: Could we gracefully shut down the transport instead? A long-lived stream could trigger this over a period of time. Unlikely given the shorter life of Android processes, but not out-of-the-question.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,173 @@
...
+```
+flags = int; (* A bitset of flags, and an optional status code. *)
+sequence no = int; (* A transaction sequence no for the RPC *)
+method ref = string; (* The full method name *)
+status = [string] (* The status code is in the flags. The description is only
+                     included if FLAG_STATUS_DESCRIPTION is set.  *);
+count = int;
+bytes data = count, [byte array]; (* byte array included iff count > 0 *)
+metadata = count, { metadata key, metadata val }; (* repeated count times *)
+metadata sentinel = int (* -1 = Parcelable, -2 = bad metadata. *)
```
ejona86, line 87: What is bad metadata?

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,173 @@
...
+server rpc data = [server prefix] (* if FLAG_PREFIX is set *)
+                  [message data]  (* if FLAG_MESSAGE_DATA is set *)
+                  [server suffix] (* if FLAG_SUFFIX is set *)
+out of band close data = status;
+rpc data = client rpc data | server rpc data | out of band close data;
+
+rpc transaction = flags, sequence no, rpc data;
+```
+
+#### Flag fields
```
ejona86, line 112: We should define what happens if an unknown field is present.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,173 @@
...
+calls to be rejected, but allows ongoing calls to complete normally.
+
+By comparison, an ungraceful shutdown closes all ongoing calls, sends a
+shutdown-transport transaction to the remote binder, and shuts down all binders
+immediately. If the remote binder dies, or any internal failure occurs, an
+ungraceful shutdown will occur. If the Android Service hosting the server is
+destroyed, a graceful shutdown will occur, though if the host process also dies,
+that will lead to a binder death in the client, shutting the client-side down
+ungracefully. If the Android component hosting the Channel is destroyed, an
+ungraceful shutdown will occur, with the status code ABORTED. The remote
```
ejona86, line 166: We can't generate ABORTED from within the library (see very end of [statuscodes.md](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md)). This would generally be exposed as UNAVAILABLE or CANCELLED (depending on implementation-specific details).

L73-java-binderchannel/wireformat.md
markb74: ignore unrecognized.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,173 @@
...
+## Transaction Types
+
+There are two types of transactions, control and stream.
+
+### Transport-level Control Transactions
+
+For general transport lifecycle, using one of 1000 reserved transaction codes.
+
+*   Setup Transport
+    *   Transaction code: BinderTransport.SETUP_TRANSPORT
```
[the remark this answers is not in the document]
markb74: Done.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,173 @@
...
+    *   Transaction code: BinderTransport.SHUTDOWN_TRANSPORT
+    *   From ClientTransport to server binder, or ServerTransport to client
+        binder.
+*   Acknowledge bytes.
+    *   Transaction code: BinderTransport.ACKNOWLEDGE_BYTES
+    *   From either transport to its peer's binder. Reports reception of
+        transaction data for flow control.
+*   Ping
+    *   Transaction code: BinderTransport.PING
+    *   From ClientTransport to server binder.
```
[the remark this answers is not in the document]
markb74: Updated to indicate both sides should respond to a ping message.

L73-java-binderchannel/wireformat.md
markb74: yep, do.

L73-java-binderchannel/wireformat.md
markb74: Cool, I'll go with CANCELLED since that seems the most appropriate.

L73-java-binderchannel/wireformat.md
markb74: Good point, this is easy to handle gracefully as well. Updated to shutdown gracefully and fail the new call (the one which led to the collision) with UNAVAILABLE.

L73-java-binderchannel.md
```diff
@@ -0,0 +1,321 @@
...
+
+```java
+BinderChannelBuilder.create(
+      applicationContext,
+      AndroidComponentAddress.forRemoteComponent("pkg", "pkg.ServiceClass"))
+    .build();
+
+BinderChannelBuilder.create(
+      activity,
+      activity.getLifecycle(),
```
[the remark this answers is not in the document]
markb74: As discussed offline, I ended up deciding to keep it as an option, since avoiding cross-component memory leaks is important on Android.

L73-java-binderchannel/wireformat.md
markb74: Added a comment to say we'll shut down the transport gracefully if this happens.

I also modified the version below to include optional flag fields we discussed.

L73-java-binderchannel/wireformat.md
markb74: As discussed, it's unnecessary, and it looks like I can remove it without disruption. Removed.

L73-java-binderchannel.md
```diff
@@ -0,0 +1,321 @@
...
+    .build();
+
+OnDeviceChannelBuilder.create(
+        activity,
+        activity.getLifecycle(),
+        "ondevice://example")
+    .build();
+```
+
+Creating a channel directly to an external AndroidComponentAddress is also
```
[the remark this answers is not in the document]
markb74: Oh good point, it's not. Removed.

L73-java-binderchannel.md
```diff
@@ -0,0 +1,321 @@
...
+
+### BinderChannelBuilder
+
+BinderChannelBuilder is used to create a channel to a BinderServer, and takes an
+AndroidComponentAddress as target. Channels can be created either globally for
+the entire application, or tied to the lifecycle of one component via an
+optional [Lifecycle] instance.
+
+```java
+BinderChannelBuilder.create(
```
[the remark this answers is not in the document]
markb74: Done. I used "forAndroidComponent" here and "forTarget" with OnDeviceChannelBuilder.

L73-java-binderchannel/wireformat.md
```diff
@@ -25,38 +25,60 @@ There are two types of transactions, control and stream.
...
 ping response transaction = ping id;
 ```
 
+The setup transaction sends version information as a major version followed by
+optional extras, which are only includede if the major version is at
+least 2.
+
+The extras are a minor version number, and a set of protocol extension flags.
+
+During transport setup, the client transport should provide the set of
```
ejona86, line 75: I think we should just have the server respond with all flags it supports. That would allow a feature which is non-symmetric on the connection. For example, let's say we wanted a feature bit for "allows parcelable metadata" (made up example) we could have the client allow it and the server not; the server could then send parcelables if it wanted but wouldn't have to receive. To do so with the current approach would require two bits. It's really no more work for the client or server; they just read the bits they care about and send the bits they support.

L73-java-binderchannel/wireformat.md
```diff
@@ -25,38 +25,60 @@ There are two types of transactions, control and stream.
...
+minor version = int;
+protocol extension flags = int;
+version extras = minor version, protocol extension flags
 num bytes = int;
 ping id = int;
+version info = major version, [version extras] (* If major version is at least 2. *)
+
 
-setup transport transaction = version, binder;
+setup transport transaction = version info, binder;
```
ejona86, line 62: As discussed separately, we could add `[version extras]` after `binder` and avoid the need to bump the protocol version.

L73-java-binderchannel/wireformat.md
ejona86, line 75: As discussed separately, if the client's version is too high, let's have the server still respond with a SETUP_TRANSACTION but with a version it does support.

L73-java-binderchannel/wireformat.md
```diff
@@ -25,38 +25,60 @@ There are two types of transactions, control and stream.
...
+it will shutdown the transport gracefully.
 
 #### Transaction Data
 
 ```
-version = int; (* Currently always 1 *)
+major version = int;
+minor version = int;
+protocol extension flags = int;
+version extras = minor version, protocol extension flags
```
ejona86, line 56: Thinking about the server auto-downgrading to an older major version, made me think about if we missed anything else. We don't have any magic bytes right now. Do we need them? I'm thinking they don't add value for us, but just wanted to check.

We could add them to `version extras`, but it sort of seems like magic bytes are unlikely to be helpful if `version extras` is after `binder`.

L73-java-binderchannel/wireformat.md
markb74: Sounds good, done.

L73-java-binderchannel/wireformat.md
```diff
@@ -25,38 +25,60 @@ There are two types of transactions, control and stream.
...
 ```
 
+The setup transaction sends version information as a major version followed by
+optional extras, which are only includede if the major version is at
+least 2.
+
+The extras are a minor version number, and a set of protocol extension flags.
+
+During transport setup, the client transport should provide the set of
+extension flags it supports, and the server should respond with the intersecion
```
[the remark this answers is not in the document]
markb74: Done.

L73-java-binderchannel/wireformat.md
```diff
@@ -25,38 +25,60 @@ There are two types of transactions, control and stream.
...
-setup transport transaction = version, binder;
+setup transport transaction = version info, binder;
 shutdown transport transaction = nothing;
 acknowledge bytes transaction = num bytes;
 ping transaction = ping id;
 ping response transaction = ping id;
 ```
 
+The setup transaction sends version information as a major version followed by
+optional extras, which are only includede if the major version is at
```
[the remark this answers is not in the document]
markb74: Done.

L73-java-binderchannel/wireformat.md
markb74: I went with just putting the protocol extension flags after binder, since we discussed dropping the minor version.

L73-java-binderchannel/wireformat.md
markb74: Done.

L73-java-binderchannel/wireformat.md
markb74: Oh good catch, we don't need something new, since we effectively have an equivalent in the intent we use to bind in the first place. "grpc.io.action.BIND". Adding that.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,200 @@
...
+ping transaction = ping id;
+ping response transaction = ping id;
+```
+
+During transport setup, if a server receves a setup transaction with a higher
+protocol version, it should respond with the protocol version _it_ supports.
+
+Both client and server transport may also include protocol extension flags at
+the end of their setup transport transaction. This is reserved for potential
+future protocol extensions, though no flags are currently specified. Since some
```
ejona86, line 70: I'm not sure the "Since some features may..." sentence actually adds anything, since the expected behavior is _always_ specific to that one flag. So you may just drop it.

It is probably more important to say that when unknown flags are received, they should be ignored.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,200 @@
...
+    *   From SeverTransport to client binder or ClientTransport to server binder,
+        in response to a ping message.
+
+If either side of a transport receives an unrecognized control transacton,
+it will shutdown the transport gracefully.
+
+#### Transaction Data
+
+```
+major version = int;
```
ejona86, line 53: nit: Do we still want "major version." I think it was just "version" before. Doesn't matter.

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,200 @@
...
+ping id = int;
+
+setup transport transaction = major version, binder, [protocol extension flags];
+shutdown transport transaction = nothing;
+acknowledge bytes transaction = num bytes;
+ping transaction = ping id;
+ping response transaction = ping id;
+```
+
+During transport setup, if a server receves a setup transaction with a higher
```
ejona86, line 65: "higher protocol version" may easily be misinterpreted. I know you mean a version higher than it supports. And actually we don't say what should happen if the version is lower than the max supported by the server.

Not perfect, but maybe good enough:
> `setup transport transaction` is guaranteed to be backward-compatible and so is safe to process even if the version number is unknown to an implementation. During transport setup, the version sent by the client should be the largest version it supports. The server should respond with the actual version to be used, which will be equal to or less than the client-advertised version. The server must not fail if the client-advertised version is higher than supported. The server may fail if the client-advertised version is too low to be supported. The client may fail if the server-selected version is too low to be supported.

(I said "fail" there. It may be good to be more specific that the server would respond with SHUTDOWN_TRANSPORT and the client would simply release the binder and notify the client.)

s/receves/receives/

L73-java-binderchannel/wireformat.md
```diff
@@ -0,0 +1,200 @@
...
+    responds will multiple messages _anyway_, the hint will be ignored, and messages will be
+    delivered normally.
+*   FLAG_STATUS_DESCRIPTION (0x20)- Indicates the status description string is
+    included.
+*   FLAG_MESSAGE_DATA_IS_PARCELABLE (0x40) - Indicates the message is a parcelable
+    object.
+*   status - If a status is included in the data, the status code will be
+    present in the top 16 bits of the flags value.
+
+To allow for potential additions to the protocol, any any unrecognized flags may
```
ejona86, line 147: s/any any/any/

s/may/must/ ?

ejona86 (approved):

L73-java-binderchannel.md
```diff
@@ -0,0 +1,310 @@
+# BinderChannel for native cross-process communication on Android
+
+*   Author(s): markb74
+*   Approver:
+*   Status: In Review
```
ejona86, line 5: Let's update this and the "last updated." And put me as the approver.

ejona86:
@markb74, tell me when you think it is ready and I can merge it.

L73-java-binderchannel.md
markb74: done.

L73-java-binderchannel/wireformat.md
markb74: done.

L73-java-binderchannel/wireformat.md
markb74: Yeah, that sentence annoyed me while writing it. Happy to drop, and added the 'ignored' part.

L73-java-binderchannel/wireformat.md
markb74: good point, changed.

L73-java-binderchannel/wireformat.md
markb74: You've written these things before. 8) Copied shamelessly with the SHUTDOWN_TRANSPORT detail.

L73-java-binderchannel/wireformat.md
```diff
@@ -50,26 +50,32 @@ it will shutdown the transport gracefully.
...
+setup transport transaction = version, binder, [protocol extension flags];
 shutdown transport transaction = nothing;
 acknowledge bytes transaction = num bytes;
 ping transaction = ping id;
 ping response transaction = ping id;
 ```
 
-During transport setup, if a server receves a setup transaction with a higher
-protocol version, it should respond with the protocol version _it_ supports.
+==setup transport transaction== is guaranteed to be backward-compatible and so
```
ejona86, line 65: Did you mean to use backtick or `**` here? `==` doesn't format into anything here.

ejona86 (approved):
One small formatting issue

L73-java-binderchannel/wireformat.md
markb74: Oops, fixed.

abdelhalim:
What is the status of this proposal?

ejona86:
@abdelhalim, see https://github.com/grpc/grpc-java/releases/tag/v1.42.0

paulocoutinhox:
There is something similar to Swift or Obj-C?

PR #192, opened 2020-07-13
A31: xDS Timeout Support and Config Selector Design

dfawley:
This proposal includes details for implementing support for xDS RouteActions
which configure RPC timeouts, retries, etc.


A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+
+![xDS Architecture](A31_graphics/xds_architecture.png "xDS Architecture")
+
+Initially, only timeout-related `RouteAction` settings will be supported:
+<code>[timeout](http://google3/third_party/envoy/src/api/envoy/api/v2/route/route_components.proto?l=886&rcl=309630292)</code>
+and
+<code>[max_grpc_timeout](http://google3/third_party/envoy/src/api/envoy/api/v2/route/route_components.proto?l=979&rcl=309630292)</code>. <code>[grpc_timeout_offset](http://google3/third_party/envoy/src/api/envoy/api/v2/route/route_components.proto?l=988&rcl=309630292)</code>
+is used for shortening timeouts set by clients to ensure the timeout is reached
+before the client cancels the RPC, so supporting it in the client would be
+counter-productive.  If <code>max_grpc_timeout</code> is not present, the
```
markdroth, line 150: We also need to clarify the relationship between these values and the deadline set by the application.

Just to make this super-clear to readers, it might be a good idea to include a table to show what timeout will actually be used with various combinations of these values.  For example, the following entries might be useful:

- application deadline unset, max_grpc_timeout unset, timeout unset: actual=15s
- application deadline unset, max_grpc_timeout unset, timeout=10s: actual=10s
- application deadline unset, max_grpc_timeout=0: actual=infinite
- application deadline unset, max_grpc_timeout=10s: actual=10s
- application deadline=20s, max_grpc_timeout unset, timeout unset: actual=15s
- application deadline=20s, max_grpc_timeout unset, timeout=10s: actual=10s
- application deadline=20s, max_grpc_timeout=0: actual=20s
- application deadline=20s, max_grpc_timeout=10s: actual=10s

@htuch may also be interested in reviewing this piece.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+these settings.
+
+## Rationale
+
+### Alternatives Considered
+
+Many alternatives were considered as part of this design.  Below are several of them:
+
+1.  Support xDS matchers in gRPC directly.
+    1.  This could lead to performance problems for existing users and would
```
markdroth, line 165: These sub-items can probably be bullets rather than numbered items.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+The "xds" resolver will return a Config Selector which applies the xDS matchers
+to the RPC (using the Method name and RPC Header data provided to it) and uses
+those inputs to perform the initial routing steps for the RPC based upon the RDS
+Routes.  This will enable it to set the timeouts and other settings
+appropriately in the MethodConfig according to the RouteActions.  Since the
+routing decisions are made here, it will also return the cluster name to the
+top-level Load Balancing Policy in order to inform the policy which endpoint
+picker to use for the RPC.  Note that the Config Selector will also be assuming
+the responsibility of the current `weighted_target` Load Balancing Policy.  This
+means two things: 1. connectivity state is not considered when making this
```
markdroth, line 86: We should explicitly note that this is a change from the behavior specified in [gRFC A28](https://github.com/grpc/proposal/blob/master/A28-xds-traffic-splitting-and-routing.md), where we explicitly stated that we'd do this differently from Envoy.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+routing decision, and 2. once a cluster is chosen for an RPC, retry attempts
+will only use this cluster.  This matches Envoy behavior.
+
+The top-level xds Load Balancing Policy will become a simple aggregator of child
+"cds" balancers.  The Load Balancing Config provided by the Resolver will be a
```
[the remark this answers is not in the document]
dfawley: Argh, I thought I got them all.  Thanks!

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+JSON struct containing one field for each cluster name and associated values
```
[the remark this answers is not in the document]
dfawley: Hmm, so this kind of format wouldn't have a natural proto representation.

How about this instead:

```proto
message XDSClusterManager {
  repeated CDSConfig cds_config = 1;
}
```

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+    the Service Config.
+
+This wouldn't need to be encoded as an explicit two-step process -- the first
+step should happen as an immediate result of the xDS update, and the second step
+can happen when clusters are no longer referenced by anything.  E.g. if the
+Config Selector has a reference to the deleted cluster, then when there are no
+more references to the config selector, the final reference to the cluster will
+be removed, triggering the Service Config update.
+
+Note that when a cluster is deleted in xDS, if the cds balancer were to queue
```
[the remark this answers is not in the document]
dfawley: Done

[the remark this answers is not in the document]
dfawley: Correct. Fixed.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+
+When new clusters are added to xDS, they will be added to both the Config
+Selector and Service Config simultaneously.  Here, precautions must be taken to
+ensure it is not possible for the Config Selector to be used before the Load
+Balancing Policy's Picker has been updated to be aware of the new cluster.  The
+channel must update the Load Balancing Policy before applying the Config
+Selector, and the Load Balancing Policy must provide a new Picker before its
+Update method returns back to the channel. If the Picker does not recognize a
+cluster name provided to it, which should not be possible when following the
+steps above in that order, it will drop the RPC with an INTERNAL status code.
```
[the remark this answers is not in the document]
dfawley: This is a case that should never happen, so INTERNAL seems appropriate to me.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+
+The following drawing shows which parts of the system will contain which current
+Load Balancing Policies.  The blue box represents the Config Selector and the
+red box represents the top-level Load Balancing Policy and its children.
+
+![xDS Architecture](A31_graphics/xds_architecture.png "xDS Architecture")
+
+Initially, only timeout-related `RouteAction` settings will be supported:
+<code>[timeout](http://google3/third_party/envoy/src/api/envoy/api/v2/route/route_components.proto?l=886&rcl=309630292)</code>
```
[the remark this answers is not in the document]
dfawley: Oops, thanks.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+and
+<code>[max_grpc_timeout](http://google3/third_party/envoy/src/api/envoy/api/v2/route/route_components.proto?l=979&rcl=309630292)</code>. <code>[grpc_timeout_offset](http://google3/third_party/envoy/src/api/envoy/api/v2/route/route_components.proto?l=988&rcl=309630292)</code>
+is used for shortening timeouts set by clients to ensure the timeout is reached
```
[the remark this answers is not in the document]
dfawley: Done

A31-xds-routeaction-support.md
dfawley: Table added.

A31-xds-routeaction-support.md
dfawley: Done.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,184 @@
...
+Selector, and the Load Balancing Policy must provide a new Picker before its
+Update method returns back to the channel. If the Picker does not recognize a
+cluster name provided to it, which should not be possible when following the
+steps above in that order, it will drop the RPC with an INTERNAL status code.
+
+The following drawing shows which parts of the system will contain which current
+Load Balancing Policies.  The blue box represents the Config Selector and the
+red box represents the top-level Load Balancing Policy and its children.
+
+![xDS Architecture](A31_graphics/xds_architecture.png "xDS Architecture")
```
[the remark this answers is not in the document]
dfawley: I'll take a stab at this after the rest of the doc is settled.

A31-xds-routeaction-support.md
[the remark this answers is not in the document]
markdroth: I think we should make this a little more generic, in case we want to use this with other ConfigSelectors in the future.  How about this:

```proto
message XdsClusterManagerConfig {
  message Child {
    repeated LoadBalancingConfig child_policy = 1;
  }
  map<string, Child> children = 1;
}
```

Note that this matches the structure we're currently using for the xds_routing policy, so this will allow us to reuse the existing code with minimal changes.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,206 @@
...
+
+Initially, only timeout-related `RouteAction` settings will be supported:
```
markdroth, line 154: Can we give this its own section header and provide a separate paragraph for each of the three options we're discussing?  I think that might make it easier for readers to see how we're going to handle each field.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,206 @@
...
+<code>[timeout](https://github.com/envoyproxy/envoy/blob/7abb0e0bbed4f6b6304403b93762614ad385f80d/api/envoy/api/v2/route/route_components.proto#L886)</code>
+and
+<code>[max_grpc_timeout](https://github.com/envoyproxy/envoy/blob/7abb0e0bbed4f6b6304403b93762614ad385f80d/api/envoy/api/v2/route/route_components.proto#L979)</code>. <code>[grpc_timeout_offset](https://github.com/envoyproxy/envoy/blob/7abb0e0bbed4f6b6304403b93762614ad385f80d/api/envoy/api/v2/route/route_components.proto#L988)</code>
```
markdroth, line 157: We should probably explicitly say something like "we will completely ignore this field" for `grpc_timeout_offset`, to make the behavior clear to readers.

A31-xds-routeaction-support.md
```diff
@@ -151,19 +151,27 @@ red box represents the top-level Load Balancing Policy and its children.
...
+timeout is reached before the gRPC client cancels the RPC due to its deadline,
+so supporting it directly in the client is unnecessary and would be
+counter-productive.
+
+If `max_grpc_timeout` is not present, the `timeout` field will specify the
+maximum RPC timeout for this route, with a default of 15s.
+
+If `max_grpc_timeout` is present, then `timeout` is ignored and
+`max_grpc_timeout` limits the maximum timeout for RPCs on this route.  A value
+of 0 indicates no limit should be applied.  In all cases, the RPC timeout set by
```
markdroth, line 172: These last two sentences should probably be in their own paragraph.

A31-xds-routeaction-support.md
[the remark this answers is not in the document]
markdroth: I think it's fine to have an xDS-specific name here even though the functionality is potentially generic.  That's essentially what we wound up doing with the xds_routing policy in the traffic splitting design.

The policy name should have an "_experimental" suffix, so we can change it later if we need to.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,214 @@
...
+    Balancing Policy's Pick method and before the end of the RPC.  This is
+    necessary for cluster lifecycle management, which will be explained in
+    detail later.
+1.  Information which will be passed to the Load Balancing Policy's Pick method
+    by the channel.  E.g. in Go this could be done via the Context, and Java may
+    use CallOptions.  If an interceptor is used for (2) it is possible that the
+    interceptor may be capable of satisfying this requirement directly.
+1.  An RPC status; if not OK, the RPC is terminated with this status.  This is
+    needed in case the RPC cannot be routed.  Like (3), this could be satisfied
+    via an interceptor.
```
lidizheng, line 68: Can delay injection be supported by Config Selector for supporting fault injection? I might be wrong here. Delay injection is tightly coupled with threading model. For Golang, we could possibly block the `selectConfig` method. But for Java and Core, it seems the `getCallOptions` or `GetCallConfig` won't allow blocking or be scheduled asynchronously.

Is it possible to consider an alternative design that uses callback instead of directly return?

Or this functionality must be implemented as an interceptor? In that case, what's our plan to support interceptors in Config Selector across languages?

markdroth: I think we should discuss fault injection separately -- I don't think that's something that should be bundled with the ConfigSelector functionality.

In Java and Go, it wouldn't be that hard to have the ConfigSelector return an interceptor -- in fact, that may be how we wind up implementing it in those languages.  But there is no reasonable way to do anything like that in C-core today.

This is a problem that we are going to have to solve eventually, because there will be use cases where people will want to configure third-party interceptors via xDS.  However, we don't currently have a good solution for that problem today, and figuring out how to do it in C-core is going to be a lot of work.  I don't think we want the fault injection design to depend on that.

Let's chat more about this offline.  I suspect that in our current architecture, we may need to build support for this directly into core, just like we've (partially) done for retries.

lidizheng: Thanks for the helpful info. I will read about retry mechanism implementation next before chat with you.

A31-xds-routeaction-support.md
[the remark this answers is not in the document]
dfawley: Image updated.

A31-xds-routeaction-support.md
dfawley: Done

A31-xds-routeaction-support.md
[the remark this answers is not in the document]
dfawley: That's fair, done.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,214 @@
...
+unset | unset | unset | 15s
+unset | unset | 10s | 10s
+unset | 0s | any | infinite
+unset | 10s | any | 10s
+10s | unset | unset | 10s
+20s | unset | unset | 15s
+20s | unset | 10s | 10s
+20s | 0s | any | 20s
+20s | 10s | any | 10s
+
```
voidzcy, line 193: [Envoy's API doc](https://github.com/envoyproxy/envoy/blob/c694e470dbc9a1dd24b296d117330f0e5becb2c9/api/envoy/api/v2/route/route_components.proto#L877-L878) also mentions that a value of 0 for `timeout` will disable the route's timeout. So may worthwhile to list that case as well.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,220 @@
...
+Selector, and the Load Balancing Policy must provide a new Picker before its
+Update method returns back to the channel. If the Picker does not recognize a
+cluster name provided to it, which should not be possible when following the
+steps above in that order, it will drop the RPC with an INTERNAL status code.
+
+The following drawing shows which parts of the system will contain which current
+Load Balancing Policies.  The blue box represents the Config Selector and the
+red box represents the top-level Load Balancing Policy and its children.
+
+![xDS Architecture](A31_graphics/xds_architecture.png "xDS Architecture")
```
markdroth, line 156: Let's include the graphic in SVG format as well, in case we need to make changes to it later.

A31-xds-routeaction-support.md
markdroth: +1

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,226 @@
...
+### New Functionality in gRPC
+
+In order to allow the xDS matchers to configure gRPC behavior, but to avoid
+implementing support for them directly in gRPC, a new "per-RPC Config Selector"
+will be added for internal-only use.  This will be provided by the Name Resolver
+component when calling `Update` on the channel in a way that does not extend our
+publicly-visible API, similarly to how the xDS Client is provided to the Load
+Balancing Policy.
+
+If present, this Config Selector will be invoked once at the beginning of every
```
ejona86, line 48: We should probably mention what pieces of data are passed to the Config Selector. It is mentioned in the xDS portion below, but it seems appropriate to include here.

dfawley:
@markdroth @ejona86 Updated to account for xDS API changes; PTAL & thanks!

A31-xds-routeaction-support.md
```diff
@@ -157,42 +157,42 @@ red box represents the top-level Load Balancing Policy and its children.
 
 #### Supported Fields
 
-Initially, only timeout-related `RouteAction` settings will be supported:
-[`timeout`](https://github.com/envoyproxy/envoy/blob/7abb0e0bbed4f6b6304403b93762614ad385f80d/api/envoy/api/v2/route/route_components.proto#L886)
+Initially, only the timeout-related `RouteAction` settings of
+[`max_stream_duration.max_stream_duration`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L775)
```
markdroth, line 161: We should also document that we will support the [max_stream_duration](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/core/v3/protocol.proto#L95) field in the [HttpConnectionManager.common_http_options](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L289) field as a default if the per-route value is unset.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,226 @@
...
+message XDSClusterManagerConfig {
+    message Child {
+      repeated LoadBalancingConfig child_policy = 1;
+    }
+    map<string, Child> children = 1;
+}
+```
+
+The "xds_cluster_manager" policy will maintain a map from cluster name (in each
+`cds_config` entry) to child Policy in order to route Picks.
```
markdroth, line 111: I'm not sure what the "in each cds_config entry" part is referring to.  I think the parenthetical can just be removed.

A31-xds-routeaction-support.md
```diff
@@ -0,0 +1,226 @@
...
+ensure it is not possible for the Config Selector to be used before the Load
+Balancing Policy's Picker has been updated to be aware of the new cluster.  The
+channel must update the Load Balancing Policy before applying the Config
+Selector, and the Load Balancing Policy must provide a new Picker before its
+Update method returns back to the channel. If the Picker does not recognize a
+cluster name provided to it, which should not be possible when following the
+steps above in that order, it will drop the RPC with an INTERNAL status code.
+
+The following drawing shows which parts of the system will contain which current
+Load Balancing Policies.  The blue box represents the Config Selector and the
```
markdroth, line 153: I think the comment about blue and red boxes is stale.

markdroth (commented):
This looks great!  Just a few minor comments.

A31-xds-routeaction-support.md
dfawley: I believe that is legacy from an older version of the design.

A31-xds-routeaction-support.md
dfawley: Removed. It's now self-explanatory in the diagram.

A31-xds-routeaction-support.md
dfawley: Done

dfawley (commented):
Apparently my last round of comments (3) did not send yesterday.  So this update contains both sets; sorry for the confusion.

ALSO: I have updated the title of the gRFC as I believe we agreed upon previously.  I'll rename the file (and PR title) to match before submitting, but I don't want to do that yet for the sake of the diffs and your email threading.

A31-xds-routeaction-support.md
```diff
@@ -150,41 +150,45 @@ cluster name provided to it, which should not be possible when following the
...
-[`max_stream_duration.max_stream_duration`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L775)
-and
-[`max_stream_duration.grpc_max_timeout`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L781)
-will be supported.  gRPC will not support
-[`timeout`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L952)
-and
-[`max_stream_duration.grpc_timeout_offset`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L789).
+Only some xDS timeout-related settings will be supported:
+
+- [`RouteActions.max_stream_duration.max_stream_duration`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L775)
```
markdroth, line 161: s/RouteActions/RouteAction/

Same thing throughout.

A31-xds-routeaction-support.md
```diff
@@ -150,41 +150,45 @@ cluster name provided to it, which should not be possible when following the
...
 
-Initially, only the timeout-related `RouteAction` settings of
-[`max_stream_duration.max_stream_duration`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L775)
-and
-[`max_stream_duration.grpc_max_timeout`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L781)
-will be supported.  gRPC will not support
-[`timeout`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L952)
-and
-[`max_stream_duration.grpc_timeout_offset`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L789).
+Only some xDS timeout-related settings will be supported:
```
markdroth, line 159: It might be a good idea to start this section by explaining that Envoy has two separate and indepenent timeouts, one that starts from the client's half-close (controlled by the `timeout` field) and another that starts from the start of the stream (controlled by the `max_stream_duration` field).  The latter matches gRPC timeout semantics, so that's the one we're supporting.

A31-xds-routeaction-support.md
```diff
@@ -150,41 +150,45 @@ cluster name provided to it, which should not be possible when following the
...
+
+- [`RouteActions.max_stream_duration.max_stream_duration`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L775)
+- [`RouteActions.max_stream_duration.grpc_max_timeout`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/route/v3/route_components.proto#L781)
+- [`max_stream_duration`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/config/core/v3/protocol.proto#L95) from the [`HttpConnectionManager.common_http_options`](https://github.com/envoyproxy/envoy/blob/6f2ad057f8655e2688a195337f7520dbd77015fa/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L289) as a default if `RouteActions.max_stream_duration.max_stream_duration` is unset.
+
+Notably, gRPC will not support
```
markdroth, line 165: How about making this a bulleted list, just like the one above where you list the fields that we *do* support?  That way, we can discuss each of these fields individually.

A31-xds-routeaction-support.md
```diff
@@ -150,41 +150,45 @@ cluster name provided to it, which should not be possible when following the
...
-table:
+If `RouteActions.max_stream_duration.grpc_max_timeout` is not set, the
+`RouteActions.max_stream_duration.max_stream_duration` field will specify the
+maximum RPC timeout for this route (or fall-back to the HTTP Connection Manager
+setting), with a default of unlimited.
+
+If `RouteActions.max_stream_duration.grpc_max_timeout` is present, then
+`RouteActions.max_stream_duration.max_stream_duration` is ignored and
+`RouteActions.max_stream_duration.grpc_max_timeout` limits the maximum timeout
+for RPCs on this route instead.  A value of 0 indicates the application's
```
markdroth, line 184: I guess this means that setting *either* `max_stream_duration` to 0 or `grpc_max_timeout` to 0 will effectively not specify any default timeout.

Should we include rows in the table below for `max_stream_duration` set to 0?

markdroth (commented):
Just a few more minor comments.

A31-xds-routeaction-support.md
dfawley: 🤦 Done

A31-xds-routeaction-support.md
dfawley: Done

A31-xds-routeaction-support.md
dfawley: Done.

dfawley (commented):
All comments addressed, thanks!

A31-xds-routeaction-support.md
dfawley: Missed this somehow.

I'm actually not sure how `max_stream_duration` = 0 would/should behave.  Theoretically it should timeout instantly since there is no need for "0" to represent "infinite" AFAICT.

dfawley: Follow-up: looks like "0" disables the timer.

markdroth: Ah, you're right -- I was thinking it was an int, so unset was the same as 0.  But the field is actually a `google.protobuf.Duration`, so I think you're right.

markdroth (approved):
This looks great!

A31-xds-routeaction-support.md
markdroth: Oh, looks like you changed it to say that 0 is unlimited.  Please check what Envoy does -- we should be consistent with it here.

dfawley:
Thanks for the review!

PR #198, opened 2020-07-17
L74: Java Channel Credentials

ejona86:
Take special note of how ManagedChannelBuilders are created.

### DIFF
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -516,6 +516,15 @@ are set. Keeping track of which features were available at each version can be
 error-prone. It's also hard to produce a user-meaningful error message, so they
 are aware of which specific configuration is causing the incompatibility.
 
+It would seems to be very convenient to be able to share the
+TlsChannelCredential API on server-side, as it is quite a wide API and _almost_
+the same. However, on deeper inspection only the trust and key management is
+likely to be the same. This is partly because not _all_ the TLS configuration is
+expected to be exposed, and instead advanced use cases would use
+transport-specific configuration or potentially "all encompassing" configuration
+like providing a pre-configured SSLContext. Using separate APIs makes the APIs
+more clear and would have limited cost.
+
 ## Implementation
 
 ejona86 will implement. TLS will initially be bare-bones, with just the
@@ -525,13 +534,4 @@ ServerCredentials.
 
 ## Open issues (if applicable)
 
-It would be very convenient to be able to share the TlsChannelCredential API on
-server-side, as it is quite a wide API and _almost_ the same. The consumer
-would need to be slightly different, but much of the configuration can
-potentially be shared. Should we make TlsChannelCredential and
-TlsServerCredential just a holder for a TlsConfiguration class? For validation,
-it would need to know whether it was for server-side or client-side. It would
-also have the union of all client- and server-specific configuration. See
-[Netty's SslContextBuilder](
-https://netty.io/4.1/api/io/netty/handler/ssl/SslContextBuilder.html)
-for a view into how client and server differ.
+N/A
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -108,9 +108,9 @@ package io.grpc;
  * possible configurations for types that are partially supported, but they
  * <em>must</em> at least fully support ChoiceChannelCredentials.
  *
- * <p>A Credential provides client identity and authenticates the server. This
- * is different from CallCredentials, which only provides client identity. They
- * can also influence types of encryption used and similar security
+ * <p>A ChannelCredential provides client identity and authenticates the server.
+ * This is different from CallCredentials, which only provides client identity.
+ * They can also influence types of encryption used and similar security
  * configuration.
  */
 public abstract class ChannelCredentials {}
@@ -123,7 +123,7 @@ public final class TlsChannelCredentials extends ChannelCredentials {
 public final class InsecureChannelCredentials extends ChannelCredentials {}
 
 /**
-  * Provides a list of ChoiceChannelCredentials, where any one may be used. The
+  * Provides a list of ChannelCredentials, where any one may be used. The
   * credentials are in preference order.
   */
 public final class ChoiceChannelCredentials extends ChannelCredentials {
@@ -195,7 +195,7 @@ public final class TlsChannelCredentials extends ChannelCredentials {
 
   public enum Feature {}
 
-  public final static class Builder {
+  public static final class Builder {
     public TlsChannelCredentials build() {...}
   }
 }
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -2,9 +2,9 @@ Java: Channel and Server Credentials
 ----
 * Author(s): [Eric Anderson](https://github.com/ejona86)
 * Approver: sanjaypujare
-* Status: In Review {Draft, In Review, Ready for Implementation, Implemented}
+* Status: Ready for Implementation {Draft, In Review, Ready for Implementation, Implemented}
 * Implemented in: <language, ...>
-* Last updated: 2020-06-17
+* Last updated: 2020-08-27
 * Discussion at: https://groups.google.com/g/grpc-io/c/GNU9svnDsQs/m/kG6XsUy9BwAJ
 
 ## Abstract
@@ -561,11 +561,6 @@ more clear and would have limited cost.
 
 ## Implementation
 
-ejona86 will implement. TLS will initially be bare-bones, with just the
-compatibility checker. Over the long-term TLS will be fleshed out with
-additional configuration. ServerCredentials will be implemented shortly after
-ChannelCredentials.
-
-## Open issues (if applicable)
-
-N/A
+ejona86 will implement. ServerCredentials will be implemented shortly after
+ChannelCredentials. ChannelCredentials will be implemented in
+https://github.com/grpc/grpc-java/pull/7294
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -5,7 +5,7 @@ Java: Channel and Server Credentials
 * Status: In Review {Draft, In Review, Ready for Implementation, Implemented}
 * Implemented in: <language, ...>
 * Last updated: 2020-06-17
-* Discussion at: <google group thread> (filled after thread exists)
+* Discussion at: https://groups.google.com/g/grpc-io/c/GNU9svnDsQs/m/kG6XsUy9BwAJ
 
 ## Abstract
 
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -381,6 +381,7 @@ public final class TlsServerCredentials extends ServerCredentials {
    */
   public TlsServerCredentials(File certChain, File privateKey)
       throws IOException {...}
+  // Will not auto-close InputStream, unlike existing transportSecurity()
   public TlsServerCredentials(InputStream certChain, InputStream privateKey)
       throws IOException {...}
 
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -201,6 +201,13 @@ public final class TlsChannelCredentials extends ChannelCredentials {
 }
 ```
 
+An example `Feature` could be `CLIENT_CERTIFICATE`. When the `Feature` is
+added, methods `getCertificateChain()`, `getPrivateKey()`, `getPassword()` can
+be added. Observing the contents of those methods would be a requirement for
+`CLIENT_CERTIFICATE`. An implementation understanding `CLIENT_CERTIFICATE`
+might not support encrypted private keys and so could consider the credential
+unsupported if `getPassword()` returned non-`null`.
+
 To support very advanced use cases, Netty will provide a credential that wraps
 a `ProtocolNegotiator`. This allows implementations like ALTS and XDS to use
 internal APIs without forcing their users to use experimental or internal APIs,
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -239,6 +239,45 @@ be added. Observing the contents of those methods would be a requirement for
 might not support encrypted private keys and so could consider the credential
 unsupported if `getPassword()` returned non-`null`.
 
+To support migration from existing builder options, Netty and OkHttp will
+provide transport-specific credentials. Note that neither transport will support
+reading the configuration after creation, except by the transport.
+
+```java
+package io.grpc.netty;
+
+/** A credential with full control over the security handshake. */
+public final class NettySslContextChannelCredentials {
+  private NettySslContextChannelCredentials() {}
+
+  /**
+   * Create a credential using Netty's SslContext as configuration. It must have
+   * been configured with {@link GrpcSslContexts}, but options could have been
+   * overridden.
+   */
+  public static ChannelCredentials create(SslContext sslContext) {...}
+}
+
+/** An insecure credential that upgrades from HTTP/1 to HTTP/2. */
+public final class InsecureFromHttp1ChannelCredentials {
+  private InsecureFromHttp1ChannelCredentials() {}
+
+  /** Creates an insecure credential that will upgrade from HTTP/1 to HTTP/2. */
+  public static ChannelCredentials create() {...}
+}
+```
+
+```java
+package io.grpc.okhttp;
+
+/** A credential with full control over the SSLSocketFactory. */
+public final class SslSocketFactoryChannelCredentials {
+  private SslSocketFactoryChannelCredentials() {}
+
+  public static ChannelCredentials create(SSLSocketFactory factory) {...}
+}
+```
+
 To support very advanced use cases, Netty will provide a credential that wraps
 a `ProtocolNegotiator`. This allows implementations like ALTS and XDS to use
 internal APIs without forcing their users to use experimental or internal APIs,
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -342,11 +342,76 @@ Server-side will be a clone of client-side, but using the `Server` instead of
 `CompositeServerCredentials` will be omitted. There is also currently no known
 use-case for `ChoiceServerCredentials`, so it will be omitted.
 
+```java
+// instead of:
+server = ServerBuilder.forPort(443)
+    .useTransportSecurity(new File("cert.pem"), new File("cert.key"))
+    .build().start();
+server = ServerBuilder.forPort(8080)
+    .build().start();
+// the user would now:
+ChannelCredentials tlsCreds = new TlsServerCredentials(
+    new File("cert.pem"), new File("cert.key"));
+server = Grpc.newServerBuilderForPort(443, tlsCreds)
+    .build().start();
+server = Grpc.newServerBuilderForPort(8080, new InsecureChannelCredentials())
+    .build().start();
+```
+
 
 ```java
 package io.grpc;
 
 public abstract class ServerCredentials {}
+```
+
+Since servers using TLS require a certificate, the `TlsServerCredentials`
+cannot have the simple no-arg configuration of `TlsChannelCredentials`. It will
+provide convenience constructors for providing a certificate and unencrypted
+key. If any other configuration is necessary, the `Builder` would be used.
+
+```java
+package io.grpc;
+
+public final class TlsServerCredentials extends ServerCredentials {
+  /**
+   * Creates an instance using provided certificate chain and private key.
+   * Generally they should be PEM-encoded and the key is an unencrypted PKCS#8
+   * key (file headers have "BEGIN CERTIFICATE" and "BEGIN PRIVATE KEY").
+   */
+  public TlsServerCredentials(File certChain, File privateKey)
+      throws IOException {...}
+  public TlsServerCredentials(InputStream certChain, InputStream privateKey)
+      throws IOException {...}
+
+  /** Same as client-side */
+  public Set<Feature> incomprehensible(EnumSet<Feature> understoodFeatures) {...}
+
+  public byte[] getCertificateChain() { return Arrays.copyOf(...); }
+  public byte[] getPrivateKey() { return Arrays.copyOf(...); }
+  public String getPassword() {...}
+
+  /* Other methods can be added to support various Features */
+
+  public enum Feature {}
+
+  public static Builder newBuilder() { return new Builder(); }
+
+  public final static class Builder {
+    public Builder keyManager(File certChain, File privateKey)
+        throws IOException {...}
+    public Builder keyManager(File certChain, File privateKey, String password)
+        throws IOException {...}
+    public Builder keyManager(InputStream certChain, InputStream privateKey)
+        throws IOException {...}
+    public Builder keyManager(
+        InputStream certChain, InputStream privateKey, String password)
+        throws IOException {...}
+
+    // Throws if no certificate was provided
+    public TlsChannelCredentials build() {...}
+  }
+}
 ```
 
 ## Rationale
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -394,7 +394,7 @@ server = ServerBuilder.forPort(443)
 server = ServerBuilder.forPort(8080)
     .build().start();
 // the user would now:
-ChannelCredentials tlsCreds = new TlsServerCredentials(
+ServerCredentials tlsCreds = new TlsServerCredentials(
     new File("cert.pem"), new File("cert.key"));
 server = Grpc.newServerBuilderForPort(443, tlsCreds)
     .build().start();
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -95,7 +95,7 @@ Note that this new API does not allow: 1) creating a channel without a
 credential and supplying the credential later and 2) changing the credential
 after creating the builder. This is not expected to be much of an issue as
 users typically have the security information available before creating the
-builder and reusing a builder is rare, especially to change the security. 
+builder and reusing a builder is rare, especially to change the security.
 
 ```java
 package io.grpc;
@@ -112,22 +112,42 @@ package io.grpc;
  * This is different from CallCredentials, which only provides client identity.
  * They can also influence types of encryption used and similar security
  * configuration.
+ *
+ * The concrete credential type should not be relevant to most users of the API
+ * and may be an implementation decision. Users should generally use the {@code
+ * ChannelCredentials} type for variables instead of the concrete type.
+ * Freshly-constructed credentials should be returned as {@code
+ * ChannelCredentials} instead of a concrete type to encourage this pattern.
+ * Concrete types would only be used after {@code instanceof} checks (which must
+ * consider {@code ChoiceChannelCredentials}!).
  */
 public abstract class ChannelCredentials {}
 
 public final class TlsChannelCredentials extends ChannelCredentials {
+  public static ChannelCredentials create() {...}
+
   /* Complicated. Described separately below. */
 }
 
 /** No client identity, authentication, or encryption is to be used. */
-public final class InsecureChannelCredentials extends ChannelCredentials {}
+public final class InsecureChannelCredentials extends ChannelCredentials {
+  private InsecureChannelCredentials() {}
+
+  public ChannelCredentials create() {
+    return new InsecureChannelCredentials();
+  }
+}
 
 /**
   * Provides a list of ChannelCredentials, where any one may be used. The
   * credentials are in preference order.
   */
 public final class ChoiceChannelCredentials extends ChannelCredentials {
-  public ChoiceChannelCredentials(ChannelCredentials... creds) {...}
+  private ChoiceChannelCredentials(ChannelCredentials... creds) {...}
+
+  public static ChannelCredentials create(ChannelCredentials... creds) {
+    return new ChoiceChannelCredentials(creds);
+  }
 
   public List<ChannelCredentials> getCredentialsList() { return creds; }
 }
@@ -146,8 +166,12 @@ the behavior in other languages.
 package io.grpc;
 
 public final class CompositeChannelCredentials extends ChannelCredentials {
-  public CompositeChannelCredentials(
+  private CompositeChannelCredentials(
     ChannelCredentials chanCreds, CallCredentials callCreds) {...}
+  public static ChannelCredentials create(
+      ChannelCredentials chanCreds, CallCredentials callCreds) {
+    return new CompositeChannelCredentials(chanCreds, callCreds);
+  }
   public ChannelCredentials getChannelCredentials() { return channelCredentials; }
   public CallCredentials getCallCredentials() { return callCredentials; }
 }
@@ -169,6 +193,13 @@ be used.
 package io.grpc;
 
 public final class TlsChannelCredentials extends ChannelCredentials {
+  private TlsChannelCredentials(Builder builder) {...}
+
+  /** Use TLS with its defaults. */
+  public static ChannelCredentials create() { return newBuilder().build(); }
+
+  public static Builder newBuilder() { return new Builder(); }
+
   /**
    * Returns an empty set if this credential can be adequately understood via
    * the features listed, otherwise returns a hint of features that are lacking
@@ -200,7 +231,7 @@ public final class TlsChannelCredentials extends ChannelCredentials {
   public enum Feature {}
 
   public static final class Builder {
-    public TlsChannelCredentials build() {...}
+    public ChannelCredentials build() {...}
   }
 }
 ```
@@ -221,16 +252,23 @@ as their users would just interact with `ChannelCredentials`.
 package io.grpc.netty;
 
 final class NettyChannelCredentials extends ChannelCredentials {
-  public NettyChannelCredentials(ProtocolNegotiator.Factory negotiator) {...}
+  private NettyChannelCredentials(ProtocolNegotiator.Factory negotiator) {...}
+
+  public static ChannelCredentials create(ProtocolNegotiator.Factory negotiator) {
+    return new NettyChannelCredentials(negotiator);
+  }
+
   public ProtocolNegotiator.Factory getNegotiator() { return negotiator; }
 }
 
 @Internal
 public final class InternalNettyChannelCredentials {
-  private InternalNettyChannelCredentials() {}
+  private InternalNettyChannelCredentials() {} // prevent instantiation
 
-  public ChannelCredentials create(InternalProtocolNegotiator.Factory negotiator)
-    {...}
+  public static ChannelCredentials create(
+      InternalProtocolNegotiator.Factory negotiator) {
+    return NettyChannelCredentials.create(negotiator);
+  }
 
   /**
    * Converts credentials to a negotiator, in similar fashion as for a new
@@ -238,8 +276,8 @@ public final class InternalNettyChannelCredentials {
    *
    * @throws IllegalArgumentException if unable to convert
    */
-  public InternalProtocolNegotiator.Factory toNegotiator(ChannelCredentials creds)
-    {...}
+  public static InternalProtocolNegotiator.Factory toNegotiator(
+    ChannelCredentials creds) {...}
 }
 ```
 
@@ -250,7 +288,7 @@ demonstration.
 
 ```java
 public final class GoogleDefaultChannelCredentials {
-  private GoogleDefaultChannelCredentials() {}
+  private GoogleDefaultChannelCredentials() {} // prevent instantiation
 
   public static ChannelCredentials create() {
     return new ChoiceChannelCredentials(
@@ -270,7 +308,7 @@ demonstration.
 
 ```java
 public final class XdsChannelCredentials {
-  private XdsChannelCredentials() {}
+  private XdsChannelCredentials() {} // prevent instantiation
 
   /**
    * Creates credentials to be configured by xDS, falling back to other
@@ -338,7 +376,7 @@ final class ProtocolNegotiators {
     }
   }
 
-  public final class Result {
+  public static final class Result {
     public static Result error(String error) {...}
     public static Result negotiator(ProtocolNegotiator.Factory factory) {...}
     public Result withCallCredentials(CallCredentials callCreds) {...}
@@ -385,16 +423,24 @@ key. If any other configuration is necessary, the `Builder` would be used.
 package io.grpc;
 
 public final class TlsServerCredentials extends ServerCredentials {
+  private TlsServerCredentials(...) {}
+
   /**
    * Creates an instance using provided certificate chain and private key.
    * Generally they should be PEM-encoded and the key is an unencrypted PKCS#8
    * key (file headers have "BEGIN CERTIFICATE" and "BEGIN PRIVATE KEY").
    */
-  public TlsServerCredentials(File certChain, File privateKey)
-      throws IOException {...}
+  public static ServerCredentials create(File certChain, File privateKey)
+      throws IOException {
+    return newBuilder().keyManager(certChain, privateKey).build();
+  }
   // Will not auto-close InputStream, unlike existing transportSecurity()
-  public TlsServerCredentials(InputStream certChain, InputStream privateKey)
-      throws IOException {...}
+  public static ServerCredentials create(
+      InputStream certChain, InputStream privateKey) throws IOException {
+    return newBuilder().keyManager(certChain, privateKey).build();
+  }
+
+  public static Builder newBuilder() { return new Builder(); }
 
   /** Same as client-side */
   public Set<Feature> incomprehensible(Set<Feature> understoodFeatures) {...}
@@ -406,8 +452,6 @@ public final class TlsServerCredentials extends ServerCredentials {
   /* Other methods can be added to support various Features */
 
   public enum Feature {}
-
-  public static Builder newBuilder() { return new Builder(); }
 
   public final static class Builder {
     public Builder keyManager(File certChain, File privateKey)
@@ -421,7 +465,7 @@ public final class TlsServerCredentials extends ServerCredentials {
         throws IOException {...}
 
     // Throws if no certificate was provided
-    public TlsChannelCredentials build() {...}
+    public ChannelCredentials build() {...}
   }
 }
 ```
@@ -452,6 +496,10 @@ To support transport credentials like ALTS and TLS-offloading, there still
 needs to be a way to provide implementations. However, these implementations
 _can_ depend on the specific transport being used. Just the users of those
 implementations should avoid depending _explicitly_ on a particular transport.
+Such "credentials" will not extend ChannelCredentials but instead create
+pre-existing credentials types with specific configuration. To reduce the
+strangeness involved, all ChannelCredentials construction returns the generic
+ChannelCredentials type instead of the concrete type.
 
 The API surface should handle generic and transport-specific credentials in the
 same manner, so the user’s code should not care about the implementation
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -23,11 +23,6 @@ This is to allow:
     credential needs to be provided by the user
 *   aligning with C core and Go implementations
 
-_The design focuses on channel credentials, since they are the most complex and
-most discussed._ Server credentials will mostly just mirror the channel
-credentials; they suffer from most of the same problems but at a smaller
-scale.
-
 ## Background
 
 Prior to gRPC v0.12.0, C core had a "unified" `Credential` type that was
@@ -54,7 +49,7 @@ credentials (`credentials.TransportCredentials`), but with a separate type to
 combine them (`credentials.Bundle`). The split credentials significantly
 predate `Bundle`.
 
-### Related Proposals: 
+### Related Proposals:
 
 N/A
 
--- a/L74-java-channel-creds.md
+++ b/L74-java-channel-creds.md
@@ -65,21 +65,27 @@ constructing a `ManagedChannelBuilder':
 
 ```java
 // instead of:
-ManagedChannelBuilder.forTarget("example.com")
-                     .build();
-ManagedChannelBuilder.forTarget("example.com").usePlaintext()
-                     .build();
+channel = ManagedChannelBuilder.forTarget("example.com")
+    .build();
+channel = ManagedChannelBuilder.forTarget("example.com").usePlaintext()
+    .build();
+channel = ManagedChannelBuilder.forAddress("example.com", 443)
+    .build();
 // the user would now:
-ManagedChannel.newBuilderForTarget("example.com", new TlsChannelCredentials())
-                     .build();
-ManagedChannel.newBuilderForTarget("example.com", new InsecureChannelCredentials())
-                     .build();
+channel = Grpc.newChannelBuilder("example.com", new TlsChannelCredentials())
+    .build();
+channel = Grpc.newChannelBuilder("example.com", new InsecureChannelCredentials())
+    .build();
+channel = Grpc.newChannelBuilderForAddress("example.com", 443, new TlsChannelCredentials())
+    .build();
 ```
 
-`newBuilderForTarget()` will iterate through the `ManagedChannelProviders`
+`newChannelBuilder()` will iterate through the `ManagedChannelProviders`
 trying to create a builder for each in turn. The first provider that succeeds
 will be used. If no provider is able to handle the credentials, then
-`newBuilderForTarget()` will throw.
+`newChannelBuilder()` will throw. `newChannelBuilderForAddress()` will be a
+convenience function that creates the target string and calls
+`newChannelBuilder()`, mirroring its current behavior.
 
 The new ChannelCredentials API cannot be mixed with the pre-existing
 `useTransportSecurity()` and `usePlaintext()` builder APIs. The old-style
@@ -393,11 +399,6 @@ additional configuration. ServerCredentials will be implemented shortly after
 ServerCredentials.
 
 ## Open issues (if applicable)
-
-Do we put the constructor in ManagedChannel instead of ManagedChannelBuilder?
-On io.grpc.Grpc? The static methods on ManagedChannelBuilder have been a
-problem, as they are exposed on all subclasses, but don’t operate within the
-scope of the subclass, unless they do.
 
 It would be very convenient to be able to share the TlsChannelCredential API on
 server-side, as it is quite a wide API and _almost_ the same. The consumer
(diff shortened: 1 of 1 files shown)


L74-java-channel-creds.md
```diff
@@ -0,0 +1,411 @@
...
+constructing a `ManagedChannelBuilder':
+
+```java
+// instead of:
+ManagedChannelBuilder.forTarget("example.com")
+                     .build();
+ManagedChannelBuilder.forTarget("example.com").usePlaintext()
+                     .build();
+// the user would now:
+ManagedChannel.newBuilderForTarget("example.com", new TlsChannelCredentials())
```
ejona86, line 73: C++ just calls ChannelCredentials "Credentials" in their name, except where they don't: `InsecureChannelCredentials`, `SslCredentials`, `GoogleDefaultCredentials`, `CompositeChannelCredentials`. ServerCredentials are more consistent, but there's few of them: `InsecureServerCredentials`, `SslServerCredentials`.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,411 @@
...
+ServerCredentials.
+
+## Open issues (if applicable)
+
+Do we put the constructor in ManagedChannel instead of ManagedChannelBuilder?
+On io.grpc.Grpc? The static methods on ManagedChannelBuilder have been a
+problem, as they are exposed on all subclasses, but don’t operate within the
+scope of the subclass, unless they do.
+
+It would be very convenient to be able to share the TlsChannelCredential API on
```
ejona86, line 402: The C++ API has very few of the full range of options defined:
https://grpc.github.io/grpc/cpp/structgrpc__impl_1_1_ssl_credentials_options.html
https://grpc.github.io/grpc/cpp/structgrpc_1_1_ssl_server_credentials_options.html

I think it would be possible to keep our configuration as terse for the most part as well, still using transport-specific APIs for the lots-of-options case. It should also be noticed that C++ allows passing many server certs, such that the only configuration in common between client and server is `pem_root_certs`.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,411 @@
...
+## Implementation
+
+ejona86 will implement. TLS will initially be bare-bones, with just the
+compatibility checker. Over the long-term TLS will be fleshed out with
+additional configuration. ServerCredentials will be implemented shortly after
+ServerCredentials.
+
+## Open issues (if applicable)
+
+Do we put the constructor in ManagedChannel instead of ManagedChannelBuilder?
```
ejona86, line 397: API review notes about this API:
 * io.grpc.Grpc was the preferred place for the builder-creation method, to avoid the current ManagedChannelBuilder issues and avoid the potential for the same issue cropping up with ManagedChannel
 * We didn't get to a point of really choosing a name for the method though. Grpc.newManagedChannelBuilderForTarget is pretty obnoxious. There was a desire expressed to keep "for target" in the name, as we will have "for address" as well. "Managed" may be easy to remove.

jiangtaoli2016:
@ZhenLian FYI

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+channel = ManagedChannelBuilder.forTarget("example.com").usePlaintext()
+    .build();
+channel = ManagedChannelBuilder.forAddress("example.com", 443)
+    .build();
+// the user would now:
+channel = Grpc.newChannelBuilder("example.com", new TlsChannelCredentials())
+    .build();
+channel = Grpc.newChannelBuilder("example.com", new InsecureChannelCredentials())
+    .build();
+channel = Grpc.newChannelBuilderForAddress("example.com", 443, new TlsChannelCredentials())
```
carl-mastrangelo, line 79: In the future, are you expecting the server builder to be constructed in the same manner?  You mentioned there were channel cred similarities above, which is why I ask

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+
+`newChannelBuilder()` will iterate through the `ManagedChannelProviders`
+trying to create a builder for each in turn. The first provider that succeeds
+will be used. If no provider is able to handle the credentials, then
+`newChannelBuilder()` will throw. `newChannelBuilderForAddress()` will be a
+convenience function that creates the target string and calls
+`newChannelBuilder()`, mirroring its current behavior.
+
+The new ChannelCredentials API cannot be mixed with the pre-existing
+`useTransportSecurity()` and `usePlaintext()` builder APIs. The old-style
```
carl-mastrangelo, line 91: Consider at some point deprecating them when this proposal is merged.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+public final class InsecureChannelCredentials extends ChannelCredentials {}
+
+/**
+  * Provides a list of ChoiceChannelCredentials, where any one may be used. The
+  * credentials are in preference order.
+  */
+public final class ChoiceChannelCredentials extends ChannelCredentials {
+  public ChoiceChannelCredentials(ChannelCredentials... creds) {...}
+
+  public List<ChannelCredentials> getCredentialsList() { return creds; }
```
carl-mastrangelo, line 132: What will happen if no credential ends up being applicable?

L74-java-channel-creds.md
ejona86: I actually have some snippets with what it will look like at the bottom `Grpc.newServerBuilderForPort()`. I didn't provide reasoning, but I purposefully did not use `Grpc.newServerBuilder()` as I want to keep that open for future expansion (I still have the long-term goal of allowing adding/removing listeners on a started server, in which case no port would be required when building the server).

TlsServerCredentials were annoying to figure out how I wanted them to look. We still have the "Java can't easily parse PKCS#8", so I wanted to keep private key parsing code in the transports and not in grpc-api.

L74-java-channel-creds.md
ejona86: Yes, plan to do so, eventually. But it will be slow-moving. There are a _lot_ of usages of this API, so I want to make sure there is some natural adoption and see that things are working well before strongly encouraging a migration.

L74-java-channel-creds.md
ejona86: Then the transport doesn't understand the credential. It would go into the normal flow of allowing other transports a try and if they all fail producing an exception.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+final class NettyChannelCredentials extends ChannelCredentials {
+  public NettyChannelCredentials(ProtocolNegotiator.Factory negotiator) {...}
+  public ProtocolNegotiator.Factory getNegotiator() { return negotiator; }
+}
+
+@Internal
+public final class InternalNettyChannelCredentials {
+  private InternalNettyChannelCredentials() {}
+
+  public ChannelCredentials create(InternalProtocolNegotiator.Factory negotiator)
```
carl-mastrangelo, line 221: not sure where you draw the boundary for credentials, but this would be a pretty big expansion.     This API is powerful to pick things like ciphersuites, TLSv1.3, session tickets, SNI support (server side), etc.  Those are all security related, but don't feel like Credentials. (e.g. keystores and truststores)

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+}
+```
+
+### GoogleDefault ChannelCredentials Example
+
+The API provided here is not part of the design. It is meant as a
+demonstration.
+
+```java
+public final class GoogleDefaultChannelCredentials {
```
carl-mastrangelo, line 241: thought: it's a little surprising to have the factory class called Credentials but not be a Credentials.

(I realize of course this is just an example).

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+```
+
+### Netty ChannelCredentials Processing Example
+
+The API provided here is not part of the design. It is meant as a
+demonstration.
+
+```java
+final class ProtocolNegotiators {
+  private static final EnumSet<> understoodTlsFeatures =
```
carl-mastrangelo, line 286: I'm having a hard time thinking of what a Feature would be.  Several of the TLS choices I have seen are more complex than an enum.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+
+```java
+// instead of:
+server = ServerBuilder.forPort(443)
+    .useTransportSecurity(new File("cert.pem"), new File("cert.key"))
+    .build().start();
+server = ServerBuilder.forPort(8080)
+    .build().start();
+// the user would now:
+ChannelCredentials tlsCreds = new TlsServerCredentials(
```
carl-mastrangelo, line 353: FWIW you would need to provide at least as many overloads as SslContextBuilder does.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+  public TlsServerCredentials(File certChain, File privateKey)
+      throws IOException {...}
+  // Will not auto-close InputStream, unlike existing transportSecurity()
+  public TlsServerCredentials(InputStream certChain, InputStream privateKey)
+      throws IOException {...}
+
+  /** Same as client-side */
+  public Set<Feature> incomprehensible(EnumSet<Feature> understoodFeatures) {...}
+
+  public byte[] getCertificateChain() { return Arrays.copyOf(...); }
```
carl-mastrangelo, line 391: I have been working on TLS proxying for the past few months and one thing I had to work around a lot was making the Server able to work with SNI.   This means picking the key/cert after most of the other config.   Having just one chain or one key would be limitting.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,478 @@
...
+are aware of which specific configuration is causing the incompatibility.
+
+## Implementation
+
+ejona86 will implement. TLS will initially be bare-bones, with just the
+compatibility checker. Over the long-term TLS will be fleshed out with
+additional configuration. ServerCredentials will be implemented shortly after
+ServerCredentials.
+
+## Open issues (if applicable)
```
carl-mastrangelo, line 467: How is the channel credential exposed to the LB Picker or the CallCredentials?

L74-java-channel-creds.md
ejona86: I was expecting users to continue providing Netty's SslContext. I will need some way to convert an SslContext to ChannelCredentials (whose concrete type is InternalNettyChannelCredentials). But yeah, there is a missing API here. Make a TlsNettyChannelCredentials container class for a static utility method? I'll have to think about it a bit.

L74-java-channel-creds.md
ejona86: To make it more concrete, one feature could be "MTLS" or "CLIENT_CERT". It would mean the caller knows how to call getCertificateChain(), getPrivateKey(), getPassword() to use the client cert. (Those methods would be added when we add the Feature.) Basically, as part of each Feature would be a specification of what it implies and the expected behavior, and there's no real limits to what could be specified.

L74-java-channel-creds.md
ejona86: I hope we can avoid that. I think this API will cater to the 80-90% case, so there is a stable API available to them. We would have some way to let people convert (wrap) a Netty SslContext to a ChannelCredential so that there would still be the same advanced-but-unstable API we have today when people call `nettyChannelBuilder.sslContext()` (ditto for server-side). We can have something equivalent for OkHttp to specify the SSLSocketFactory.

So, for example, I totally expect some stuff to support client certificates in the future, which would imply a setting for the server to {NONE, ASK, REQUIRE} client certificates and a setting to specify non-default root certificates for trust (that could be two Features, or one).

But I don't think we'll ever want to allow people to change the ciphers from this API, and we may be able to avoid letting people specify TrustManagers.

L74-java-channel-creds.md
ejona86: * For advanced use-cases we always want to allow Netty's SslContext to be available
* SNI support is actually something C core has in their SslServerCredential. We could consider supporting it as well in Java. If we did so, then there would be a new Feature that would have a Map or Function for the choosing the certificate.

L74-java-channel-creds.md
ejona86: It wouldn't be; it'd be basically the same as now. Any Attribute communication/hacks would continue to be used.

If there was need in the future (to mirror Go and C, for example), we _could_ expose them via the LB.Helper (which mirrors Go and C). I don't know of any reason we'd want to expose them to the CallCredentials; it should generally depend on the handshake results (like SSLSession) and not the specific ChannelCredential. However, they _could_ be exposed via CallCredentials.RequestInfo if necessary.

L74-java-channel-creds.md
carl-mastrangelo: One other minor thing: The SNI builder allows *dynamically* creating a SslContext based on the domain.  There may not be just one.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
+Java: Channel and Server Credentials
+----
+* Author(s): [Eric Anderson](https://github.com/ejona86)
+* Approver: sanjaypujare
+* Status: In Review {Draft, In Review, Ready for Implementation, Implemented}
+* Implemented in: <language, ...>
```
sanjaypujare, line 6: Put Java here?

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+    TLS, without polluting the general-purpose ManagedChannelBuilder API with
+    many overloads that don't apply to all transports
+*   composition of credentials, like necessary for xDS where a fallback
+    credential needs to be provided by the user
+*   aligning with C core and Go implementations
+
+_The design focuses on channel credentials, since they are the most complex and
+most discussed._ Server credentials will mostly just mirror the channel
+credentials; they suffer from most of the same problems but at a smaller
+scale.
```
sanjaypujare, line 29: There is actually a section on Server credentials but this sentence gives the impression server credentials are not covered here. How about rephrasing and moving to the server creds section?

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+channel = ManagedChannelBuilder.forTarget("example.com").usePlaintext()
+    .build();
+channel = ManagedChannelBuilder.forAddress("example.com", 443)
+    .build();
+// the user would now:
+channel = Grpc.newChannelBuilder("example.com", new TlsChannelCredentials())
+    .build();
+channel = Grpc.newChannelBuilder("example.com", new InsecureChannelCredentials())
+    .build();
+channel = Grpc.newChannelBuilderForAddress("example.com", 443, new TlsChannelCredentials())
```
sanjaypujare, line 79: I wonder if we can do away with the `ForAddress` variant in the interest of API simplification (unless there is a benefit I am not seeing).

If the caller wants to specify a port they can always use:
```
newChannelBuilder("example.com:443", new TlsChannelCredentials())
```

ejona86: We have it to deal with IPv6 addresses. IPv6 addresses have to be enclosed in `[]` since they can contain colons. `newChannelBuilderForAddress("::1", 443, creds)` → `newChannelBuilder("[::1]:443", creds)`. `forAddress(host, port)` has always been a convenience, and I'd rather not take it away since I want this API to successfully supplant the existing one.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+  * credentials are in preference order.
+  */
+public final class ChoiceChannelCredentials extends ChannelCredentials {
+  public ChoiceChannelCredentials(ChannelCredentials... creds) {...}
+
+  public List<ChannelCredentials> getCredentialsList() { return creds; }
+}
+```
+
+`ChoiceChannelCredentials` is intended to be used in cases like
```
sanjaypujare, line 136: `ChoiceChannelCredentials` don't exist in other gRPC languages (because they don't expose transports the way Java does?). Can they be made an optional part of this proposal assuming there is no absolute requirement to have them? Both `ChoiceChannelCredentials` and `CompositeChannelCredentials` allow composing and I was wondering about the complexity involved in supporting combinations e.g. first creds in `ChoiceChannelCredentials`  is a `CompositeChannelCredentials` and so on.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+  public CallCredentials getCallCredentials() { return callCredentials; }
+}
+```
+
+Unlike most of the credentials, TLS may have a considerable amount of
+configuration. And as time goes on, more configuration options may be provided.
+It could be a severe bug if certain TLS restrictions were ignored by a channel.
+To detect cases where the channel does not understand enough of the TLS
+configuration, there will be a method to check if there are required features
+that are not understood. If some are not understood, enough information will be
```
sanjaypujare, line 161: Just to confirm my understanding since I haven't seen this pattern before, the way this is going to work is as follows: 

A transport implementation such as Netty will call `incomprehensible()` with the set of `Feature`s it (Netty) supports and if the `TlsChannelCredentials` instance returns a non-empty set then there is something Netty doesn't support (that `TlsChannelCredentials` requires) and it will throw an error (with error text containing non-supported feature set).

So this pattern depends on the transport implementation verifying features thru `incomprehensible()` to prevent  a required TLS restriction/feature from being ignored by the transport.

L74-java-channel-creds.md
ejona86: It's not implemented yet, so I was leaving it as-is. I expected to update the gRFC after adding it to grpc-java to update this line. I'd also update the "Implementation" section to say which version it was added to.

L74-java-channel-creds.md
ejona86: The point of requiring it is to allow _other_ transports to define their own. It doesn't work unless everyone supports it.

If you look at the ProtocolNegotiators example for Netty, there's not combinatorics involved with the complexity. The complexity is simply the sum of the individual complexities. `CallCredentials` will actually be implemented by `ManagedChannelImpl`, so it's mainly something off-to-the-side to deal with.

L74-java-channel-creds.md
sanjaypujare: ok

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+To support very advanced use cases, Netty will provide a credential that wraps
+a `ProtocolNegotiator`. This allows implementations like ALTS and XDS to use
+internal APIs without forcing their users to use experimental or internal APIs,
+as their users would just interact with `ChannelCredentials`.
+
+```java
+package io.grpc.netty;
+
+final class NettyChannelCredentials extends ChannelCredentials {
+  public NettyChannelCredentials(ProtocolNegotiator.Factory negotiator) {...}
```
sanjaypujare, line 220: Is it [`ProtocolNegotiatorFactory`](https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyChannelBuilder.java#L524)? I didn't find any `ProtocolNegotiator.Factory`.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+  public ProtocolNegotiator.Factory getNegotiator() { return negotiator; }
+}
+
+@Internal
+public final class InternalNettyChannelCredentials {
+  private InternalNettyChannelCredentials() {}
+
+  public ChannelCredentials create(InternalProtocolNegotiator.Factory negotiator)
```
sanjaypujare, line 228: Is this actually [InternalNettyChannelBuilder.ProtocolNegotiatorFactory](https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/InternalNettyChannelBuilder.java#L43)?

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+  /**
+   * Creates credentials to be configured by xDS, falling back to other
+   * credentials if no configuration is provided.
+   *
+   * @throws IllegalArgumentException if fallback is unable to be used
+   */
+  public static ChannelCredentials createWithFallback(
+      ChannelCredentials fallback) {
+    return new InternalNettyChannelCredentials.create(
+        new XdsProtocolNegotiatorFactory(
```
sanjaypujare, line 280: Is this meant to be the existing class [SdsProtocolNegotiators.ClientSdsProtocolNegotiatorFactory](https://github.com/grpc/grpc-java/blob/master/xds/src/main/java/io/grpc/xds/internal/sds/SdsProtocolNegotiators.java#L81) - renamed for convenience? Currently it doesn't accept a fallback-negotiator (but that's to be fixed there and not in this doc)

L74-java-channel-creds.md
sanjaypujare: I am wondering about the statement "...we may be able to avoid letting people specify TrustManagers." Currently `SslContextBuilder` allows specifying a `TrustManager` on the server side so those use-cases (not sure what those are though) could benefit from the `TrustManager`, right?

ejona86: I was saying that we may have use-cases that require a `TrustManager` continue using `SslContextBuilder`. It would be like today; they would use an experimental API. If we upgraded to Netty 5 at some point we might break the API.

L74-java-channel-creds.md
ejona86: The Factory interface isn't all that interesting. Due to some builder holes I noticed that I needed a `getDefaultPort()` on the interface (basically it was a bug). It looks like it will be a new `ProtocolNegotiator.ClientFactory`. But it will be like ProtocolNegotiatorFactory in NettyChannelBuilder; basically the builder will be getting "out of the protocol negotiator business". I think we could actually remove the existing NettyChannelBuilder.ProtocolNegotiatorFactory fairly rapidly.

L74-java-channel-creds.md
ejona86: No. It would be like that though. I'd work on deleting the existing protocol negotiator stuff from InternalNettyChannelBuilder

L74-java-channel-creds.md
ejona86: It very well could be. If SdsProtocolNegotiators _only_ supports SDS, then maybe there will be some other negotiator that will wrap it and allow using SDS or Mesh CA or some other such things. The `XdsProtocolNegotiatorFactory` name here doesn't matter as it should be package-private/internal; name it whatever you want.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+```java
+// instead of:
+channel = ManagedChannelBuilder.forTarget("example.com")
+    .build();
+channel = ManagedChannelBuilder.forTarget("example.com").usePlaintext()
+    .build();
+channel = ManagedChannelBuilder.forAddress("example.com", 443)
+    .build();
+// the user would now:
+channel = Grpc.newChannelBuilder("example.com", new TlsChannelCredentials())
```
ejona86, line 75: I've sent out the changes for this proposal in https://github.com/grpc/grpc-java/pull/7294 .

One thing came out of that when using the unstable ChannelCredentials as provided by the transports: they won't have constructors. I don't care _too_ much about users that use the classes in io.grpc.netty or io.grpc.okhttp directly, since they are few and can "deal with it." But it is a bigger concern for more user-facing things like GoogleDefaultChannelCredentials. We have a few of those today and will get more, like XdsChannelCredentials. That could cause "normal users" to see a mix of both construction styles.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+
+The new ChannelCredentials API cannot be mixed with the pre-existing
+`useTransportSecurity()` and `usePlaintext()` builder APIs. The old-style
+methods will throw if called on a builder with a ChannelCredential provided.
+
+Note that this new API does not allow: 1) creating a channel without a
+credential and supplying the credential later and 2) changing the credential
+after creating the builder. This is not expected to be much of an issue as
+users typically have the security information available before creating the
+builder and reusing a builder is rare, especially to change the security.
```
ZhenLian, line 98: Just to confirm, the credential object doesn't change after creating, but we do have a way to supply changing certificates to the transport, right?

L74-java-channel-creds.md
```diff
@@ -0,0 +1,485 @@
...
+```
+
+An example `Feature` could be `CLIENT_CERTIFICATE`. When the `Feature` is
+added, methods `getCertificateChain()`, `getPrivateKey()`, `getPassword()` can
+be added. Observing the contents of those methods would be a requirement for
+`CLIENT_CERTIFICATE`. An implementation understanding `CLIENT_CERTIFICATE`
+might not support encrypted private keys and so could consider the credential
+unsupported if `getPassword()` returned non-`null`.
+
+To support very advanced use cases, Netty will provide a credential that wraps
```
ZhenLian, line 215: If we want to support the use case where users can just specify file paths and we will do the reloading for them, what could be the best approach? Shall we add a `Feature` to the current `NettyChannelCredentials`(for example `CLIENT_CERT_PATH`), or shall we create an entire new class extending ChannelCredentials?

L74-java-channel-creds.md
ZhenLian: IIUC, even if we are expecting users to provide SslContext, we still don't lose the ability to extend the API to be able to pick some security-related things, right? For example, in C core I know there are some ongoing work to choose between TLSv1.3 and TLSv1.2 through credential options(maybe @jiangtaoli2016 can provide any additional information), if one day we decide to do that in Java too, I guess we can easily do that as well?

ejona86:
API design review, just talking about whether to have `create()` methods on things like `TlsChannelCredentials`:
 * There was some concern about `*ChannelCreds` that don't actually extend ChannelCreds (like GoogleDefaultChannelCreds) (same concern raised by @carl-mastrangelo)
 * There did not seem to be a problem having a separation between producers (applications) and consumers (transports). We did not find any need for users to refer to concrete classes like TlsChannelCreds in their own APIs; they can simply use ChannelCreds and be better off for it. Consumers will pretty much always need to consume the base `ChannelCreds` class and use instanceof checks
 * **(1)** If we had a create() method for all “ChannelCreds” and had that method return `ChannelCredentials` (not the concrete type), then that helps some. This mainly impacts those in io.grpc (Tls, Insecure, etc). We are seeing it as a _feature_ that it returns ChannelCreds instead of the concrete type
 * **(2)** Going a step further, we could make ChannelCreds final and not have any extending classes. That guarantees that all credentials are uniform. To achieve that, we move the current hierarchy to a “Config” set of classes, and have ChannelCredentials just be a `final class` holder for that config.  So TlsChannelCredentials.Config would be the concrete type within a ChannelCredentials. There would be a tree of configs instead a tree of credentials
 * We did discuss the possibility of making GoogleDefaultChannelCredentials actually extend ChannelCredentials. We could make NettyChannelCredentials itself an abstract factory for ProtocolNegotiators (instead of a holder of a factory) and then GoogleDefaultChannelCredentials would extend it to provide its implementation. To support CallCredentials on that same object we’d need to add a CallCredentials.getCallCredentials() method, which could be used instead of CompositeChannelCredentials. But we had a design goal of not exposing the Netty-specific implementation detail of those classes, and this would expose users directly to ProtocolNegotiator and expose GoogleDefaultChannelCredentials forever as an NettyChannelCredential

(2) has more complexity, via an extra level of indirection, but is clear. (1) suffices, but has some awkwardness.

@carl-mastrangelo, do you want to provide a vote? We all could accept either way (1 or 2), but had different preferences.

ejona86:
I should also mention, from the API review:
 * We discussed renaming the CallCredentials-that-aren't-credentials. Using "Factory" in the same was suggested, and it make sense. But it is not a normal "Java Factory" (or producer, or whatever else), since the class would not be instantiated and it doesn't implement an interface.
 * We discussed using "Util" but were split on how we felt about that. The create() method then should probably change, to like createChannelCredentials()
 * We discussed moving the create() methods into factory methods on new classes named `OkHttp`, `Netty`, and `Alts`; for example, Alts.googleDefaultChannelCreds(). However, we would prefer to keep the classes separate. We've experienced cases where users want to limit usage of certain classes (like InsecureChannelCreds, for security reasons) and multiple credentials in one class makes that difficult.

So those were things considered, but did not settle on them.

L74-java-channel-creds.md
ZhenLian: Just FWIW, I think in C++ we are trying to add a lot of security-related configurations to tls_credentials_options(https://github.com/grpc/grpc/blob/master/include/grpcpp/security/tls_credentials_options.h), instead of ssl_credentials_options. That's probably the reason why C++ ssl API has very few of options.
Also I think in Go, users can also pass multiple server certs by specifying the `Certificates` field of tls.config(https://golang.org/pkg/crypto/tls/#Config), and use SNI to choose one for the client. I guess it would be great if Java can also support this feature :-)

L74-java-channel-creds.md
ejona86: With the current TlsChannelCredentials, no; they would be for fixed-certificate use-cases. But NettyChannelCredentials (and even OkHttp's SslSocketFactoryChannelBuilder) can change the credentials over time. It's mostly about what format the configuration is in. If we wanted to allow TlsChannelCredentials to change its creds over time, that would be possible by passing a function as configuration to it; the transport could then call that function as appropriate.

L74-java-channel-creds.md
ejona86: `Feature` was only for `TlsChannelCredentials`. `NettyChannelCredentials` can change _very_ freely. Easiest would be to offer a way to create a `NettyChannelCredentials` that does the reloading. See the `GoogleDefaultChannelCredentials` as an example; they are `NettyChannelCredentials` but the negotiator is able to fully control the TLS configuration.

L74-java-channel-creds.md
ejona86: We can create new netty negotiators at any time and expose them with an API. They would have full reign over what they want to do.

I'm not sure what we'd want to do about TLSv1.2 vs TLSv1.3; on that front I'd expect we'd just start supporting TLSv1.3 as soon as it works. Deciding when to disable TLSv1.2 may need a Feature in TlsChannelCredentials (or maybe it is optional, only observed in the transport supports it).

L74-java-channel-creds.md
ejona86: So the new way for users to use TLS is to construct a `TlsCredentialsOptions` and pass it to `TlsCredentials` or `TlsServerCredentials`?

tls.Config in Go supports SNI directly. The Java equivalent would be to let the user pass a SSLContext, but we both know that is a PITA to configure. (It's also not clear to me how we'd consume a raw SSLContext directly in Netty.)

ZhenLian: Yes, exactly.
+ @jiangtaoli2016 in case he has something to add on.

L74-java-channel-creds.md
ZhenLian: I see. That makes sense. Thanks for the explanation!

L74-java-channel-creds.md
ejona86: Looking at TlsCredentialsOptions fields:
 * cert_request_type: server-only
 * server_verification_option: server-only
 * key_materials_config: trust and key configuration. Seems like it'd work on both, although it's not clear how multiple keys would be handled on client-side.
 * credentials_reload_config: allows reloading key_materials_config
 * server_authorization_check_config: client-only

So only trust and key configuration is shared today. That's fairly easy to just duplicate and handle twice, as long as we keep the number of representations low (which seems feasible).

ZhenLian: Yeah. IIRC, for key_material_config, client side is not able to pass in multiple key-identity cert chain pairs; If it does, either the connection would fail, or we pick the first one(I don't remember the details; maybe @jiangtaoli2016 
 or @yihuazhang could add).
From the API's point of view, I think it's probably better to separate at least the  identity key-cert pair part of key_material_config for client and server, to speak the difference out explicitly. For client it would be only one pair; for server it would be multiple pairs, and if multiples are specified, we will use SNI to decide.

ejona86: The details don't really matter, so that's fine.

In short, it seems like "there's very little expected to be shared configuration-wise between client and server." I know there's a laundry list of other TLS features, like which certs and TLS version, but we tend to manage that ourselves and are less likely to need a public, stable API. I am still fine allowing people to configure those things, but they can use the transport-specific APIs.

I'll go ahead and remove the note about sharing configuration. It seems like there's little gain available and probably not worth the confusion cost of mixing both pieces of configuration together.

carl-mastrangelo:
My leaning is more towards (2).   I didnt have any other feedback at this point.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,489 @@
...
+channel/server construction.
+
+This is to allow:
+
+*   implementing new transport credentials without creating a new
+    channel/server type, which allows users of the API to separate concerns
+*   providing stable APIs for advanced credential configuration like mutual
+    TLS, without polluting the general-purpose ManagedChannelBuilder API with
+    many overloads that don't apply to all transports
+*   composition of credentials, like necessary for xDS where a fallback
```
sanjaypujare, line 22: Even without xDS the `CompositeChannelCredentials`  feature also provides composition of credentials which is quite useful.

ejona86: CompositeChannelCredentials does not compose ChannelCredentials. But that's the opposite of what this is talking about. This is listing requirements/goals, and xDS benefits from composition. CompositeChannelCredentials is a _solution_, so by definition it isn't useful to describe requirements.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,489 @@
...
+  }
+}
+```
+
+### ServerCredentials
+
+Server-side will be a clone of client-side, but using the `Server` instead of
+`Channel`. `ServerCredentials` cannot contain `CallCredentials`, so
+`CompositeServerCredentials` will be omitted. There is also currently no known
+use-case for `ChoiceServerCredentials`, so it will be omitted.
```
sanjaypujare, line 354: I am wondering if `ChoiceServerCredentials` can be the API for the SNI use-case. So each ChannelCredentials in the list will be some kind of TlsServerCredentials with the associated hostname (`javax.net.ssl.SNIServerName` ?) and the implementation will pick the correct TlsServerCredentials for the associated hostName in the incoming connection. An alternative is to have a special feature in TlsServerCredentials that takes a map of SNIServerName to key/cert pair but looks more complex (and less flexible?) to me.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,489 @@
...
+```java
+package io.grpc;
+
+public final class TlsServerCredentials extends ServerCredentials {
+  /**
+   * Creates an instance using provided certificate chain and private key.
+   * Generally they should be PEM-encoded and the key is an unencrypted PKCS#8
+   * key (file headers have "BEGIN CERTIFICATE" and "BEGIN PRIVATE KEY").
+   */
+  public TlsServerCredentials(File certChain, File privateKey)
```
sanjaypujare, line 393: nit: `keyPassword` param needed for this and the next one?

L74-java-channel-creds.md
```diff
@@ -0,0 +1,489 @@
...
+
+```java
+// instead of:
+server = ServerBuilder.forPort(443)
+    .useTransportSecurity(new File("cert.pem"), new File("cert.key"))
+    .build().start();
+server = ServerBuilder.forPort(8080)
+    .build().start();
+// the user would now:
+ChannelCredentials tlsCreds = new TlsServerCredentials(
```
sanjaypujare, line 364: nit: You probably meant to say
```
ServerCredentials tlsCreds = new TlsServerCredentials(
```
since ServerCreds are not ChannelCreds?

L74-java-channel-creds.md
```diff
@@ -0,0 +1,489 @@
...
+  public final static class Builder {
+    public Builder keyManager(File certChain, File privateKey)
+        throws IOException {...}
+    public Builder keyManager(File certChain, File privateKey, String password)
+        throws IOException {...}
+    public Builder keyManager(InputStream certChain, InputStream privateKey)
+        throws IOException {...}
+    public Builder keyManager(
+        InputStream certChain, InputStream privateKey, String password)
+        throws IOException {...}
```
sanjaypujare, line 421: We'll also need the equivalent of netty's `ClientAuth` for making client authentication optional vs required when root-certs are provided.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,489 @@
...
+OkHttp uses blocking I/O with a byte[]-backed Buffer whereas Netty uses
+nonblocking I/O with a reference-counted, ByteBuffer-backed ByteBuf with an
+event loop. Go and C use a code-based API where credentials implement an
+interface and use a single I/O model. It is not practical to have one
+code-based API to support both OkHttp and Netty, as it would have a
+considerable amount of complexity in the implementations, be hard to optimize,
+and have a wide public API surface. Instead, Java needs a data-based API, where
+configuration values are communicated and each transport can have its own
+implementation. A data-based API can also be used by InProcess and Binder
+transports.
```
sanjaypujare, line 442: Great rationale. A diagram showing the OkHttp and Netty blocks, and blocks for Channel credentials and how the transport blocks (Netty and OkHttp) fetch configs from the channel creds to implement security implied by the config. Just a suggestion.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,489 @@
...
+are set. Keeping track of which features were available at each version can be
+error-prone. It's also hard to produce a user-meaningful error message, so they
+are aware of which specific configuration is causing the incompatibility.
+
+## Implementation
+
+ejona86 will implement. TLS will initially be bare-bones, with just the
+compatibility checker. Over the long-term TLS will be fleshed out with
+additional configuration. ServerCredentials will be implemented shortly after
+ServerCredentials.
```
sanjaypujare, line 476: => "ServerCredentials will be implemented shortly after ChannelCredentials."

L74-java-channel-creds.md
sanjaypujare: How about creating a limited hierarchy:
```
abstract class TlsCredentials 
class TlsChannelCredentials extends TlsCredentials
class TlsServerCredentials extends TlsCredentials
```
This gives you some sharing.

ejona86:
I made the changes for approach (1), which is a prerequisite for (2) anyway.

I thought I had been convinced to go with approach (2), and I started with making the changes for (2). But it seemed to mostly add confusion and goo. The documentation in particular had to become more precise in a way that seemed unhelpful. And users were going to need to "see" the Config object places and know to just ignore it.

I'm going to try again tomorrow with the changes to (2) and see how it compares a bit more directly.

L74-java-channel-creds.md
ejona86: @sanjaypujare, then `TlsChannelCredentials`/`TlsServerCredentials` would not extend `ChannelCredentials`/`ServerCredentials`, as the base credential classes are also abstract classes (not interfaces). We'd need to swap them to interfaces so we could do `class TlsChannelCredentials extends TlsCredentials implements ChannelCredentials`. We could do that, but it encourages more complex hierarchies.

L74-java-channel-creds.md
ejona86: Probably not. I think we should stick with the pattern where the credentials are converted into some internal representation at channel creation time. I'm not excited about the conversion occurring in the data path for performance and complexity reasons. Also, SNI we'd have serious issues supporting SNI if it could change arbitrary TLS configuration. I know xDS has some stuff along those lines, but I think we'd handle anything like that with a low-level API (protocol negotiator) and not at a high level here.

L74-java-channel-creds.md
ejona86: No, this is mirroring the current `serverBuilder.transportSecurity(...)` methods which don't support the password and are pretty common. If you need to provide a Key, use the Builder. I don't find specifying the password to be more rare, since it doesn't generally provide much protection because the password is generally plaintext and no more difficult to acquire than the private key file.

L74-java-channel-creds.md
ejona86: That will be part of a client authentication Feature of some sort. You need trust roots and ASK/REQUIRE. My goal here is to reproduce what we already have in the new API and have new features be introduced afterward.

ejona86: I only added keyPassword support now because it is relatively easy to handle and is closely associated with what we already support; it'd be more work for implementations to deal with a Feature than just have it be assumed initially.

L74-java-channel-creds.md
ejona86: It isn't clear to me what this would look like. Maybe we should discuss "in person"?

L74-java-channel-creds.md
ejona86: > and if the TlsChannelCredentials instance returns a non-empty set then there is something Netty doesn't support (that TlsChannelCredentials requires) and it will throw an error (with error text containing non-supported feature set).

If "it" means "Netty," yes.

> So this pattern depends on the transport implementation verifying features thru incomprehensible() to prevent a required TLS restriction/feature from being ignored by the transport.

Yes.

L74-java-channel-creds.md
ejona86: I chose to just delete this paragraph.

L74-java-channel-creds.md
ejona86: I've added notes about `NettySslContextChannelCredentials` API to convert a netty `SslContext` to `ChannelCredentials`.

L74-java-channel-creds.md
sanjaypujare: yes, I do understand where you are coming from and I sort of expected that. I remember there was some kind of comment exchange with @ZhenLian where he said gRPC core/C++ TLS credentials API has a way (or may be a proposal) of providing a map of hostname->certs to support SNI so I thought we could match that.But low level API sounds good for now

L74-java-channel-creds.md
```diff
@@ -0,0 +1,571 @@
...
+public abstract class ChannelCredentials {}
+
+public final class TlsChannelCredentials extends ChannelCredentials {
+  public static ChannelCredentials create() {...}
+
+  /* Complicated. Described separately below. */
+}
+
+/** No client identity, authentication, or encryption is to be used. */
+public final class InsecureChannelCredentials extends ChannelCredentials {
```
sanjaypujare, line 128: Sorry if I have asked this before. Is there a plan to support `LocalCredentials` like grpc-core/C++ does https://github.com/grpc/grpc/blob/master/include/grpcpp/security/credentials.h#L311 ? I think local creds can be used only for "local" connections such as UDS or "localhost".

L74-java-channel-creds.md
```diff
@@ -0,0 +1,571 @@
...
+ * They can also influence types of encryption used and similar security
+ * configuration.
+ *
+ * The concrete credential type should not be relevant to most users of the API
+ * and may be an implementation decision. Users should generally use the {@code
+ * ChannelCredentials} type for variables instead of the concrete type.
+ * Freshly-constructed credentials should be returned as {@code
+ * ChannelCredentials} instead of a concrete type to encourage this pattern.
+ * Concrete types would only be used after {@code instanceof} checks (which must
+ * consider {@code ChoiceChannelCredentials}!).
```
sanjaypujare, line 117: We are not hiding concrete types (like `ChoiceChannelCredentials` or `CompositeChannelCredentials`) but expect users to do `instanceof` and a type-cast to be able to use methods like `getChannelCredentials()` of `CompositeChannelCredentials` (for example). How does it support the view that "concrete type should not be relevant to API users and may be an implementation decision"?

Wouldn't it then be better for the `create()` methods to return the actual public concrete type, but making sure that most of the functionality is at the `ChannelCredentials` level and we only expose the minimum necessary functionality in the exposed concrete type?

L74-java-channel-creds.md
ejona86: I still have no plan to support `LocalCredentials`. Java still doesn't support UDS and the use of "localhost" isn't near as important for Java as it is in C++, as we have in-process transport for testing.

We may end up having it later to reduce the number of cases that have to use "unsafe" APIs. Basically, every instance of `InsecureChannelCredentials` may be a security issue so it may be useful to have `LocalCredentials` which reduces the risk (it isn't _safe_, but it is less bad) and have fewer instances of `InsecureChannelCredentials` to continually monitor. But right now Java internally has very few usages of `usePlaintext()` because our testing story has been so different and UDS is a PITA in Java.

L74-java-channel-creds.md
ejona86: Most users don't need to call any methods on the credentials. They just 1) create them and 2) pass them around. The transports are the main thing that needs to call methods on the credentials.

> Wouldn't it then be better for the create() methods to return the actual public concrete type

Better in what way?

> but making sure that most of the functionality is at the ChannelCredentials level and we only expose the minimum necessary functionality in the exposed concrete type?

￼I don't understand what that would look like. Could you provide a sketch of what ChannelCredentials would look like in that approach?

sanjaypujare: > > Wouldn't it then be better for the create() methods to return the actual public concrete type
> 
> Better in what way?

Compared to what was described in the original text "...Concrete types would only be used after {@code instanceof} checks..."

So with the current proposal one would do:
```
ChannelCredentials channelCreds = ChoiceChannelCredentials.create(...);
...
// somewhere later in the code
if (channelCreds instanceof ChoiceChannelCredentials) {
    ChoiceChannelCredentials choiceChannelCreds = (ChoiceChannelCredentials)channelCreds;
    // do something with choiceChannelCreds
}
```
Whereas if `create` returned the actual concrete type you wouldn't need the check and the cast:
```
ChoiceChannelCredentials choiceChannelCreds = ChoiceChannelCredentials.create(...);
...
// somewhere later in the code
// do something with choiceChannelCreds
```

All I am saying is that exposing/making public concrete subclasses but not have `create` return them looks a little inconsistent but I do understand the argument about encouraging the pattern.

> 
> > but making sure that most of the functionality is at the ChannelCredentials level and we only expose the minimum necessary functionality in the exposed concrete type?
> 
> ￼I don't understand what that would look like. Could you provide a sketch of what ChannelCredentials would look like in that approach?

Nothing different from what is currently being proposed. As you said users would mostly create creds and pass them around because the `ChannelCredentials` interface/functionality should work for most cases.

L74-java-channel-creds.md
sanjaypujare: Sounds good.

L74-java-channel-creds.md
```diff
@@ -0,0 +1,571 @@
...
+To support very advanced use cases, Netty will provide a credential that wraps
+a `ProtocolNegotiator`. This allows implementations like ALTS and XDS to use
+internal APIs without forcing their users to use experimental or internal APIs,
+as their users would just interact with `ChannelCredentials`.
+
+```java
+package io.grpc.netty;
+
+final class NettyChannelCredentials extends ChannelCredentials {
+  private NettyChannelCredentials(ProtocolNegotiator.Factory negotiator) {...}
```
sanjaypujare, line 290: nit: I haven't found any `ProtocolNegotiator.Factory` but `NettyChannelBuilder.ProtocolNegotiatorFactory` unless it's a new class proposed.

L74-java-channel-creds.md
ejona86: There is a very strong producer/consumer split with this API. The producer _doesn't have any need_ to look at the configuration; it provided it. The consumer is what needs to look at the configuration, and if it only supports one specific credential type that sort of defeats the purpose. So I don't see your example as reflecting any actual use case for this API.

> As you said users would mostly create creds and pass them around because the ChannelCredentials interface/functionality should work for most cases.

No, I don't understand what configuration you think should be within ChannelCredentials.

L74-java-channel-creds.md
ejona86: The existing class isn't quite right for this usage, simply because the old class is on the builder. So yes, it'd be some new class, but it would be similar to NettyChannelBuilder.ProtocolNegotiatorFactory. That's what was done in the implementation PR

L74-java-channel-creds.md
```diff
@@ -0,0 +1,571 @@
...
+    return newBuilder().keyManager(certChain, privateKey).build();
+  }
+
+  public static Builder newBuilder() { return new Builder(); }
+
+  /** Same as client-side */
+  public Set<Feature> incomprehensible(Set<Feature> understoodFeatures) {...}
+
+  public byte[] getCertificateChain() { return Arrays.copyOf(...); }
+  public byte[] getPrivateKey() { return Arrays.copyOf(...); }
```
sanjaypujare, line 483: Couple of comments:
- shouldn't the same or similar methods be available in `TlsChannelCredentials` to get the identity and root certificates? I didn't see them defined above. The transport layers will need these methods right?
- for the mTLS case , `getRootCertificate` may also need to be defined

L74-java-channel-creds.md
```diff
@@ -0,0 +1,571 @@
...
+KeyManagers, TrustManagers, and session cache configuration. While versions
+safely allow adding features in the future and is simpler than the Features
+enum approach, it means a consumer that only supports a subset of features at a
+version must check _all_ the features at a particular version to see if they
+are set. Keeping track of which features were available at each version can be
+error-prone. It's also hard to produce a user-meaningful error message, so they
+are aware of which specific configuration is causing the incompatibility.
+
+It would seems to be very convenient to be able to share the
+TlsChannelCredential API on server-side, as it is quite a wide API and _almost_
```
sanjaypujare, line 554: I was also thinking about it. May be have a `CommonTlsCredentials` class with `TlsChannelCredentials` and `TlsServerCredentials` subclasses with `CommonTlsCredentials` encapsulating the common TLS functionality.

L74-java-channel-creds.md
ejona86: Client-side: eventually they will be added as added Features, but the basic functionality doesn't require it. Today there is not a stable API to specify the identity and root certs client-side.

Server-side: ditto, really. mTLS is not supported by the pre-existing stable TLS APIs. You have to use transport-specific APIs. Eventually that will likely be added, but not as a first cut.

L74-java-channel-creds.md
sanjaypujare: ok

L74-java-channel-creds.md
ejona86: How would TlsChannelCredentials extend ChannelCredentials in that case?

L74-java-channel-creds.md
sanjaypujare: ok

L74-java-channel-creds.md
sanjaypujare: `SdsProtocolNegotiators` is currently misnamed (like many other classes in the package) because it's not tied to SDS but rather to xDS-provided security info so it supports all possible credential-types provided by xDS (file-based, SDS based and the new cert-provider plugin based).

L74-java-channel-creds.md
sanjaypujare: SG

L74-java-channel-creds.md
sanjaypujare: Yes, makes sense. I keep thinking of mTLS as basic functionality :-) On the client side - with just TLS - I understand there is a default trust-store.

L74-java-channel-creds.md
sanjaypujare: Oops, never mind!

PR #201, opened 2020-08-13
A33: Client-Side Fault Injection

lidizheng:
gRFC: https://github.com/lidizheng/proposal/blob/fault-injection/A33-Fault-Injection.md

### DIFF
--- a/A33-Fault-Injection.md
+++ b/A33-Fault-Injection.md
@@ -4,7 +4,7 @@ Client-Side Fault Injection
 * Approver: markdroth, ejona86, dfawley
 * Status: In Review
 * Implemented in: all languages
-* Last updated: Feb 9th, 2021
+* Last updated: Feb 10th, 2021
 * Discussion at: https://groups.google.com/g/grpc-io/c/d9dI5Hy9zzk
 
 ## Abstract
@@ -193,6 +193,11 @@ Envoy uses the token bucket algorithm to throttle the speed of letting response 
 Envoy implements fault injection is an `HTTPFilter`, which can choose to operate on bytes or HTTP messages or events. It is unclear if gRPC can perform the precise same behavior, since gRPC generally operates on proto messages (only Core has visibility to bytes, not Java/Golang). More importantly, there is not enough consensus for if the timeout timer should start at the beginning of an RPC, or the receipt of the first bytes from peer, or after the entire message is received.
 
 
+### Environment variable protection
+
+`GRPC_XDS_EXPERIMENTAL_FAULT_INJECTION` as an environment variable will be used to guard this feature for the initial release. Once it is set, gRPC will start to interpret xDS HTTP filters ([A39: xDS HTTP Filters](https://github.com/grpc/proposal/pull/219)) and enforce fault injection if configured.
+
+
 ## Implementation
 
-TBD
+gRPC Core: https://github.com/grpc/grpc/pull/24354
--- a/A33-Fault-Injection.md
+++ b/A33-Fault-Injection.md
@@ -5,7 +5,7 @@ Client-Side Fault Injection
 * Status: In Review
 * Implemented in: all languages
 * Last updated: August 13th, 2020
-* Discussion at: TBD
+* Discussion at: https://groups.google.com/g/grpc-io/c/d9dI5Hy9zzk
 
 ## Abstract
 
--- a/A33-Fault-Injection.md
+++ b/A33-Fault-Injection.md
@@ -20,59 +20,44 @@ Please refer to [HTTP connection management](https://www.envoyproxy.io/docs/envo
 
 Please refer to [HTTP filters](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_filters) for how HTTP filters process traffic in xDS.
 
-Please refer to [A39: xDS HTTP Filters](https://github.com/grpc/proposal/pull/219) for how xDS HTTP Filters is processed in gRPC.
+Please refer to [A39: xDS HTTP Filters](https://github.com/grpc/proposal/pull/219) for how xDS HTTP Filters are processed in gRPC.
 
 
-### Fault Injection in Envoy xDS API
+### Fault Injection in xDS
 
-Fault injection in xDS API is an HTTPFilter named `HTTPFault`. As mentioned above, the fault injection config can be provided by **LDS** (HTTP Connection Manager) and **RDS** (VirtualHost, Route, ClusterWeight). Besides, the fault injection filter offers [global Runtime settings](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter#runtime) that can override global default fault injection filter config. For the explanation of Envoy Runtime, see section below.
-
-This design doc focuses on getting fault injection filter config from LDS and RDS.
+Please refer to [Fault Injection](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter) section in Envoy document.
 
 
-### Fault Injection Features
+## Proposal
 
-Fault Injection Options
-The full definition of Envoy v3 fault injection proto can be found [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/fault/v3/fault.proto#extensions-filters-http-fault-v3-httpfault). In short, it allows users to configure following options:
+### Overview
 
-1. Fixed delay before forwarding the request;
-1. Overridden gRPC status code;
-1. Traffic matcher - headers;
-1. Maximum active faults;
-1. Per-Stream Response rate limit (kbps, and selected by percentage);
-1. Options to override global runtime settings. 
-1. Traffic matcher - upstream cluster name, downstream node names
-
-Feature 1-4 should be implemented to make gRPC compatible with OSS control planes, and we should aim to ship them in the initial implementation. Feature 5 is postponed until we have a consensus on its behavior (see timeout issue below). Feature 6 is ignored since there is no plan to support Runtime settings. Feature 7 is postponed until we have a clear ask for it.
-
-To support all features above, it requires an API as powerful as the interceptors, which can intercept each individual RPC in their early stages and apply fault injection policy.
-
-**Definition of active faults:** Active fault number is a gauge representing the number of ongoing fault injected RPCs. The gauge increases when an RPC is injected by one or more faults, and it decreases when the injected RPC is finished (see [FaultFilter::onDestroy](**https://source.corp.google.com/piper///depot/google3/third_party/envoy/src/source/extensions/filters/http/fault/fault_filter.cc;l=411;bpv=1;bpt=1**)).
-
-**Response rate limit algorithm:** Envoy uses the token bucket algorithm to throttle the speed of letting response pass through this filter. It will try to write as much as possible from the buffer, then it will schedule another write when the next token is available.
-
-Each token permits to write `(max_kbps * 1024) / SecondDivisor` of bytes. The `SecondDivisor` is a constant 16, which splits 1 second into 16 segments (~63ms apart).
-
-Only the body of the HTTP stream is throttled by the response rate limit algorithm. Not headers or trailers.
-
-**Response rate limit and timeout:** Envoy implements fault injection is an `HTTPFilter`, which can choose to operate on bytes or HTTP messages or events. It is unclear if gRPC can perform the precise same behavior, since gRPC generally operates on proto messages (Core has visibility to bytes). The timeout timer may start at the beginning of an RPC, or the receipt of the first bytes from peer, or after the entire message is received.
+The full definition of Envoy v3 fault injection proto can be found at [HTTPFault](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/fault/v3/fault.proto#extensions-filters-http-fault-v3-httpfault). The provided feature can be categorized as 7 items. This proposal will talk about the ones that gRPC plans to support, the ones that will be deferred to future releases, and ones that will not be supported.
 
 
-### Absence of Envoy Runtime in gRPC
+#### Supported Fault Injection Features
 
-Envoy has a mechanism to allow reload of configuration without restarting the entire process named [Runtime](https://www.envoyproxy.io/docs/envoy/latest/operations/runtime). It can be loaded from a static file, updated with a local file system, or overridden from an Admin console ([reference](https://www.envoyproxy.io/docs/envoy/latest/configuration/operations/runtime)).
+1. Injecting delays before starting a request;
+1. Failing RPC with a particular gRPC status code;
+1. Imposing a limit on the maximum number of currently active faults;
+1. Controlling fault injection policies via HTTP headers (or gRPC metadata).
 
-However, as a library, gRPC doesn’t have a decent standing to maintain an “Runtime” state. Available options are 1) ignore Runtime settings; 2) allow initial Runtime setting (from a file or env variables). According to previous design decisions, the Runtime settings are usually ignored (see [gRPC xDS Traffic Splitting and Routing](https://github.com/grpc/proposal/blob/master/A28-xds-traffic-splitting-and-routing.md)).
 
-### Fault Injection HTTP Headers in Envoy
+#### Deferred Fault Injection Features
+
+1. Restricting per-stream response rate limit (in kbps, and selected by percentage) (see alternative section below for deferring rationale).
+1. Selecting fault injecting traffic based on upstream cluster name, downstream node names, and headers (postponed until a valid use-case presents itself).
+
+
+#### Unsupported Fault Injection Features
+
+1. Configuring fault injection based on global runtime settings (because gRPC doesn't have the concept of a global runtime);
+
+
+#### Fault Injection HTTP Headers in Envoy
 
 Envoy allows applications to specify [HTTP headers](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter#controlling-fault-injection-via-http-headers) to enable fault injection. This feature still needs a fault injection filter to be specified, and the delay or the abort field needs to explicitly say `HeaderDelay` or `HeaderAbort`.
 
-### Prior Art: Envoy HTTP Filter
-
-Envoy filters have very granular logic injection points at different stages of a request (or stream). For example, while decoding headers / data / trailer, encoding headers / data / trailer, or completion of decoding / encoding. In each injection point, the return value dictates whether Envoy should proceed to other filters or stop. This design gives immense flexibility for filters. The delay injection could be a headache since it can’t block the thread. But fault injection filters utilize this design and are able to schedule callback (rest of request processing logic) asynchronous.
-
-## Proposal
 
 ### Priority Between Fault Injection And Retry
 
@@ -81,18 +66,6 @@ In Envoy, HTTP filters (including fault injection filter) are executed before ro
 Based on a local experiment, if an RPC is delayed and failed by peer (not fault injected), Envoy will retry without any additional delay. There is an ongoing GitHub thread about the conflict between fault injection and retry at [istio/istio#13705](https://github.com/istio/istio/issues/13705).
 
 In summary, **if retry and delay both take effect, only delay once before sending any traffic; if retry and abort both take effect, retry will be disabled; if retry, delay and abort all take effect, the RPC will be delayed then aborted without retry.**
-
-
-### Enforce Fault Injection
-
-Take Envoy as examples, it utilizes a powerful traffic interception API to implement fault injection. Here are possible solutions to implement FI:
-
-1. **Returning an interceptor from the ConfigSelector.** In the [ConfigSelector design doc](https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md), supporting interceptor is one of the proposed features. But it’s not available in actual implementations yet. (Viable for Java, Go)
-2. **Defining a fault injection global interceptor.** The interceptor inspects the method config of each RPC. Then it makes the decision whether to delay or to abort. (Viable for Java, Go, possibly for filters in Core)
-3. **Implementing fault injection as a new RPC feature controlled by service config.** In this way, fault injection will have similar architecture as timeout, wait for ready, and retry (gRFC A6). The logic needs to be implemented in native RPC call processing code path. (Viable for Core, Java, Go)
-4. **Using load balancer policy.** Load balancer is capable of aborting an RPC, and able to mimic delay by returning “no subchannel, but wait for X seconds before trying again”. (Viable for Core, Java, Go)
-
-**Java and Go will implement via solution 1 (ConfigSelector)**, and **Core will implement fault injection in client channel** with an non-service-config mechanism to fetch the FI config.
 
 
 ### Fault Injection Features
@@ -107,7 +80,9 @@ This feature aborts RPC with given status code (see status code mapping section 
 
 Also, once an RPC is aborted, the enforcer should push back on following retry attempts.
 
-#### Max active fault
+#### Maximum active faults
+
+Active fault number is a gauge representing the number of ongoing fault injected RPCs. The gauge increases when an RPC is injected by one or more faults, and it decreases when the injected RPC is finished (see [FaultFilter::onDestroy](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/fault/fault_filter.cc#L427)).
 
 This feature limits the maximum number of concurrent active fault injected RPC for the entire process. Apparently, this requires a global counter (gauge). This value is designed to be updated via multiple channels. So, **it needs to be thread-safe**, but the actual implementation depends on each stack. For example, volatile variable, atomic variable, memory order, locking. Anything works best for the language.
 
@@ -116,7 +91,7 @@ The way to implement the tracking of RPC lifespan also depends on the language. 
 
 #### Header matching
 
-This doc proposes to support as many metadata as we can (see [detailed definitions](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter.html?highlight=fault%20injection#controlling-fault-injection-via-http-headers)):
+This doc proposes to support as many metadata as we can (see [detailed definitions](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter.html?#controlling-fault-injection-via-http-headers)):
 
 ```bash
 # Aborts HTTP requests with status code 503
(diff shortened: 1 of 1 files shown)


A33-Fault-Injection.md
```diff
@@ -0,0 +1,268 @@
...
+On the other hand, TD has a released support for fault injection via route configuration for HTTP. So, **this design doc focuses on getting fault injection filter config from LDS and RDS**.
+
+### Fault Injection Features
+
+Fault Injection Options
+The full definition of Envoy v3 fault injection proto can be found [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/fault/v3/fault.proto#extensions-filters-http-fault-v3-httpfault). In short, it allows users to configure following options:
+
+1. Fixed delay before forwarding the request;
+1. Overridden gRPC status code;
+1. Traffic matcher (upstream cluster name, downstream cluster name, and headers);
```
dfawley, line 46: There isn't any detail in here as to what is covered by this part of the feature.

In the envoy docs, I see:

  "upstream_cluster": "...",
  "headers": [],
  "downstream_nodes": [],

`upstream_cluster` makes sense - If set, we'd match on the cluster chosen by the config selector.

`headers` seems straightforward based on the envoy docs: "A match will happen if all the headers in the config are present in the request with the same values (or based on presence if the value field is not in the config)."

`downstream_nodes`, though, I'm not sure what to do with.  Is this what you meant by "downstream cluster name"?  For this it sounds like we would just match the value(s? - what if there are multiple) of the header "x-envoy-downstream-service-node" with the settings in "downstream_nodes" in the fault injection config.

But where does this header come from?  It looks like it's set by envoy, not the user.  Based on "the value is taken from the `--service-node` option" [[ref]](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-envoy-downstream-service-node).  We aren't setting this today as far as I'm aware, so does it make sense to support this feature?

lidizheng: For the traffic matcher, I updated the design doc that I think that only header-matching is common enough. Both the upstream cluster or downstream nodes features will postponed until users are asking for them.

The headers (or our metadata) can be set either by upstream Envoy or the application. I'm not sure how much will that make.

dfawley: > The headers (or our metadata) can be set either by upstream Envoy or the application.

The ones for the header matchers may be set by the application.  But the one for the "downstream node" feature (which I now see we aren't supporting here) seems to be set only by Envoy, per the ref above.  I'm not clear on what else this may be used for, and I am wondering if setting it is something we need to do in gRPC (or will need to do soon for other reasons).

lidizheng: To support this feature, I think we will need to add this feature as well. Technically, gRPC's XdsClient can add `x-envoy-downstream-service-node` or `service-node` header with it's Node info for all processed requests?

But, for this proposal, I hope to focus on the most common usages of the fault injection filter, which is delay, abort, match header, and max active faults.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,269 @@
...
+* Last updated: August 13th, 2020
+* Discussion at: https://groups.google.com/g/grpc-io/c/d9dI5Hy9zzk
+
+## Abstract
+
+Fault injection is an essential feature that assists developers in building fault-tolerant applications. To keep gRPC’s competitive advantage, this doc aims to design a fault injection mechanism on the client side for gRPC while allowing transparent transition from Envoy configuration.
+
+## Background
+
+### Envoy HTTPFilter
```
markdroth, line 16: I think this section should just be replaced with a reference to #219.  We should explicitly state that the FI functionality depends on xDS HTTP filter support.

lidizheng: Removed most descriptions and linked to A39.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,262 @@
...
+
+Envoy has a network level filter named HTTP Connection Manager. It is responsible to handle common functionalities against HTTP connections and requests. In its `http_filters` field, users can set up multiple HTTPFilters. Each HTTPFilter requires at least a name. The name will be the identifier that assists future updates of filter configuration. For example, users can specify an `{"name": "envoy.cors"}` filter in `http_filters` with no content, then update the CORS filter’s configuration with the `typed_per_filter_config` field of a VirtualHost in RDS.
```
sanjaypujare, line 18: Do we need this or can we just reference an Envoy page that describes the HTTP CM and HTTP filter?

A33-Fault-Injection.md
```diff
@@ -0,0 +1,262 @@
...
+
+Please refer to [A39: xDS HTTP Filters](https://github.com/grpc/proposal/pull/219) for how xDS HTTP Filters is processed in gRPC.
+
+### Fault Injection in Envoy xDS API
+
+Fault injection in xDS API is an HTTPFilter named `HTTPFault`. As mentioned above, the fault injection config can be provided by **LDS** (HTTP Connection Manager) and **RDS** (VirtualHost, Route, ClusterWeight). Besides, the fault injection filter offers [global Runtime settings](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter#runtime) that can override global default fault injection filter config. For the explanation of Envoy Runtime, see section below.
+
+Currently, the Traffic Director (TD) isn’t forwarding HTTPFilter configs in HTTP Connection Manager from LDS, and there isn’t an exposed API allowing users to directly config fault injection for all HTTP connections. To pave gRPC’s way for more extensive control plane compatibility, we need to support fetching filter config from HTTP Connection Manager among LDS responses.
```
sanjaypujare, line 26: Should we reference TD in the gRFC? Can we say certain control planes or certain xDS servers?

A33-Fault-Injection.md
```diff
@@ -0,0 +1,262 @@
...
+
+1. Fixed delay before forwarding the request;
+1. Overridden gRPC status code;
+1. Traffic matcher - headers;
+1. Max active faults;
+1. Per-Stream Response rate limit (kbps, and selected by percentage);
+1. Options to override global runtime settings. 
+1. Traffic matcher - cluster name
+
+Feature 1-4 should be implemented to make gRPC compatible with OSS control planes, and we should aim to ship them in the initial implementation. Feature 5 is postponed until we have a consensence on its behavior (see timeout issue below). Feature 6 is ignored since there is no plan to support Runtime settings. Feature 7 is postponed until we have a clear ask for it.
```
sanjaypujare, line 45: s/consensence/consensus/

A33-Fault-Injection.md
```diff
@@ -0,0 +1,262 @@
...
+
+**Response rate limit algorithm:** Envoy uses the token bucket algorithm to throttle the speed of letting response pass through this filter. It will try to write as much as possible from the buffer, then it will schedule another write when the next token is available.
+
+Each token permits to write `(max_kbps * 1024) / SecondDivisor` of bytes. The `SecondDivisor` is a constant 16, which splits 1 second into 16 segments (~63ms apart).
+
+Only the body of the HTTP stream is throttled by the response rate limit algorithm. Not headers or trailers.
+
+**Response rate limit and timeout:** Envoy implements fault injection is an `HTTPFilter`, which can choose to operate on bytes or HTTP messages or events. It is unclear if gRPC can perform the precise same behavior, since gRPC generally operates on proto messages (Core has visibility to bytes). The timeout timer may start at the beginning of an RPC, or the receival of the first bytes from peer, or after the entire message is received.
```
sanjaypujare, line 57: s/receival/receipt/

A33-Fault-Injection.md
```diff
@@ -0,0 +1,262 @@
...
+ // triggered, the RPC will be delayed first then abort with the status code.
+ message FaultInjectionPolicy {
+   // The fixed delay before starting the RPC.
+   //
+   // No traffic should be sent until the delay phase ends. By default, the
+   // delay duration is 0 (0 seconds and 0 nanos). If the delay is 0, the RPC
+   // delay feature is disabled and delay_fraction will be ignored.
+   google.protobuf.Duration delay = 1;
+
+   // The possibility that this RPC will be delayed.
```
sanjaypujare, line 227: Should it say probability?

A33-Fault-Injection.md
```diff
@@ -0,0 +1,262 @@
...
+   // feature and ignores abort_percent. If the control plane only offers
+   // HTTP status code, the application should map it to cardinal RPC status
+   // code.
+   //
+   // By default the status message is empty. If the status message is
+   // specified with non-OK status code, then the status message will be added
+   // to the aborted RPC.
+   google.rpc.Status abort_status = 3;
+
+   // The possibility that this RPC will be aborted.
```
sanjaypujare, line 246: Probability?

A33-Fault-Injection.md
```diff
@@ -0,0 +1,262 @@
...
+
+Based on a local experiment, if an RPC is delayed and failed by peer (not fault injected), Envoy will retry without any additional delay. There is an ongoing GitHub thread about the conflict between fault injection and retry at [istio/istio#13705](https://github.com/istio/istio/issues/13705).
+
+In summary, **if retry and delay both take effect, only delay once before sending any traffic; if retry and abort both take effect, retry will be disabled; if retry, delay and abort all take effect, the RPC will be delayed then aborted without retry.**
+
+### Enforce Fault Injection
+
+Take Envoy as examples, it utilizes a powerful traffic interception API to implement fault injection. Here are possible solutions to implement FI:
+
+1. **Returning an interceptor from the ConfigSelector.** In the [ConfigSelector design doc](https://github.com/grpc/proposal/pull/192/files/b2fe83ad515e8a72f669dd3f84d185b7f00c07f3#diff-560d8f55a3a659fca6dfaa0823d4e1d1R56), supporting interceptor is one of the proposed features. But it’s not available in actual implementations yet. (Viable for Java, Go)
```
sanjaypujare, line 108: Point to the actual doc https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md instead of the PR

A33-Fault-Injection.md
lidizheng: Thanks for the suggestion. I have removed this paragraph, and added two links instead.

A33-Fault-Injection.md
lidizheng: Good catch, removed. After Mark's proposal, this restriction is outdated, there we will match our filter config accessing features with Envoy. So, there is no need to mention TD.

A33-Fault-Injection.md
lidizheng: Done.

A33-Fault-Injection.md
lidizheng: Done.

A33-Fault-Injection.md
lidizheng: Good catch!

A33-Fault-Injection.md
lidizheng: You are right, probability is more accurate here. Updated.

A33-Fault-Injection.md
lidizheng: Done.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+* Implemented in: all languages
+* Last updated: Feb 4th, 2021
+* Discussion at: https://groups.google.com/g/grpc-io/c/d9dI5Hy9zzk
+
+## Abstract
+
+Fault injection is an essential feature that assists developers in building fault-tolerant applications. To keep gRPC’s competitive advantage, this doc aims to design a fault injection mechanism on the client side for gRPC while allowing transparent transition from Envoy configuration.
+
+
+## Background
```
markdroth, line 15: I think most of the material in this section is redundant with what comes later in the doc.  I suggest removing most of it.

I think that all we really need in the "background" section is the following:

- The text you currently have in the "Envoy HTTPFilter" subsection, immediately below.
- A link to https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter, as a reference for the Envoy fault injection filter functionality.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Fault Injection HTTP Headers in Envoy
+
+Envoy allows applications to specify [HTTP headers](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter#controlling-fault-injection-via-http-headers) to enable fault injection. This feature still needs a fault injection filter to be specified, and the delay or the abort field needs to explicitly say `HeaderDelay` or `HeaderAbort`.
+
+### Prior Art: Envoy HTTP Filter
+
+Envoy filters have very granular logic injection points at different stages of a request (or stream). For example, while decoding headers / data / trailer, encoding headers / data / trailer, or completion of decoding / encoding. In each injection point, the return value dictates whether Envoy should proceed to other filters or stop. This design gives immense flexibility for filters. The delay injection could be a headache since it can’t block the thread. But fault injection filters utilize this design and are able to schedule callback (rest of request processing logic) asynchronous.
+
+## Proposal
```
markdroth, line 75: I think this section should start out with a high-level overview of what this gRFC is proposing.  This should be similar to the material that you currently have in the "background" section above, but much more concise.  Specifically, I think we should briefly touch on the following high-level points:

- We will support delays before starting a request, failing with a particular gRPC status code, and imposing a limit on the maximum number of currently active faults.
- We will allow these features to be configured directly from the xDS config.  We will also allow the xDS config to tell us to look at request headers for configuration.
- We will not support configuration based on global runtime settings, because unlike Envoy, gRPC does not have the concept of a global runtime.
- We will not initially support response rate limiting, since that will require more design and implementation work.  We will wait for a use-case to present itself before we do that work.
- We will not support selecting faults based on upstream or downstream peers until a use-case presents itself.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Priority Between Fault Injection And Retry
```
markdroth, line 77: I think this subsection can be removed, since gRPC does not yet support retries configured by xDS.  In the future, when we publish a gRFC covering xDS retry support, it can cover this kind of interaction.  @dapengzhang0

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+
+In Envoy, HTTP filters (including fault injection filter) are executed before router filters. Retry is part of the router filter. If an RPC is aborted, the router filter won’t see this RPC and the RPC won’t be retried. In gRPC, we will align our behavior with Envoy.
+
+Based on a local experiment, if an RPC is delayed and failed by peer (not fault injected), Envoy will retry without any additional delay. There is an ongoing GitHub thread about the conflict between fault injection and retry at [istio/istio#13705](https://github.com/istio/istio/issues/13705).
+
+In summary, **if retry and delay both take effect, only delay once before sending any traffic; if retry and abort both take effect, retry will be disabled; if retry, delay and abort all take effect, the RPC will be delayed then aborted without retry.**
+
+
+### Enforce Fault Injection
```
markdroth, line 86: I think this section is no longer relevant.  To the extent that these details need to be discussed, they are addressed in the HTTP filter gRFC.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+4. **Using load balancer policy.** Load balancer is capable of aborting an RPC, and able to mimic delay by returning “no subchannel, but wait for X seconds before trying again”. (Viable for Core, Java, Go)
+
+**Java and Go will implement via solution 1 (ConfigSelector)**, and **Core will implement fault injection in client channel** with an non-service-config mechanism to fetch the FI config.
+
+
+### Fault Injection Features
+
+#### RPC delay injection
+
+The injection of RPC delay translates to **injecting delay before sending any traffic for an RPC**. As discussed above, the delay should be injected only once regardless of how many following retries await.
```
markdroth, line 102: As per above, I suggest removing discussion of retries.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+X-envoy-fault-throughput-response=''
+X-envoy-fault-throughput-response-percentage=''
+```
+
+**Fault inject config in HTTP filter and headers won’t be conflicted**, because the header only being evaluated if the delay or abort is set to `HeaderDelay` or `HeaderAbort` instead of concrete values.
+
+
+### Status Code Mapping
+
+`grpc_status` is only available in xDS v3 API. In xDS v2 API, no matter TD or Envoy, they all only allow `http_status`. Mapping between HTTP status code and gRPC status code is well defined (see [doc](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md)), but that means users are limited to inject only 7 out of 16 gRPC status codes (including `OK`). The underlying schema for abort looks like [FaultAbort](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/fault/v3/fault.proto#extensions-filters-http-fault-v3-faultabort). **This doc proposes to update TD configuration to allow more expressive status code injection, once xDS v3 API is rolled out.**
```
markdroth, line 149: We are planning to support HTTP filters in v3 only, so I think discussion of v2 can be removed.

A33-Fault-Injection.md
sanjaypujare: Even probability doesn't quite sound right in this context. What we are really talking about is, for example, out of 1 million RPCs delay will be introduced for these many RPCs (`delay_fraction`). If this is done on a random basis then probability applies to a given RPC being delayed because of this feature. However the number being described here is the "fraction" of RPCs (out of a million) that will be delayed by this feature.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+
+#### RPC delay injection
+
+The injection of RPC delay translates to **injecting delay before sending any traffic for an RPC**. As discussed above, the delay should be injected only once regardless of how many following retries await.
+
+#### RPC abort injection
+
+This feature aborts RPC with given status code (see status code mapping section below for more details). In Envoy, the abort behavior in HTTP/2 happens immediately after the stream reaches the fault injection filter and the injected delay. In gRPC, it translates to that for **both unary and streaming RPCs the abort will happen before sending any traffic**.
+
+Also, once an RPC is aborted, the enforcer should push back on following retry attempts.
```
markdroth, line 108: As per above, suggest removing discussion of retries.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+#### Max active fault
+
+This feature limits the maximum number of concurrent active fault injected RPC for the entire process. Apparently, this requires a global counter (gauge). This value is designed to be updated via multiple channels. So, **it needs to be thread-safe**, but the actual implementation depends on each stack. For example, volatile variable, atomic variable, memory order, locking. Anything works best for the language.
+
+The way to implement the tracking of RPC lifespan also depends on the language. For example, Core & Java may use the deallocation method of the RPC object or enforcer. Go’s Finalizer might not work well enough, so the enforcer (interceptor) needs to update the active fault counter when RPC is finished.
+
+
+#### Header matching
+
+This doc proposes to support as many metadata as we can (see [detailed definitions](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/fault_filter.html?highlight=fault%20injection#controlling-fault-injection-via-http-headers)):
```
markdroth, line 119: I think you can remove the `?highlight=fault%20injection` part of this URL.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+
+
+### Status Code Mapping
+
+`grpc_status` is only available in xDS v3 API. In xDS v2 API, no matter TD or Envoy, they all only allow `http_status`. Mapping between HTTP status code and gRPC status code is well defined (see [doc](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md)), but that means users are limited to inject only 7 out of 16 gRPC status codes (including `OK`). The underlying schema for abort looks like [FaultAbort](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/fault/v3/fault.proto#extensions-filters-http-fault-v3-faultabort). **This doc proposes to update TD configuration to allow more expressive status code injection, once xDS v3 API is rolled out.**
+
+We still accept fault injection settings with only HTTP status code. **The ConfigSelector should translate the HTTP status code into gRPC status code** before sending the config to the fault injection enforcer.
+
+
+### No multiple fault injection filter
```
markdroth, line 154: This section can be removed.  As per the HTTP filter design, we *will* support multiple instances of the same filter.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+
+Multiple fault injection is allowed in Envoy. Fault injection filters are normal HTTP filters in Envoy. All HTTP filters are executed consequently. Confirmed with experiment that 2 fault injection filters work fine. Also, if Envoy specified any Runtime fault injection parameter, it overrides all HTTP fault injection filters and only one fault injection will be working.
+
+In TD and Istio, there could only be one fault injection setting in routing configuration. Each RPC can only have 1 matched routing configuration. The fault injection policy in the routing configuration will be executed. For simplicity, **this doc proposes to only allow one fault injection filter**.
```
markdroth, line 158: There should be no discussion of TD or Istio in this doc.  This is about OSS gRPC features only.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+After ConfigSeletor calculated the final fault injection config, it can encode the config proto and set it as metadata for that RPC. This approach is straightforward and allows more cohesive logic in the fault injection enforcer. On the other hand, it requires additional serialization and deserialization of proto messages, and it can reduce debuggability.
+
+```bash
+grpc-fault-injection-policy-bin=<...>  # The encoded FI config proto message
+```
+
+However, setting additional metadata may leak implementation details onto wire, which is undesired.
+
+
+### Alternative 2: Fault Injection Config via Multiple Metadata
```
markdroth, line 179: We basically are already going to support this due to supporting configuring FI via headers.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+   // 1 million (1000000). For example, 0.625% will be 6250, 12.5% will be
+   // 125000. Its value should be clamped into the interval [0, 1000000].
+   uint32 abort_fraction = 4;
+ }
+
+ FaultInjectionPolicy fault_injection_policy = 8;
+}
+```
+
+This approach requires more engineering effort, but provides the fault injection feature to non-xDS gRPC applications.
```
markdroth, line 248: Might be worth noting that we may consider doing something like this in the future.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+1. Maximum active faults;
+1. Per-Stream Response rate limit (kbps, and selected by percentage);
+1. Options to override global runtime settings. 
+1. Traffic matcher - upstream cluster name, downstream node names
+
+Feature 1-4 should be implemented to make gRPC compatible with OSS control planes, and we should aim to ship them in the initial implementation. Feature 5 is postponed until we have a consensus on its behavior (see timeout issue below). Feature 6 is ignored since there is no plan to support Runtime settings. Feature 7 is postponed until we have a clear ask for it.
+
+To support all features above, it requires an API as powerful as the interceptors, which can intercept each individual RPC in their early stages and apply fault injection policy.
+
+**Definition of active faults:** Active fault number is a gauge representing the number of ongoing fault injected RPCs. The gauge increases when an RPC is injected by one or more faults, and it decreases when the injected RPC is finished (see [FaultFilter::onDestroy](**https://source.corp.google.com/piper///depot/google3/third_party/envoy/src/source/extensions/filters/http/fault/fault_filter.cc;l=411;bpv=1;bpt=1**)).
```
markdroth, line 50: This is an implementation detail.  It can be moved down next to the "Evaluate Possibility Fraction" section below.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,253 @@
...
+
+**Response rate limit algorithm:** Envoy uses the token bucket algorithm to throttle the speed of letting response pass through this filter. It will try to write as much as possible from the buffer, then it will schedule another write when the next token is available.
```
markdroth, line 52: These two paragraphs about response rate limiting can be moved into the "Alternatives" section as an explanation for why it's complicated for us to support rate limiting.

markdroth (changes requested):
Thanks for doing this, Lidi.  My comments here are mostly about overall organization and updates based on the HTTP filter design.

A33-Fault-Injection.md
lidizheng: This section is removed. It suggested an implementation alternative to always translate fault injection policy into gRPC metadata. It's similar to our config-FI via headers.

lidizheng (commented):
@markdroth Thanks for the advices! I have visited all comments, but would like to discuss whether to keep mentioning retry or not in the comment threads.

A33-Fault-Injection.md
```diff
@@ -1,10 +1,10 @@
 Client-Side Fault Injection
 ----
 * Author(s): lidiz
-* Approver: roth, ejona, dfawley
+* Approver: markdroth, ejona, dfawley
```
markdroth, line 4: ejona86

A33-Fault-Injection.md
markdroth: I think the only reason we had discussions about it was because we were trying to understand the structure of the xDS protos.  Once you understand that, I think the ordering here is obvious: fault injection is provided by the FI filter, and retries are provided by the router filter, which always comes after the FI filter.  I don't think this is something that needs to be explicitly called out for users to understand.

A33-Fault-Injection.md
markdroth: As above, I don't think this is necessary to call out.

A33-Fault-Injection.md
```diff
@@ -107,7 +75,9 @@ This feature aborts RPC with given status code (see status code mapping section
 
 Also, once an RPC is aborted, the enforcer should push back on following retry attempts.
 
-#### Max active fault
+#### Maximum active faults
+
+Active fault number is a gauge representing the number of ongoing fault injected RPCs. The gauge increases when an RPC is injected by one or more faults, and it decreases when the injected RPC is finished (see [FaultFilter::onDestroy](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/fault/fault_filter.cc#L427)).
```
markdroth, line 80: I don't think we need the link to the Envoy source code.

A33-Fault-Injection.md
```diff
@@ -245,7 +194,14 @@ message MethodConfig {
 }
 ```
 
-This approach requires more engineering effort, but provides the fault injection feature to non-xDS gRPC applications.
+Turned out, this approach requires more engineering effort, but makes the fault injection feature to non-xDS gRPC applications. gRPC team does consider to improve service config with fault injection settings, and will carry-on this work in a separate proposal in future.
```
markdroth, line 197: s/feature to/feature available to/

A33-Fault-Injection.md
markdroth: This confusion is confusion about xDS, not about gRPC's xDS support.  I don't think it's our responsibility to clear this up.

markdroth (commented):
This looks pretty good!  Just a few comments remaining.

A33-Fault-Injection.md
lidizheng: Removed. The statement about fault injection filter and router filter was the first sentence in this section. I guess users could find out about xDS retry from xDS website... with a few search, they should find the same issue as I did [istio/istio#13705](https://github.com/istio/istio/issues/13705).

A33-Fault-Injection.md
lidizheng: Removed the sentence mentioning retry.

A33-Fault-Injection.md
lidizheng: Removed.

markdroth:
This looks great!  Just one more thing: Please add a small section about the env var guard, similar to what I added in the HTTP filter gRFC in https://github.com/grpc/proposal/pull/219/commits/7eaa156ada13aab38f52ddccd52ce38d994a80e4.

lidizheng:
@markdroth The env guard section is added. It also mentioned A39 since this flag will flip both features on.

A33-Fault-Injection.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+### Support for Response Rate Limit Algorithm
+
+Envoy uses the token bucket algorithm to throttle the speed of letting response pass through this filter. It will try to write as much as possible from the buffer, then it will schedule another write when the next token is available. Each token permits to write `(max_kbps * 1024) / SecondDivisor` of bytes. The `SecondDivisor` is a constant 16, which splits 1 second into 16 segments (~63ms apart). Only the body of the HTTP stream is throttled by the response rate limit algorithm. Not headers or trailers.
+
+Envoy implements fault injection is an `HTTPFilter`, which can choose to operate on bytes or HTTP messages or events. It is unclear if gRPC can perform the precise same behavior, since gRPC generally operates on proto messages (only Core has visibility to bytes, not Java/Golang). More importantly, there is not enough consensus for if the timeout timer should start at the beginning of an RPC, or the receipt of the first bytes from peer, or after the entire message is received.
+
+
+### Environment variable protection
```
markdroth, line 196: This shouldn't be in the "Alternatives" section.  Please move it up to the "Proposal" section, probably line 120.

lidizheng: Updated. Good catch!

A33-Fault-Injection.md
```diff
@@ -117,6 +117,11 @@ gRPC should accept fault injection settings with either HTTP status code or gRPC
 According to the [Envoy implementation](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/fault/fault_filter.cc#L222), each feature of fault injection is checked with a different random number. Compared to using a single random number for all features, this affects the probability of an RPC being affected by fault injection. For example, the fault injection has 20% chance to delay and 5% chance to abort. Under the single random number solution, only 20% RPC will be affected. But if two features are independent, 24% RPC will be affected. So, **this doc suggests to make each feature independent by using a different random number each time.**
 
 
+### Environment Variable Protection
```
markdroth, line 120: Please make changes similar to what I did in https://github.com/grpc/proposal/pull/219/commits/af7b012a38e677f5721f1b021b2e829809802aa9 to reflect the fact that this is just temporary.

markdroth (approved):
One minor change left, but otherwise this looks great!  I'll go ahead and approve.

Thanks for doing this!

A33-Fault-Injection.md
lidizheng: Done. Using the same words as your commit.

lidizheng:
@dfawley @dapengzhang0 Added a section about the fault injection header percentage is capped by the effective fault injection filter config. Link to the implementation PR https://github.com/envoyproxy/envoy/pull/10724

PR #202, opened 2020-08-17
A34 `weighted_round_robin` lb_policy for per endpoint weight from `ClusterLoadAssignment` response

yishuT:


markdroth:
Please fix the commit to include your user ID, as per the EasyCLA check above.

yishuT:
Hey, would like to hear an update even you think this proposal is far from the roadmap. I personally have some cycles in coming weeks to make this proposal aligned with grpc roadmap.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
+`weighted_round_robin` lb_policy for per endpoint weight from `ClusterLoadAssignment` response
+----
+* Author(s): Yi-Shu Tai (echo80313@gmail.com)
+* Approver: a11r, Mark D. Roth (roth@google.com)
```
markdroth, line 4: You can set this to `markdroth`.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Overview of `weighted_round_robin` policy
```
markdroth, line 28: I'd prefer to call this policy something other than "weighted_round_robin", since we are planning to eventually provide a policy with that name that works slightly differently: instead of getting the endpoint weights along with the addresses, the policy will automatically compute the endpoint weights based on utilization data reported by the endpoints.

I can't offhand come up with a really ideal suggestion for a better name here.  I can think of the following options, none of which seem quite ideal:

- address_weighted_round_robin
- injected_weighted_round_robin
- weighted_round_robin_with_injected_weights

I'm open to other options here.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+`weighted_round_robin` policy is powered by [EDF](https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling) picker. EDF picker maintains a priority queue of `EdfEntry`. On the top of the queue, it’s the entry with lowest `deadline`. `order_offset` is the tie breaker when two entries have same deadline to maintain FIFO order.
+
+Proposed `EdfEntry`
+```
+struct EdfEntry {
+    // primary key for the priority queue. The entry with least deadline is the top of the queue.
+    double deadline;
+    // secondary key for the priority queue. Used as a tiebreaker for same deadline to maintain FIFO order.
+    uint64 order_offset;
+    // `load_balancing_weight` of this endpoint from the latest `ClusterLoadAssignment`.
```
markdroth, line 39: I would prefer that this LB policy be implemented in a way that is not xDS-specific.  It should basically just use the endpoint weight attributes passed into it, and it should not care where those weights come from.  That way, it will be possible to also use this policy without xDS, by writing a custom resolver that provides the same endpoint weight attributes.

To say this another way, I think there are two distinct parts to this proposal: (1) changing the xDS code to populate the addresses' endpoint weight attributes based on the EDS data, and (2) implementing the new LB policy that uses the addresses' endpoint weight attributes.  I suggest structuring this document in a way that makes it clear that these are two separate changes.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+    double weight;
+    // Subchannel data structure of this endpoint.
+    Subchannel subchannel;
+}
+```
+
+- On each call to the `Pick`, EDF picker picks the entry `e` on the top of the queue, returns the subchannel associated with the entry. After that, picker updates the `deadline` of `e` to `e.deadline + 1/weight` and either performs a pop and push the entry back to the queue or key increase operation.
+- `weighted_round_robin` updates the entries in EDF priority queue on any change of subchannel `ConnnectiveState` or endpoint `load_balancing_weight`.
+- If all endpoints have the same `load_balancing_weight`, `EDF` picker degenerates to `round_robin` picker. It's easier to reason and consistent with envoy.
+- Endpoints do not have `load_balancing_weight` assigned are discarded.
```
markdroth, line 49: I don't think this is the behavior we want.  The comment on the `load_balancing_weight` field says "If unspecified, each host is presumed to have equal weight in a locality."

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+- `weighted_round_robin` should always be updated to the lastest `ClusterLoadAssignment`. It's xDS server's responsibility to maintain consistency.
+
+#### EDF picker interface
+```
+/*
+Add new SubChannel to EDF picker.
+
+weighted_round_robin sees a new SubChannel in the most recent ClusterLoadAssignment or ConnectivityState of SubChannel is READY again.
+*/
+Add(SubChannel, weight)
```
markdroth, line 59: The intent of our LB policy API is that a picker object should generally be immutable from the outside.  When the routing state changes, the LB policy creates a new picker and gives it to the channel, which will use it to replace the previous picker.  So there's no need for these kinds of mutable methods on the picker.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+*/
+Update(SubChannel, newWeight)
+
+/*
+`Pick` picks the EdfEntry on top of the queue, returns the underlying SubChannel and updates deadline of the EdfEntry.
+*/
+PickResult Pick(PickArgs args)
+```
+
+### On new `ClusterLoadAssignment`
```
markdroth, line 82: When an EDS update is received, an update will be sent to the new weighted_round_robin policy (whatever we decide to call it). The policy will then create a new picker and return that new picker to the channel, and that new picker will be used for all subsequent picks.

This same flow will be used for any change to EDS data, which includes all three of the changes described below.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+### On endpoint ConnectiveState Update
+- Udpate EDF priority queue by adding/removing entries to reflect the change immediately.
+
+## Rationale
+
+Several applications can be built upon this feature, e.g. utilization load balancing, blackhole erroring endpoints, load testing,... etc.
+
+The reason to refresh EDF picker even there is only weight change on some endpoints which is different from envoy is because we'd like real time traffic shift for use cases like load testing, blackhole erroring endpoints.
+
+The reasons to introduce a new algorithm instead of using existing `weighted_target` policy are
```
markdroth, line 100: The `weighted_target` policy would not work for this anyway, since it's designed to distribute traffic across child policies, not across endpoints.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+- [EDF](https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling) maintains FIFO order for endpoints with same weight which is easier to reason.
+- We want to be consistent with the behavior of Envoy.
+
+## Implementation
+
+N/A
+
+## Open issues (if applicable)
+
+- Replace `round_robin` with `weighted_round_robin`?
```
markdroth, line 110: There will still be cases where `round_robin` is useful, so I don't think there's any need to replace it.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+- Following first issue, if not, do we want to use `weighted_round_robin` as default lb_policy for eDS response?
```
markdroth, line 111: Since this is more in-line with Envoy's behavior, I think we can use this policy when the CDS response sets the LB policy to `ROUND_ROBIN` (which is the only value we accept today, although that may change in the not-too-distant future).

markdroth (commented):
Sorry for taking so long to get to this!

Please let me know if you have any questions about any of this.

A34-weighted-round-robin.md
dfawley: How about using "utilization_weighted_round_robin" for the other policy?  That policy could potentially be implemented as a wrapper of this policy, FWIW.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+
+`weighted_round_robin` policy is powered by [EDF](https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling) picker. EDF picker maintains a priority queue of `EdfEntry`. On the top of the queue, it’s the entry with lowest `deadline`. `order_offset` is the tie breaker when two entries have same deadline to maintain FIFO order.
+
+Proposed `EdfEntry`
+```
+struct EdfEntry {
+    // primary key for the priority queue. The entry with least deadline is the top of the queue.
+    double deadline;
+    // secondary key for the priority queue. Used as a tiebreaker for same deadline to maintain FIFO order.
+    uint64 order_offset;
```
dfawley, line 38: How this works needs to be documented.  Is the smaller value preferred?  What if there is a tie here?  Is that illegal?  Do we validate that?

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+    // `load_balancing_weight` of this endpoint from the latest `ClusterLoadAssignment`.
+    double weight;
+    // Subchannel data structure of this endpoint.
+    Subchannel subchannel;
+}
+```
+
+- On each call to the `Pick`, EDF picker picks the entry `e` on the top of the queue, returns the subchannel associated with the entry. After that, picker updates the `deadline` of `e` to `e.deadline + 1/weight` and either performs a pop and push the entry back to the queue or key increase operation.
+- `weighted_round_robin` updates the entries in EDF priority queue on any change of subchannel `ConnnectiveState` or endpoint `load_balancing_weight`.
```
dfawley, line 47: Should that be `ConnectivityState`?

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+- If all endpoints have the same `load_balancing_weight`, `EDF` picker degenerates to `round_robin` picker. It's easier to reason and consistent with envoy.
```
dfawley, line 48: Do we use the `order_offset` to determine the order in this case?

A34-weighted-round-robin.md
dfawley: That doesn't really make sense either.  If only one address has a missing weight, what do we assign it?  The average of the defined weights?

dfawley:
cc @menghanl

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+Update the weight of SubChannel
+
+There is weight change on Subchannel in the most recent ClusterLoadAssignment comparing to last ClusterLoadAssignment
+*/
+Update(SubChannel, newWeight)
+
+/*
+`Pick` picks the EdfEntry on top of the queue, returns the underlying SubChannel and updates deadline of the EdfEntry.
+*/
+PickResult Pick(PickArgs args)
```
menghanl, line 79: Pick would be only public API of picker. The other "update"s should be internal to the balancing policy.

A34-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+### On new `ClusterLoadAssignment`
+
+#### endpoints have weight change
```
menghanl, line 84: It would be helpful to add more details of how the change will be handled. E.g., only the weight is updated, but not the deadline, so the entry at top of the queue is still at the top?

Similar for the following two updates. This helps to keep the implementations consistent cross languages.

A34-weighted-round-robin.md
markdroth: That's an interesting idea -- I hadn't thought of wrapping one with the other.

In order to do this, we'd have to use the same WRR algorithm in both cases.  @menghanl, you've looked into how our internal WRR algorithm works.  How does that compare to the EDF algorithm proposed here?

A34-weighted-round-robin.md
markdroth: Good point.  I suggest checking the Envoy implementation to see how they handle this.  I suspect they just assume a weight of 1 if the field is not set.  We can also ask them to improve the comment.

A34-weighted-round-robin.md
menghanl: Wrapping one with another means the wrapper needs to convert the loads into weights, and send the new weights along with the addresses to the child whenever the load changes?

I would think if we make an EDF scheduler that's capable of handling both situations, the two balancers can reuse this scheduler implementation.

The WRR algorithm is similar to the internal one. I will double check the details.

A34-weighted-round-robin.md
yishuT: https://github.com/envoyproxy/envoy/blob/5d95032baa803f853e9120048b56c8be3dab4b0d/source/common/upstream/upstream_impl.cc#L1243 and https://github.com/envoyproxy/envoy/blob/5d95032baa803f853e9120048b56c8be3dab4b0d/source/common/upstream/upstream_impl.cc#L359

I think you are right. it uses default value which is 0, and in the constructor it's adjusted to 1. I am not sure it's good default behavior because the endpoints with unspecified weight receive same amount of traffic with endpoints specified with 1. However, I don't know how to assign it properly either and I think it's control plane's job to assign proper weight for each endpoint. I will make the proposal same as envoy behavior if nobody has other concerns

A34-weighted-round-robin.md
yishuT: I have no preference toward the name. Per discussion here, I would leave it as `weighted_round_robin` and other policy can build upon it. Or maybe `weighted_round_robin_by_edf` to indicate underlying impl?

A34-weighted-round-robin.md
yishuT: kk. I will update the doc as suggested structure with two parts (1) and (2).

yishuT:
Thanks for all the feedback. I will post another update to address all of them next couple of days. Please feel free to leave any comments in the meantime.

yishuT:
soft ping. any udpates?

A34-weighted-round-robin.md
```diff
@@ -1,103 +1,101 @@
-`weighted_round_robin` lb_policy for per endpoint weight from `ClusterLoadAssignment` response
+`weighted_round_robin` lb_policy for per endpoint `load_balancing_weight` from `ClusterLoadAssignment` response
 ----
 * Author(s): Yi-Shu Tai (echo80313@gmail.com)
-* Approver: a11r, Mark D. Roth (roth@google.com)
-* Status: Draft
+* Approver: a11r, markdroth
```
markdroth, line 4: You can remove a11r.  I can approve this when it's ready.

A34-weighted-round-robin.md
markdroth: @menghanl, did you have a chance to look up how the algorithm compares to the internal WRR algorithm?

I guess for now, let's call this `edf_weighted_round_robin`.

A34-weighted-round-robin.md
```diff
@@ -1,103 +1,101 @@
...
+    // maintain FIFO order. If there is a tie on deadline of two entries, the one with
+    // smaller `order_offset` will have higher priority. `order_offset` is assigned to this
+    // entry on constructing the priority queue for the first time and it's immutable.
+    // Also the `order_offset` assigned to entries is strictly increasing, in other words,
+    // no two entries have same `order_offset`.
     uint64 order_offset;
-    // `load_balancing_weight` of this endpoint from the latest `ClusterLoadAssignment`.
+
+    // `load_balancing_weight` of this endpoint from address attribute of this endpoint.
     double weight;
```
markdroth, line 52: Why is this a `double` instead of a `uint32_t`?  The endpoint weight in the xDS API is an integer, and integers are usually a lot less error-prone to work with, because there's no rounding error.

Same question for the `deadline` field.  Is there a way we can represent is as an integer instead of a double?  For example, can we simply multiply by 1000 and say that we are storing milli-units?

A34-weighted-round-robin.md
```diff
@@ -1,103 +1,101 @@
...
+
     // Subchannel data structure of this endpoint.
     Subchannel subchannel;
 }
 ```
+Initialization
+- At the very beginning, `deadline` of an entry `e` is equal to `1/e.weight`.
+- We assign `order_offset` to each entry while constructing the priority queue. `order_offset` assigned to an entry is distinct and nonnegative interger. During the whole lifecycle of this picker, `order_offset` of an entry is unchanged.
```
markdroth, line 60: s/interger/integer/

A34-weighted-round-robin.md
```diff
@@ -1,103 +1,101 @@
...
-
-Subchannel is removed from the most recent ClusterLoadAssignment comparing to last ClusterLoadAssignment or
-ConnectivityState of SubChannel becomes not READY.
-*/
-Remove(SubChannel)
-
-/*
-Update the weight of SubChannel
+#### Service Config
+The service config for `weighted_round_robin` is very similar to `round_robin`
```
markdroth, line 70: I think this can just say something like "the config for the edf_weighted_round_robin LB policy is an empty proto message".

A34-weighted-round-robin.md
```diff
@@ -106,6 +104,4 @@ The reasons to introduce a new algorithm instead of using existing `weighted_tar
 N/A
 
 ## Open issues (if applicable)
-
-- Replace `round_robin` with `weighted_round_robin`?
-- Following first issue, if not, do we want to use `weighted_round_robin` as default lb_policy for eDS response?
\ No newline at end of file
+- Do we need a way to opt out WRR even per endpoint weight is assigned by ClusterLoadAssignment?
```
markdroth, line 107: If the new policy behaves exactly like round_robin when no endpoint weights are provided by the xDS server, then we probably won't need an opt-out mechanism.  But we'll need to verify that that's the case.

A34-weighted-round-robin.md
```diff
@@ -1,103 +1,101 @@
...
 ## Rationale
 
 Several applications can be built upon this feature, e.g. utilization load balancing, blackhole erroring endpoints, load testing,... etc.
 
 The reason to refresh EDF picker even there is only weight change on some endpoints which is different from envoy is because we'd like real time traffic shift for use cases like load testing, blackhole erroring endpoints.
 
-The reasons to introduce a new algorithm instead of using existing `weighted_target` policy are
+The reasons to introduce a new algorithm instead of re-using the same algorithm of `weighted_target` policy are
 - [EDF](https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling) maintains FIFO order for endpoints with same weight which is easier to reason.
 - We want to be consistent with the behavior of Envoy.
```
markdroth, line 100: There is one important architectural difference between Envoy and gRPC to be aware of here.  In Envoy, the LB policy doesn't handle connection management; it just chooses from the set of configured hosts, and once it does that, Envoy figures out if it has an existing connection to that host or if it needs to create a new one.  But in gRPC, the LB policy is responsible for deciding when to connect to the endpoints.  (The gRPC LB policy also needs to monitor the connectivity state of all subchannels, and it must tell the channel to use only subchannels that are actually connected.  This generally means that the LB policy needs to create a new picker whenever the connectivity state of a subchannel changes.)

The `round_robin` LB policy's connection management behavior is that it always tries to keep one connection open to each address in the address list at all times.  When the policy is first instantiated, it immediately tries to connect to all addresses, and whenever a subchannel becomes disconnected, it immediately tries to reconnect.

This behavior is different from how Envoy's connection pools work.  But given that the main purpose of the `round_robin` policy is to spread load evenly across the backends, that implies that the client will be generating a fair amount of traffic, which implies that proactively connecting this way is generally better.  The Envoy approach actually causes increased latency, because the first RPC sent to each endpoint winds up needing to wait for it to connect, even if there's an existing connection to a different endpoint that could handle the traffic.

I think that if we want this new policy to be a drop-in replacement for the existing `round_robin` policy, we'll need to use this same connection management behavior.

markdroth (commented):
Sorry for the delay on this!

A34-weighted-round-robin.md
yishuT: I make weight double because of deadline is double. I am not sure we can use `uint32` for deadline.  On updating deadline of an entry after a pick, one way to implement (or at least envoy does)
```
new_deadline = deadline + 1/weight
```
if we want to use `uint32`, we likely need to 
```
new_deadline = deadline + 1000/weight
```
or similar. This will need the assumption that per endpoint weight is never larger than 1000.

A34-weighted-round-robin.md
yishuT: based on
```
 Endpoints do not have `load_balancing_weight` is assigned to 1 (the smallest possible weight). 
```

yes, it will behave same as simple RR.

A34-weighted-round-robin.md
yishuT: Thanks, this is really helpful. I added a section `Subchannel connectivity management`. I just slightly modify how rr manages connectivity from this comment. Hope there is no copyright issue : )

A34-weighted-round-robin.md
markdroth: If the deadlines and weights are all multiplied by 1000, then there's no requirement that endpoint weight is never larger than 1000.  We are doing exactly the same calculation as we would be doing with doubles, except that we're storing milliunits instead of whole units.  The only thing we're losing is the ability to store more 3 decimal places of precision, and that seems like it shouldn't really affect the algorithm in a material way.

yishuT: hmmm, sry, I am lost. how to handle the inverse of weight `1/weight`, anything larger than the numerator is rounded to 0?

markdroth: Okay, sorry, let me try this again.

The idea here is that you can do the actual division using a double, just don't use doubles for storage.  I think we don't actually need to store the weight as a double at all; it comes as an integer in xDS, and we can keep it that way.  We can then do the calculation as a double but immediately round the result for storage.

I think this can be done using something like this:

```
int32_t weight = <weight from xDS>;
int32_t deadline += std::lround(1000.0 / weight);
```

Then we continue to use the lowest deadline value as the sorting key.

Does that seem reasonable, or is there something I'm missing about this algorithm?  The latter is quite possible, since I haven't grokked it fully -- for example, if deadline is constantly increasing, how does it avoid wrap-around?  Is it depending on the `double` representation to avoid that?

yishuT: ah, I got the point, and it totally makes sense. I tried to avoid type casting, but yeah, we can avoid storing double. 

I don't remember there is any handling of wrap around out of head, so it depends on the double representation, and I think we need uint64 to stand against wrap around.

uint32 ~ `4*10^9 / 1000 ~ 4*10^6` requests
uint64 ~ `1.8*10^19 / 1000 ~ 1.8*10^16` requests

markdroth: We should probably figure out how the wrap around is handled, just to make sure we understand how this is supposed to work.  Can you figure out how Envoy does that and add info about it to this gRFC?

yishuT: ugh, Monday is not suitable for brainstorming work. 

```
std::lround(1000.0 / weight);
```
still goes to 0 if `weight` > 1000. I am sorry I keep back and forth on this. The issue I am confused is when `weight` is larger we need more precision in `deadline`.

re wrapping deadline:
per chat with @htuch:
it takes forever for double deadline to overflow, so no logic to handle it. Same argument can apply to uint64 i think.

Also, @htuch mentioned that envoy is reconsidering the use of EDF.
It's hard to avoid all grpc clients perform synchronized pick.
I will add this to open issues

markdroth: I see, so the problem is not the multiplying by 1000; it's the fact that the algorithm is actually counting on way more than 3 decimal digits' worth of precision.  Okay, then I guess we have no choice but to use a double here.  That's unfortunate, but I guess the use of it is fairly constrained here, so it shouldn't cause too many bugs.

What is Envoy considering using instead of EDF?  It seems like it would be better to not implement something that we'll immediately move away from.

A34-edf-weighted-round-robin.md
```diff
@@ -0,0 +1,111 @@
...
+The reasons to introduce a new algorithm instead of re-using the same algorithm of `weighted_target` policy are
+- [EDF](https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling) maintains FIFO order for endpoints with same weight which is easier to reason.
+- We want to be consistent with the behavior of Envoy on how lb_policy picks the backend for sending the request. However, there is a difference between `edf_weighted_round_robin` and `edf_scheduler` of Envoy. `edf_weighted_round_robin` actively monitors the connectivity of each subchannel but `edf_scheduler` of Envoy does not.
+
+## Implementation
+
+N/A
+
+## Open issues (if applicable)
+* How to desync to avoid all clients do synchronized pick?
```
markdroth, line 111: I don't understand what this is referring to.  Can you explain the problem here?

yishuT: updated. Please let me know it's still not clear

A34-weighted-round-robin.md
yishuT: @htuch do you have timeline if envoy is going to discard EDF?

A34-edf-weighted-round-robin.md
markdroth: Okay, I understand the problem.  I see two possible solutions here:

- The simple approach is, as you suggest, randomize the starting point in the list whenever we create a new picker, just like `round_robin` does.  This will avoid having all clients hammer the same endpoints at the same time (and also, keep in mind that the control plane could actually send different results to different clients -- each client could see different endpoints or the same endpoints in a different order or with different weights).  But it will disrupt the expected scheduling whenever the picker changes.

- A more complicated solution would be to maintain the current scheduler state in the LB policy and use a mutex to synchronize access to it between the picker and the LB policy.  This is more complicated to implement, and the synchronization imposes a performance penalty (because you have to acquire the mutex for every pick), but it should work.

A34-weighted-round-robin.md
htuch: @tonya11en is currently investigating alternatives, the current leading candidate is IWRR. The motivation is to provide a robust solution around desyncing, which is something that came up in work done by @mattklein123 and @snowp. See https://github.com/tonya11en/envoy/pull/1 for some early investigation, there may be other performance advantages of IWRR as well.

yishuT: If Envoy will have a decision soon, say this Q, let's put this PR on hold?

htuch: I think we'll know in the next few weeks.

tonya11en: I'm planning to make a more rigorous simulation over the weekend. Hopefully something we can just check in and use for other LBs. Main focus is on what the hosts in the pool experience w.r.t. syncing behavior. The perf numbers are more favoriable to IWRR, so we'll just need to see if there's a catch or unwanted change in behavior.

Stay tuned.

yishuT: Howdy folks! Happy New Year

@tonya11en do we have a final decision that envoy will use iwrr to replace edf?

tonya11en: It looks like we'll be replacing EDF with WRSQ (in review https://github.com/envoyproxy/envoy/pull/14681) for weighted RR among other things. More details on where it'll be used in https://github.com/envoyproxy/envoy/issues/14597.

yishuT: Gotcha, thanks

yishuT:
@markdroth  I want to revisit this proposal.  Does gRPC community have any plan for per endpoint weight LB policy?

markdroth:
We don't have any other plan to do this, so if it's still something you'd like, please feel free to continue moving this design forward.

s-matyukevich:
I am confused about the status of this proposal. It is not yet merged, but looks like we are already using endpoint weights, at least in grpc-go.

* [Here](https://github.com/grpc/grpc-go/blob/08479c5e2ecb7d30f5210fc74127179745ffca46/xds/internal/xdsclient/xdsresource/unmarshal_eds.go#L85)  EDS response is parsed and weights are stored.
* Next we use the weight property to configure weighted round robin LB in the EDS (clusterresolver) load balancer [link](https://github.com/grpc/grpc-go/blob/f51d21267df53c7f4e3533e18fdf5b61c76e005d/xds/internal/balancer/clusterresolver/configbuilder.go#L393)
* We also use those weights in the ringhash  EDS implementation [link](https://github.com/grpc/grpc-go/blob/f51d21267df53c7f4e3533e18fdf5b61c76e005d/xds/internal/balancer/clusterresolver/configbuilder.go#L360)

yishuT:
Hey @s-matyukevich

afaik, weights are stored but grpc client does not have weighted round robin implemented for every language. The proposal is to have a unified protocol for implementation.

PR #203, opened 2020-08-20
L75: Proposal to remove grpc_channel_ping from core surface API

yashykt:


L75-core-remove-grpc-channel-ping
```diff
@@ -0,0 +1,37 @@
+Title
```
markdroth, line 1: Please set the title.

Also, please rename this file so that it has a ".md" extension.

yashykt: Done.

PR #205, opened 2020-09-08
L46: C-core: New TLS Credentials API

ZhenLian:


dfawley:
Please include the language in the title for "L" proposals.  Thanks!

ZhenLian:
> Please include the language in the title for "L" proposals. Thanks!

Sure, sounds good!

jiangtaoli2016 (commented):
A few comments after quick look. Will do detailed review later today.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
+Title
```
markdroth, line 1: Please change the title to "L46: C-core: New TLS Credentials API".

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+----
+* Author(s): ZhenLian
+* Approver: markdroth, yashykt, jiangtaoli2016
+* Status: Draft
+* Implemented in:
+* Last updated: September 7, 2020
+
+## Abstract
+
+This proposal aims to enhance the security features provided by [this already implemented proposal](https://github.com/grpc/proposal/pull/98).
```
markdroth, line 11: It looks like this proposal is designed to describe just the deltas relative to that earlier PR.  However, that earlier PR was never approved or merged, which is why the existing TLS creds API is still marked experimental.

I have closed the earlier PR in favor of this one.  This PR should be rewritten such that instead of referring to the earlier one, it completely replaces it.  In other words, it should describe the entire TLS creds API, not just the recent changes.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+
+This proposal aims to enhance the security features provided by [this already implemented proposal](https://github.com/grpc/proposal/pull/98). 
+The improvements include:
+1. replacing the per-connection callback-based credential reloading mechanism with the new distributor/provider mechanism
+2. changing the server authorization check to an authorization check that could be applied to both the client side and the server side
+3. adding options for users to configure their TLS version 
+
+## Background
+
+The current [TLS credential API](https://github.com/grpc/grpc/blob/master/src/core/lib/security/credentials/tls/tls_credentials.h) in gRPC C core implemented the following two features:
```
markdroth, line 19: This link points to internal implementation details, not to the C-core API.  While it does make sense for this proposal to discuss implementation details that are relevant to understanding the C-core API, we should focus more on the latter and include only those implementation details that are important for understanding the API.

markdroth (commented):
Please let me know if you have any questions about any of this.  Thanks!

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+### Related Proposals:
+
+[L46: Create a new TLS credential API that supports SPIFFE mutual TLS](https://github.com/grpc/proposal/pull/98)
+
+## Proposal
+
+### Changes around TLS credentials and Options
+This part of the proposal introduces changes made to existing TLS credential APIs. Changes include:
+1. Replacing all the occurrence of "*_server_authorization_*" to "*_authorization_"
+2. Removing `grpc_tls_key_materials_config`, `grpc_tls_credential_reload_config` and `grpc_tls_credential_reload_arg`
```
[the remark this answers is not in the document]
ZhenLian: Yes, currently wrap languages would directly use `const char *` for root certs and `grpc_ssl_pem_key_cert_pair` for identity certs(See grpc_tls_certificate_distributor_set_key_materials under Distributor section).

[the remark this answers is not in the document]
jiangtaoli2016: okay, by reading the gRFC, it infers that gRPC only provides a default certificate provider for reloading credentials from filesystem.

L46-core-tls-credential-API.md
jiangtaoli2016: +1 Agree with Mark. Let's describe the entire TLS creds API.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+    handshake.  */
+GRPCAPI void grpc_tls_server_credentials_set_min_tls_version(grpc_tls_credentials_options* options,
+    grpc_tls_version min_tls_version);
+
+/** Sets the maximum TLS version that will be negotiated during the TLS
+    handshake. */
+GRPCAPI void grpc_tls_server_credentials_set_max_tls_version(grpc_tls_credentials_options* options, 
+    grpc_tls_version max_tls_version);
+
+typedef struct grpc_tls_authorization_check_arg grpc_tls_authorization_check_arg;
```
jiangtaoli2016, line 106: Please explain how `grpc_tls_authorization_check_config` and `grpc_tls_authorization_check_arg` work.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+    const char* identity_cert_name);
+
+/** Sets the credential provider. */
+GRPCAPI int grpc_tls_credentials_options_set_certificate_provider(
+    grpc_tls_credentials_options* options,
+    grpc_tls_certificate_provider* provider);
+
+/** Sets the minimum TLS version that will be negotiated during the TLS
+    handshake.  */
+GRPCAPI void grpc_tls_server_credentials_set_min_tls_version(grpc_tls_credentials_options* options,
```
jiangtaoli2016, line 98: why `grpc_tls_server_credentials_set_min_tls_version`?
It applies to both client and server. And we are setting options, but on credentials.
It should be `grpc_tls_credentials_options_set_min_tls_version`. Same for one below.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+
+typedef struct grpc_tls_error_details grpc_tls_error_details;
+
+typedef struct grpc_tls_authorization_check_config grpc_tls_authorization_check_config;
+
+typedef struct grpc_tls_credentials_options grpc_tls_credentials_options;
+
+GRPCAPI grpc_tls_credentials_options* grpc_tls_credentials_options_create(void);
+
+GRPCAPI int grpc_tls_credentials_options_set_cert_request_type(
```
jiangtaoli2016, line 71: please specify the default value if it is not called. Same for `grpc_tls_credentials_options_set_server_verification_option` below

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+GRPCAPI int grpc_tls_credentials_options_set_identity_cert_name(
+    grpc_tls_credentials_options* options,
+    const char* identity_cert_name);
+
+/** Sets the credential provider. */
+GRPCAPI int grpc_tls_credentials_options_set_certificate_provider(
+    grpc_tls_credentials_options* options,
+    grpc_tls_certificate_provider* provider);
+
+/** Sets the minimum TLS version that will be negotiated during the TLS
```
jiangtaoli2016, line 96: Please comment that if this function is not called, then the underlying SSL library will pick min and max TLS version automatically.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+external_provider->destroy = Destroy;
+/* The ownership will be taken by provider. */
+grpc_tls_certificate_provider* provider grpc_tls_certificate_provider_external_create(external_provider);
+
+```
+
+## Wrap Languages
+Each wrap language might have their own way to wrap these core APIs, but in general, there are a few things to consider:
+1. create an interface-like class `ProviderInterface` which contains two functions GetDistributor() and Destroy()
+2. create a concrete class `FileCertificateProvider` that implements `ProviderInterface` and wraps the provider created by `grpc_tls_certificate_provider_file_watcher_create`
```
jiangtaoli2016, line 248: We need a concrete class `StaticCertificateProvider` that takes static certificates from memory, passed by application. So that we will be feature parity with `SslCredentialsOptions`.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,339 @@
...
+
+GRPCAPI int grpc_tls_credentials_options_set_server_verification_option(
+    grpc_tls_credentials_options* options,
+    grpc_tls_server_verification_option server_verification_option);
+
+GRPCAPI int grpc_tls_credentials_options_set_authorization_check_config(
+    grpc_tls_credentials_options* options,
+    grpc_tls_authorization_check_config* config);
+
+GRPCAPI int grpc_tls_credentials_options_set_root_cert_name(
```
jiangtaoli2016, line 83: Please clarify what cert_name means. What if certificate does not have a name. Most of users may not know or care what certificate name is and how it is used. You can clarify this in the background section.

L46-core-tls-credential-API.md
```diff
@@ -1,236 +1,328 @@
...
+3. The credentials reloading should be efficient. We shall reload only when needed rather than reloading per handshake.
 
-This proposal introduces a new credential API model that allows flexible addition of TLS-related features that can be implemented in either a synchronous or asynchronous manner. It then proposes new credential API's for both HTTPS and SPIFFE-based TLS that complies with the API model.
+Based on the requirements, we are going to propose a new distributor/provider API that will meet these requirements. 
+The distributor is responsible for caching the credentials and distributing them in each handshake. 
+The users will need to specify a particular provider implementation when creating their credentials, and we will provide one implementation for reloading from files.
+If that doesn't meet users' requirements, they could implement their own providers.
+
+Besides major credential reloading change, we are also planning to:
+1. make the server authorization original designed for client side only applicable to both the client side and the server side.
```
ak800i, line 40: original -> originally

L46-core-tls-credential-API.md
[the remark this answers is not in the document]
ZhenLian: Done, added `StaticCertificateProvider`.

L46-core-tls-credential-API.md
ZhenLian: Sure. I revised this PR. Could you please take a look again? Thanks!

L46-core-tls-credential-API.md
ZhenLian: Yeah, renamed these two functions.

ZhenLian:
@yihuazhang Hey Yihua, can you please take a look at the `Custom Authorization` section? That part remains almost unchanged. I included some examples a the end, and it would be great to have you double check that the logic is correct. Thank you so much!

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+
+## Background
+
+The current TLS implementation in gRPC C core has some restrictions:
+1. lack of the credential reloading support. Once an application starts, it can never reload its identity key/cert-chain and root certificates without shutting down
+2. inflexibility in the peer verification. In a mutual TLS scenario, servers would always perform certificate verification, and clients would always perform certificate verification as well as default hostname check. No customized checks can be introduced.
+3. inability to choose the TLS version that will be used
+
+There have always been some ad hoc requests on GitHub Issue Page for supporting these features. The demands for a more flexible and configurable API have increased with the emerging needs in cloud platform.
+One common example is to support SPIFFE ID as the identity format, which requires us to disable the default hostname check on the client side.
```
yihuazhang, line 24: SPIFFE ID and hostname verification check are orthogonal to each other. Nothing prevents a certificate containing both SPIFFE ID as URI and hostname as dnsName SAN fields, in which case, hostname verification check can still be performed on the dnsName field.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+1. credential reloading: reload transport credentials periodically from various sources, e.g. from credential files, or users' customized Provider implementations, without shutting down
+2. custom authorization check: optionally bypass the certificate check, bypass the hostname check on the client side, and perform an additional per-connection authorization callback at the end of authentication
+3. TLS version configuration: optionally configure the minimum and maximum of the TLS version that will be used in the handshake
+
+## Background
+
+The current TLS implementation in gRPC C core has some restrictions:
+1. lack of the credential reloading support. Once an application starts, it can never reload its identity key/cert-chain and root certificates without shutting down
+2. inflexibility in the peer verification. In a mutual TLS scenario, servers would always perform certificate verification, and clients would always perform certificate verification as well as default hostname check. No customized checks can be introduced.
+3. inability to choose the TLS version that will be used
```
yihuazhang, line 21: Maybe rephrase it to "inability to choose the maximum TLS version supported in OpenSSL/boringSSL for use""? as I believe users always prefer TLS 1.3 over TLS 1.2 when both are available?

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+
+There have always been some ad hoc requests on GitHub Issue Page for supporting these features. The demands for a more flexible and configurable API have increased with the emerging needs in cloud platform.
+One common example is to support SPIFFE ID as the identity format, which requires us to disable the default hostname check on the client side.
+Another example is to reload the identity certificates without shutting down the server, because certificates will eventually expire and the ability to rotate the certificates is essential to a cloud application.
+
+Hence we proposed an API that will meet the following requirements:
```
yihuazhang, line 27: nit. s/proposed/propose

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+## Background
+
+The current TLS implementation in gRPC C core has some restrictions:
+1. lack of the credential reloading support. Once an application starts, it can never reload its identity key/cert-chain and root certificates without shutting down
+2. inflexibility in the peer verification. In a mutual TLS scenario, servers would always perform certificate verification, and clients would always perform certificate verification as well as default hostname check. No customized checks can be introduced.
+3. inability to choose the TLS version that will be used
+
+There have always been some ad hoc requests on GitHub Issue Page for supporting these features. The demands for a more flexible and configurable API have increased with the emerging needs in cloud platform.
+One common example is to support SPIFFE ID as the identity format, which requires us to disable the default hostname check on the client side.
+Another example is to reload the identity certificates without shutting down the server, because certificates will eventually expire and the ability to rotate the certificates is essential to a cloud application.
```
yihuazhang, line 25: nit. s/identity certificates/credential  Same for other occurrences in the same paragraph.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+* Approver: markdroth, yashykt, jiangtaoli2016
+* Status: Draft
+* Implemented in:
+* Last updated: September 9, 2020
+
+## Abstract
+
+This proposal aims to provide the following security features in a new TLS credential API:
+1. credential reloading: reload transport credentials periodically from various sources, e.g. from credential files, or users' customized Provider implementations, without shutting down
+2. custom authorization check: optionally bypass the certificate check, bypass the hostname check on the client side, and perform an additional per-connection authorization callback at the end of authentication
```
yihuazhang, line 13: How about re-phrasing this paragraph as follows?

Support custom verification on server identity: 

- Perform both certificate and hostname verification checks (default option).
- Perform certificate verification but skip hostname verification check.
- Skip both certificate and hostname verification checks.

In the last two cases, custom server authorization check MUST be performed.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+3. inability to choose the TLS version that will be used
+
+There have always been some ad hoc requests on GitHub Issue Page for supporting these features. The demands for a more flexible and configurable API have increased with the emerging needs in cloud platform.
+One common example is to support SPIFFE ID as the identity format, which requires us to disable the default hostname check on the client side.
+Another example is to reload the identity certificates without shutting down the server, because certificates will eventually expire and the ability to rotate the certificates is essential to a cloud application.
+
+Hence we proposed an API that will meet the following requirements:  
+
+1. Flexible enough to support multiple use cases. Users should be able to use their own implementations if the provided ones don't fit their needs
+2. C core API needs to be simple, so that it can be easily wrapped by different languages
```
yihuazhang, line 30: Can we be more specific on it? It's not clear to me how "simple" the C-core API should be in order to get wrapped by other languages.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+3. The credentials reloading should be efficient. We shall reload only when needed rather than reloading per handshake
```
yihuazhang, line 31: Maybe change it to "per TLS connection"? More than one handshakes are performed per gRPC connection, and it's not clear which handshake it refers to.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+----
+* Author(s): ZhenLian
+* Approver: markdroth, yashykt, jiangtaoli2016
+* Status: Draft
+* Implemented in:
+* Last updated: September 9, 2020
+
+## Abstract
+
+This proposal aims to provide the following security features in a new TLS credential API:
```
yihuazhang, line 11: nit. s/security features/features? as I believe most of the proposed features are either related to flexibility (e.g., custom server identity verification) or performance (e.g., credential reloading), but not security.

yihuazhang (commented):
Thanks Zhen for the PR. I left some comments on all sections but proposal, which I will review in more details later.

L46-core-tls-credential-API.md
ZhenLian: Yeah by saying "default hostname check", I meant the exact check between the hostname sent from client and the identity shown on the server's certificate. If using SPIFFE, the identity is a SPIFFE ID and doesn't match the hostname sent from client. I added some explanations. Does it look better now?

L46-core-tls-credential-API.md
ZhenLian: Sure, I added "maximum/minimum" instead of "maximum" though since that's the API implies...

L46-core-tls-credential-API.md
ZhenLian: SG! That's a better format of wording.
I reworded it a bit because we are now also doing custom authorization on server side, for reasons to support CRL in the future, etc. And that's the reason I replaced all the "server authorization" with "authorization" throughout this proposal.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,450 @@
...
+2. inflexibility in the peer verification. In a mutual TLS scenario, servers would always perform certificate verification, and clients would always perform certificate verification as well as default hostname check. No customized checks can be introduced.
+3. inability to choose the TLS version that will be used
+
+There have always been some ad hoc requests on GitHub Issue Page for supporting these features. The demands for a more flexible and configurable API have increased with the emerging needs in cloud platform.
+One common example is to support SPIFFE ID as the identity format, which requires us to disable the default hostname check on the client side.
+Another example is to reload the identity certificates without shutting down the server, because certificates will eventually expire and the ability to rotate the certificates is essential to a cloud application.
+
+Hence we proposed an API that will meet the following requirements:  
+
+1. Flexible enough to support multiple use cases. Users should be able to use their own implementations if the provided ones don't fit their needs
```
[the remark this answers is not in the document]
ZhenLian: I revised a bit. Please take a look, thanks!

L46-core-tls-credential-API.md
ZhenLian: That's a previous requirement. I think as time goes, this API becomes a bit more complicated, and some designs might be needed when wrapping the Core API. 
@jiangtaoli2016 
Would it be appropriate to remove this point here? I think our goal here is to make the API wrappable with other languages, but the simplicity is a bit vague and we have no way to measure it.

jiangtaoli2016: I'd rephrase this as 
C core API needs to be simple enough, so that most of existing SSL credentials in wrapped languages can migrate to call this new API.

L46-core-tls-credential-API.md
yihuazhang: It still does not sound correct. Notice that in the hostname verification check, a client verifies that the dnsName in the subjectAltName field (i.e., hostname) of the certificate matches the host portion of the URL used to make the request. In other words, it does not compare SPIFFE ID with URL, and does not matter if they are different.

L46-core-tls-credential-API.md
yihuazhang: I think the purpose of specifying max and min TLS is a little different. For max, it should be set to the maximum version a client advocates (e.g., TLS 1.3) so that if both sides support the same maximum, the maximum will be selected and used. For min, it is the bottom line a user can tolerate and if the negotiated version is smaller than it, the TLS handshake should fail. @jiangtaoli2016 for confirm. 
How about saying "inability to set the minimum and maximum supported protocol versions for SSL CTX"?  Sorry for the pickiness, but I just want to make it clear.

L46-core-tls-credential-API.md
yihuazhang: Oh I see. Thanks for the clarification. For the purpose of CRL check, does it matter if the check is done after TLS handshake finishes? I expect it should be done before the TLS handshake starts, but the current server authorization check is designed to be performed after the handshake. @jiangtaoli2016

L46-core-tls-credential-API.md
jiangtaoli2016: No all customers prefer TLS 1.3. The purpose of setting max/min TLS version is to give users some control on TLS version, e.g., if user only wants TLS 1.3 or only wants TLS 1.2. Of course, such API is not guarantee to be working. If the underlying OpenSSL library uses 1.0.2, set min/max TLS version won't take effect.

L46-core-tls-credential-API.md
jiangtaoli2016: The standard revocation check is during TLS handshake (not before handshake :-) However, given gRPC core does not support CRL check yet. This custom authorization check provides an opportunity to check revocation before gRPC sends actual payload.

yihuazhang: Understood. My concern is that as it deviates from the standard practice (i.e., perform check after the handshake finishes), we need to be more careful about its implication before introducing this feature because once the feature is there, nothing can prevent OSS users from using it, and it is also more difficult to retire it once we have a standard CRL check.

L46-core-tls-credential-API.md
yihuazhang: Understood. So I will take it as a confirmation on my understanding.

L46-core-tls-credential-API.md
jiangtaoli2016: There are feature requests to support custom authz check on server side. see https://github.com/grpc/grpc/issues/23498. Today, on client side, we expose custom authorization check callback for the purpose of server authorization, but many customers use it for other legitimate purpose. E.g., Microsoft uses it for validating cert chain.

By exposing a method to allow application to validate certificate (e.g., certificate check, revocation check) adds significant security value. Revocation check during handshake vs revocation check after handshake has little security implication -- as far as revocation is before actual RPC payload.

L46-core-tls-credential-API.md
ZhenLian: Ah I see what you mean. IIUC, in typical SPIFFE semantics, a physical server can have multiple workload ids, and a valid SPIFFE cert might not contain dnsName in the SAN field. If we do default hostname verification on a SPIFFE cert, the check will fail. Isn't that the original problem we are going to solve? I might misunderstand something.
How about rewording as "a valid certificate using SPIFFE ID doesn't necessarily contain the dnsName in the SAN field, which might cause default hostname verification to fail"?

yihuazhang: That sounds much better. More importantly, I think we need to highlight that we need an ability of skipping the hostname verification check if a client has a different way of validating the server's identity.

L46-core-tls-credential-API.md
yihuazhang: Thanks for the clarification. I misunderstood the way CRL check will be realized. I thought the intention was for gRPC to implement CRL checks (after TLS handshake finishes) and expose an API for users to use, but instead we want to introduce a similar authorization check mechanism at the server side and it's up to users to implement the callback that may do the CRL check (if they want to). If it's the latter, I am fine with it.

L46-core-tls-credential-API.md
ZhenLian: SG. Revised! Please take a look, thanks!

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,462 @@
...
+## Abstract
+
+This proposal aims to provide the following features in a new TLS credential API:
+1. credential reloading: reload transport credentials periodically from various sources, e.g. from credential files, or users' customized Provider implementations, without shutting down
+2. custom authorization check: support the following cases
+    - on the client side:
+        - perform both the certificate and default hostname check(default option)
+        - perform certificate check but skip default hostname check(a custom authorization check is mandatory)
+        - skip both certificate check and default hostname check(a custom authorization check is mandatory)
+        - perform a custom authorization check, if needed
```
yihuazhang, line 18: It is a little confusing to have `custom authorization check` in both title and bullet items (client and server side). 
Also since this gRFC does not propose any API to specify the first three cases (of client-side), it does not make much sense to explain them in details (sorry for the previous comment). WDYT?

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,462 @@
...
+        - perform a custom authorization check, if needed 
+    - on the server side:
+        - perform a custom authorization check, if needed    
+3. TLS version configuration: optionally configure the minimum and maximum of the TLS version that will be used in the handshake
+
+## Background
+
+The current TLS implementation in gRPC C core has some restrictions:
+1. lack of the credential reloading support. Once an application starts, it can never reload its identity key/cert-chain and root certificates without shutting down
+2. inflexibility in the peer verification. In a mutual TLS scenario, servers would always perform certificate verification, and clients would always perform certificate verification as well as default hostname check. No customized checks can be introduced.
```
yihuazhang, line 27: This paragraph seems to contain necessary information to put in the abstract for `custom authorization check
' (per my comment above).

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,462 @@
...
+    grpc_ssl_client_certificate_request_type type);
+
+/**
+ * Set grpc_tls_server_verification_option. 
+ * If not set, we will use the default value GRPC_TLS_SERVER_VERIFICATION.
+ * This should be called only on the client side.
+ */
+GRPCAPI int grpc_tls_credentials_options_set_server_verification_option(
+    grpc_tls_credentials_options* options,
+    grpc_tls_server_verification_option server_verification_option);
```
yihuazhang, line 78: Should we define it somewhere in this gRFC?

yihuazhang, line 78: Do you plan to introduce `grpc_tls_client_verification_option` enum for the client-side custom authorization check?

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,462 @@
...
+
+/**
+ * Sets the minimum TLS version that will be negotiated during the TLS handshake.
```
yihuazhang, line 81: How about rephrase it to "Sets the minimum supported TLS protocol version for SSL CTX"? The current description is a little confusing.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,462 @@
...
+ * If not set, the underlying SSL library will pick the version automatically.
+ */
+GRPCAPI void grpc_tls_credentials_options_set_min_tls_version(grpc_tls_credentials_options* options,
+    grpc_tls_version min_tls_version);
+
+/**
+ * Sets the maximum TLS version that will be negotiated during the TLS handshake.
```
yihuazhang, line 88: Same as above.

L46-core-tls-credential-API.md
```diff
@@ -0,0 +1,462 @@
...
+
+ private:
+  struct CertsWatching {
+    bool watching_root = false;
+    bool watching_identity = false;
+  };
+  std::map<std::string /*cert_name*/, CertsWatching> certs_watching_;
+};
+```
+### Custom Authorization
```
yihuazhang, line 341: @jiangtaoli2016 What's the plan for supporting async authorization check? The current server authorization check supports both sync and async, which is pretty complicated due to async support and goes against the goal of simple C-core API. The API's in this section can be simplified a lot if we decide to only support sync authorization check.

L46-core-tls-credential-API.md
ZhenLian: Yeah, Jiangtao and Mark wanted to me write a full introduction about the TLS credential to be a replacement of our previous gRFC, so I am trying to include the all features here, no matter whether it is already implemented or not. 
I do agree the current abstract contains more information than needed. Revised. Please take a look, thanks!

L46-core-tls-credential-API.md
ZhenLian: I think async authz check has its values, especially considering the fact that a check can take a long time and we don't want to block the main thread. 
I was thinking for a long time if we can make `grpc_tls_authorization_check_arg` a class instead of a struct? Currently it contains a lot of information that the caller(from the wrap languages) doesn't need to know. That may simplify the logic a little bit, but I am not sure how long the refactoring would take...

L46-core-tls-credential-API.md
ZhenLian: I can do either way. Having a `grpc_tls_client_verification_option` is what we are doing in Go.

If server side doesn't want to verify the certificate, it can choose the corresponding `GRPC_SSL_REQUEST_AND_REQUIRE_CLIENT_CERTIFICATE_BUT_DONT_VERIFY` value in `grpc_ssl_client_certificate_request_type`, which I think is equivalent here. But I am also fine adding a `grpc_tls_client_verification_option` to make it (probably) easier for users to understand, etc.

yihuazhang: +1 on having `grpc_tls_client_verification_option` for extendability.

L46-core-tls-credential-API.md
yihuazhang: If we decide to support async, we first need to flush out some details on how to wrap the C-core API's related to server authorization check (e.g., `grpc_tls_authorization_check_arg` and others).

ZhenLian: OK. How about me writing a one pager briefly describe the necessary changes this would take?

yihuazhang: Yeah, but it also needs to be signed off by TL's of each wrapped language, which may take longer than we expect. Do we have an immediate use case for async authorization check @jiangtaoli2016? If we do not, how about we start with a simple sync version in this gRFC and introduce an async version in another gRFC when its design is mature?

jiangtaoli2016: We  need to support async authz check, e.g., if customer needs to validate the cert chain with a remote server or fetching server authz ACLs remotely. gRPC core cannot be blocked. This is a requirement from @markdroth .

Yihua: we can re-use your previous server authorization check cb API, which was reviewed by Mark. Any concern with it?

yihuazhang: At the time I worked on async server authz check, I did not consider much about whether it can be easily wrapped by languages other than C++, while gRFC aims for the API simplicity for ALL wrapped languages. I am not sure if `struct` and function pointer fields in `grpc_tls_authorization_check_arg` can be easily wrapped.

jtattermusch:
CC @jtattermusch

7hacker:
Hi, this looks great, however we were wondering if this can also support TLS-PSK (Pre-Shared Keys)?

For TLS 1.3 the TLS-PSK mode enables 0-RTT handshakes, reducing overhead for resource constrained use-cases

ZhenLian:
> Hi, this looks great, however we were wondering if this can also support TLS-PSK (Pre-Shared Keys)?
> 
> For TLS 1.3 the TLS-PSK mode enables 0-RTT handshakes, reducing overhead for resource constrained use-cases

No, I don't think so. This API is mainly for "common" TLS with asymmetric keys.

mankhakb:
Proposal - grpc library to provide support to configure TLS version at run time? Such that application can opt TLS version (1.2 or 1.3) at run time. Basically user can configure TLS version to use.

rockspore:
This proposal has APIs for setting TLS version and JFYI I have https://github.com/grpc/grpc/pull/31368 pending review for this.

PR #208, opened 2020-09-14
A35-grpc-gateway initial proposal

Talador12:
See A35 proposal included in PR 😎

### DIFF
--- a/A35-grpc-gateway.md
+++ b/A35-grpc-gateway.md
@@ -1,8 +1,8 @@
 Integrate the grpc-gateway as a multi-language option to allow for REST endpoint support
 ----
 * Author(s): Keith Adler (Talador12)
-* Approver: a11r
-* Status: Draft --- {Draft, In Review, Ready for Implementation, Implemented}
+* Approver: wenbozhu and/or gnossen
+* Status: In Review --- {Draft, In Review, Ready for Implementation, Implemented}
 * Implemented in: Core (C++), Java, Python, and Go.
 * Last updated: 2020-09-14
 * Discussion at: https://groups.google.com/u/1/g/grpc-io/c/rEfqlLL9Emw (filled after thread exists)
@@ -15,7 +15,11 @@ Based on this [gRPC issue](https://github.com/grpc/grpc/issues/23979), the goal 
 grpc-gateway is a separate binary/process from your server. It can be deployed alongside a server of any language, which (I assume) is what's happening in the gist you linked above. We have talked about in-library support for this sort of thing in the past. That way, you wouldn't have to orchestrate a separate gateway process. But that would be a pretty substantial effort.
 ```
 
-As gRPC has grown in popularity and usage, an option to run in-library support would be a huge win for many teams. Instead of a sidecar, this gateway could be run alongside the gRPC server instance itself. As it would be a substantial effort, the background section will detail out the complexity of implementing across all languages and the steps required to start this process.
+As gRPC has grown in popularity and usage, an option to run in-library support would be a huge win for many teams.
+
+Instead of a separate package, this gateway could be run alongside the gRPC server instance itself. This would allow for support to be built into packages like KFServing, BentoML, and Seldon Core for both gRPC and REST/JSON traffic (that is ultimately routed to gRPC through this Gateway). The above packages are just ones used in a machine learning space, but there are many other microservices that have clients that could request via REST/JSON if the gateway functionality was enabled. For teams using gRPC, it is uncommon to build out the separate gateway, as it is a separate feature level effort for each team to implement. Generally, adding the `grpc-gateway` as a package to an open source project as an entirely different component is not intuitive or as accessible as a built into gRPC option would be. If this component was built into gRPC, it would be much easier to add an optional HTTP/JSON endpoint to your existing gRPC implementation. This should save a lot of duplicated development effort for many teams, projects, and libraries.
+
+As this feature request would be a substantial effort, the background section will detail out the complexity of implementing across all languages and the steps required to start this process.
 
 
 ## Background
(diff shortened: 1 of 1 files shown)


A35-grpc-gateway.md
```diff
@@ -0,0 +1,83 @@
+Integrate the grpc-gateway as a multi-language option to allow for REST endpoint support
+----
+* Author(s): Keith Adler (Talador12)
+* Approver: a11r
+* Status: Draft --- {Draft, In Review, Ready for Implementation, Implemented}
+* Implemented in: Core (C++), Java, Python, and Go.
```
gnossen, line 6: There are two big chunks to this feature: codegen and the runtime HTTP/1 server. Codegen is definitely language-specific, but the runtime aspect doesn't have to be. If possible, the runtime aspect should be implemented entirely in Core, so that it won't have to be *re*implemented for Python, Ruby, PHP, and C#.

Is there any particular reason why you called Python out as the only wrapped language here?

I don't think it's an absolute must that this feature be implemented in all languages from the get-go, but I think we should design it in such a way that adding support to additional languages in the future is as easy as possible.

A35-grpc-gateway.md
```diff
@@ -0,0 +1,83 @@
...
+The first step would be for the interested party to start a gRFC in https://github.com/grpc/proposal. There, you'd want to get relevant people from all of the languages to chime in on the requirements for the cross-language feature. Then, we'd need to get into per-language details. Finally, we'd need to ensure that support is added across the language matrix, which will entail at least three different implementations: Core (C++), Java, and Go.
+```
+
+### Related Proposals: 
+
+* N/A - none was referenced by the gRPC team, although they have discussed implementing this feature informally in the past.
+
+## Proposal
+
+Add a [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) like component that is a separate binary/process from the gRPC server. It can be deployed alongside a server of any language. This should be in-library support, and not a `grpc-ecosystem` project. That way, you wouldn't have to orchestrate a separate gateway process. To implement this, it is necessary to ensure that support is added across the language matrix, which will entail four different implementations: Core (C++), Java, Python, and Go.
```
gnossen, line 56: Is it actually necessary that this be implemented as a subprocess? I think the real value is in translating from JSON+REST to gRPC, regardless of *how* it happens under the hood.

A35-grpc-gateway.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Abstract
+
+Based on this [gRPC issue](https://github.com/grpc/grpc/issues/23979), the goal is to include the functionality of the [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) directly into the gRPC codebase across all languages. From @gnossen:
+
+```
+grpc-gateway is a separate binary/process from your server. It can be deployed alongside a server of any language, which (I assume) is what's happening in the gist you linked above. We have talked about in-library support for this sort of thing in the past. That way, you wouldn't have to orchestrate a separate gateway process. But that would be a pretty substantial effort.
+```
+
+As gRPC has grown in popularity and usage, an option to run in-library support would be a huge win for many teams. Instead of a sidecar, this gateway could be run alongside the gRPC server instance itself. As it would be a substantial effort, the background section will detail out the complexity of implementing across all languages and the steps required to start this process.
```
gnossen, line 18: Can you add a little more detail to this? I'm interested in your perspective on the usage of `grpc-gateway` out in the wild.

A35-grpc-gateway.md
Talador12: I'll clarify, the value of JSON+REST to gRPC is the goal

A35-grpc-gateway.md
Talador12: I agree on Core first. I called out Python first because I am working primarily in python projects/libraries and think it is a decent option.

```
I don't think it's an absolute must that this feature be implemented in all languages from the get-go, but I think we should design it in such a way that adding support to additional languages in the future is as easy as possible.
```
Agreed. Better to do this the right way than a rush job.

I'll add more detail about those 2 facets.

Talador12:
^^ Updated approver on the proposal

gnossen:
CC @ejona86 @dfawley @markdroth as tech leads for Java, Go, and C++ respectively.

srini100:
I think @wenbozhu is the right person to evaluate this gRFC and see if this is the right direction and features to have in gRPC.

Talador12:
@wenbozhu Did you get a chance to review this proposal?

wenbozhu:
Sorry I missed this. Will take a look asap. Thanks.

A35-grpc-gateway.md
```diff
@@ -0,0 +1,99 @@
...
+
+* N/A - none was referenced by the gRPC team, although they have discussed implementing this feature informally in the past.
+
+## Proposal
+
+Add a [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) like component that able to REST+JSON towards the gRPC server. It can be deployed alongside a server of any language. This should be in-library support, and not a `grpc-ecosystem` project. To implement this, it is necessary to ensure that support is added across the language matrix, which will entail four different implementations: Core (C++), Java, Python, and Go. Core should come first so that as much as possible is language agnostic.
+
+## Rationale
+
+Using the `grpc-gateway` as a separate project is not always an option based on library usage. Projects that build on top of `grpc` often want the capability to offer HTTP1 traffic to clients, but do not have the ability to use in library solutions. Currently, most simply do not offer HTTP1 traffic. This excludes usage by several projects and companies.
```
wenbozhu, line 66: Could you elaborate on this? Ignoring the overhead of grpc-gateway as a separate proxy/process, why is it impossible to deploy grpc-gateway for solutions that expose grpc endpoints?

A35-grpc-gateway.md
```diff
@@ -0,0 +1,99 @@
...
+
+More implementation details in `Implementation`.
+
+### Related Proposals: 
+
+* N/A - none was referenced by the gRPC team, although they have discussed implementing this feature informally in the past.
+
+## Proposal
+
+Add a [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway) like component that able to REST+JSON towards the gRPC server. It can be deployed alongside a server of any language. This should be in-library support, and not a `grpc-ecosystem` project. To implement this, it is necessary to ensure that support is added across the language matrix, which will entail four different implementations: Core (C++), Java, Python, and Go. Core should come first so that as much as possible is language agnostic.
```
wenbozhu, line 62: (I assume this gRFC only concerns server-side implementations.)

The benefits of grpc-gateway are that it's very extensible and it uses a grpc client to address "combability" concerns. 

For in-library support, we still have to decide if we are talking about in-process proxying, or a new transport for gRPC runtime, or maybe something in-between. 

Another alternative is to create a list of  language-specific server-side interceptors to support REST endpoints alike; and use the http/1.1 stack provided by the language-platform.

wenbozhu: I may be wrong about the interceptor part (for servers). I vaguely remembered that it was proposed to support non-standard transports, as a lightweight solution.   @gnossen

Talador12:
Closing - this is from 2020 and is no longer being actively pursued. Thanks to anyone who reviewed at the time.

PR #219, opened 2021-01-28
A39: xDS HTTP Filters

markdroth:


markdroth:
CC @lidizheng

jiangtaoli2016:
cc @ashithasantosh FYI

A39-xds-http-filters.md
```diff
@@ -0,0 +1,134 @@
...
+  field to determine the actual type; this allows for third-party
+  plugins to encode their configurations as a `google.protobuf.Struct`
+  proto instead of using a concrete proto message type that the client
+  may not be able to handle.
+
+When gRPC receives the `HttpConnectionManager` proto in the LDS
+response, it will validate the list of filters as follows:
+- Every filter in the list must be of a known type, as identified by the
+  `type_url` field.  Any unknown config type will cause the Listener
+  resource to be NACKed.
```
lidizheng, line 82: Filters in `HttpConnectionManager` could only have name and the `typed_config` is empty. So, we might not be able to detect the filter type automatically (defer the type detection works but seems a bit over complex). We might need to keep a map between filter name and type.

markdroth: Given that Envoy uses the type in the `typed_config` to indicate which filter implementation to use, I don't think that field can be empty.  @htuch, can you confirm?

htuch: Should not be empty since v3.

A39-xds-http-filters.md
```diff
@@ -0,0 +1,145 @@
...
+this filter configuration and treat it similarly to the way that Envoy
+does.
+
+Specifically, if the router filter is not present, gRPC will fail all
+requests.  In addition, any filters in the list after the router filter
+will not actually be used, although their configurations will still be
+validated when the config is received from the xDS server.
+
+For now, gRPC will not support any of the fields in the router filter's
+config.  All fields will be ignored.
```
sanjaypujare, line 139: Should we just ignore the fields or NACK the config if it contains values we don't support? e.g. if `dynamic_stats` is true?

markdroth: We ignore them.  This is no different than any other individual xDS field we don't (or don't *yet*) support.

A39-xds-http-filters.md
```diff
@@ -0,0 +1,152 @@
...
+When gRPC receives the `HttpConnectionManager` proto in the LDS
+response, it will validate the list of filters as follows:
+- Every filter in the list must be of a known type, as identified by the
+  `type_url` field.  Any unknown config type will cause the Listener
+  resource to be NACKed.
+- Every filter in the list must have a unique instance name.  Any
+  duplicate instance names will cause the Listener resource to be NACKed.
+
+Note that gRPC will support configuring multiple instances of the same
+filter implementation, each with its own name and config.
```
lidizheng, line 95: Core might have slightly more challenges comparing to Java/Golang, since Core is using service config to transport the filter config and Java/Golang can map each filter to an interceptor.

lidizheng (approved):

A39-xds-http-filters.md
markdroth: That's true, but it's an implementation detail that doesn't need to be covered in the gRFC.  I'll deal with that as part of grpc/grpc#25310.

ejona86:
CC @dapengzhang0, this impacts fault injection.

A39-xds-http-filters.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Related Proposals: 
+
+- [A27: xDS-Based Global Load Balancing](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)
+- [A30: xDS v3 Support](https://github.com/grpc/proposal/blob/master/A30-xds-v3.md)
+- [A31: gRPC xDS Timeout Support and Config Selector Design](https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md)
+
+## Proposal
+
+In the gRPC client channel, the xDS HTTP filters will run in between
```
sanjaypujare, line 38: What about on the gRPC server side? Should we say something about whether we are/will be planning to support HTTP filters on the server side e.g. [rbac_filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/rbac_filter) which is a server side one?

A39-xds-http-filters.md
```diff
@@ -0,0 +1,152 @@
...
+  resource to be NACKed.
+- Note that gRPC will *not* fail validation if the map key specifies a
+  filter instance name that does not exist in the `HttpConnectionManager`
+  filter list.  This is because during an update, the xDS client code
+  cannot know which `HttpConnectionManager` config is currently being
+  used.
+
+Unlike Envoy, which provides an API that filters must call to access their
+per-{VirtualHost,Route,ClusterWeight} config overrides, gRPC will
+require filters to construct a merged configuration, applying any
```
sanjaypujare, line 119: In case of update of HttpFilters by the xDS server (e.g. a new filter being added) wouldn't the Envoy approach be better in terms of complexity and performance as compared to merged configuration?

A39-xds-http-filters.md
```diff
@@ -0,0 +1,152 @@
...
+Note that gRPC will not actually have a discrete filter implementation
+for the router filter; instead, this filter will simply trigger the
+built-in routing functionality that we have already implemented in the
+resolver and LB policy plugins to the gRPC client channel.  However,
+in order to retain compatibility with Envoy, gRPC will still accept
+this filter configuration and treat it similarly to the way that Envoy
+does.
+
+Specifically, if the router filter is not present, gRPC will fail all
+requests.  In addition, any filters in the list after the router filter
```
sanjaypujare, line 141: Is this a breaking change assuming the current gRPC xDS implementation does not look for the router filter?

A39-xds-http-filters.md
```diff
@@ -0,0 +1,152 @@
...
+
+gRPC will add support for the following fields in the xDS APIs:
+- [`HttpConnectionManager.http_filters`](https://github.com/envoyproxy/envoy/blob/2ee9543fbb0fccdb69d4dcd8cb57a19743afb94b/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L273)
+- [`VirtualHost.typed_per_filter_config`](https://github.com/envoyproxy/envoy/blob/2ee9543fbb0fccdb69d4dcd8cb57a19743afb94b/api/envoy/config/route/v3/route_components.proto#L142)
+- [`Route.typed_per_filter_config`](https://github.com/envoyproxy/envoy/blob/2ee9543fbb0fccdb69d4dcd8cb57a19743afb94b/api/envoy/config/route/v3/route_components.proto#L252)
+- [`WeightedCluster.ClusterWeight.typed_per_filter_config`](https://github.com/envoyproxy/envoy/blob/2ee9543fbb0fccdb69d4dcd8cb57a19743afb94b/api/envoy/config/route/v3/route_components.proto#L366)
+
+The `HttpConnectionManager.http_filters` field is an ordered list of
+filters.  In each entry, gRPC will look at the following fields:
+- [`name`](https://github.com/envoyproxy/envoy/blob/2ee9543fbb0fccdb69d4dcd8cb57a19743afb94b/api/envoy/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#L833):
```
dfawley, line 68: Should we NACK if the name is the empty string, or is this a valid setting as long as only one filter entry has it?

A39-xds-http-filters.md
```diff
@@ -0,0 +1,152 @@
...
+necessary overrides, when the xDS config is applied.  The filters will
+then use this merged configuration at run-time.  However, this approach
+is an implementation detail that could change in the future.
+
+### Initially Supported Filter: Router
+
+Initially, gRPC will support only one filter, the [router
+filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter),
+which is used for the proto message type
+[`extensions.filters.http.router.v3.Router`](https://github.com/envoyproxy/envoy/blob/18db4c90e3295fb2c39bfc7b2ce641cfd6c3fbed/api/envoy/extensions/filters/http/router/v3/router.proto#L23).
```
dfawley, line 129: Do we need to support the router config being encoded inside a `udpa.type.v1.TypedStruct`, or will it always be directly in the `typed_config` field?

A39-xds-http-filters.md
markdroth: Yes, it could be, which is why it's called out here.

If the xDS server is not currently populating any HTTP filters, upgrading gRPC without changing the xDS server to populate the router filter will cause failures.  And if the xDS server is populating HTTP filters meant for Envoy that gRPC does not support (which the current version of gRPC is ignoring), upgrading gRPC without changing the xDS server to populate only the router filter will cause failures.  So the right upgrade path is to first change the xDS server to populate only the router filter (or the router filter and maybe the fault injection filter, since we'll be releasing both features at the same time), and then upgrade their clients.

Adding support for HTTP filters isn't really fundamentally different than adding support for any new xDS field.  Whenever we start to support a new feature, existing clients may break.  For example, when we add circuit breaking support, the default for that is a max of 1024 concurrent requests, so an existing client that is busier than that will start breaking.  Unfortunately, I think this is just something people need to be used to with xDS.

sanjaypujare: Understood but the consequences are quite drastic "...if the router filter is not present, gRPC will fail all
requests...".

At the same time most current users of gRPC with xDS probably use Envoy-conforming xDS config which always has the router filter so in practice won't break anybody.

A39-xds-http-filters.md
dfawley: > both Envoy and in my code for C-core in grpc/grpc#25310 ... just uses the resulting type to look up the filter implementation that should parse the proto

So what's the end result for Envoy/C right now when this happens?

You validate LDS successfully, but then the actual filter implementation fails because it finds the wrong message type for the config?  So...RPCs fail?  Or since we don't actually support any of the configuration for the `Router` type anyway, the whole config can be safely ignored, and it "works" (for now)?  Or I could also see a possible implementation where we end up processing any filters after this one, and if we find another `Router` entry, everything "works" but this oddly-encoded filter is just ignored.

Most importantly: what do we _want_ to do?  I can implement it any of these ways with roughly the same effort - I just need to know what to do.

A39-xds-http-filters.md
dfawley: Ah, you're right - I missed that.

A39-xds-http-filters.md
markdroth: Currently, the router filter impl does not expect its config as a `TypedStruct`, so the config will be rejected.  But it won't be rejected by the general HCM parsing code; it will be rejected by the router filter.  Specifically, here's what the HCM parsing code will do:

1. Look at the type of the filter config.  The type is `TypedStruct`, so it will look inside of the `TypedStruct` for the real name, where it will see `envoy.extensions.filters.http.router.v3.Router`.
2. Ask the registry what filter to use for config type `envoy.extensions.filters.http.router.v3.Router`.  The registry will return the router filter.
3. Ask the router filter to parse the `TypedStruct`.  The router filter will return an error, because it doesn't know how to parse the `TypedStruct`.
4. Because of that error, the HCM parsing code will reject the config.

dfawley: It's not specified in the proposal when the configs should be parsed by the filters.  I was assuming we would wait until the entire config was known (the merge of the 4 sources) before passing it to the filter.  Is that wrong?  Can you document exactly how and when parsing should occur?

markdroth: I think the proposal already states this explicitly.  It says:

> When gRPC receives the `HttpConnectionManager` proto in the LDS
> response, it will validate the list of filters as follows:
> - Every filter in the list must be of a known type, as identified by the
>   `type_url` field.  Any unknown config type will cause the Listener
>   resource to be NACKed.
> - Every filter in the list must have a unique instance name.  Any
>   duplicate instance names will cause the Listener resource to be NACKed.

And then later:

> When gRPC receives the `RouteConfiguration` proto, either in the LDS or
> RDS response, it will validate the contents of each
> `typed_per_filter_config` map as follows:
> - Every config in the list must be of a known type, as identified by the
>   map value proto message type.  Any unknown config type will cause the
>   resource to be NACKed.
> - Note that gRPC will *not* fail validation if the map key specifies a
>   filter instance name that does not exist in the `HttpConnectionManager`
>   filter list.  This is because during an update, the xDS client code
>   cannot know which `HttpConnectionManager` config is currently being
>   used.

markdroth: As we discussed offline, I've clarified that the filter config itself will be validated, and I've added a section about the override config merging precedence rules.

A39-xds-http-filters.md
```diff
@@ -0,0 +1,165 @@
...
+
+When gRPC receives the `HttpConnectionManager` proto in the LDS
+response, it will validate the list of filters as follows:
+- Every filter in the list must be of a known type, as identified by the
+  `type_url` field.  Any unknown config type will cause the Listener
+  resource to be NACKed.
+- The fields in the filter config should be validated by the filter
+  implementation.  Any validation error will cause the Listener resource
+  to be NACKed.
+- Every filter in the list must have a unique instance name.  Any
```
markdroth, line 94: @htuch, can you confirm that this is consistent with Envoy?  @lidizheng has pointed out to me that there are examples like the following in the Envoy docs that show two instances of the same filter using the legacy well-known filter names, so both instances have the same name:

https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/decompressor_filter#using-different-decompressors-for-requests-and-responses

Should we allow this kind of config?  If we do, and we see a `typed_per_filter_config` override entry with this same name, which of the two filters do we apply the override to?

htuch: I'm willing to believe we don't enforce this property (name uniqueness) today. I think we have to in order for per-route config overrides to make sense, so suggest we make this official in the API.

The one concern is that this might break a lot of folks if we just rollout the enforcement. Probably would need a runtime override in Envoy.

markdroth: Okay, sounds like this is just another side-effect of the fact that Envoy hasn't yet fixed envoyproxy/envoy#12274, which is already noted in this gRFC.

A39-xds-http-filters.md
```diff
@@ -0,0 +1,165 @@
...
+then use this merged configuration at run-time.  However, this approach
+is an implementation detail that could change in the future.
+
+Note that, for a given request, the filter's configuration will come
+from the top-level filter list in the `HttpConnectionManager` config
+and the most specific override.  In other words, if a given filter
+instance has an override in the `ClusterWeight` proto, that will be
+used; otherwise, if it has an override in the `Route` proto, that will
+be used; otherwise, if it has an override in the `VirtualHost` proto,
+that will be used; otherwise, no override will be used.
```
dfawley, line 135: This sounds like a limitation vs. the Envoy API which allows a filter to query for any and all of the three overrides.  Is this okay?  I could imagine a system where the config is:

- LDS (partial config that applies to all virtual hosts behind this listener)
- Virtual Host (more config that applies to all routes behind this virtual host)
- Route (more config that applies to all weighted clusters behind this route)
- WeightedCluster (the remaining config that applies to this cluster)

markdroth: The behavior described here is exactly how the Envoy implementation works.  Filters don't have APIs that let them access each override location independently; they just have an API that says "give me the most specific override config".

dfawley: My mistake, then.  This made it sound like they could access them independently:

> Unlike Envoy, which provides an API that filters must call to access their
per-{VirtualHost,Route,ClusterWeight} config overrides

A39-xds-http-filters.md
dfawley: What happens when a single filter name appears with multiple, incompatible types?

I.e.:
```
LDS:
HttpConnectionManager.http_filters: {
  name: "FOO_FILTER"
  typed_config: extensions.filters.http.fault.v3.HTTPFault {
    // fault injection config
  }
}

RDS:
VirtualHost.typed_per_filter_config["FOO_FILTER"]: extensions.filters.http.health_check.v3.HealthCheck {
  // health check config
}
```

IIUC, with this proposal, assuming gRPC has implemented fault injection and health checking:

- LDS will validate, as the fault injection filter is able to parse the HTTPFault config.
- RDS will validate, as the health check filter is able to parse the HealthCheck config.
- When an RPC happens which uses that VirtualHost, we will call the fault injection HTTP filter, passing it the FI config from HCM and the HC config from the VH.  This is not valid - so RPCs will fail with an appropriate error (something like "unexpected config type") when this happens?

Is this what happens in Envoy?

Or should it be caught earlier and nack the whole config (LDS, I guess)?

markdroth: Yeah, unfortunately, I don't think we'll be able to NACK in that case.

Unfortunately, as we've discussed in other contexts, the xDS ACK/NACK is really just about the validation of an individual resource in isolation, not about whether it actually works in the context of the rest of the client's configuration.  xDS does not currently have any mechanism for the client to indicate its overall config health, so there's really no way to handle this kind of case.  One obvious example of where this happens in Envoy is if it receives two Listeners for the same port.  Another case that affects both Envoy and gRPC would be if a RouteConfiguration refers to a Cluster that does not actually exist.  In these kinds of cases, we'll ACK but not actually work, and the case you describe here is just another example of that general problem.  See envoyproxy/envoy#11396 for discussion of one possible solution.

I think the case you cite here is another example of this problem.  In terms of layering, the validation code (inside of the `XdsClient`) doesn't actually know which LDS resource a given RDS resource will be used with -- and even if there weren't layering issues to consider, if both the LDS and RDS resources are being updated at around the same time, it would be really hard to figure out *which* LDS resource the RDS resource is supposed to match up with.  This is basically why validation only looks at each individual resource in isolation.

I think Envoy will do the same thing once they implement the changes described in envoyproxy/envoy#12274, and I've already added a note about this to that issue.

dfawley: That all sounds reasonable. So what should happen, then?  RPCs that end up invoking this configuration combination just fail?

markdroth: I think it's ultimately up to the filter how to handle it.  In C-core, we'll just pass the two configs to the filter's `GenerateServiceConfig()` method, which will notice that the override config is of the wrong type.

My off-the-cuff guess is that the right thing to do here is to (a) log an error indicating the problem and (b) ignore the override config.  But individual filters might choose to do something different for their own reasons.

dfawley: So for the specific design for the router filter in this gRFC, we should log an error but ignore the misconfiguration?

markdroth: In the specific case of the router filter, it doesn't accept per-route overrides, so its `GenerateFilterConfigOverride()` method should unconditionally return an error, regardless of what its input was.

A39-xds-http-filters.md
markdroth: We've decided that we will support HTTP filters on the server side.  I've updated this gRFC accordingly, and @ejona86 will update gRFC A36 to (a) cover L4 filters, (b) refer to this gRFC for discussion of HTTP filters, and (c) cover handling of RouteConfiguration on the server side.

A39-xds-http-filters.md
markdroth: Yeah, I don't think this is likely to affect too many people.  There may be some that are not sending HTTP filter configs at all, but it shouldn't be too hard for them to identify and fix the problem.

dfawley (approved):

A39-xds-http-filters.md
```diff
@@ -0,0 +1,268 @@
...
+
+Note that gRPC will not actually have a discrete filter implementation
+for the router filter; instead, this filter will simply trigger the
+built-in routing functionality that we have already implemented in the
+resolver and LB policy plugins to the gRPC client channel.  However,
+in order to retain compatibility with Envoy, gRPC will still accept
+this filter configuration and treat it similarly to the way that Envoy
+does.
+
+Specifically, if the router filter is not present, gRPC will fail all
```
ejona86, line 235: In the discussion earlier today it sounded like we'd allow filters earlier in the list to still "have a go" at processing the request. But if the request gets to the end we'd fail it. That's a bit different than what's written here.

A39-xds-http-filters.md
```diff
@@ -0,0 +1,268 @@
...
+  may not be able to handle.
+
+Note that Envoy currently keys these maps by legacy filter names instead
+of by filter instance names, but this is expected to change, as per
+https://github.com/envoyproxy/envoy/issues/12274.  gRPC will directly
+implement the desired semantics here.
+
+When gRPC receives the `RouteConfiguration` proto, either in the LDS or
+RDS response, it will validate the contents of each
+`typed_per_filter_config` map as follows:
```
ejona86, line 184: Where's the stuff about "fail validation if using a config in the wrong context (listener filter list vs override config)" that was discussed earlier today? If that's implied by "The fields in the filter config should be validated by the filter implementation", then that needs to be strongly clarified. This applies to `HttpConnectionManager.http_filters` as well.

ejona86 (approved):

A39-xds-http-filters.md
```diff
@@ -0,0 +1,266 @@
...
+  `type_url` field) that is present in the filter registry.  Note that the
+  set of known filter config types will be different on the gRPC client and
+  server; if the `HttpConnectionManager` proto is inside an HTTP API Listener,
+  it will look only at filters registered for the gRPC client, whereas if it
+  is inside a TCP Listener, it will look only at filters registered for
+  the gRPC server.  Any unknown config type will cause the Listener
+  resource to be NACKed, unless the `is_optional` field is true, in which
+  case the individual filter will be ignored.
+- The fields in the filter config will be validated by the filter
+  implementation.  Any validation error will cause the Listener resource
```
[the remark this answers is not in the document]
markdroth: Done.

A39-xds-http-filters.md
markdroth: Clarified this in both places.

A39-xds-http-filters.md
markdroth: Clarified.

markdroth (commented):
Thanks for the review!

PR #222, opened 2021-02-10
template: add section for environment variable protection

dfawley:


markdroth (approved):

GRFC-TEMPLATE.md
```diff
@@ -23,6 +23,10 @@ Title
 
 [A precise statement of the proposed change.]
 
+### Environment variable protection
```
ejona86, line 26: "Temporary" or "Experimental" environment variable protection? The description paragraph could also mention when such things are removed.

ejona86 (approved):

GRFC-TEMPLATE.md
lidizheng: +1 If gRFC as proposals describes how a feature is used, then this section might suggest the feature will be guarded in the long term... If this section describes a temporary guard, maybe we will need a start release version and an end release version (when guard removed)? Should we consider the alternatives of mentioning the environment variable in our `doc/environment_variables.md` or release notes?

dfawley: Done; PTAL.

dfawley: > If this section describes a temporary guard, maybe we will need a start release version and an end release version (when guard removed)? 

The implication is the "start release version" is "the same PR as when the feature is added", and "end release version" is unknown when the gRFC is written.

> Should we consider the alternatives of mentioning the environment variable in our doc/environment_variables.md or release notes?

Since these are temporary, I don't think they should be in `doc/`.  For release notes, we *usually* keep these features out of there until they are ready for end users (no env var needed), but it may depend on the feature.

ejona86 (approved):

PR #225, opened 2021-02-24
Patch the gRFC A33 with details around 200/OK and headers

lidizheng:
Add details about how to handle aborting with 200/OK status code, and several header corner cases.

### DIFF
--- a/A33-Fault-Injection.md
+++ b/A33-Fault-Injection.md
@@ -122,7 +122,7 @@ gRPC should accept fault injection settings with either HTTP status code or gRPC
 
 ### 200/OK Error Injection
 
-xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field.
+xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field (see [doc]((https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md))). So, if the filter config sets 200 as `http_status`, gRPC will abort with gRPC status `UNKNOWN`.
 
 However, a successful RPC not only requires the final status code to be `OK`, it also requires minimum header or message frames (see [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)). Merely aborting the RPC with the status code `OK` immediately won't necessarily result in an `OK` status. A gRPC client may treat this kind of RPC as `INTERNAL_ERROR` or `UNIMPLEMENTED` depending on implementation details of each language.
 
--- a/A33-Fault-Injection.md
+++ b/A33-Fault-Injection.md
@@ -122,7 +122,9 @@ gRPC should accept fault injection settings with either HTTP status code or gRPC
 
 ### 200/OK Error Injection
 
-xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field. However, in case of injection `OK` status, gRPC libraries should abort the RPC with the status code `OK`, just like any other status code. A successful RPC not only requires the final status code to be `OK`, it also requires minimum header or message frames (see [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)). A gRPC client may treat this kind of RPC as `INTERNAL_ERROR` or `UNIMPLEMENTED` depending on implementation details of each language.
+xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field.
+
+However, a successful RPC not only requires the final status code to be `OK`, it also requires minimum header or message frames (see [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)). Merely aborting the RPC with the status code `OK` immediately won't necessarily result in an `OK` status. A gRPC client may treat this kind of RPC as `INTERNAL_ERROR` or `UNIMPLEMENTED` depending on implementation details of each language.
 
 
 ### Evaluate Possibility Fraction
--- a/A33-Fault-Injection.md
+++ b/A33-Fault-Injection.md
@@ -108,7 +108,7 @@ Note that for both `X-envoy-fault-abort-request-percentage` and `x-envoy-fault-d
 
 Note that if any header is invalid, for example, passing non-numerical value as percentage header, the invalid headers will be ignored.
 
-Note if there are multiple valid headers with the same key name, gRPC doesn't make no promise about which one will be picked at the end.
+Note if there are multiple valid headers with the same key name, gRPC implementations may choose any one of them.
 
 **Fault inject config in HTTP filter and headers won’t be conflicted**, because the header only being evaluated if the delay or abort is set to `HeaderDelay` or `HeaderAbort` instead of concrete values.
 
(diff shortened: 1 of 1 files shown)


A33-Fault-Injection.md
```diff
@@ -106,6 +106,10 @@ Note that for both `X-envoy-fault-abort-request-percentage` and `x-envoy-fault-d
 
 > ...the percentage of request to apply aborts to and must be greater or equal to 0 and its maximum value is capped by the value of the numerator of percentage field.
 
+Note that if any header is invalid, for example, passing non-numerical value as percentage header, the invalid headers will be ignored.
+
+Note if there are multiple valid headers with the same key name, gRPC doesn't make no promise about which one will be picked at the end.
```
markdroth, line 111: The phrasing here is very hard to understand.  Suggest changing "gRPC doesn't make no promise about which one will be picked at the end" to "gRPC implementations may choose any one of them".

A33-Fault-Injection.md
```diff
@@ -116,6 +120,11 @@ Both `http_status` and `grpc_status` are available in xDS v3 API. Mapping betwee
 gRPC should accept fault injection settings with either HTTP status code or gRPC status code. **The ConfigSelector should translate the HTTP status code into gRPC status code** before sending the config to the fault injection enforcer.
 
 
+### 200/OK Error Injection
+
+xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field. So, in case of injection `OK` status, gRPC libraries should abort the RPC with the status code `OK`, just like any other status code. However, a successful RPC not only requires the final status code to be `OK`, it also requires minimum header or message frames (see [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)). A gRPC client may treat this kind of RPC as `INTERNAL_ERROR` or `UNIMPLEMENTED` depends on implementation details of each language.
```
markdroth, line 125: s/depends on/depending on/

markdroth, line 125: s/So,/However,/

A33-Fault-Injection.md
lidizheng: Thanks. This is cleaner.

A33-Fault-Injection.md
lidizheng: Done.

lidizheng: Replaced the "So," and removed the existing "However,".

(I thought the surprising part is simply abort with OK won't work.)

lidizheng (commented):
Thanks for the review!

A33-Fault-Injection.md
```diff
@@ -116,6 +120,11 @@ Both `http_status` and `grpc_status` are available in xDS v3 API. Mapping betwee
 gRPC should accept fault injection settings with either HTTP status code or gRPC status code. **The ConfigSelector should translate the HTTP status code into gRPC status code** before sending the config to the fault injection enforcer.
 
 
+### 200/OK Error Injection
+
+xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field. However, in case of injection `OK` status, gRPC libraries should abort the RPC with the status code `OK`, just like any other status code. A successful RPC not only requires the final status code to be `OK`, it also requires minimum header or message frames (see [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)). A gRPC client may treat this kind of RPC as `INTERNAL_ERROR` or `UNIMPLEMENTED` depending on implementation details of each language.
```
markdroth, line 125: I think the wording here is still a little hard to follow, even with the latest changes.  Maybe we should split this into two paragraphs, one talking about the case of HTTP status 200 (which we fail with status UNKNOWN) and the other talking about gRPC status OK (which we handle as if the server had returned OK immediately).

lidizheng: Reworded into:

> xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field.
> 
> However, a successful RPC not only requires the final status code to be `OK`, it also requires minimum header or message frames (see [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)). Merely aborting the RPC with the status code `OK` immediately won't necessarily result in an `OK` status. A gRPC client may treat this kind of RPC as `INTERNAL_ERROR` or `UNIMPLEMENTED` depending on implementation details of each language.

markdroth: The wording still does not make it clear that when HTTP status is set to 200, we will actually fail with UNKNOWN instead of OK.

lidizheng: Updated the first paragraph to:

> xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field (see [doc]((https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md))). So, if the filter config sets 200 as `http_status`, gRPC will abort with gRPC status `UNKNOWN`.

A33-Fault-Injection.md
```diff
@@ -116,6 +120,13 @@ Both `http_status` and `grpc_status` are available in xDS v3 API. Mapping betwee
 gRPC should accept fault injection settings with either HTTP status code or gRPC status code. **The ConfigSelector should translate the HTTP status code into gRPC status code** before sending the config to the fault injection enforcer.
 
 
+### 200/OK Error Injection
+
+xDS allows users to set `http_status` to 200 and `grpc_status` to `OK`, and still performs fault injection. There are valid use cases for injecting an OK status with no payload. Notably, in above HTTP-gRPC status mapping spec, 200 is mapped to `UNKNOWN` because gRPC libraries expect the application to set true gRPC status in the `grpc_status` field (see [doc]((https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md))). So, if the filter config sets 200 as `http_status`, gRPC will abort with gRPC status `UNKNOWN`.
```
markdroth, line 125: The text here still seems confusing.  The first part of the paragraph talks about the status being 200 or OK, but then the second part talks about the HTTP status and says that it will *not* be treated as OK.

I think we need to be very clear about the behavior here, which means talking about each of the two cases separately.  I propose the following text for this section:

"""
In order for gRPC to provide behavior consistent with Envoy, there are two counter-intuitive edge cases that are worth calling out explicitly.

If the filter config sets `http_status` to 200, gRPC will abort with gRPC status `UNKNOWN`.  This is consistent with what gRPC would do if it received an HTTP 200 status with no `grpc-status` header, because this is the normal mapping for the HTTP 200 status in the HTTP-gRPC status mapping mentioned above.

If the filter config sets `grpc_status` to `OK`, gRPC will react as if it has immediately received trailing metadata from the server with the status `OK`.  Note that in this case, the exact behavior seen by the application depends on the type of the RPC.  For unary and client-streaming RPCs, the client will return a non-OK status to the application, because it was expecting to receive exactly one message from the server, which never arrived.  For server-streaming and bidi-streaming RPCs, the client will return status `OK` to the application without returning any messages from the server.
"""

@dfawley and @voidzcy, please confirm that I'm not lying here about the behavior in Java or Go.  Note that I intentionally side-stepped the question of which status code is used in the unary and client-streaming cases. :)

dfawley: That suggestion sounds great to me.

lidizheng: To be sure how Envoy behaves here, I setup a gRPC Python backend with Envoy proxy. For unary calls, Core seems fine with the missing message, and will simply treat it as received an empty message. This may broke the ProtoBuf parsing at wrapper layer, but the RPC status is indeed OK. For streaming calls (server-streaming and bidi), Core works fine with missing operations, and will end the RPC with OK status.

Here is the trace log for unary call:

```
I0301 20:34:03.620949200      71 call.cc:1971]               grpc_call_start_batch(call=0x125ece0, ops=0x10a8830, nops=6, tag=0x7f7f022df858, reserved=(nil))
I0301 20:34:03.621050500      71 call.cc:1569]               ops[0]: SEND_INITIAL_METADATA(nil)
I0301 20:34:03.621066200      71 call.cc:1569]               ops[1]: SEND_MESSAGE ptr=0x1218a10
I0301 20:34:03.621075000      71 call.cc:1569]               ops[2]: SEND_CLOSE_FROM_CLIENT
I0301 20:34:03.621082300      71 call.cc:1569]               ops[3]: RECV_INITIAL_METADATA ptr=0x7f7f0230f948
I0301 20:34:03.621088700      71 call.cc:1569]               ops[4]: RECV_MESSAGE ptr=0x7f7f02345920
I0301 20:34:03.621098300      71 call.cc:1569]               ops[5]: RECV_STATUS_ON_CLIENT metadata=0x7f7f02ad1a18 status=0x7f7f02ad1a30 details=0x7f7f02ad1a38
I0301 20:34:03.621383500      71 socket_utils_common_posix.cc:428] Disabling AF_INET6 sockets because ::1 is not available.
I0301 20:34:03.622419400      75 subchannel.cc:1113]         New connected subchannel at 0x7f7eec002530 for subchannel 0x1079bc0
I0301 20:34:03.623555400      75 completion_queue.cc:702]    cq_end_op_for_next(cq=0x10d8490, tag=0x7f7f022df858, error="No Error", done=0x7f7f036312c0, done_arg=0x125f7e0, storage=0x125f828)
I0301 20:34:03.623698900      75 completion_queue.cc:1083]   RETURN_EVENT[0x10d8490]: OP_COMPLETE: tag:0x7f7f022df858 OK
```

I'm not sure about Golang/Java. If needed, I can spend more time find out. But the conclusion here is that 200/OK may be a case that is costly to behave unanimously across all gRPC languages, even if the status code is returned as OK, the upper layer may failed with other reasons like failed ProtoBuf parsing, received empty metadata, etc.. This case is a bit like returning messages with non-OK in unary calls. It's valid HTTP/2, but not entirely valid gRPC. To find out the gap, we have to validate all gRPC languages with broken traffic tagged as OK.

If we don't feel like to guarantee the behavior of the OK case, I updated the description to be uncertain about the outcome:

> ...Note that in this case, the exact behavior seen by the application depends on the type of the RPC **and implementation of gRPC**. For unary and client-streaming RPCs, the client **MAY** return a non-OK status to the application, because it was expecting to receive exactly one message from the server, which never arrived. For server-streaming and bidi-streaming RPCs, the client **MAY** return status `OK` to the application without returning any messages from the server.

A33-Fault-Injection.md
```diff
@@ -122,9 +122,11 @@ gRPC should accept fault injection settings with either HTTP status code or gRPC
...
+In order for gRPC to provide behavior consistent with Envoy, there are two counter-intuitive edge cases that are worth calling out explicitly.
 
-However, a successful RPC not only requires the final status code to be `OK`, it also requires minimum header or message frames (see [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)). Merely aborting the RPC with the status code `OK` immediately won't necessarily result in an `OK` status. A gRPC client may treat this kind of RPC as `INTERNAL_ERROR` or `UNIMPLEMENTED` depending on implementation details of each language.
+If the filter config sets `http_status` to 200, gRPC will abort with gRPC status `UNKNOWN`.  This is consistent with what gRPC would do if it received an HTTP 200 status with no `grpc-status` header, because this is the normal mapping for the HTTP 200 status in the [HTTP-gRPC status mapping]((https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md)) mentioned above.
```
markdroth, line 127: The inner parens in the link are not needed.

A33-Fault-Injection.md
markdroth: Yes, I'm aware that core will return status OK.  But in the unary and client-side streaming cases, the C++ layer will convert that to a non-OK status because the expected message from the server is not present.  I think all languages should behave that way.

But I'm okay with hedging our bets in the wording here.

markdroth (approved):
Thanks for updating this!

PR #228, opened 2021-03-03
Add env guard for header-control fault injection

lidizheng:
`GRPC_XDS_EXPERIMENTAL_FAULT_INJECTION_HEADER_CONTROL` will be used to guard the fault injection header config overriding feature. As discussed offline with @dfawley, we currently don't have an easy way to test this feature in our interop tests, and we don't want to risk producing an untested feature that might stuck with old client forever. This env should be lifted once we have proper interop tests for it.

@dapengzhang0 @dfawley CC @srini100

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/proposal/228)

### DIFF
--- a/A33-Fault-Injection.md
+++ b/A33-Fault-Injection.md
@@ -138,6 +138,8 @@ According to the [Envoy implementation](https://github.com/envoyproxy/envoy/blob
 
 `GRPC_XDS_EXPERIMENTAL_FAULT_INJECTION` as an environment variable will be used to guard this feature for the initial release. Once it is set, gRPC will start to interpret xDS HTTP filters ([A39: xDS HTTP Filters](https://github.com/grpc/proposal/pull/219)) and enforce fault injection if configured. This environment variable protection will be removed once the new feature has proven to be stable.
 
+`GRPC_XDS_EXPERIMENTAL_FAULT_INJECTION_HEADER_CONTROL` as an environment variable will be used to guard the header control feature of client-side fault injection. If this feature is not enabled, gRPC should reject filter configs containing `HeaderDelay` and `HeaderAbort`. This environment variable protection will be removed once the header control feature is proven to be stable in interop tests between gRPC implementations.
+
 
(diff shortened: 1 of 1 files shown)


A33-Fault-Injection.md
```diff
@@ -138,6 +138,8 @@ According to the [Envoy implementation](https://github.com/envoyproxy/envoy/blob
 
 `GRPC_XDS_EXPERIMENTAL_FAULT_INJECTION` as an environment variable will be used to guard this feature for the initial release. Once it is set, gRPC will start to interpret xDS HTTP filters ([A39: xDS HTTP Filters](https://github.com/grpc/proposal/pull/219)) and enforce fault injection if configured. This environment variable protection will be removed once the new feature has proven to be stable.
 
+`GRPC_XDS_EXPERIMENTAL_FAULT_INJECTION_HEADER_CONTROL` as an environment variable will be used to guard the header control feature of client-side fault injection. If this feature is not enabled, gRPC should reject filter configs containing `HeaderDelay` and `HeaderAbort`. This environment variable protection will be removed once the header control feature is proven to be stable in interop tests between gRPC implementations.
```
dapengzhang0, line 141: Reject the filter config if not marked isOptional?

lidizheng: In the filter config parsing logic, the presence of `HeaderDelay` and `HeaderAbort` will cause error. Whether to propagate this error as NACK should depend on the xDS HTTP filter setting, which is controlled by the `is_optional` field.

srini100:
@lidizheng, is it possible to have the control plane configure a fault injection policy with 0% fault and then configure test client to apply header based FI on top of it? That will allow you to test the feature.

markdroth:
I would think that even with a non-zero FI config, we can still test the header-based controls, just by setting a very different numerator -- e.g., if the filter config says to abort 10% and the header says to abort 100%, we can easily tell which one is being used.

PR #229, opened 2021-03-06
A36: Allow missing Router in http_filters for L4-only servers

ejona86:
If no HTTP filters are implemented, it'd require plumbing
interceptors/filters _just_ for failing RPCs if Router is missing.
That places unnecessary burden on implementations for relatively small
benefit. We want the behavior, but it is only required when
implementations will already have the plumbing available to implement it
trivially.

CC @easwars, @yashykt, @sanjaypujare, @dfawley

### DIFF
--- a/A36-xds-for-servers.md
+++ b/A36-xds-for-servers.md
@@ -185,11 +185,12 @@ used nor validated at this time, but may be in the future.
 
 If the `envoy.extensions.filters.http.router.v3.Router` is not present in
 `http_filters`, A39 calls for inserting a special filter that fails all RPCs. If
-an XdsServer implementation does not support any HTTP filters other than the
-hard-coded Router behavior called for in A39, then the special filter that fails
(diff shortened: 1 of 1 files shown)


A36-xds-for-servers.md
```diff
@@ -183,6 +183,14 @@ HttpConnectionManager support is required. HttpConnectionManager must have valid
 RouteConfiguration (via `route_config` or indirect fields like `rds`) is not
 used nor validated at this time, but may be in the future.
 
+If the `envoy.extensions.filters.http.router.v3.Router` is not present in
```
markdroth, line 186: Might be a good idea to explicitly state that this is something that will probably change in the future when we add support for RouteConfigurations on the server side.

ejona86: Good point. I added "implementation isn't using RouteConfiguration" to the list of conditions necessary to ignore the fail-everything filter.

markdroth (approved):

PR #233, opened 2021-04-29
P5: Require a gRFC for large scale changes to gRPC core

ctiller:


P5-core-large-scale-changes.md
```diff
@@ -0,0 +1,57 @@
...
+* Provide increased visibility to the community on upcoming changes and the design considerations around them.
+* Provide the ability to reason about larger "sets" of changes that are too big to be covered either in an Issue or in a PR.
+* Establish a consistent process for structured participation by the community on large changes, especially those that impact multiple runtimes and implementations.
+
+There are now multiple teams contributing to gRPC Core, each with different motivations and organizational aims.
+These teams form a community that needs to find consensus on overall direction of the gRPC core codebase.
+As such, the core API should no longer be considered a cut point for gRFC's, and instead any larger change ought to be considered for such a process.
+
+### Related Proposals: 
+* This change extends P3 to expect gRFC's for changes to gRPC Cores internals and not just it's public API.
```
chwarr, line 26:
```suggestion
* This change extends P3 to expect gRFC's for changes to gRPC Core's internals and not just its public API.
```

Minor typos.

P5-core-large-scale-changes.md
```diff
@@ -0,0 +1,57 @@
...
+* Changes that require downstream consumers of gRPC Core to modify their code.
+* Changes with a significant performance impact.
+* Changes to vocabulary types - types that are used broadly within the library.
+* Changes that constrain future development.
+* Changes that modify system architecture.
+
+## Rationale
+
+This gRFC will slow down progress on some kinds of changes for gRPC Core.
+At this point in gRPC's lifetime this is probably appropriate: gRPC is a widely used product, multiple teams contribute to it's development, and any single person cannot cover all of the impact of changes on the myriad use cases for which gRPC is deployed.
```
chwarr, line 45:
```suggestion
At this point in gRPC's lifetime this is probably appropriate: gRPC is a widely used product, multiple teams contribute to its development, and any single person cannot cover all of the impact of changes on the myriad use cases for which gRPC is deployed.
```

chwarr (commented):
Noticed a few typos.

ctiller:
> Noticed a few typos.

Thanks for the edits!

ctiller:
@markdroth per our conversation I've modified this to include a lighter weight update process.

P5-core-large-scale-changes.md
```diff
@@ -0,0 +1,63 @@
...
+* This change extends P3 to expect gRFC's for changes to gRPC Core's internals and not just its public API.
+
+## Proposal
+
+Any large scale change to gRPC Core should not be implemented without following the gRFC process.
+
+A large scale change is one that has a wide-ranging impact on the implementation of gRPC Core.
+
+Such changes include:
+* Changes to extensibility points, or the types used by extensibility points.
```
markdroth, line 35: I think this bullet in particular needs to be more nuanced.  I do absolutely agree that a gRFC is appropriate for large structural changes to these kinds of APIs, but I think we also have a lot of changes to these APIs that are smaller in scope, where requiring a gRFC doesn't seem like a positive cost/benefit trade-off.

In particular, the resolver and LB policy APIs have been seeing a fair amount of small changes over time, some as part of the xDS work and others just as we evolve the client channel architecture, which we've currently been free to do since they are not yet public APIs.  It's been very convenient to make these kinds of changes happen slowly over time, and I think that would be a much heavier-weight approach if every small change required a gRFC.

To make this more concerete, here are some examples of actual changes we've made to these APIs over the years, grouped based on whether or not I think the change should require a gRFC moving forward:

First, the changes that I think should require a gRFC:

- grpc/grpc#17987 (restructured resolver API to add `ResultHandler` interface) and grpc/grpc#18357 (restructure how addresses flow from resolver to LB policy)
- grpc/grpc#17770 (introduced the LB policy picker API)

Here are a few that I think are borderline; the scope or impact of the changes probably justifies a gRFC, but the fact that there don't happen to be any external LB policy implementations today mean that the changes didn't affect anyone:

- grpc/grpc#24172 (pass address to LB policy helper `CreateSubchannel()` separately instead of encoding in channel args)
- grpc/grpc#19405 (add metadata abstraction to LB policy API) and grpc/grpc#20443 (improve metadata iteration interface used by LB policies)
- grpc/grpc#18917 (use an interface to hide real subchannel impl from LB policies) and grpc/grpc#19390 (hide `ConnectedSubchannel` from LB policy API)
- grpc/grpc#19066 (remove channelz from LB policy API)
- grpc/grpc#19049 (various LB policy API improvements)
- grpc/grpc#19036 (add API for LB policies to access backend metric data)
- grpc/grpc#18081 (restructure how LB policies get their config)

And here are the ones that I think are small enough that they should not require a gRFC:

- grpc/grpc#25215 (changed resolver base class to no longer take ownership of `WorkSerializer` or `ResultHandler`)
- grpc/grpc#23417 (make request path more efficient to access in LB policies)
- grpc/grpc#23024 (add experimental call attributes API for LB policies)
- grpc/grpc#20441 (use `std::function<>` for LB policy callback to intercept trailing metadata)
- grpc/grpc#20090 (some cosmetic changes to the LB policy API suggested by Sanjay)
- grpc/grpc#19394 (pass error to LB policy trailing metadata callback)
- grpc/grpc#19389 (simplified shutdown of both resolvers and LB policies)
- grpc/grpc#19038 (remove `CreateChannel()` from LB policy helper API)
- grpc/grpc#18628 (remove error from connectivity state reporting) and then later grpc/grpc#23480 (add error back in using `absl::Status`)
- grpc/grpc#18251 (LB policy config includes policy name)
- grpc/grpc#18094 (remove call context from LB policy API)
- grpc/grpc#16342 (remove requirement that resolver `RequestReresolutionLocked()` must immediately return a new result)
- grpc/grpc#16225 (add experimental API for resetting connection backoff)
- grpc/grpc#14733 (fix resolver `NextLocked()` semantics to differentiate between fatal and transient failures)

(I think there are more in that last category, but I didn't take the time to go back further into the history.)

Given the above, I propose something like the following for these extension points:
- If the change is a major structural change or fundamental API redesign, it requires a gRFC.
- If the change affects almost all implementations of an API that is used externally, it requires a gRFC.
- If the change adds, removes, or modifies a significant part of the API surface (i.e., more than just an individual parameter), it requires a gRFC.
- Any smaller change not in the above categories does not require a gRFC.

WDYT?

P5-core-large-scale-changes.md
```diff
@@ -0,0 +1,63 @@
...
+* Changes that require downstream consumers of gRPC Core to modify their code.
+* Changes with a significant performance impact.
+* Changes to vocabulary types - types that are used broadly within the library.
+* Changes that constrain future development.
+* Changes that modify system architecture.
```
markdroth, line 40: Same question here, an example would help.

P5-core-large-scale-changes.md
```diff
@@ -0,0 +1,63 @@
...
+
+Since implementation experience can affect how a large scale change may proceed, it's additionally proposed that LSC gRFC's may be updated by later PR's against the approved change.
```
markdroth, line 42: It might be a good idea to split this part into a separate P-series gRFC.  In practice, we've actually been just sending PRs to update existing gRFCs when we need to, without any real review period at all.  If we want to formalize this, then this process should apply to more than just C-core changes.

markdroth (commented):
Sorry for the delay in reviewing!  Please let me know what you think.

P5-core-large-scale-changes.md
ctiller: Not opposed, but wanted to throw out an alternative:

This gRFC includes a light-weight update procedure for prior gRFC's that fall into it's domain.
Suppose we retroactively document the interfaces we're worried about here and publish them as gRFC's and then going forward we use the update process (and tune the timing to something that makes sense).

P5-core-large-scale-changes.md
ctiller: Yeah agree.

ctiller: I was trying to not re-trigger the review period :)

P5-core-large-scale-changes.md
```diff
@@ -0,0 +1,63 @@
...
+Any large scale change to gRPC Core should not be implemented without following the gRFC process.
+
+A large scale change is one that has a wide-ranging impact on the implementation of gRPC Core.
+
+Such changes include:
+* Changes to extensibility points, or the types used by extensibility points.
+* Changes that require downstream consumers of gRPC Core to modify their code.
+* Changes with a significant performance impact.
+* Changes to vocabulary types - types that are used broadly within the library.
+* Changes that constrain future development.
```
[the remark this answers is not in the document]
ctiller: One that's happened before: L6 introduced the restriction that a C++11 compiler was required to compile Core, L59 required a standard library be dynamically linkable. Neither formally required a gRFC (no API changes), but we did so anyway because it felt right - this tries to formalize that.

New changes that might fall into that: taking a large dependency, refactoring in a way that prevents a known use case... things that would require a large scale change to undo.

P5-core-large-scale-changes.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Proposal
+
+Any large scale change to gRPC Core should not be implemented without following the gRFC process.
+
+A large scale change is one that has a wide-ranging impact on the implementation of gRPC Core.
+
+Such changes include:
+* Changes to extensibility points, or the types used by extensibility points.
+* Changes that require downstream consumers of gRPC Core to modify their code.
```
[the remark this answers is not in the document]
ctiller: I think I was thinking about internal extensibility... I'll clean this one up.

P5-core-large-scale-changes.md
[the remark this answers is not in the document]
ctiller: Changed text.

P5-core-large-scale-changes.md
ctiller: Added some text.

ctiller:
What are next steps here?

ctiller:
@a11r this has now been idle for 2 months - what are the next steps here?

PR #234, opened 2021-04-30
Add xds-enabled server creation details for Core and C++

yashykt:
I've updated the A36-xds-for-servers.md doc with API details for Core and C++. Please let me know if a new gRFC would be preferred.

### DIFF
--- a/A36-xds-for-servers.md
+++ b/A36-xds-for-servers.md
@@ -79,7 +79,7 @@ SETTINGS). With either behavior, client connection attempts will quickly fail
 and clients would not perform further attempts without a backoff. Load balancing
 policies like `pick_first` would naturally attempt connections to any remaining
 addresses to quickly find an operational backend. However, the
-`accept()`+`listen()` approach will be improperly detected as server liveness
+`accept()`+`close()` approach will be improperly detected as server liveness
 for TCP heath checking.
 
 If the xDS bootstrap is missing or invalid, implementations would ideally fail
@@ -290,7 +290,7 @@ validation.
 ### Language-specifics
 
 The overall "wrapping" server API has many language-specific ramifications. We
-show each individual language's approach. For C++ and wrapped languages we just
+show each individual language's approach. For wrapped languages we just
 show a rough sketch of how they would be done.
 
 XdsClients may be shared without impacting this design. If shared, any mentions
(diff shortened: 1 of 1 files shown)


A36-xds-for-servers.md
```diff
@@ -299,36 +299,94 @@ reference" and "release a reference" on the shared instance, or similar
...
+the server. Additionally, when using a server config fetcher, the bound port
+might not be available immediately as a return value of the
+`grpc_server_add_secure_http2_port()` or the
+`grpc_server_add_insecure_http2_port()` API when using a wildcard port input.
+This arises from implementation details where Core currently invokes both
+`bind()` and `listen()` as part of `grpc_tcp_server_add_port()`, and as detailed
+earlier, since we would ideally not want to invoke `listen()` until we have a
+valid xDS configuration, `grpc_tcp_server_add_port()` is only invoked when the
+server is first ready to serve. (Note that any future transitions to the not
+serving state are still dealt with the `accept()`+`close()` method.) This
```
markdroth, line 343: s/dealt with the/dealt with via the/

A36-xds-for-servers.md
```diff
@@ -299,36 +299,94 @@ reference" and "release a reference" on the shared instance, or similar
...
-templates. If this becomes a problem more methods could become `virtual`.
+/** Sets the server's config fetcher.  Takes ownership. Must be called before 
+    adding ports. */
+GRPCAPI void grpc_server_set_config_fetcher(
+    grpc_server* server, grpc_server_config_fetcher* config_fetcher);
+```
+
+The server needs to be configured with the config fetcher before we add ports to
+the server. Additionally, when using a server config fetcher, the bound port
+might not be available immediately as a return value of the
```
markdroth, line 335: I think we should say this a little more strongly: it means that it is not supported to use an ephemeral port with and xDS-enabled gRPC server.

A36-xds-for-servers.md
```diff
@@ -299,36 +299,94 @@ reference" and "release a reference" on the shared instance, or similar
...
+class XdsServerBuilder : public ::grpc::ServerBuilder {
+ public:
+  // It is the responsibility of the application to make sure that \a notifier
+  // outlasts the life of the server. Notifications will start being made
+  // asynchronously once `BuildAndStart()` has been called. Note that it is
+  // possible for notifications to be made before `BuildAndStart()` returns.
+  void set_status_notifier(XdsServerServingStatusNotifierInterface* notifier);
+```
 
+The same consideration for bound port number not being available for wildcard
```
markdroth, line 388: Same as above: We should explicitly state that ephemeral ports are not supported via `XdsServerBuilder`.

A36-xds-for-servers.md
```diff
@@ -299,36 +299,94 @@ reference" and "release a reference" on the shared instance, or similar
...
+port inputs provided via `AddListeningPort` applies to `XdsServerBuilder`.
 
 ### Wrapped Languages
```
markdroth, line 391: The second paragraph of this section can now be removed (the one starting with "C will expose an API for wrapped languages to utilize").

markdroth (commented):
Thanks for doing this!  Just a few minor comments.

A36-xds-for-servers.md
```diff
@@ -299,36 +299,94 @@ reference" and "release a reference" on the shared instance, or similar
...
+`grpc_server_add_secure_http2_port()` or the
+`grpc_server_add_insecure_http2_port()` API when using a wildcard port input.
+This arises from implementation details where Core currently invokes both
+`bind()` and `listen()` as part of `grpc_tcp_server_add_port()`, and as detailed
+earlier, since we would ideally not want to invoke `listen()` until we have a
+valid xDS configuration, `grpc_tcp_server_add_port()` is only invoked when the
+server is first ready to serve. (Note that any future transitions to the not
+serving state are still dealt with the `accept()`+`close()` method.) This
+behavior might change in the future by splitting `grpc_tcp_server_add_port` so
+that `bind()` and `listen()` are done separately, allowing `bind()` to be
```
ejona86, line 345: I thought we decided that the wildcard port was chosen by listen(). So this isn't applicable. I removed the language about splitting bind+listen long before merging the gRFC. The choices are "not listen()" or "accept()+close()".

A36-xds-for-servers.md
yashykt: The reason for the not-so-strong language is to allow for flexibility to add support for wildcard ports, unless we want to say that we are never going to consider doing that.

A36-xds-for-servers.md
yashykt: This is the relevant comment from @markdroth in the original PR
```
I think it's worth pointing out that calling listen() before the xDS resource is obtained is actually worse for clients, because even without calling accept(), it will still cause timeout delays before pick_first can fail over to the next address in the list. So I think implementations should prefer to defer calling listen() until after the xDS resource is obtained. That's what we're planning to do in C-core.

Also, the language here seems to imply that bind() should be done immediately even if listen() is delayed. Is there actually any reason for that? Based on what @yashykt and I are looking at in C-core, it seems like the only case where it would matter that bind() happens before we get the xDS resource is a server using an ephemeral port. And that seems like such a niche case that I think I'd be okay saying that we just don't support it for the xDS-enabled server API. (To be clear, it does seem better to call bind() earlier if possible, so that we can support ephemeral ports, but I think it's a nice-to-have, not a hard requirement. In C-core, it looks like that may be hard to de-couple, so at least initially we will probably start by deferring both bind() and listen() until after we have the xDS resource. We can revisit that later if someone actually wants to use ephemeral ports with xDS.)
```
From this I'm interpreting that splitting "bind()" and "listen()" is still a possible strategy.

A36-xds-for-servers.md
```diff
@@ -299,36 +299,94 @@ reference" and "release a reference" on the shared instance, or similar
...
+                                   grpc_status_code code,
+                                   const char* error_message);
+  void* user_data;
+} grpc_server_xds_status_notifier;
+
+typedef struct grpc_server_config_fetcher grpc_server_config_fetcher;
+
+/** Creates an xDS config fetcher. */
+GRPCAPI grpc_server_config_fetcher* grpc_server_config_fetcher_xds_create(
+    grpc_server_xds_status_notifier notifier, const grpc_channel_args* args);
```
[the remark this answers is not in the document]
yashykt: These channel args are used for the creation of the xds client. At present, we only seem to use this to pass in ```GRPC_ARG_TEST_ONLY_DO_NOT_USE_IN_PROD_XDS_BOOTSTRAP_CONFIG```. I'm not sure if we would need this for anything else in the future. @markdroth

A36-xds-for-servers.md
ejona86: There was a lot more discussion than that[1][2][3], but in the end all the reference to bind() was removed in favor of focusing on listen() and "XdsServer will not support ephemeral ports as part of this gRFC but may be enhanced in the future."

I guess this text is fine as-is. It initially seemed like a lot about a topic that wasn't part of the gRFC, but it is really only one mention of splitting bind()+listen(), but another mention to bind() to make that make sense. That's fair. Let's keep it as-is.

1. https://github.com/grpc/proposal/pull/214#issuecomment-749280531
2. https://github.com/grpc/proposal/pull/214#discussion_r552172048
3. after other discussions, https://github.com/grpc/proposal/pull/214/commits/a933f33729a327fc153274b71fdd6a41b340b25b

A36-xds-for-servers.md
```diff
@@ -299,36 +299,94 @@ reference" and "release a reference" on the shared instance, or similar
 behavior. Such sharing does not avoid the need of shutting down XdsClients when
 no longer in use; they are a resource and must not be leaked.
 
-### C++
+### Core
```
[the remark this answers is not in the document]
markdroth: +1 -- this should say "C-core".

A36-xds-for-servers.md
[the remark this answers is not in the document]
markdroth: That arg is the reason we plumbed the channel args in there, but in principle they could be used for other things later.  They will just be whatever args were passed to the server.

A36-xds-for-servers.md
markdroth: I think the current language is so fuzzy that it really doesn't tell the reader anything useful.  If the concern is that we want to be able to change this later, then let's just say something like "Note that we will initially not support ephemeral ports on an xDS-enabled server, but we might add that later."

markdroth (approved):
Looks great!

ejona86 (approved):

yashykt:
Thanks for the reviews!
I'll merge this next Friday assuming there are no objections on the API.

yashykt:
Don't see any objections on the discussion thread. Merging this. Thanks for reviewing!

PR #236, opened 2021-05-06
A37 update: DNS name to resolve is encoded in CDS resource for LOGICAL_DNS

markdroth:


A37-xds-aggregate-and-logical-dns-clusters.md
```diff
@@ -59,6 +59,7 @@ New fields will be added to the data structure reported to the ClusterWatcher
 to represent logical DNS and aggregate clusters.  Specifically:
 
 - cluster type (EDS, logical DNS, or aggregate)
+- for logical DNS clusters, the DNS name to resolve (in "host:port" form)
```
ejona86, line 62: If our client is doing the concatenation, we should probably have an IPv6-literal test... Seems very unlikely, but still. Just something to think about.

ejona86 (approved):

A37-xds-aggregate-and-logical-dns-clusters.md
markdroth: Good idea.  @apolcyn, please consider this in the interop tests.

(In C-core, we use an existing library function to join the host and port, and that function handles IPv6, so I'd be surprised if there are any problems here.)

PR #238, opened 2021-05-13
L80: Add a proposal for ClientAsyncResponseReaderInterface destruction.

jacobsa:
See previous discussion in https://github.com/grpc/grpc/pull/26233.

@ctiller I've set you as the approver after the discussion in the thread above. Feel free to re-assign if that's not appropriate. Thanks!

### DIFF
--- a/L80-cpp-async-response-reader-destruction.md
+++ b/L80-cpp-async-response-reader-destruction.md
@@ -1,0 +1,115 @@
+API changes for `ClientAsyncResponseReaderInterface` destruction
+----
+* Author: jacobsa
+* Approver: soheilhy
+* Status: Draft
+* Implemented in: C++
+* Last updated: 2021-05-14
+* Discussion at: https://groups.google.com/g/grpc-io/c/IzRZVDo4W8c
+
+## Abstract
+
+Currently although
+[`grpc::ClientAsyncResponseReaderInterface`](https://github.com/grpc/grpc/blob/7e14d23ab46e1b0924d6c3e40797ed1d587aee7a/include/grpcpp/impl/codegen/async_unary_call.h#L37-L72)
+has a virtual destructor, `std::default_delete` is specialized such that the
+destructor never actually runs. So the only subclasses that can be implemeneted
+without hacks are those like `grpc::ClientAsyncResponseReader`, which knows it
+is always allocated on an arena and doesn't care about its destructor running.
+This proposal provides a plan for guaranteed destructor invocation, such that
+subclasses can count on their destructor running and their backing memory being
+freed, as with any other C++ class.
+
+This proposal requires a bump of the major version.
+
+## Background
+
+We currently [specialize](https://github.com/grpc/grpc/blob/7e14d23ab46e1b0924d6c3e40797ed1d587aee7a/include/grpcpp/impl/codegen/async_unary_call.h#L404-L408)
+`std::default_delete` for
+`ClientAsyncResponseReaderInterface` to do nothing. This means that, despite the
+fact that async stub methods return
+`std::unique_ptr<ClientAsyncResponseReaderInterface>`, it's impossible to have a
+subclass of that interface that is allocated on the heap or whose destructor has
+side effects without a hack that manages ownership _outside of_ the `unique_ptr`
+returned by the stub.
+
+This was introduced by
+[`dd36b153`](https://github.com/grpc/grpc/commit/dd36b15315cd691e86a94d4574bd9f3e3a33633f),
+which added the specialization for `grpc::ClientAsyncResponseReader` itself, and
+[`e8a61d63`](https://github.com/grpc/grpc/commit/e8a61d63b5a6db0f81b688481a9485d412e5a41e),
+which did so for the interface. These commits didn't document their intent, but
+I suspect it was a combination of the following facts:
+
+*   In production uses, `ClientAsyncResponseReaderHelper::Create`
+    [allocates](https://github.com/grpc/grpc/blob/7e14d23ab46e1b0924d6c3e40797ed1d587aee7a/include/grpcpp/impl/codegen/async_unary_call.h#L96-L99)
+    the `ClientAsyncResponseReader` on an arena.
+
+*   Although it's technically non-trivial, the `ClientAsyncResponseReader`
+    destructor doesn't actually have any important side effects.
+
+*   Therefore it's not necessary to run the destructor, and it's not correct to
+    use `delete` to free the backing memory.
+
+From this point of view, the original commit (`dd36b153`) is technically
+correct: `std::unique_ptr<ClientAsyncResponseReader>` doesn't actually need to
+do anything. But:
+
+*   Of course the objects may be deleted other ways. If I have a
+    `std::unique_ptr<ClientAsyncResponseReaderInterface>` (i.e. without a
+    specific deleter type), it would be natural for me to assume that I can do
+    `delete p.release()` to get the same effect as destroying the unique
+    pointer. But that is not true, and doing so will be undefined behavior.
+
+    So from this point of view `dd36b153` is not correct, or at least makes for
+    a surprising API.
+
+*   `ClientAsyncResponseReader` is not the only subclass of the interface. Even
+    though subclasses can override the virtual destructor to add a side effect,
+    their destructors do not actually run. This results in subtle bugs and
+    confusing behavior, specially because `delete p.release()` works differently
+    than just letting a `std::unique_ptr p` go out of scope.
+
+    So from this point of view `e8a61d63` is also not correct.
+
+This is not a theoretical concern: there are multiple hacks in user code working
+around this with a grumbling tone. For example, see
+[this hack](https://github.com/googleapis/google-cloud-cpp/blob/74de20227371119ea8b8c788d37ae4e24be8ed5d/google/cloud/bigtable/testing/mock_response_reader.h#L77-L106)
+in the cloud bigtable library that uses the phrase "terrible, horrible, no
+good, very bad hack" that involves managing a
+`ClientAsyncResponseReaderInterface` object with **two** `unique_ptr`s (making
+them not so unique).
+
+## Proposal
+
+The plan is to land this change a single commit that does the following:
+
+*   Add a `virtual void Destroy()` method to
+    `ClientAsyncResponseReaderInterface`, and update the `std::default_delete`
+    specializations to call it.
+
+*   Make the interface itself implement the method by doing `delete this`, a
+    suitable default behavior that frees future subclasses from having to think
+    about this problem at all.
+
+*   Make `ClientAsyncResponseReader` implement the method by doing nothing,
+    with an explanation that it knows it is allocated on an arena and has
+    historically never run its destructor.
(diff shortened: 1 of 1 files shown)


L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,123 @@
...
+
+Currently although
+[`grpc::ClientAsyncResponseReaderInterface`](https://github.com/grpc/grpc/blob/7e14d23ab46e1b0924d6c3e40797ed1d587aee7a/include/grpcpp/impl/codegen/async_unary_call.h#L37-L72)
+has a virtual destructor, `std::default_delete` is specialized such that the
+destructor never actually runs. So the only subclasses that can actually be
+implemeneted without hacks are those like `grpc::ClientAsyncResponseReader`,
+which knows it is always allocated on an arena and doesn't care about its
+destructor running. This proposal provides a plan for fixing this so that other
+subclasses can count on their destructor running and their backing memory being
+freed, as with any other C++ class.
```
soheilhy, line 20: >This proposal provides a plan for fixing this so that other
> subclasses can count on their destructor running and their backing memory being
> freed, as with any other C++ class.

I would replace this with: 

```
This proposal provides a plan for guaranteed destructor invocation, such that 
subclasses can count on their destructor running and their backing memory being
freed, as with any other C++ class.
```

L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,123 @@
...
+`ClientAsyncResponseReaderInterface` to do nothing. This means that, despite the
+fact that async stub methods return
+`std::unique_ptr<ClientAsyncResponseReaderInterface>`, it's impossible to have a
+subclass of that interface that is allocated on the heap or whose destructor has
+side effects without a hack that manages ownership _outside of_ the `unique_ptr`
+returned by the stub.
+
+This was introduced by
+[`dd36b153`](https://github.com/grpc/grpc/commit/dd36b15315cd691e86a94d4574bd9f3e3a33633f),
+which introduced the specialization for `grpc::ClientAsyncResponseReader`
```
soheilhy, line 35: nit: I would replace `introduced` with `included` to avoid repetition.

L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,123 @@
...
+    specific deleter type), it would be natural for me to assume that I can do
+    `delete p.release()` to get the same effect as destroying the unique
+    pointer. But that is not true, and doing so will be undefined behavior.
+
+    So from this point of view `dd36b153` is not correct, or at least makes for
+    a surprising API.
+
+*   `ClientAsyncResponseReader` is not the only subclass of the interface. The
+    interface's destructor is virtual, so other subclasses may have destructors
+    with side effects. However they are never actually run! This is a huge
```
soheilhy, line 66: > The interface's destructor is virtual, so other subclasses may have destructors
>  with side effects. However they are never actually run! ...

I would replace this with:

```
Even though subclasses can override the virtual destructor to add a side effect, their dtor does not actually run. This results in subtle bugs and confusing behavior, specially because   `delete p.release()` works differently than
 just letting a `std::unique_ptr p` go out of scope.
```

L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,123 @@
...
+    surprise: it took me hours yesterday to figure out that I wasn't actually
+    crazy: `delete p.release()` really was working differently than just letting
+    a `std::unique_ptr p` go out of scope.
+
+    So from this point of view `e8a61d63` is also not correct.
+
+This is not a theoretical concern: I've found multiple hacks in user
```
soheilhy, line 73: nit: There are multiple hacks ...

L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,123 @@
...
+code working around this with a grumbling tone. For example, see
+[this hack](https://github.com/googleapis/google-cloud-cpp/blob/74de20227371119ea8b8c788d37ae4e24be8ed5d/google/cloud/bigtable/testing/mock_response_reader.h#L77-L106)
+in the cloud bigtable library that uses the phrase "terrible, horrible, no
+good, very bad hack" that involves managing a
+`ClientAsyncResponseReaderInterface` object with **two** `unique_ptr`s (making
+them not so unique).
+
+## Proposal
+
+My plan is split into two commits:
```
soheilhy, line 83: nit: I would reword this as: The plan is to land this change into steps

L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,123 @@
...
+    Thus with this commit there is no behavior change for anybody, except in
+    one edge case: if a hack similar to the one mentioned above didn't ensure
+    that the `unique_ptr` to the interface didn't outlive the unique pointer to
+    the concrete subclass, then the former's call to `Destroy` would be
+    undefined behavior after the latter destroyed the object itself (and its
+    vtable). This is unfortunate, but not likely to be common, and the whole
+    point of this proposal is to make such hacks unnecessary.
+
+2.  Change the interface's implementation of the `Destroy` method to do `delete
+    this`, a more suitable default behavior that frees future subclasses from
```
soheilhy, line 106: This can break some random subclass that was relying on the dtor being a no-op.  I believe it needs a major gRPC version bump when we merge the commit.

We should clarify that here in the proposal.

L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,123 @@
...
+    only because it may or may not be more controversial. It also allows time
+    to update subclasses in Google's monorepo after (1).
+
+## Rationale
+
+I came across this problem while trying to write mocking support for async
+methods. It works just fine once this problem is fixed, but this problem makes
+it extremely awkward otherwise.
+
+The obvious alternative is to leave things as-is. But this requires awful hacks
```
soheilhy, line 120: nit: as-is, but this requires ...

soheilhy (commented):
Thank you for creating the gRFC.

L79-cpp-async-response-reader-destruction.md
jacobsa: That is in theory also true of the change above; see the last paragraph. Does it need a major bump? Are major version bumps actually done, or does that make this a no-go?

L79-cpp-async-response-reader-destruction.md
jacobsa: Done.

L79-cpp-async-response-reader-destruction.md
jacobsa: Done.

L79-cpp-async-response-reader-destruction.md
jacobsa: Done.

L79-cpp-async-response-reader-destruction.md
jacobsa: Done.

L79-cpp-async-response-reader-destruction.md
jacobsa: Done (except I used `added`).

L79-cpp-async-response-reader-destruction.md
jacobsa: Done.

L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,123 @@
...
+* Implemented in: C++
+* Last updated: 2021-05-14
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+Currently although
+[`grpc::ClientAsyncResponseReaderInterface`](https://github.com/grpc/grpc/blob/7e14d23ab46e1b0924d6c3e40797ed1d587aee7a/include/grpcpp/impl/codegen/async_unary_call.h#L37-L72)
+has a virtual destructor, `std::default_delete` is specialized such that the
+destructor never actually runs. So the only subclasses that can actually be
```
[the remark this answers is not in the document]
jacobsa: Done.

jacobsa:
@soheilhy Thanks for looking. I've addressed all your comments except the one I left open. PTAL.

(Sorry if I'm doing this wrong; I have no idea what conventions are for PRs, so am just treating this like Critique.)

jacobsa:
Oh, I also set you to the approver in the doc. Also not sure what the process is there.

L79-cpp-async-response-reader-destruction.md
soheilhy: We call "major version" bump as really going from gRPC 1.x to 1.x+1 not gRPC 2. Yes this is something that's done often and there is no issue doing so, once the gRFC is approved.

L79-cpp-async-response-reader-destruction.md
```diff
@@ -0,0 +1,121 @@
+API changes for `ClientAsyncResponseReaderInterface` destruction
+----
+* Author: jacobsa
+* Approver: soheilhy
+* Status: Draft
+* Implemented in: C++
+* Last updated: 2021-05-14
+* Discussion at: <google group thread> (filled after thread exists)
```
soheilhy, line 8: Before merging the PR, we usually send an email to the mailing list and then merge this PR.

soheilhy (approved):
Please add the group link before merging the PR.

L79-cpp-async-response-reader-destruction.md
jacobsa: Okay, I added text saying each should result in a version bump. PTAL.

L79-cpp-async-response-reader-destruction.md
jacobsa: Thanks. Do you want me to start the thread? Can you tell me how/give me an example?

soheilhy (changes requested):
Sorry I am confused. I shouldn't have clicked approve.

We need two changes:
1) The title of the PR should change to include the suffix L79: ...
2) We need to post this PR to the group and post the thread link here for discussion.

L79-cpp-async-response-reader-destruction.md
jacobsa: Started the thread and left a link.

jacobsa:
Thanks, both done.

soheilhy:
> Thanks, both done.

All look great thanks!

jacobsa:
@soheilhy: I believe the 10 business days timeout has been hit. Are you okay with merging this now?

ctiller (approved):

soheilhy (approved):

soheilhy (changes requested):
@jacobsa Can you please update it to L80? Looks like L79 is already merged. Thanks!

jacobsa:
Done; it's now L80. I don't know how to "resolve" this comment in github though. :-/

jacobsa:
@soheilhy Sorry, meant to tag you before. As far as I can tell you need to clear the "requested changes" bit.

soheilhy (approved):
@jacobsa thanks for the update! yes, I have to click approve to make it ready to submit.

PR #242, opened 2021-06-05
L80: change to only a minor version bump.

jacobsa:
See discussion on https://github.com/grpc/grpc/pull/26389.

@soheilhy


soheilhy (approved):
Thank you1

PR #243, opened 2021-06-11
A29: xDS-Based Security for gRPC Clients and Servers

sanjaypujare:
CC @ejona86 @dfawley @yashykt @easwars pls review.

### DIFF
--- a/A29-xds-tls-security.md
+++ b/A29-xds-tls-security.md
@@ -410,29 +410,22 @@ The following fields from `CertificateValidationContext` are ignored:
 * `allow_expired_certificate`
 * `trust_chain_verification`
 
-[CPI]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L165
-[TCCPI]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L256
 [TCPI]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L247
-[VCCPI1]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L205
 [TLS-CERT]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L226
 [CERT-SDS]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L239
-[CERT-PROVIDER]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L251
 [VAL-SDS]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L265
-[VAL-PROVIDER]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L278
 [TLS-PARAMS]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L212
 [CUSTOM-HS]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L301
 [ALPN-PROTOCOLS]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L297
 [UTC]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L26
 [server-authz]: #server-authorization-aka-subject-alt-name-checks
-[VCCP1]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L199
-[VCSSC1]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L193
 [CCPI]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L315
 [CPPI]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L241
 
 ### Certificate Provider Plugin Framework
 
 xDS supports many different configurations for endpoints to obtain certificates but gRPC supports only
-[`CertificateProviderPluginInstance`][CPPI] (or its deprecated version [`CertificateProviderInstance`][CPI]).
+[`CertificateProviderPluginInstance`][CPPI].
 In this approach, the control plane tells the xDS client the [name of a certificate][CERT-NAME] and the
 ["instance" name][INST-NAME] of a provider to use to obtain a certificate. This provider
 [instance name][INST-NAME] is translated into a provider implementation and a configuration
--- a/A29-xds-tls-security.md
+++ b/A29-xds-tls-security.md
@@ -384,10 +384,10 @@ not take into account whether Xds credentials are in effect for the respective c
 A `CertificateValidationContext` (either the [`validation_context`][validation_context] field or
 the [`default_validation_context`][default_validation_context] field inside of the
 [`combined_validation_context`][CVC]) is validated in any CDS/LDS update as follows:
-The field [match_subject_alt_names][] is used only on the client side i.e. inside
-`UpstreamTlsContext` and is processed as described [above][server-authz]; it is ignored on the
-server side. If any of the other fields (listed below) are present, the update is NACKed because
-ignoring these fields compromises security:
+The field [match_subject_alt_names][] is used on the client side i.e. inside `UpstreamTlsContext`
+as described [above][server-authz]. On the server side an implementation may support this field
+and its semantics, or NACK the update it if it doesn't. If any of the other fields (listed below)
+are present, the update is NACKed because ignoring these fields compromises security:
 
 * `verify_certificate_spki`
 * `verify_certificate_hash`
--- a/A29-xds-tls-security.md
+++ b/A29-xds-tls-security.md
@@ -478,7 +478,8 @@ For example, the bootstrap file might contain the following:
 
 With this configuration, when the xDS server tells the client to get a certificate from
 plugin instance `"google_cloud_private_spiffe"` (the value of [`instance_name`][INST-NAME]),
-the client will load the certificate data from the specified files.
+the client will load the certificate data from the specified files. Note the
+[certificate_name][CERT-NAME] value is currently ignored by this plugin.
 
 [DURATION-JSON]: https://developers.google.com/protocol-buffers/docs/proto3#json.
 
--- a/A29-xds-tls-security.md
+++ b/A29-xds-tls-security.md
@@ -147,8 +147,9 @@ Use of XdsServerCredentials:
 xDS v3 is a pre-requisite for this proposal because the required fields were added to xDS v3. 
 gRPC uses CDS to configure client side security and LDS for server side security as described in the
 sections below. The sections describe how the CDS or LDS updates are processed by gRPC. Note that gRPC
-always validates all security configurations in CDS or LDS regardless of whether an XdsChannelCredentials
-or XdsServerCredentials is in force. If validation fails on an update, the update is NACKed.
+always validates all security configurations in CDS or LDS regardless of whether the application used
+`XdsChannelCredentials` or `XdsServerCredentials`. If validation fails on an update, the update is
+NACKed.
 
 An unsupported field is normally ignored. However if ignoring a field compromises security, or
 if the unsupported field affects how we interpret other fields, we NACK the update when the
@@ -170,23 +171,29 @@ load balancer policy information down to child policies.
 [`common_tls_context`][CTC] in the `UpstreamTlsContext` contains the required security
 configuration. See below for [`CommonTlsContext`][CTC-type] processing.
 
-A secure client at a minimum requires a CA root certificate to be able to validate the server
-certificate in the TLS mode and this is provided by [ca_certificate_provider_instance][CCPI]
-inside the [`default_validation_context`][default_validation_context] of
-[combined_validation_context][CVC]. For backward compatibility, gRPC will use the deprecated
-field [`validation_context_certificate_provider_instance`][VCCPI1] inside
-[`combined_validation_context`][CVC] if [ca_certificate_provider_instance][CCPI] is not
-present. If [`UpstreamTlsContext`][UTC] is present but
-[`combined_validation_context`][CVC] is not present or neither of
-[ca_certificate_provider_instance][CCPI] and [`validation_context_certificate_provider_instance`][VCCPI1]
-is set, then gRPC will NACK the CDS update. The client identity certificate is configured via
-[`tls_certificate_provider_instance`][TCPI] (or the deprecated
-[`tls_certificate_certificate_provider_instance`][TCCPI] if [`tls_certificate_provider_instance`][TCPI]
-is not set). Note that the identity certificate on the client side is optional. The
-client certificate is sent to the server if the server requests or requires it in the
-TLS handshake. If the server requires the client certificate but is not configured on the
-client side then the TLS handshake will fail because of the client's inability to send the
-certificate.
+A secure client performs validation of the server's certificate. This validation is configured
+via the `CertificateValidationContext` message, which is present in either the
+[`validation_context`][validation_context] field or in the
+[`default_validation_context`][default_validation_context] field inside of the
+[`combined_validation_context`][CVC] field. If neither of those fields are set, gRPC will NACK
+the CDS update.
+
+At minimum, a secure client requires a CA root certificate to be able to validate the server
+certificate. The CA root certificate is obtained using the certificate provider instance configured
+via the [`ca_certificate_provider_instance`][CCPI] field of the `CertificateValidationContext`
+message. If this field is not present, or if it specifies a certificate provider instance that
+is not configured in the bootstrap file, gRPC will NACK the CDS update.
+
+The client's identity certificate, if configured, is obtained using the certificate provider instance
+configured via the [`tls_certificate_provider_instance`][TCPI] field of the `CommonTlsContext` message.
+If this field is set, the client certificate is sent to the server if the server requests or requires
+it in the TLS handshake. If the server requires the client certificate but is not configured on the
+client side then the TLS handshake will fail because of the client's inability to send the certificate.
+If this field specifies a certificate provider instance that is not configured in the bootstrap file,
+gRPC will NACK the CDS update. Note that gRPC does not support any of the other mechanisms in xDS for
+configuring the client's identity certificate. If the [`tls_certificate_provider_instance`][TCPI] field
+is unset but either of the [`tls_certificates`][TLS-CERT] or [`tls_certificate_sds_secret_configs`][CERT-SDS]
+fields are set, gRPC will NACK the CDS update.
 
 The following fields in a CDS update are ignored:
 
@@ -210,10 +217,9 @@ a server for the connection as described below. This check is in lieu of the
 canonical [hostname check in Web PKI][RFC6125-6].
 
 If [`match_subject_alt_names`][match_subject_alt_names] in
-[`default_validation_context`][default_validation_context] of
-[combined_validation_context][CVC] of the received `UpstreamTlsContext` is populated
-then gRPC checks the SAN entries in the leaf server certificate against the
-`match_subject_alt_names` values as follows:
+the `CertificateValidationContext` is populated then gRPC checks the SAN entries
+in the leaf server certificate against the
+[`match_subject_alt_names`][match_subject_alt_names] values as follows:
 
 * if the `match_subject_alt_names` list is empty then there is nothing to match and the
 check succeeds.
@@ -254,6 +260,7 @@ implementation through the `SslContext` provided to the client sub-channel.
 [no-leading-0]: https://datatracker.ietf.org/doc/html/rfc5952#section-2.1
 [lower-case]: https://datatracker.ietf.org/doc/html/rfc5952#section-2.3
 [DNS-wildcard]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L392-L394
+[validation_context]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L261
 
 #### LDS for Server Side Security
 
@@ -275,22 +282,31 @@ is not present) then gRPC uses the fallback credentials for the incoming connect
 [`common_tls_context`][CTC1] in the `DownstreamTlsContext` contains the required
 security configuration. See below for [`CommonTlsContext`][CTC-type] processing.
 
-A secure server at a minimum requires an identity certificate which is provided
-by [`tls_certificate_provider_instance`][TCPI] (or in its absence the deprecated
-field [`tls_certificate_certificate_provider_instance`][TCCPI]). If both fields are
-empty, gRPC will NACK an update as part of validating all LDS updates. If
-[ca_certificate_provider_instance][CCPI] inside the
-[`default_validation_context`][default_validation_context] of [combined_validation_context][CVC]
-(or in its absence the deprecated field
-[`validation_context_certificate_provider_instance`][VCCPI1] inside
-[`combined_validation_context`][CVC]) is provided then the server will request
-the client certificate. When [`require_client_certificate`][RCC] is `true`,
-gRPC requires the client certificate and will reject a connection without a valid
-client certificate. When [`require_client_certificate`][RCC] is `true`, the
-[ca_certificate_provider_instance][CCPI] field (or in its absence
-[`validation_context_certificate_provider_instance`][VCCPI1]) should be
-present otherwise gRPC will NACK an LDS update as part of validating all LDS
-updates.
+The server's identity certificate is obtained using the certificate provider instance
+configured via the [`tls_certificate_provider_instance`][TCPI] field of the
+`CommonTlsContext` message. If this field is not present, or if it specifies a certificate
+provider instance that is not configured in the bootstrap file, gRPC will NACK the LDS
+update.
+
+If a secure server is configured for mTLS, it will need configuration for how to validate
+the client's certificate. This validation is configured via the `CertificateValidationContext`
+message, which is present in either the [`validation_context`][validation_context] field or
+in the [`default_validation_context`][default_validation_context] field inside of the
+[`combined_validation_context`][CVC] field. If neither of those fields are set, the server
+will not request the client's certificate during the TLS handshake (i.e., it will use TLS
+instead of mTLS).
+
+If a `CertificateValidationContext` is provided, then the server requires a CA root certificate
+to be able to validate the client certificate. The CA root certificate is obtained using the
+certificate provider instance configured via the [`ca_certificate_provider_instance`][CCPI]
+field. If this field is not present, or if it specifies a certificate provider instance that
+is not configured in the bootstrap file, gRPC will NACK the LDS update.
+
+If the [`require_client_certificate`][RCC] field in the `DownstreamTlsContext` message is set
+to true, gRPC requires the client certificate and will reject a connection without a client
+certificate that is successfully validated as per the `CertificateValidationContext`. Note
+that if [`require_client_certificate`][RCC] is true, and if no `CertificateValidationContext`
+is provided, gRPC will NACK the LDS resource.
 
 The following fields in an LDS update are ignored:
 
@@ -302,8 +318,18 @@ The following fields in an LDS update are ignored:
 If any of the following fields are present, gRPC will NACK an LDS update as
 part of validating all LDS updates:
 
-* [require_sni][REQ-SNI]
-* [ocsp_staple_policy][OCSP-STAPLE-POLICY]: any value other than `LENIENT_STAPLING` causes NACK.
+* [require_sni][REQ-SNI]: when this is set to `true`, it requires the server to
+["reject connections without a valid and matching SNI"][sni-comment]. However SNI is
+almost always used for "name-based virtual hosting" web-servers and is not applicable
+to gRPC. Silently ignoring the `true` value will result into gRPC not rejecting connections
+without a valid and matching SNI thereby making the implementation less secure than
+what the control plane intended, hence gRPC will NACK such an LDS update.
+
+* [ocsp_staple_policy][OCSP-STAPLE-POLICY]: Instead of mandating the implementation of
+[`STRICT_STAPLING`][STRICT_STAPLING] and [`MUST_STAPLE`][MUST_STAPLE] in gRPC,
+we NACK any value other than [`LENIENT_STAPLING`][LENIENT_STAPLING] because ignoring
+the unsupported values would make the implementation less secure than what the control
+plane intended.
 
 [A36]: A36-xds-for-servers.md
 [filter-chains]: https://github.com/envoyproxy/envoy/blob/45ec050f91407147ed53a999434b09ef77590177/api/envoy/config/listener/v3/listener.proto#L120
@@ -321,54 +347,38 @@ part of validating all LDS updates:
 [DIS-STATELESS-SESS-RES]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L109
 [SESSION-TIMEOUT]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L116
 [OCSP-STAPLE-POLICY]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L124
+[sni-comment]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L90
+[STRICT_STAPLING]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L73
+[MUST_STAPLE]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L80
+[LENIENT_STAPLING]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L65
 
 #### CommonTlsContext Processing
 
 `CommonTlsContext` is present in both the `UpstreamTlsContext` and `DownstreamTlsContext` protos and contains the
 certificate and key configuration information needed on the client and server side respectively. The configuration
 tells gRPC how to obtain certificates and the keys for the TLS handshake. Although there are various ways to obtain
-certificates as per this proto (which are supported by Envoy), gRPC supports only two of them. The preferred mechanism
-is the [`CertificateProviderPluginInstance`][CPPI] proto and the less preferred mechanism is
-[`CertificateProviderInstance`][CPI] which is a deprecated proto but looks identical to
-[`CertificateProviderPluginInstance`][CPPI]. They are both based on the notion of CertificateProvider plugin framework
-(described later). The field `instance_name` defines a CertificateProvider "instance" that gRPC looks up in the bootstrap
-file (described later) to obtain the CertificateProvider configuration. This configuration along with the CertificateProvider
-plugin framework enables gRPC to acquire the certificates necessary for the TLS handshake.
+certificates as per this proto (which are supported by Envoy), gRPC supports only one of them and that is
+the [`CertificateProviderPluginInstance`][CPPI] proto. This proto is based on the notion of the CertificateProvider
+plugin framework (described later). The field `instance_name` defines a CertificateProvider "instance" that gRPC looks
+up in the bootstrap file (described later) to obtain the CertificateProvider configuration. This configuration along
+with the CertificateProvider plugin framework enables gRPC to acquire the certificates necessary for the TLS handshake.
 
-The field [`tls_certificate_provider_instance`][TCPI] (or in its absence the deprecated
-field [`tls_certificate_certificate_provider_instance`][TCCPI]) is used for the identity certificate and the
-private key. And [`ca_certificate_provider_instance`][CCPI] inside the
-[`default_validation_context`][default_validation_context] of [`combined_validation_context`][CVC]
-(or in its absence the deprecated field [`validation_context_certificate_provider_instance`][VCCPI1] inside
-[`combined_validation_context`][CVC]) is used for the root certificates for validating peer certificates.
+The field [`tls_certificate_provider_instance`][TCPI] is used for the identity certificate and the
+private key. And [`ca_certificate_provider_instance`][CCPI] inside a `CertificateValidationContext`
+is used for the root certificates for validating peer certificates.
 
 gRPC does not support the other certificate acquisition mechanisms specified by the following fields:
 
-* [tls_certificates][TLS-CERT]
-* [tls_certificate_sds_secret_configs][CERT-SDS]
-* [tls_certificate_certificate_provider][CERT-PROVIDER]
-* [validation_context_sds_secret_config][VAL-SDS]
-* [validation_context_certificate_provider][VAL-PROVIDER]
+* [tls_certificates][TLS-CERT]: if this is present but [`tls_certificate_provider_instance`][TCPI] is
+not present, then gRPC will NACK the update.
+* [tls_certificate_sds_secret_configs][CERT-SDS]: similar to above, if this is present but
+[`tls_certificate_provider_instance`][TCPI] is not present, then gRPC will NACK the update.
+* [validation_context_sds_secret_config][VAL-SDS]: if this is present but a
+`CertificateValidationContext` (either the [`validation_context`][validation_context] field or
+the [`default_validation_context`][default_validation_context] field inside of the
+[`combined_validation_context`][CVC]) is not present, then gRPC will NACK the update.
 
-When fields of type [`CertificateProviderPluginInstance`][CPPI] (or the deprecated
-[`CertificateProviderInstance`][CPI]) are present gRPC ignores the respective
-unsupported fields but if the unsupported fields are present by themselves gRPC will NACK an update
-as part of validating all CDS/LDS updates.
-
-For example, in [`UpstreamTlsContext`][UTC] (on the client side), the identity certificate is not
-needed for the TLS mode, which is indicated by the absence of both [`tls_certificate_provider_instance`][TCPI]
-and [`tls_certificate_certificate_provider_instance`][TCCPI]. However in this case if any of the other
-unsupported identity certificate fields ([tls_certificates][TLS-CERT], [tls_certificate_sds_secret_configs][CERT-SDS],
-[tls_certificate_certificate_provider][CERT-PROVIDER]) are present then gRPC will NACK a CDS update.
-But if [`tls_certificate_provider_instance`][TCPI] (or in its absence the deprecated
-[`tls_certificate_certificate_provider_instance`][TCCPI]) is present, then the unsupported
-fields are ignored. This allows a control plane to formulate a single `CommonTlsContext` that works
-for both gRPC (which only supports [`CertificateProviderPluginInstance`][CPPI] or the deprecated
-[`CertificateProviderInstance`][CPI]) and Envoy (which supports neither [`CertificateProviderPluginInstance`][CPPI]
-nor [`CertificateProviderInstance`][CPI]) to achieve consistent security configuration between
-gRPC and Envoy.
-
-The following fields are unsupported and if present will cause a NACK from gRPC for any update
+The following fields are unsupported and if present will cause a NACK from gRPC
 because ignoring these fields compromises security:
 
 * [tls_params][TLS-PARAMS]
@@ -379,11 +389,13 @@ on the client side (inside [`UpstreamTlsContext`][UTC]) is ignored if present.
 The processing and parsing of `CommonTlsContext` inside any particular CDS or LDS response does
 not take into account whether Xds credentials are in effect for the respective channel or server.
 
-[`combined_validation_context`][CVC] is validated in any CDS/LDS update as follows:
-The field [match_subject_alt_names][] inside [default_validation_context][] is accepted only on
-the client side i.e. inside `UpstreamTlsContext` and is processed as described [above][server-authz].
-If any of the other fields (listed below) are present in [default_validation_context][], the update
-is NACKed because ignoring these fields compromises security:
+A `CertificateValidationContext` (either the [`validation_context`][validation_context] field or
+the [`default_validation_context`][default_validation_context] field inside of the
+[`combined_validation_context`][CVC]) is validated in any CDS/LDS update as follows:
+The field [match_subject_alt_names][] is accepted only on the client side i.e. inside
+`UpstreamTlsContext` and is processed as described [above][server-authz]. If any of the other
+fields (listed below) are present, the update is NACKed because ignoring these fields compromises
+security:
 
 * `verify_certificate_spki`
 * `verify_certificate_hash`
@@ -391,19 +403,12 @@ is NACKed because ignoring these fields compromises security:
 * `crl`
 * `custom_validator_config`
 
-The following fields from [default_validation_context][] are ignored:
+The following fields from `CertificateValidationContext` are ignored:
 
 * `trusted_ca`
 * `watched_directory`
 * `allow_expired_certificate`
 * `trust_chain_verification`
-
-In [`combined_validation_context`][CVC], (other than [default_validation_context][]) only
-[`validation_context_certificate_provider_instance`][VCCPI1] is accepted as a deprecated alternative to
-[ca_certificate_provider_instance][CCPI]. When one of those fields ([ca_certificate_provider_instance][CCPI]
-or [`validation_context_certificate_provider_instance`][VCCPI1]) is present, the other fields
-[`validation_context_certificate_provider`][VCCP1] and [`validation_context_sds_secret_config`][VCSSC1]
-are ignored otherwise the presence of any of those fields will cause a NACK.
 
 [CPI]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L165
 [TCCPI]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L256
@@ -428,9 +433,9 @@ are ignored otherwise the presence of any of those fields will cause a NACK.
 
 xDS supports many different configurations for endpoints to obtain certificates but gRPC supports only
 [`CertificateProviderPluginInstance`][CPPI] (or its deprecated version [`CertificateProviderInstance`][CPI]).
-In this approach, the control plane tells the xDS client the [name of a certificate][CERT-NAME] (which
-is currently ignored) and the ["instance" name][INST-NAME] of a provider to use to obtain a certificate.
-This provider [instance name][INST-NAME] is translated into a provider implementation and a configuration
+In this approach, the control plane tells the xDS client the [name of a certificate][CERT-NAME] and the
+["instance" name][INST-NAME] of a provider to use to obtain a certificate. This provider
+[instance name][INST-NAME] is translated into a provider implementation and a configuration
 for that implementation using the client's own configuration instead of being sent by the control plane.
 This enables flexibility in heterogeneous deployments where different clients can use different
 implementations for the same provider instance name. For example, an on-prem client may get its certificate
--- a/A29-xds-tls-security.md
+++ b/A29-xds-tls-security.md
@@ -161,13 +161,13 @@ All the other fields are ignored. See below for [`CommonTlsContext`][CTC-type]
 processing.
 
 [A36]: A36-xds-for-servers.md
-[filter-chains]: https://github.com/envoyproxy/envoy/blob/45ec050f91407147ed53a999434b09ef77590177/api/envoy/config/listener/v3/listener.proto#120
+[filter-chains]: https://github.com/envoyproxy/envoy/blob/45ec050f91407147ed53a999434b09ef77590177/api/envoy/config/listener/v3/listener.proto#L120
 [filter-chain-match]: A36-xds-for-servers.md#filterchainmatch
 [transport-socket]: https://github.com/envoyproxy/envoy/blob/45ec050f91407147ed53a999434b09ef77590177/api/envoy/config/listener/v3/listener_components.proto#L246
 [DTC]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L57
 [transport-socket-comment]: https://github.com/envoyproxy/envoy/blob/45ec050f91407147ed53a999434b09ef77590177/api/envoy/config/listener/v3/listener_components.proto#L241
-[CTC1]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#84
-[RCC]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#88
+[CTC1]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L84
+[RCC]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L88
 [CTC-type]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L129
 
 #### CommonTlsContext Processing
@@ -194,10 +194,10 @@ TLS certificates and keys. Also the processing and parsing of `CommonTlsContext`
 CDS or LDS response does not take into account whether the Xds credentials are in effect for the
 respective channel or server.
 
-[CPI]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#154
-[TCCPI]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#234
-[VCCPI]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#259
-[VCCPI1]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#200
+[CPI]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L154
+[TCCPI]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L234
+[VCCPI]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L259
+[VCCPI1]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L200
 
 ### Use of SPIFFE IDs in Certificates
 
@@ -365,7 +365,7 @@ languages use similar hooks provided by the underlying TLS framework to implemen
 authorization.
 
 [match_subject_alt_names]: https://github.com/envoyproxy/envoy/blob/c94e646e0280e4c521f8e613f1ae2a02b274dbbf/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L373
-[CVC]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#251
+[CVC]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L251
 [StringMatcher]: https://github.com/envoyproxy/envoy/blob/6321e5d95f7e435625d762ea82316b7a9f7071a4/api/envoy/type/matcher/string.proto#L20
 
 #### Server Side Flow
--- a/A29-xds-tls-security.md
+++ b/A29-xds-tls-security.md
@@ -213,17 +213,19 @@ check succeeds.
 the check fails.
 
 * each SAN entry of type DNS, URI and IP address is considered for the below match
-logic. If an entry matches as per the logic, the check succeeds and it exits the check
-logic.
+logic. An IP address is converted to its canonical string representation. For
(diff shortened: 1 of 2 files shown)


yayapichfficial (approved):

yayapichfficial (approved):

A29-xds-tls-security.md
```diff
@@ -0,0 +1,411 @@
...
+
+The gRPC server xDS flow is described in
+[gRPC Server ](./A27-xds-global-load-balancing.md#grpc-client-architecture). The server
+configuration comes from the
+[`Listener`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/listener/v3/listener.proto#L39)
+resource which contains a list of
+[`FilterChain`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/listener/v3/listener_components.proto#L193)s.
+The `FilterChain` contains the `DownstreamTlsContext` that is applied to an incoming connection
+after selecting the best-matching `FilterChain` as described in
+[FilterChainMatch](A36-xds-for-servers.md#filterchainmatch). If the server is using `XdsServerCredentials`,
```
ejona86, line 372: The server doesn't know if it is using XdsServerCredentials.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,411 @@
...
+
+## Rationale
+
+gRPC has been steadily adding support for various xDS features including load balancing and
+advanced traffic management. xDS-based security was next in the roadmap and this proposal
+addresses that. gRPC makes a proxyless service mesh (PSM) possible. A secure proxyless
+service mesh requires the security features described in this proposal.
+
+The Certificate Provider Plugin Framework does not depend on an agent (SDS server). The
+SDS server/agent based solution compromised security (by sharing the private key with the
```
ejona86, line 390: There is still an agent here, in order to dump files on disk, so this seems questionable. The security property here is the same independent of whether it dumps files on disk or a UDS socket. I can understand if we say we want to _allow_ for systems that don't share the private key, although I think we believe the best security is for the workload to not have the private key at all and keep it solely within an agent.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,411 @@
...
+agent). The combination of Envoy sidecar and the SDS server was also somewhat tied to the
+Kubernetes architecture where Envoy runs in a sidecar container and the SDS server runs
+as a Node agent in the cluster. The Certificate Provider Plugin Framework has eliminated that
```
ejona86, line 393: I wouldn't say "eliminated that dependency" as the plugin framework didn't do that at all; swapping to file-based cert did that and it could be done outside of the plugin framework. The plugin framework allows different environments to have different dependencies and it allows the control plane to be agnostic to the specific environment the client is running (since generally the control plane has either hard-coded paths assuming a single environment or requires the client to specify values in metadata and have it echoed back).

A29-xds-tls-security.md
```diff
@@ -0,0 +1,411 @@
...
+
+At this stage the CDS policy uses the flow described above in
+[Use of Bootstrap File by the Certificate Provider Plugin Framework](#use-of-bootstrap-file-by-the-certificate-provider-plugin-framework)
+to fetch the requisite `CertificateProvider`s (one or two as the case may be). These `CertificateProvider`s,
+in one form or another, need to be made available to all the sub-channels of the cluster. How this is
+done is language dependent. For example, C-core uses the channel args to pass the individual
+`CertificateProvider`s whereas Java constructs a dynamic `SslContextProvider` that is directly
+usable by the sub-channel's TLS handshaker to build an `SslContext` for the impending handshake.
+
+##### Server Authorization aka Subject-Alt-Name Checks
```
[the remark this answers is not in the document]
markdroth: Agreed.  This should be moved up to the "CDS for Client Side Security" section.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,422 @@
...
+`tls_certificate_certificate_provider_instance`. gRPC looks up
+`"google_cloud_private_spiffe"` in the `"certificate_providers"` map loaded from the
+bootstrap file and use the JSON object value to extract the plugin name (value of
+`"plugin_name"` as a string) and configuration (value of `"config"` as a JSON object).
+It passes on these values to the `CertificateProviderStore` to retrieve
+(after instantiating if necessary) the associated `CertificateProvider` plugin.
+The caller of the framework registers a watcher with the plugin to receive
+certificate updates as described below.
+
+### Implementing Security in the xDS Flow
```
markdroth, line 326: I don't think we really need to rehash the overall flow here; that should already be covered earlier in the document.  Instead, I think this section should be converted into an "Implementation Details" section, which includes things like the internals of the CertificateProvider API (e.g., flow from CertificateProvider to distributor to TLS creds) or how data flows through LB policies in various languages.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,422 @@
...
+
+In any case the addition of xDS based security has brought gRPC xDS support closer to
+Envoy in terms of capabilities. This enables various use-cases such as migration to
+proxyless and interop or co-existence with Envoy.
+
+## Implementation
+
+Java had an early prototype implementation of xDS-based security that used SDS similar to Envoy. It was
+then modified and refined to match this spec. Other languages followed soon after to complete
+the implementation. The implementations (in Java, C++ and Go) are currently "hidden" behing the
```
markdroth, line 420: s/behing/behind/

markdroth (changes requested):

A29-xds-tls-security.md
```diff
@@ -0,0 +1,422 @@
...
+The `DownstreamTlsContext` thus obtained is used for the incoming connection.
+
+[`common_tls_context`][CTC1] in the `DownstreamTlsContext` contains the required
+security configuration. The [`require_client_certificate`][RCC] field determines
+whether the client certificate is required (mTLS mode if `true`, TLS if `false`).
+All the other fields are ignored. See below for [`CommonTlsContext`][CTC-type]
+processing.
+
+[A36]: A36-xds-for-servers.md
+[filter-chains]: https://github.com/envoyproxy/envoy/blob/45ec050f91407147ed53a999434b09ef77590177/api/envoy/config/listener/v3/listener.proto#120
```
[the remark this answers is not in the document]
sanjaypujare: Oops, I'll fix this first thing. Thanks for finding it.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,411 @@
...
+bootstrap file (described later) to obtain the CertificateProvider configuration. This
+configuration along with the CertificateProvider plugin framework enables gRPC to acquire
+the certificates necessary for the TLS handshake.
+
+The field [`tls_certificate_certificate_provider_instance`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L234)
+is used for the local certificate and the private key. And either
+[`validation_context_certificate_provider_instance`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L259)
+or [`validation_context_certificate_provider_instance`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L200)
+inside [`combined_validation_context`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L251)
+is used for the root certificate-chain for validating peer certificates. For mTLS both the local certificate
```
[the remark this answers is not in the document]
yashykt: We should probably also mention the behavior if the instance_name is not recognized, i.e., not present in the bootstrap file. 
Core's behavior - Don't check at parsing time, so no NACKing. It's checked later when it needs to be used, at which point an error is propagated. I believe that results in a TransientFailurePicker on the client side and closing the connection on the server side.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,422 @@
...
+
+#### LDS for Server Side Security
+
+Server-side xDS processing is described in [A36: xDS-Enabled Servers][A36].
+An LDS may have one or more [`filter_chains`][filter-chains] and one of those
+`filter_chain`s is matched against an incoming connection in order to get and apply
+the security (and other) configuration for that connection as described in the
+[`FilterChainMatch`][filter-chain-match]. The [`transport_socket`][transport-socket]
+of the matched (selected) `filter_chain` is used to extract the
+[`DownstreamTlsContext`][DTC] as described [here][transport-socket-comment].
```
yashykt, line 154: gRPC's behavior on decoding `transport_socket` is slightly different where if the name is not `envoy.transport_sockets.tls`, the overall behavior is to use the fallback creds instead of plaintext

A29-xds-tls-security.md
```diff
@@ -0,0 +1,422 @@
...
+### Programming API
+
+For each language, the channel and server credentials have been extended to allow a gRPC
+channel and server to use xDS provided security configurations. A Xds- Channel or Server credentials
+also needs to be provided with a "fallback credentials". The fallback credentials
+is used in the following cases:
+
+- when xDS is not in use such as when the `xds:` scheme is not used on the client side
+
+- xDS is in use but the control plane does not provide security configuration.
```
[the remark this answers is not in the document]
sanjaypujare: Thanks I will do that. But I am thinking that sometime in the not so distant future we will stop using Envoy as some kind of a baseline and compare our behavior with it. And my thinking (which could be wrong) is that most gRPC users may not know or care what Envoy behavior is.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,422 @@
...
+      .build()
+      .start();
+```
+
+#### Go
+
+Use of XdsChannelCredentials:
+
+```Go
+   // TODO: easwars to add
```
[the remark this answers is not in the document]
easwars:
```go

import (
	xdscreds "google.golang.org/grpc/credentials/xds"
)

if creds, err := xdscreds.NewClientCredentials(xdscreds.ClientOptions{FallbackCreds: insecure.NewCredentials()}); err != nil {
	log.Fatalf("Failed to create xDS client credentials: %v", err)
}
conn, err := grpc.Dial(*target, grpc.WithTransportCredentials(creds))
```

A29-xds-tls-security.md
```diff
@@ -0,0 +1,427 @@
...
+```
+
+Use of XdsServerCredentials:
+
+```Go
```
easwars, line 123:
```go
import (
	xdscreds "google.golang.org/grpc/credentials/xds"
	"google.golang.org/grpc/xds"
)

if creds, err := xdscreds.NewServerCredentials(xdscreds.ServerOptions{FallbackCreds: insecure.NewCredentials()}); err != nil {
	log.Fatalf("Failed to create xDS server credentials: %v", err)
}
server := xds.NewGRPCServer(grpc.Creds(creds))
```

sanjaypujare: Thanks. I have made the changes. PTAL

sanjaypujare: included this. PTAL

A29-xds-tls-security.md
sanjaypujare: Good point. Ideally we should fix the xDS proto file to include the gRPC variation (and make all of them less Envoy centric in general).

A29-xds-tls-security.md
```diff
@@ -0,0 +1,446 @@
...
+           "google.golang.org/grpc/xds"
+   )
+
+   if creds, err := xdscreds.NewServerCredentials(xdscreds.ServerOptions{FallbackCreds: insecure.NewCredentials()}); err != nil {
+           log.Fatalf("Failed to create xDS server credentials: %v", err)
+   }
+   server := xds.NewGRPCServer(grpc.Creds(creds))
+```
+
+### xDS Protocol
```
easwars, line 146: Sorry for being late to the party. But the sub-sections underneath and hard to read since they are structured as paragraphs. What do you think about using bullet points instead?

A29-xds-tls-security.md
```diff
@@ -0,0 +1,446 @@
...
+gRPC uses CDS to configure client side security and LDS for server side security as described below.
+The required fields were added to xDS v3 hence xDS v3 is a pre-requisite for this proposal.
+
+#### CDS for Client Side Security
+
+The CDS policy is the top level LB policy applied to all the connections under that policy.
+[A27:CDS][] describes the flow. The field [`transport_socket`][CL-TS] is used to extract the
+[`UpstreamTlsContext`][UTC] as described [here][CL-TS-comment].
+Note that we don't (currently) support [`transport_socket_matches`][CL-TS-matches].
+The `UpstreamTlsContext` thus obtained is passed down to all the child policies and connections.
```
easwars, line 157: The actual `UpstreamTlsContext` message is not passed down to the child policies. Instead whatever information is of use to us is extracted and that is what is passed down. Now sure if that distinction is useful to highlight here.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,446 @@
...
+### xDS Protocol
+
+gRPC uses CDS to configure client side security and LDS for server side security as described below.
+The required fields were added to xDS v3 hence xDS v3 is a pre-requisite for this proposal.
+
+#### CDS for Client Side Security
+
+The CDS policy is the top level LB policy applied to all the connections under that policy.
+[A27:CDS][] describes the flow. The field [`transport_socket`][CL-TS] is used to extract the
+[`UpstreamTlsContext`][UTC] as described [here][CL-TS-comment].
```
easwars, line 155: The comment talks about setting the `name` field to `tls` and the `typed_config` field to an `UpstreamTlsContext` proto. 

gRPC-Go NACKs resources which contain other values in these fields. Do the other two languages do the same? 
And if so, we should highlight that here.
And if not, what are y'all doing?

easwars: Same applies to the server side as well.

A29-xds-tls-security.md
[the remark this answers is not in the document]
easwars: gRPC-Go NACKs the resource if `require_client_certificate` is set and a certificate provider instance name for the root certificate is not provided. I think we discussed this long back and decided to NACK in this case.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,422 @@
...
+certificate and key configuration information needed on the client and server side respectively. The configuration
+tells gRPC how to obtain certificates and the keys for the TLS handshake. Although there are various ways to obtain
+certificates as per this proto (which are supported by Envoy), gRPC supports only one of them and that is defined by the
+[`CertificateProviderInstance`][CPI] proto which is based on the notion of CertificateProvider plugin framework
+(described later). The field `instance_name` defines a CertificateProvider "instance" that gRPC looks up in the
+bootstrap file (described later) to obtain the CertificateProvider configuration. This configuration along with
+the CertificateProvider plugin framework enables gRPC to acquire the certificates necessary for the TLS handshake.
+
+The field [`tls_certificate_certificate_provider_instance`][TCCPI] is used for the local certificate and the private
+key. And either [`validation_context_certificate_provider_instance`][VCCPI] or
```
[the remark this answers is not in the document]
sanjaypujare: As discussed, I agree with you. I'll change the text to only support `validation_context_certificate_provider_instance` inside of `combined_validation_context`.

A29-xds-tls-security.md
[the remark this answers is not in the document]
easwars: > We should probably also mention the behavior if the instance_name is not recognized, i.e., not present in the bootstrap file.

In gRPC-Go, we do something similar to Core. In fact, we cannot verify at parsing time as to whether the given certificate provider instance name is present in the bootstrap config or not. So, we simply accept it. But when the update it given to the CDS balancer, it is able to verify if the instance name is present in the bootstrap config or not. If the instance name is not found in the bootstrap config, the CDS balancer propagates such errors down to its child policy, which eventually leads to the channel being put in transient failure.

On the server side, since we are not required to put the server in `not-serving` mode for bad configuration (we are only required to do that when the configuration is deleted), we simply fail the connection at handshake time if the provider instance is not found in the bootstrap file.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,411 @@
...
+[`validation_context_certificate_provider_instance`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L259)
+or [`validation_context_certificate_provider_instance`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L200)
+inside [`combined_validation_context`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L251)
+is used for the root certificate-chain for validating peer certificates. For mTLS both the local certificate
+and root certificate are needed at both client and server. For TLS the server needs only the local certificate
+(the [`tls_certificate_certificate_provider_instance`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L234))
+and the client needs only the root certificate
+(the [`validation_context_certificate_provider_instance`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L200)).
+
+gRPC silently ignores the other fields in `CommonTlsContext` pertaining to other ways of fetching
```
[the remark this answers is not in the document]
easwars: Same behavior in Go as in Core.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,446 @@
...
+Another example is a plugin that periodically "mints" new certificates by creating
+CSRs at regular intervals and getting those signed by a Certification Authority (CA) that is
+configured into the plugin.
+
+A client registers itself as a **`Watcher`**  of the plugin to receive certificate updates.
+The plugin supports multiple **`Watchers`** to be registered and uses a **`DistributorWatcher`**
+internally to propagate a single update to multiple **`Watchers`**. A plugin caches the latest
+certificate (and key) it has fetched (or minted) and delivers them to every newly registered
+watcher.
+
```
easwars, line 279: We probably need to add a line saying something to the effect:

Go uses a fetch-style API instead of a watch-style API, where the component using the certificates and keys calls into the `CertificateProvider` whenever it needs them. The high-level concepts are identical except for the API style.

A29-xds-tls-security.md
sanjaypujare: > ...Why do you use fallback creds in this case?

There are arguments for both sides although we should have consistent behavior in all languages. One could argue that we don't recognize anything other than `envoy.transport_sockets.tls` so we NACK the update (like Go is doing). On the other hand we can define for gRPC the behavior that the fallback creds are used when we don't understand the `transport_socket`. My vote is for using the fallback creds but will agree to NACK if there are more convincing arguments.

A29-xds-tls-security.md
sanjaypujare: Yes, you have the same point made for the comment for line 178. Let's discuss it there and close this thread?

A29-xds-tls-security.md
sanjaypujare: Is that because you are trying to view it as a raw text file? How about using the markdown mode of the browser? Also what's the exact issue of "hard to read" due to paragraphs?

A29-xds-tls-security.md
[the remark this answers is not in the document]
markdroth: I was talking with @ejona86 about this just last week.  I think we misunderstood how the xDS validation config worked when we added those fields.  The [`validation_context_certificate_provider`](https://github.com/envoyproxy/envoy/blob/ea063c0dfebe57877de7968d51b650c1991a1c8d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L255) and [`validation_context_certificate_provider_instance`](https://github.com/envoyproxy/envoy/blob/ea063c0dfebe57877de7968d51b650c1991a1c8d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L259) fields don't actually make any sense, because cert providers don't return a complete [`CertificateValidationContext` proto](https://github.com/envoyproxy/envoy/blob/ea063c0dfebe57877de7968d51b650c1991a1c8d/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L236); they just provide the CA certs, which are basically just the [`trusted_ca` field](https://github.com/envoyproxy/envoy/blob/ea063c0dfebe57877de7968d51b650c1991a1c8d/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L282) within that proto.

What we probably should have done was simply add the new cert provider fields in `CommonTlsContext` but not inside of the [`validation_context_type` oneof](https://github.com/envoyproxy/envoy/blob/ea063c0dfebe57877de7968d51b650c1991a1c8d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L236).  The idea would be that the `CertificateValidationContext` is obtained from one of [`validation_context`](https://github.com/envoyproxy/envoy/blob/ea063c0dfebe57877de7968d51b650c1991a1c8d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L238), [`validation_context_sds_secret_config`](https://github.com/envoyproxy/envoy/blob/ea063c0dfebe57877de7968d51b650c1991a1c8d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L242), or [`combined_validation_context`](https://github.com/envoyproxy/envoy/blob/ea063c0dfebe57877de7968d51b650c1991a1c8d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L251), fully populated *except* for the `trusted_ca` field.  Then, as a second step, the cert provider is used to populate the `trusted_ca` field.

I've put together envoyproxy/envoy#17201 for this.  We'll need to update all 3 languages and TD.

CC @jaychenatr

A29-xds-tls-security.md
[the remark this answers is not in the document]
markdroth: Why can't we validate that the certificate provider instance is valid at config parsing time?  I think that would be a trivial change in C-core, and it seems like the right thing to do.  We basically do the same thing for supported HTTP filters (the only difference being that those are not configured in the bootstrap file, they're set at compile time).

A29-xds-tls-security.md
easwars: I was viewing the markdown, the issue is not that. I feel that each paragraph is just a stream of words, with references to a lot of protos and fields within them. If we split them into points, and show relevant proto snippets with relevant fields, I feel it would make it easier to read.

A29-xds-tls-security.md
easwars: Using fallback creds when the name field is anything other than `envoy.transport_sockets.tls` seems like a special-case behavior, while NACK'ing is consistent with what we do everywhere else (wherein, if we receive something that we don't like or support, we NACK the resource).

A29-xds-tls-security.md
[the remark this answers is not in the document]
easwars: > Why can't we validate that the certificate provider instance is valid at config parsing time?

That's how the code is modularized. The component which handles receiving the message and parsing knows only to do that. It does not have knowledge of the contents of the bootstrap file. I don't think it will super hard to give it that information though. But I think we discussed this in some form when we were implementing, and decided not to NACK, but instead to fail things somewhere up the stack where the required information is already available.

I don't remember the exact the pros and cons of the argument though. @yashykt @sanjaypujare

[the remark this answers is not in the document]
yashykt: Yes, checking the instance name is possible for Core, but I think we decided to go with <current_behavior> because of limitations from Go and to avoid inconsistencies between languages.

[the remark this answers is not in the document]
markdroth: If it's not super hard to do this in any language (and I don't see why it would be; it's just passing in a list of supported instance names to the config validation code), then I think we should do this at config-validation time.  We should NACK if we are told to use a provider that is not configured.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,446 @@
...
+a plaintext or insecure credentials is used as the fallback credentials.
+
+
+#### C++ 
+
+Use of XdsChannelCredentials:
+
+```C++
+    std::shared_ptr<grpc::ChannelCredentials> credentials =
+        grpc::experimental::XdsCredentials(grpc::InsecureChannelCredentials());
```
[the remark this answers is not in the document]
sanjaypujare: I am removing it

A29-xds-tls-security.md
sanjaypujare: Okay NACK it is then. I'll update accordingly.

A29-xds-tls-security.md
sanjaypujare: Yes, your comment applies to the file_watcher plugin but the "Certificate Provider Plugin Framework" itself does not assume the existence of an agent. For example, a plugin that mints CSRs and gets them signed by a CA. But I will modify the text to address your comment.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,411 @@
...
+[`CommonTlsContext`](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L129)
+processing.
+
+
+#### CommonTlsContext Processing
+
+`CommonTlsContext` is present in both the `UpstreamTlsContext` and `DownstreamTlsContext` protos and contains the
+certificate and key configuration information needed on the client and server side respectively. The configuration
+tells gRPC how to obtain certificates and the keys for the TLS handshake. Although there are various ways to obtain
+certificates as per this proto (which are supported by Envoy), gRPC supports only one of them and that is defined by the
```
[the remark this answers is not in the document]
sanjaypujare: Added.

A29-xds-tls-security.md
sanjaypujare: Okay, good point and I will convert this into an "Implementation Details" section. Regarding your suggestion

> ...or how data flows through LB policies in various languages.

I thought it would be better to leave out language specific implementation details from this gRFC similar to how other gRFCs have done. What do you think?

A29-xds-tls-security.md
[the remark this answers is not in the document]
sanjaypujare: Done

A29-xds-tls-security.md
sanjaypujare: fixed. PTAL

A29-xds-tls-security.md
sanjaypujare: fixed

A29-xds-tls-security.md
sanjaypujare: True. I understand what you are saying and our implementation does incorporate that. But let me know how this should be reworded here.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,411 @@
...
+[Use of Bootstrap File by the Certificate Provider Plugin Framework](#use-of-bootstrap-file-by-the-certificate-provider-plugin-framework)
+to fetch the requisite `CertificateProvider`s (one or two as the case may be). These `CertificateProvider`s,
+in one form or another, need to be made available to all the sub-channels of the cluster. How this is
+done is language dependent. For example, C-core uses the channel args to pass the individual
+`CertificateProvider`s whereas Java constructs a dynamic `SslContextProvider` that is directly
+usable by the sub-channel's TLS handshaker to build an `SslContext` for the impending handshake.
+
+##### Server Authorization aka Subject-Alt-Name Checks
+
+If [match_subject_alt_names](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L373)
```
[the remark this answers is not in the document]
sanjaypujare: Don't know how I missed this comment earlier.

> What if it isn't present?  Is no verification of the server performed? 

The server cert trust verification is performed of course. But there is no verification (or rather matching) of the ID presented in the certificate because we don't know what to match it to.

>  ... verify_certificate_hash, verify_certificate_spki....

verify_certificate_hash, verify_certificate_spki are purely cryptographic verification of the peer certificate based on out of band values supplied by the control plane and I don't think is a substitute for either hostname verification in Web PKI or the subject-alt-name check here.

However I should probably add here that we should NACK an update that has these fields populated. Thanks for pointing that out.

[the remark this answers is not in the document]
sanjaypujare: Done. PTAL

A29-xds-tls-security.md
```diff
@@ -0,0 +1,422 @@
...
+
+To facilitate service mesh secure communication, we support [SPIFFE][] based certificates and peer certificate
+verification following the [SPIFFE][] specification and the [match_subject_alt_names][] field in xDS.
+
+[SPIFFE]: https://github.com/spiffe/spiffe
+[match_subject_alt_names]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/common.proto#envoy-v3-api-field-extensions-transport-sockets-tls-v3-certificatevalidationcontext-match-subject-alt-names
+
+### Certificate Provider Plugin Framework
+
+The pluggable Certificate Provider framework in gRPC underlies the use of `CertificateProviderInstance`
```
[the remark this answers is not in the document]
sanjaypujare: Done. PTAL

sanjaypujare:
All the reviewers: thanks for your comments. I have addressed all your comments. Pls take a look.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,509 @@
...
+[CL-TS-comment]: https://github.com/envoyproxy/envoy/blob/9aca65c395f01020080166e4795455addde167fa/api/envoy/config/cluster/v3/cluster.proto#L932
+[CL-TS-matches]: https://github.com/envoyproxy/envoy/blob/9aca65c395f01020080166e4795455addde167fa/api/envoy/config/cluster/v3/cluster.proto#L680
+[CTC]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L37
+[SNI]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L40
+[ALLOW-RENEG]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L47
+[MAX-SESS-KEYS]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L53
+
+##### Server Authorization aka Subject-Alt-Name Checks
+
+"Server authorization" is performed by the client where the client authorizes
```
sergiitk, line 192: Why "Server authorization" is in quotes?

A29-xds-tls-security.md
sergiitk: +1, had the same feeling. It's not a blocker, but bullet points would've been easier to follow.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,509 @@
...
+`CertificateProviderInstance`. Through this object the control plane tells the xDS client the name of
+a certificate and the "instance" name of a provider to use to obtain a certificate. This provider
+instance name is translated into a provider implementation and a configuration for that implementation
+using the client's own configuration instead of being sent by the control plane. This enables flexibility
+in heterogeneous deployments where different clients can use different implementations for the same
+provider instance name. For example, an on-prem client may get its certificate using a local certificate
+provider whereas a client running in cloud may use a cloud-vendor provided implementation to obtain
+the certificate.
+
+gRPC provides a `CertificateProvider` plugin API that can support multiple implementations. The plugin API
```
sergiitk, line 378: editorial nit: tautology 
```suggestion
gRPC offers a `CertificateProvider` plugin API that can support multiple implementations. The plugin API
```

A29-xds-tls-security.md
```diff
@@ -0,0 +1,509 @@
...
+in this map is the certificate_provider_instance name and the value is a JSON object
+having exactly 2 fields:
+
+* `"plugin_name"` which is the name of the plugin implementation (a string value), and
+* `"config"` which is the configuration for the plugin. The value of `"config"` is a JSON object
+whose schema is defined by that plugin.
+
+The structure in the bootstrap file is as follows:
+
+```
```
sergiitk, line 396:
```suggestion
```jsonc
```

hint: use `jsonc` to render json with comments:

```jsonc
"certificate_providers": {
  "instance_name": {
    "plugin_name": "implementation_name",
    "config": {
      // ...config for implementation_name plugin...
    }
  }
}
```

A29-xds-tls-security.md
```diff
@@ -0,0 +1,509 @@
...
+- `private_key_file`: The path to the file containing the private key. The file should contain a
+PEM-formatted PKCS encoded private key.
+- `ca_certificate_file`: The path to the file containing the root certificate. The file should contain
+a PEM-formatted X.509 conforming certificate chain.
+- `refresh_interval`: Specifies how frequently the plugin should read the files. The value must be
+in the [JSON format described for a `Duration`][DURATION-JSON] protobuf message.
+
+For example, the bootstrap file might contain the following:
+
+```
```
sergiitk, line 427:
```suggestion
```jsonc
```

```jsonc
  "certificate_providers": {
    "google_cloud_private_spiffe": { // certificate_provider_instance name
      "plugin_name": "file_watcher", // name of the plugin instance
      "config": {                    // config to be supplied to the plugin instance
        "certificate_file": "/var/run/gke-spiffe/certs/certificates.pem",
        "private_key_file": "/var/run/gke-spiffe/certs/private_key.pem",
        "ca_certificate_file": "/var/run/gke-spiffe/certs/ca_certificates.pem",
        "refresh_interval": "60s"
      }
    }
  }
```

A29-xds-tls-security.md
```diff
@@ -0,0 +1,509 @@
...
+extensible certificate providers.
+
+In any case the addition of xDS based security has brought gRPC xDS support closer to
+Envoy in terms of capabilities. This enables various use-cases such as migration to
+proxyless and interop or co-existence with Envoy.
+
+## Implementation Status
+
+Java had an early prototype implementation of xDS-based security that used SDS similar to Envoy. It was
+then modified and refined to match this spec. Other languages followed soon after to complete
```
sergiitk, line 506: 1. Should we clarify exact versions when this happened? 
2. Or should we just drop this part and describe current state of things instead?

A29-xds-tls-security.md
```diff
@@ -0,0 +1,509 @@
...
+The framework has also made the certificate provider functionality extensible
+and pluggable which enables support for new certificate providers without requiring
+changes to the xDS protocol or the control plane. These advantages plus the extensions
+made to the xDS protocol are providing an impetus to the Envoy community so that Envoy
+will hopefully have a similar framework in the near future to support pluggable and
+extensible certificate providers.
+
+In any case the addition of xDS based security has brought gRPC xDS support closer to
+Envoy in terms of capabilities. This enables various use-cases such as migration to
+proxyless and interop or co-existence with Envoy.
```
sergiitk, line 501:
```suggestion
proxyless and interop or coexistence with Envoy.
```

sergiitk (approved):
LGTM.

nit: I'd recommend to put it though a spell checker. My Grammarly shows a bunch of complaints on missing commas, etc.

A29-xds-tls-security.md
sanjaypujare: Describing the current state includes how the implementation came about and the reason for including this context. I don't know the value of adding versions - also there are no "exact" versions since this happened incrementally

ejona86 (commented):
Still working through it. Sending what I have, since it is the end of the day.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,509 @@
...
+
+## Proposal
+
+### Programming API
+
+We need a mechanism by which a programmer can "opt in" to allow the use of xDS provided security
+configuration for a gRPC channel or server. This is achieved by supplying an `XdsChannelCredentials`
+(to a channel) or an `XdsServerCredentials` (to a server). These two new credentials types are an extension
+of the existing channel or server credentials types in each language. Each `Xds*Credentials` needs to
+be provided with a "fallback credentials". The fallback credentials is used in the following cases:
```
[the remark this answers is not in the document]
sanjaypujare: In this group (or may be Google in general) "credentials" is treated as a singular.

A29-xds-tls-security.md
```diff
@@ -130,45 +153,124 @@ The CDS policy is the top level LB policy applied to all the connections under t
...
+Java constructs a dynamic `SslContextProvider` that is directly usable by a sub-channel's
+TLS handshaker to build an `SslContext` for the pending TLS handshake.
+
+Only when a channel is using `XdsChannelCredentials`, gRPC uses the security configuration
+in `UpstreamTlsContext`. [`common_tls_context`][CTC] in the `UpstreamTlsContext` contains
+the required security configuration. See below for [`CommonTlsContext`][CTC-type] processing.
+
+A secure client at a minimum requires [`validation_context_certificate_provider_instance`][VCCPI1]
+inside [`combined_validation_context`][CVC] to be able to validate the server certificate
+in the TLS mode. If this field is not set, gRPC will NACK the CDS update. The client identity
```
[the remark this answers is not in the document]
sanjaypujare: That's the minimum requirement but not sufficient i.e. if `combined_validation_context` is present but `validation_context_certificate_provider_instance` inside it is not present then it is NACKed. But there are further validations done as described in the following sentences even if this condition is satisfied

A29-xds-tls-security.md
```diff
@@ -130,45 +153,124 @@ The CDS policy is the top level LB policy applied to all the connections under t
...
+certificate is configured via the [`tls_certificate_certificate_provider_instance`][TCCPI]
+field which is optional. The client certificate is sent to the server if the server requests
+or requires it in the TLS handshake. If the server requires the client certificate but is not
```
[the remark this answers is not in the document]
sanjaypujare: Good question and I was actually thinking about this (whether I should include stuff that's not part of this spec) when I wrote this. 

The stuff about TLS handshake failing when client doesn't send the cert when server requires it, is not part of this spec but just informative about the mismatch that happens when client doesn't send the cert (since `tls_certificate_certificate_provider_instance` is not configured) even when server requires it. Let me see if I can reword to remove stuff that's not a requirement of this spec.

A29-xds-tls-security.md
sanjaypujare: Because I thought this is not a well-defined or understood term in the industry and instead of providing a long or technical definition here I wanted to convey the intuitive meaning of "server authorization".

A29-xds-tls-security.md
```diff
@@ -130,45 +153,124 @@ The CDS policy is the top level LB policy applied to all the connections under t
...
-[`common_tls_context`][CTC] in the `UpstreamTlsContext` contains the required security configuration.
-All other fields are ignored. See below for [`CommonTlsContext`][CTC-type] processing.
+The security configuration extracted from the `UpstreamTlsContext` thus obtained
+is passed down to all the child policies and connections. How this is done is language dependent
+and is similar to how the implementations pass policy information down to child policies.
+For example, C-core uses channel args to pass security configuration down, whereas
+Java constructs a dynamic `SslContextProvider` that is directly usable by a sub-channel's
+TLS handshaker to build an `SslContext` for the pending TLS handshake.
+
+Only when a channel is using `XdsChannelCredentials`, gRPC uses the security configuration
```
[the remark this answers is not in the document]
sanjaypujare: fixed

A29-xds-tls-security.md
```diff
@@ -130,45 +153,124 @@ The CDS policy is the top level LB policy applied to all the connections under t
...
+
+##### Server Authorization aka Subject-Alt-Name Checks
+
+"Server authorization" is performed by the client where the client authorizes
+a server for the connection as described below. This check is in lieu of the
+canonical [hostname check in Web PKI][RFC6125-6].
+
+If [match_subject_alt_names][] in [default_validation_context][]
+of [combined_validation_context][CVC] of the received `UpstreamTlsContext` is populated
+then gRPC validates the SAN entries in the server certificate by matching the
```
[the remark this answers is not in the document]
sanjaypujare: Yes, you are right - reading it back I realize this needs to be expanded and explained better. Will rewrite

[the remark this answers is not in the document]
sanjaypujare: done

A29-xds-tls-security.md
```diff
@@ -130,45 +153,124 @@ The CDS policy is the top level LB policy applied to all the connections under t
...
+[MAX-SESS-KEYS]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L53
+
+##### Server Authorization aka Subject-Alt-Name Checks
+
+"Server authorization" is performed by the client where the client authorizes
+a server for the connection as described below. This check is in lieu of the
+canonical [hostname check in Web PKI][RFC6125-6].
+
+If [match_subject_alt_names][] in [default_validation_context][]
+of [combined_validation_context][CVC] of the received `UpstreamTlsContext` is populated
```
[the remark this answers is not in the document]
sanjaypujare: fixed. PTAL

sanjaypujare:
@markdroth @ejona86 @sergiitk all comments addressed. I will also mark it "ready for review" so it can be merged

A29-xds-tls-security.md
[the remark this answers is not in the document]
sanjaypujare: So I looked at what Envoy does (as we discussed offline) for IP addresses and it looks like it converts IP addresses to their canonical format using [`inet_ntop`](https://man7.org/linux/man-pages/man3/inet_ntop.3.html) which has zero ellision. So I am going to include that in the spec here - no other open issues for IP addresses. 

Refs:

https://github.com/envoyproxy/envoy/blob/main/source/extensions/transport_sockets/tls/cert_validator/default_validator.cc#L317
https://github.com/envoyproxy/envoy/blob/main/source/extensions/transport_sockets/tls/utility.cc#L192
https://github.com/envoyproxy/envoy/blob/main/source/common/network/address_impl.cc#L296
https://github.com/envoyproxy/envoy/blob/main/source/common/network/address_impl.cc#L234

A29-xds-tls-security.md
markdroth: It's kind of a gray area.  I think that for a gRFC in the `A` category (as opposed to those in the `L` category), the overall design and functional specification needs to be described in a language-agnostic way.  However, it's also fine to include some language-specific details to help implementors understand how things need to work, as long as those details are not the core of the document.

In this case, I do think it would be useful to include a high-level description of how data flows through the LB policies.  The comments you have earlier in the doc about how C-core uses channel args and Java uses an `SslContextProvider` are appropriate to mention as part of that.

A29-xds-tls-security.md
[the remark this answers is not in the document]
markdroth: I think the open issue of potential changes to the CertificateProvider fields in xDS is the biggest open issue in this document.  We really need to get this nailed down before we finalize this gRFC.

A29-xds-tls-security.md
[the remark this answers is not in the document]
markdroth: Sanjay, please clarify in the text that we will NACK if the specified certificate provider instance name is not one of the instances configured in the bootstrap file.

markdroth (changes requested):

A29-xds-tls-security.md
```diff
@@ -0,0 +1,543 @@
...
+then gRPC checks the SAN entries in the leaf server certificate against the
+`match_subject_alt_names` values as follows:
+
+* if the `match_subject_alt_names` list is empty then there is nothing to match and the
+check succeeds.
+
+* if there is no leaf server certificate, or there are no SAN entries in the certificate
+the check fails.
+
+* each SAN entry of type DNS, URI and IP address is considered for the below match
```
ejona86, line 215: Envoy also supports RFC 822 (email) names. Are we purposefully not supporting them?
https://github.com/envoyproxy/envoy/blob/0fae6970ddaf93f024908ba304bbd2b34e997a51/source/extensions/transport_sockets/tls/utility.cc#L174

A29-xds-tls-security.md
```diff
@@ -0,0 +1,543 @@
...
+IPv6 this includes [maximum zero compression][zero-compr], [no leading zeros][no-leading-0]
+and [lower case][lower-case] e.g. `2001:db8::1` but *not* `2001:DB8:0::01`.
+If an entry matches as per this logic, the check completes successfully.
+
+  * if a SAN entry is empty then the match fails for that entry.
+
+  * a SAN entry is matched against each value of
+[`match_subject_alt_names`][match_subject_alt_names] as follows.
+    * A `match_subject_alt_names` value is a [`StringMatcher`][StringMatcher] and
+    the match is performed as per the semantics described for each `match_pattern`
```
ejona86, line 226: It seems DNS and exact match is supposed to do subdomain matching (e.g., `*.example.com`) as is typically done for certs.
https://github.com/envoyproxy/envoy/blob/c94e646e0280e4c521f8e613f1ae2a02b274dbbf/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L358-L359

Envoy impl: https://github.com/envoyproxy/envoy/blob/6b575e3e353e2fcbf21f362e7c6edcc52f970846/source/extensions/transport_sockets/tls/cert_validator/default_validator.cc#L319

A29-xds-tls-security.md
```diff
@@ -0,0 +1,543 @@
...
+This defines a plugin instance for the instance name `instance_name` which consists of a
+plugin implementation called `implementation_name` with the corresponding configuration
+defined as the value for the key `config`. When the xDS server tells the client to obtain
+a certificate from `instance_name`, the client will use this plugin instance.
+
+#### `file_watcher` Certificate Provider
+
+As mentioned before, we currently only have one plugin implementation called `file_watcher`.
+The configuration for this plugin has the following fields:
+- `certificate_file`: The path to the file containing the identity certificate. The file should contain
```
ejona86, line 450: s/certificate/certificate and certificate chain/ ?

A29-xds-tls-security.md
```diff
@@ -0,0 +1,543 @@
...
+a PEM-formatted X.509 conforming certificate.
+- `private_key_file`: The path to the file containing the private key. The file should contain a
+PEM-formatted PKCS encoded private key.
+- `ca_certificate_file`: The path to the file containing the root certificate. The file should contain
```
ejona86, line 454: s/certificate/certificates/. "Certificate chain" doesn't make sense here, as that is supposed to chain a leaf cert to a root cert.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,543 @@
...
+  }
+}
+```
+
+This defines a plugin instance for the instance name `instance_name` which consists of a
+plugin implementation called `implementation_name` with the corresponding configuration
+defined as the value for the key `config`. When the xDS server tells the client to obtain
+a certificate from `instance_name`, the client will use this plugin instance.
+
+#### `file_watcher` Certificate Provider
```
ejona86, line 446: Somewhere should we make it explicit that the file_watcher plugin ignores the certificate name that comes from xds?

A29-xds-tls-security.md
sanjaypujare: Should that be treated as a bug in Envoy? When the `StringMatcher` supports SUFFIX match why should exact match perform subdomain matching?

A29-xds-tls-security.md
sanjaypujare: Envoy being a general purpose proxy and supporting other protocols (http1 etc) I thought they see more use cases for email names etc. I am not aware of use-cases or usage of email names in gRPC so I thought it will be okay to omit them.

A29-xds-tls-security.md
```diff
@@ -145,31 +144,42 @@ Use of XdsServerCredentials:
...
+in the TLS mode. As part of validating all CDS updates, if this field is not set in a CDS,
+gRPC will NACK the CDS update. The client identity certificate is configured via the
+[`tls_certificate_certificate_provider_instance`][TCCPI] field which is optional. The
+client certificate is sent to the server if the server requests or requires it in the
+TLS handshake. If the server requires the client certificate but is not
+configured on the client side then the TLS handshake will fail because of the client's inability
+to send the certificate.
 
-If any of the following fields are present, gRPC will NACK the CDS update:
+As part of validating all CDS updates, if any of the following fields are present, gRPC
```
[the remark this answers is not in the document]
sanjaypujare: Eric and I talked a bit about this: we can ignore a field (instead of NACKint it as unsupported) if that doesn't make the implementation any more insecure. Will re-evaluate the list with that goal.

A29-xds-tls-security.md
ejona86: As discussed separately, the behavior is clearly defined in the proto and in the code. It is on purpose and allows using conventional certificates with hostname verification which may need to process wildcards. The default should be we do the same. In our discussion, it sounded like we'd handle the wildcards in this specific case as documented in the proto.

A29-xds-tls-security.md
ejona86: As discussed separately, I wasn't saying we _should_ support it but think we should either support it or call out that we are purposefully not supporting (or requiring support for) this. It seems very easy to support, so I'd just assume to support it, but I accept it isn't much of a mainstream feature and if we didn't support it we may never hear from a user noticing it isn't available.

A29-xds-tls-security.md
sanjaypujare: Agreed, will modify text to match the behavior described in proto and what Envoy does

A29-xds-tls-security.md
sanjaypujare: I think it's almost identical to URI matching so I don't see any reason to exclude support since the implementation cost is almost zero. Will update the text.

A29-xds-tls-security.md
easwars: I thought we had discussed this during implementation, and I even copied the code for doing the wildcard matching from some Java implementation that you pointed to :)

sanjaypujare: Easwar: Correct, that's what Java had. Once we decided to support other StringMacher types (other than EXACT) I think I removed the wildcard logic from the Exact one. Anyway it looks like Go already has what Eric describes

sanjaypujare: Just looked at Envoy and the proto and realized it is not a generalized wildcard match that is needed. There is only case that is supported and that is (as an example): `*.example.com` in the cert's DNS SAN name matching `api.example.com` in the matcher present in `match_subject_alt_names`. So if the cert's DNS SAN name has an in initial `*.` then it matches a subdomain present in the exact matcher provided the rest of the domain strings match. There are no other wildcard semantics supported

easwars: Why do you say that it is not a generalized wildcard match?

```
  // When a certificate has wildcard DNS SAN entries, to match a specific client, it should be
  // configured with exact match type in the :ref:`string matcher <envoy_v3_api_msg_type.matcher.v3.StringMatcher>`.
  // For example if the certificate has "\*.example.com" as DNS SAN entry, to allow only "api.example.com",
  // it should be configured as shown below.
  //
  // .. code-block:: yaml
  //
  //  match_subject_alt_names:
  //    exact: "api.example.com"
```

This is just an example.

sanjaypujare: Easwar: the "*" is only allowed in the initial position followed by "." as in "*.example.com". "*" in any other place is not processed as wildcard. So "api.*.com" is not processed as wildcard . Also "*" in the StringMatcher string has no significance for wildcard processing so the only "*." that matters is in the cert SAN string.

This is what one sees in the Envoy code

easwars: Yes, DNS wildcard match supports a wildcard only in the first component.

ejona86: > This is what one sees in the Envoy code

I'm very concerned that generalized wildcard matching was considered. It is _not_ a string-based wildcard. This should be normal HTTP TLS matching rules. I don't think we should re-define it or reimplement it if it can be avoided. There are many edge cases like `*.example.com` does not match `sub.sub.example.com`. And some implementations appropriately disallow wildcards with TLDs like `*.com`.

https://datatracker.ietf.org/doc/html/rfc2818#page-5 talks about it some, but it is out-of-date without any updated text. https://codereview.chromium.org/762013002 and https://bugzilla.mozilla.org/show_bug.cgi?id=1107791 are the browsers removing the `f*.example.com` ability.

easwars: Sorry for the confusion. We did *not* consider generalized wildcard matching. The implementation in Go does DNS wildcard matching. See [here](https://github.com/grpc/grpc-go/blob/master/internal/credentials/xds/handshake_info.go#L256-L305). I just got confused by Sanjay's [comment](https://github.com/grpc/proposal/pull/243#discussion_r678746033).

sanjaypujare: Sorry for causing the confusion in the first place. And in fact not considering the proper subdomain matching spec that was evident in the proto and Envoy code.

A29-xds-tls-security.md
```diff
@@ -260,29 +328,51 @@ gRPC does not support the other certificate acquisition mechanisms specified by
 * [validation_context_sds_secret_config][VAL-SDS]
 * [validation_context_certificate_provider][VAL-PROVIDER]
 
-When the supported fields of type [`CertificateProviderInstance`][CPI] are present gRPC ignores the respective
-unsupported fields but if the unsupported fields are present by themselves gRPC will NACK the update.
+When fields of type [`CertificateProviderInstance`][CPI] are present gRPC ignores the respective
```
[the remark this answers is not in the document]
sanjaypujare: Ack. But no change needed.

A29-xds-tls-security.md
sanjaypujare: Mentioned on line 404 in the section "Certificate Provider Plugin Framework"

A29-xds-tls-security.md
sanjaypujare: done

A29-xds-tls-security.md
```diff
@@ -0,0 +1,543 @@
...
+[CUSTOM-HS]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L276
+[ALPN-PROTOCOLS]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L272
+[UTC]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L26
+[server-authz]: #server-authorization-aka-subject-alt-name-checks
+[VCCP1]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L193
+[VCSSC1]: https://github.com/envoyproxy/envoy/blob/7d4b2cae486b66b62ba0d3e1e348504699bea1bf/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L184
+
+### Use of SPIFFE IDs in Certificates
+
+To facilitate service mesh secure communication, we support [SPIFFE][] based certificates and peer certificate
```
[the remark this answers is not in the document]
sanjaypujare: Agreed. Removed the whole section.

A29-xds-tls-security.md
```diff
@@ -0,0 +1,543 @@
...
+* `watched_directory`
+* `verify_certificate_spki`
+* `verify_certificate_hash`
+* `require_signed_certificate_timestamp`
+* `crl`
+* `allow_expired_certificate`
+* `trust_chain_verification`
+* `custom_validator_config`
+
+Only [`validation_context_certificate_provider_instance`][VCCPI1] is accepted. When that value is present
```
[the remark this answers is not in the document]
sanjaypujare: Clarified that the field references are for `combined_validation_context`

A29-xds-tls-security.md
sanjaypujare: fixed the text

A29-xds-tls-security.md
sanjaypujare: Added support for email type

A29-xds-tls-security.md
```diff
@@ -145,31 +144,42 @@ Use of XdsServerCredentials:
...
 
 The CDS policy is the top level LB policy applied to all the connections under that policy.
 [A27:CDS][] describes the flow. The field [`transport_socket`][CL-TS] is used to extract the
 [`UpstreamTlsContext`][UTC] as described [here][CL-TS-comment].
 Note that we don't (currently) support [`transport_socket_matches`][CL-TS-matches].
-The `UpstreamTlsContext` thus obtained is passed down to all the child policies and connections.
-How this is done is language dependent and is similar to how the implementations pass policy
-information down to child policies.
+The security configuration extracted from the `UpstreamTlsContext` thus obtained
+is passed down to all the child policies and connections. How this is done is language dependent
```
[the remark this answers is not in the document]
sanjaypujare: Moved

A29-xds-tls-security.md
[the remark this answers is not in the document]
sanjaypujare: Thanks for pointing it out. Corrected the text

A29-xds-tls-security.md
```diff
@@ -0,0 +1,562 @@
...
+The following fields in an LDS update are ignored:
+
+* [disable_stateless_session_resumption][DIS-STATELESS-SESS-RES]
+* [session_timeout][SESSION-TIMEOUT]
+
+If any of the following fields are present, gRPC will NACK an LDS update as
+part of validating all LDS updates:
+
+* [require_sni][REQ-SNI]
+* [session_ticket_keys][SESSION-TICKET-KEYS]
```
yashykt, line 292: as discussed, it seems fine to `ignore session_ticket_keys` and `session_ticket_keys_sds_secret_config`

ejona86 (commented):
Sending what I have

A29-xds-tls-security.md
sanjaypujare: yes, changed

A29-xds-tls-security.md
sanjaypujare: thx. Updated on line 482

A29-xds-tls-security.md
```diff
@@ -0,0 +1,588 @@
...
+is passed down to all the child policies and connections similar to how gRPC passes
+load balancer policy information down to child policies.
+
+[`common_tls_context`][CTC] in the `UpstreamTlsContext` contains the required security
+configuration. See below for [`CommonTlsContext`][CTC-type] processing.
+
+A secure client at a minimum requires a CA root certificate to be able to validate the server
+certificate in the TLS mode and this is provided by [ca_certificate_provider_instance][CCPI]
+inside the [`default_validation_context`][default_validation_context] of
+[combined_validation_context][CVC]. For backward compatibility gRPC will use the deprecated
```
ejona86: With the field cleanup, I had expected we'd use `validation_context`, not `combined_validation_context`. This seems to suggest using `combined_validation_context` with only `default_validation_context` set which defeats the point of `combined_validation_context`. I don't see a reason to do that from a compatibility standpoint.

ejona86 (approved):

A29-xds-tls-security.md
sanjaypujare: Yes I considered that. An envoy-compatible config from a control plane currently looks like this:

```
        "commonTlsContext": {
...
          "combinedValidationContext": {
            "defaultValidationContext": {
              "matchSubjectAltNames": [{
                "exact": "spiffe://foo.example.com/ns/example-grpc-server/sa/example-grpc-server"
              }]
            },
            "validationContextSdsSecretConfig": {
              "name": "validation_context_sds",
              "sdsConfig": {
                "path": "/etc/envoy/validation_context_sds_secret.yaml"
              }
            },
            "validationContextCertificateProviderInstance": {
              "instanceName": "google_cloud_private_spiffe",
              "certificateName": "ROOTCA"
            }
          },
...
        }
```

With the restructured fields and as described here it will change to:

```
        "commonTlsContext": {
...
          "combinedValidationContext": {
            "defaultValidationContext": {
              "matchSubjectAltNames": [{
                "exact": "spiffe://foo.example.com/ns/example-grpc-server/sa/example-grpc-server"
              }],
              "caCertificateProviderInstance": {
                 "instanceName": "foobar",
                 "certificateName": "ROOTCA"
              }
            },
            "validationContextSdsSecretConfig": {
              "name": "validation_context_sds",
              "sdsConfig": {
                "path": "/etc/envoy/validation_context_sds_secret.yaml"
              }
            }
          },
...
        }
```

Whereas what you are proposing wouldn't be possible if you still want to maintain Envoy compatibility. Because of the `oneof` the control plane can send only either `combined_validation_context` (which works for Envoy and gRPC) or `validation_context` (which wouldn't work for Envoy).

I guess I can make a small change here to look for `validation_context` (if present) otherwise `combined_validation_context.default_validation_context` to get the validation context information.

markdroth: Yeah, I think the `oneof` requires us to support `combined_validation_context` for Envoy compatibility, unfortunately.  But I think we can basically ignore the SDS part of it, because we have an existing rule that we NACK if `ca_certificate_provider_instance` is not set.

See my other comment with suggested replacement wording for this paragraph.

markdroth (changes requested):
This is getting closer, but I think we still need to make some of the wording clearer, and I think we need to ditch all of the references to the deprecated fields.

A29-xds-tls-security.md
```diff
@@ -328,93 +350,101 @@ gRPC does not support the other certificate acquisition mechanisms specified by
 * [validation_context_sds_secret_config][VAL-SDS]
 * [validation_context_certificate_provider][VAL-PROVIDER]
 
-When fields of type [`CertificateProviderInstance`][CPI] are present gRPC ignores the respective
+When fields of type [`CertificateProviderPluginInstance`][CPPI] (or the deprecated
```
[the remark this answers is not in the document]
sanjaypujare: > These next two paragraphs are confusing. They also seem overly broad, because I think this applies only to the client side, not to the server side; on the server side, we can simply NACK if the `tls_certificate_provider_instance` field is unset, regardless of whether any of the unsupported fields are set.

On the server side we have the issue of validation_context and the control plane sending us only the unsupported field(s). Something like this:

```
            "combinedValidationContext": {
              "defaultValidationContext": {
              },
              "validationContextSdsSecretConfig": {
                "name": "validation_context_sds",
                "sdsConfig": {
                  "path": "/etc/envoy/validation_context_sds_secret.yaml"
                }
              }
            }
```

Instead of sending something like this:

```
            "combinedValidationContext": {
              "defaultValidationContext": {
                 "caCertificateProviderInstance": {
                    "instanceName": "google_cloud_private_spiffe",
                    "certificateName": "ROOTCA"
                 } 
              },
              "validationContextSdsSecretConfig": {
                "name": "validation_context_sds",
                "sdsConfig": {
                  "path": "/etc/envoy/validation_context_sds_secret.yaml"
                }
              }
            }
```

The first one should be NACKed but the second one we can process. Isn't that true?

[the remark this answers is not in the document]
markdroth: Your example is correct, but I think that's already covered by the text I suggested above to describe the server-side behavior:

"""
If a secure server is configured for mTLS, it will need configuration for how to validate the client's certificate.  This validation is configured via the `CertificateValidationContext` message, which is present in either the `validation_context` field or in the `default_validation_context` field inside of the `combined_validation_context` field.  If neither of those fields are set, the server will not request the client's certificate during the TLS handshake (i.e., it will use TLS instead of mTLS).

If a `CertificateValidationContext` is provided, then the server requires a CA root certificate to be able to validate the client certificate.  The CA root certificate is obtained using the certificate provider instance configured via the `ca_certificate_provider_instance` field.  If this field is not present, or if it specifies a certificate provider instance that is not configured in the bootstrap file, gRPC will NACK the LDS update.
"""

More generally, I think the document will be easier to understand if you document the requirements for the client and server separately rather than trying to combine them.

A29-xds-tls-security.md
```diff
@@ -256,69 +269,78 @@ As part of validating all LDS updates, if a `transport_socket` name is not
...
 
-* [require_sni][REQ-SNI]
+* [disable_stateless_session_resumption][DIS-STATELESS-SESS-RES]
 * [session_ticket_keys][SESSION-TICKET-KEYS]
 * [session_ticket_keys_sds_secret_config][SESSION-TICKET-KEYS-SDS]
-* [disable_stateless_session_resumption][DIS-STATELESS-SESS-RES]
 * [session_timeout][SESSION-TIMEOUT]
-* [ocsp_staple_policy][OCSP-STAPLE-POLICY]
+
+If any of the following fields are present, gRPC will NACK an LDS update as
```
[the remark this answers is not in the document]
sanjaypujare: done

sanjaypujare:
> This is getting closer, but I think we still need to make some of the wording clearer, and I think we need to ditch all of the references to the deprecated fields.
> 
> Please let me know if you have any questions. Thanks!

I have addressed all the comments. PTAL

A29-xds-tls-security.md
```diff
@@ -321,54 +347,38 @@ part of validating all LDS updates:
...
-[`CertificateProviderInstance`][CPI]) and Envoy (which supports neither [`CertificateProviderPluginInstance`][CPPI]
-nor [`CertificateProviderInstance`][CPI]) to achieve consistent security configuration between
-gRPC and Envoy.
-
-The following fields are unsupported and if present will cause a NACK from gRPC for any update
+* [tls_certificates][TLS-CERT]: if this is present but [`tls_certificate_provider_instance`][TCPI] is
+not present, then gRPC will NACK the update.
+* [tls_certificate_sds_secret_configs][CERT-SDS]: similar to above, if this is present but
+[`tls_certificate_provider_instance`][TCPI] is not present, then gRPC will NACK the update.
+* [validation_context_sds_secret_config][VAL-SDS]: if this is present but a
```
markdroth, line 376: This is true only on the server side.  On the client side, if there is no `CertificateValidationContext` from either `validation_context` or `combined_validation_context.default_validation_context`, then we NACK the CDS update without ever looking at the `validation_context_sds_secret_config` field.

I think we should remove this bullet and instead note this in the server-side section above.  I suggest rewriting the paragraph starting on line 291 to read as follows:

"""
If a secure server is configured for mTLS, it will need configuration for how to validate the client's certificate. This validation is configured via the `CertificateValidationContext` message, which comes from one of the options in the `validation_context_type` oneof.  If the `validation_context_sds_secret_config` field is set, gRPC will NACK the CDS update, since we do not support SDS.  If the [`validation_context`][validation_context] field is set, we get the `CertificateValidationContext` from there.  If the [`combined_validation_context`][CVC] field is set, we get the `CertificateValidationContext` from its [`default_validation_context`][default_validation_context] field.  If there is no `CertificateValidationContext`, then the server will not request the client's certificate during the TLS handshake (i.e., it will use TLS instead of mTLS).
"""

A29-xds-tls-security.md
```diff
@@ -379,58 +389,46 @@ on the client side (inside [`UpstreamTlsContext`][UTC]) is ignored if present.
...
 
-[`combined_validation_context`][CVC] is validated in any CDS/LDS update as follows:
-The field [match_subject_alt_names][] inside [default_validation_context][] is accepted only on
-the client side i.e. inside `UpstreamTlsContext` and is processed as described [above][server-authz].
-If any of the other fields (listed below) are present in [default_validation_context][], the update
-is NACKed because ignoring these fields compromises security:
+A `CertificateValidationContext` (either the [`validation_context`][validation_context] field or
+the [`default_validation_context`][default_validation_context] field inside of the
+[`combined_validation_context`][CVC]) is validated in any CDS/LDS update as follows:
+The field [match_subject_alt_names][] is accepted only on the client side i.e. inside
```
markdroth, line 395: The wording here implies that if `match_subject_alt_names` is specified on the server side, we NACK.  But I think we should actually just ignore it in that case, right?

markdroth (commented):
Just a few minor changes remaining here.

A29-xds-tls-security.md
sanjaypujare: You are right: we should clearly specify whether we NACK or just ignore. In terms of what we should do: I am okay to ignore it but it might be worthwhile to see what Envoy does. If I can figure it out, I'll update this thread

A29-xds-tls-security.md
```diff
@@ -321,54 +347,38 @@ part of validating all LDS updates:
...
-But if [`tls_certificate_provider_instance`][TCPI] (or in its absence the deprecated
-[`tls_certificate_certificate_provider_instance`][TCCPI]) is present, then the unsupported
-fields are ignored. This allows a control plane to formulate a single `CommonTlsContext` that works
-for both gRPC (which only supports [`CertificateProviderPluginInstance`][CPPI] or the deprecated
-[`CertificateProviderInstance`][CPI]) and Envoy (which supports neither [`CertificateProviderPluginInstance`][CPPI]
-nor [`CertificateProviderInstance`][CPI]) to achieve consistent security configuration between
-gRPC and Envoy.
-
-The following fields are unsupported and if present will cause a NACK from gRPC for any update
+* [tls_certificates][TLS-CERT]: if this is present but [`tls_certificate_provider_instance`][TCPI] is
```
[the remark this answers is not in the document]
sanjaypujare: done

A29-xds-tls-security.md
sanjaypujare: done

A29-xds-tls-security.md
sanjaypujare: I didn't spend too much time on analyzing the Envoy code but I didn't see Envoy NACKIng this on the server side. So I just modified the wording to say "we ignore it on the server side"

sanjaypujare:
> Just a few minor changes remaining here.
> 
> Please let me know if you have any questions. Thanks!

Addressed the latest comments

A29-xds-tls-security.md
ejona86: I think we should either support this field on server-side or NACK. Following the principal of "NACKing things that are unsupported and would reduce security," this would certainly reduce security.

I agree it may be rarely used, but it can be used very similarly to RBAC (especially L4 RBAC).

> but I didn't see Envoy NACKIng this on the server side

I'd assume they actually support it on server-side. I see no reason for them to special-case that code.

sanjaypujare:
> > I've just announced this on the Envoy slack instance, so I'll let this sit open for another couple of days, in case we get some feedback from there. But otherwise, I think this is ready to go!
> 
> Thanks Mark! I'll plan on merging this on Monday (unless there is feedback).

I'll be merging this today as planned assuming there are no further comments.

PR #245, opened 2021-06-17
L82: gRPC Core EventEngine API

drfloob:


L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+further API changes.
+
+The EventEngine rollout is planned to be on a per-platform basis. Once we
+establish that the default EventEngine implementation is sufficient for a
+specific platform, we will change the build to replace iomgr with EventEngine on
+that platform.
+
+## Limitations
+
+### Wrapped languages cannot provide a custom EventEngine at runtime
```
ctiller, line 408: Can we have a discussion on why?

@nathanielmanistaatgoogle made a solid case for allowing such integrations years ago, and I'd hoped that EventEngine could fill the gap in allowing gRPC IO to be implemented by non-C/C++ languages.

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+ServerBuilder builder;
+std::shared_ptr<EventEngine> engine = std::make_shared<MyEngine>(...);
+builder.SetEventEngine(engine);
+std::unique_ptr<Server> server(builder.BuildAndStart());
+server->Wait();
+```
+
+#### Providing a default EventEngine factory
+
+TODO: this has not been thoroughly discussed.
```
ctiller, line 369: I'd still prefer to see no default event engine in core - no concept of it at all in fact!

Instead, move this cost out to the wrapped languages and allow them to bind in an implementation that makes sense to them.

Doing so means that we can keep core *core* - nothing extraneous - and also likely helps push us to better point solutions in wrapped languages down the track.

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+  /// The \a fn callback's \a status argument is used to indicate whether it was
+  /// executed normally. For example, the status may be CANCELLED if
+  /// \a TryCancel was called, or if the EventEngine is being shut down.
+  virtual TaskHandle Run(Callback fn, RunOptions opts) = 0;
+  /// Synonymous with scheduling an alarm to run at time \a when.
+  ///
+  /// The callback \a fn will execute when either when time \a when arrives
+  /// (receiving status OK), or when the \a fn is cancelled (reveiving status
+  /// CANCELLED). The callback is guaranteed to be called exactly once.
+  virtual TaskHandle RunAt(absl::Time when, Callback fn, RunOptions opts) = 0;
```
ctiller, line 299: Would prefer to see this be some EventEngineTime type as we've discussed in the past, along with a Now() function which vends them to grpc core - or a rationale in this document for why we are not doing so.

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+    /// that the callback has access to the buffer when executed later.
+    /// Ownership of the buffer is not transferred. Valid slices *may* be placed
+    /// into the buffer even if the callback is invoked with a non-OK Status.
+    ///
+    /// For failed read operations, implementations should pass the appropriate
+    /// statuses to \a on_read. For example, callbacks might expect to receive
+    /// DEADLINE_EXCEEDED when the deadline is exceeded, and CANCELLED on
+    /// endpoint shutdown.
+    virtual void Read(Callback on_read, SliceBuffer* buffer,
+                      absl::Time deadline) = 0;
```
ctiller, line 122: Please explain the rationale for adding a timeout to read?

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+    /// Slices within the \a data buffer may be mutated at will by the Endpoint
+    /// until \a on_writable is called. The \a data SliceBuffer will remain
+    /// valid after calling \a Write, but its state is otherwise undefined.
+    ///
+    /// For failed write operations, implementations should pass the appropriate
+    /// statuses to \a on_writable. For example, callbacks might expect to
+    /// receive DEADLINE_EXCEEDED when the deadline is exceeded, and CANCELLED
+    /// on endpoint shutdown.
+    virtual void Write(Callback on_writable, SliceBuffer* data,
+                       absl::Time deadline) = 0;
```
ctiller, line 135: Again, why the deadline?

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+  ///
+  /// If this method returns a Listener, then \a on_shutdown will be invoked
+  /// exactly once, when the Listener is shut down. The status passed to it will
+  /// indicate if there was a problem during shutdown.
+  ///
+  /// The provided \a SliceAllocatorFactory is used to create \a SliceAllocators
+  /// for Endpoint construction.
+  virtual absl::StatusOr<std::unique_ptr<Listener>> CreateListener(
+      Listener::AcceptCallback on_accept, Callback on_shutdown,
+      const EndpointConfig& args,
```
ctiller, line 196: Where is the gRFC for this type?

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+  ///
+  /// Implementation Note: it is important that the \a slice_allocator be used
+  /// for all read/write buffer allocations in the EventEngine implementation.
+  /// This allows gRPC's \a ResourceQuota system to monitor and control memory
+  /// usage with graceful degradation mechanisms. Please see the \a
+  /// SliceAllocator API for more information.
+  virtual absl::Status Connect(OnConnectCallback on_connect,
+                               const ResolvedAddress& addr,
+                               const EndpointConfig& args,
+                               SliceAllocator slice_allocator,
```
ctiller, line 214: Where is the gRFC for this type?

ctiller (commented):
A few notes based on a quick first run through

L82-core-event-engine.md
drfloob: I had considered this an implementation detail, the current code can be found here: https://github.com/grpc/grpc/blob/master/include/grpc/event_engine/endpoint_config.h. I'll add it to this gRFC.

L82-core-event-engine.md
drfloob: Same justification. It can currently be found here: https://github.com/grpc/grpc/blob/master/include/grpc/event_engine/slice_allocator.h

L82-core-event-engine.md
drfloob: I hadn't seen any input from @nathanielmanistaatgoogle when researching this project. I'd be interested to hear it!

The answers to "why not" are: early hypotheses about poor performance given the way many wrapped languages are implemented, low ROI (relatively few reports of people wanting to implement custom I/O in wrapped languages), and our overall priorities for the work - we make compromises to minimize complexity and increase the likelihood that this project is shipped where other similar efforts have not.

I'm happy to have the discussion again, I think @markdroth and @nicolasnoble should be involved as they have the most context on previous work and current needs.

L82-core-event-engine.md
drfloob: I'm working on that proposal for that now, and will add it to this API in the next week or so.

L82-core-event-engine.md
nicolasnoble: The main limitations here were mainly dll-barrier crossing wrt to memory management, and, well, C++.

L82-core-event-engine.md
drfloob: @markdroth your input is requested here as well, please.

L82-core-event-engine.md
drfloob: I believe @markdroth and Sanjay were the key proponents of it. Care to weigh in, Mark?

L82-core-event-engine.md
ctiller: Let's add these types to this gRFC yeah... if they're named in this interface, then they're part of this interface and we should apply the same level of thought to them.

L82-core-event-engine.md
ctiller: Just to check my assumptions while I wait for that: my belief is that it's uncommon for operating systems to provide a primitive that will perform an asynchronous read with a deadline, and so we're forcing all (or at least most) implementations to emulate it. Further, most reads in gRPC don't need a deadline, so it looks we're forcing adding functionality we don't need (and the associated performance cost), as well as implementation complexity to every event engine that's unwarranted.

L82-core-event-engine.md
ctiller: The proposals were around sans-io and allowing more event loops to leverage gRPC in a performant manner.
I'm not sure any of the stated objections would actually bear out preventing an implementation, and so my preference here would be to simply omit this section - maybe with a note that we don't plan on exposing any of this to wrapped languages other than C++ at this point in time.

L82-core-event-engine.md
lidizheng: Maybe this section is talking about providing an API to let users set a default factory for this process?

L82-core-event-engine.md
lidizheng: Python has many heterogeneous asynchronous library implementation, they use different IO libraries. There were asks from time to time. Based on the event-engine-setting API above, what is blocking wrapper languages to provide custom EventEngine?

Nico mentioned memory management issue, then if several EventEngines are provided by Python, users still have the chance to select at the runtime right?

If above works, the wrapper language can create a shell EventEngine to invoke user-provided logic, then custom EventEngine is doable?

L82-core-event-engine.md
ctiller: Please add it to this grfc

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+namespace experimental {
+class EventEngine {
+ public:
+  /// A basic callable function. The first argument to all callbacks is an
+  /// absl::Status indicating the status of the operation associated with this
+  /// callback. Each EventEngine method that takes a callback parameter, defines
+  /// the expected sets and meanings of statuses for that use case.
+  using Callback = std::function<void(absl::Status)>;
+  /// A callback handle, used to cancel a callback.
+  struct TaskHandle {
```
ctiller, line 66: This likely needs to have some more space allocated to it.

Right now an implementation atop the internal EventManager2 would be forced to allocate per timer, which is probably unacceptable.

Suggest a handle size of at least 2 intptr_t's.

ctiller: Pinging @soheilhy to verify my understanding

soheilhy: +1 please allow another QWORD in addition to the key so that we can use it for a more efficient implementation internally.

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+    /// For failed write operations, implementations should pass the appropriate
+    /// statuses to \a on_writable. For example, callbacks might expect to
+    /// receive DEADLINE_EXCEEDED when the deadline is exceeded, and CANCELLED
+    /// on endpoint shutdown.
+    virtual void Write(Callback on_writable, SliceBuffer* data,
+                       absl::Time deadline) = 0;
+    /// These methods return an address in the format described in DNSResolver.
+    /// The returned values are owned by the Endpoint and are expected to remain
+    /// valid for the life of the Endpoint.
+    virtual const ResolvedAddress* GetPeerAddress() const = 0;
```
soheilhy, line 139: To match the style guide: Does this have to be a pointer, or can it be a ref?

soheilhy (commented):
Thanks for CCing me @ctiller . Will review the gRFC tonight.

L82-core-event-engine.md
drfloob: Added. Please take a look!

L82-core-event-engine.md
drfloob: Added. Please take a look!

L82-core-event-engine.md
```diff
@@ -0,0 +1,564 @@
...
+  virtual Setting Get(absl::string_view key) const = 0;
+};
+```
+
+The `SliceAllocator` provides memory buffers for EventEngine read/write
+operations, allowing EventEngines to seamlessly participate in the
+C++ `ResourceQuota` system.
+
+```cpp
+class SliceAllocator {
```
ctiller, line 83: Could we turn this into an interface perhaps?

I could imagine a future pivot where we implement SliceAllocator directly in resource user - but also a nearer future where we increase our testability by substituting in a fake SliceAllocator to verify behavior of EventEngine implementations.

L82-core-event-engine.md
```diff
@@ -0,0 +1,564 @@
...
+  grpc_resource_user* resource_user_;
+};
+```
+
+The `SliceAllocatorFactory` allows `EventEnigine::Listener`s to create and
+provide `SliceAllocator`s to `EventEngine::Endpoint`s as incoming connections
+are accepted.
+
+```cpp
+class SliceAllocatorFactory {
```
ctiller, line 118: Again, probably worth five minutes discussion around why this needs to be a concrete type and not an interface: likely we can test better with an interface.

L82-core-event-engine.md
```diff
@@ -0,0 +1,564 @@
...
+  SliceAllocatorFactory(SliceAllocatorFactory& other) = delete;
+  SliceAllocatorFactory& operator=(const SliceAllocatorFactory& other) = delete;
+  // Moveable
+  SliceAllocatorFactory(SliceAllocatorFactory&& other) noexcept;
+  SliceAllocatorFactory& operator=(SliceAllocatorFactory&& other) noexcept;
+  ~SliceAllocatorFactory();
+
+  /// On Endpoint creation, call \a CreateSliceAllocator with the name of the
+  /// endpoint peer (a URI string, most likely). Note: \a peer_name must outlive
+  /// the Endpoint.
```
ctiller, line 132: Probably just pass a string here, or have the factory copy from the string view - the requirement that peer_name outlive another thing is brittle, and an extra allocation along this path is tolerable.

L82-core-event-engine.md
```diff
@@ -0,0 +1,564 @@
...
+```
+
+### Task Execution
+
+```cpp
+  /// Retrieves an instance of a DNSResolver.
+  virtual absl::StatusOr<std::unique_ptr<DNSResolver>> GetDNSResolver() = 0;
+
+  /// Intended for future expansion of Task run functionality.
+  struct RunOptions {};
```
ctiller, line 373: What behavior should an EventEngine implementation have when it sees a RunOptions struct it does not understand?

How could we verify that RunOptions is understood by an implementation?

IOW: if I upgrade grpc and an option is added to this struct, how can calling code guarantee it's observed (this is the minimum bar to making a useful addition here), and how an EventEngine implementor know that the interface has expanded and some modifications are required?

ctiller: One proposal that could help here would be to add a `static constexpr int kVersion = 1;` and encourage implementations to take up a static_assert on the version, with documentation around why.

L82-core-event-engine.md
```diff
@@ -0,0 +1,564 @@
...
+```
+
+### DNS Resolution
+
+```cpp
+  /// The DNSResolver that provides asynchronous resolution.
+  class DNSResolver {
+   public:
+    /// A task handle for DNS Resolution requests.
+    struct LookupTaskHandle {
```
ctiller, line 311: Let's make this 2x intptr_t too, for the same reasons as TaskHandle?

L82-core-event-engine.md
```diff
@@ -0,0 +1,564 @@
...
+  /// \a on_connect callback *will not* have been executed. Otherwise, it is
+  /// expected that the \a on_connect callback will be asynchronously executed
+  /// exactly once by the EventEngine.
+  ///
+  /// Implementation Note: it is important that the \a slice_allocator be used
+  /// for all read/write buffer allocations in the EventEngine implementation.
+  /// This allows gRPC's \a ResourceQuota system to monitor and control memory
+  /// usage with graceful degradation mechanisms. Please see the \a
+  /// SliceAllocator API for more information.
+  virtual absl::Status Connect(OnConnectCallback on_connect,
```
ctiller, line 297: Why are there two ways to observe failure here (return value and callback)?

Wouldn't it be easier for implementors and users to always use the callback and make this infallible?

L82-core-event-engine.md
ctiller: I think we agreed on Friday to remove the deadline... Thinking about this some more, should this return a ReadTaskHandle so that we can offer a cancellation mechanism? (Maybe we can do that later)

Should it take a ReadOptions akin to RunOptions for future API expansion?

L82-core-event-engine.md
drfloob: I've weakened the statement here to state that wrapped language support for writing and providing custom EventEngines is not an immediate goal of this work. I'd like to leave this section in, since the scope of wrapped language support is an obvious question that I figure folks would want an answer to.

@lidizheng I think there are mostly just a broader set of concerns to address if we want to move EventEngine support up closer to the wrapped language layers. gRPC will allow implementations of the C++ EventEngine API to be plugged in at runtime; if the Python layer has support for multiple EventEngines written in C/C++ (I assume compiled in and distributed with our python packages), you could write a Python API to apply them as needed. To your second question on whether a proxying C++ EventEngine could work by ushering C-core requests to/from EventEngine-like things written in Python: probably yes.

L82-core-event-engine.md
drfloob: @lidizheng that's right.

@ctiller I see your argument. That's not really related to this section. I think you're engaging a much bigger discussion which might obviate this gRFC. Let's have that discussion between team leads and review an alternative proposal if you'd like to put one together. I can't really estimate how big a change it would be in the team's direction and investment over the past few quarters.

L82-core-event-engine.md
drfloob: Correct, I've removed the deadlines from the API. I didn't find any discussion on the deadlines at all, going back to early reviews, so I think they were at most hypothetically useful from the beginning.

Regarding `ReadOptions`, that sounds potentially useful. Can you think of any way it would be used that couldn't otherwise be solved by writing a custom EventEngine?

For `RunOptions`, we had discussed maybe needing a `priority` field to be added for efficiency of some implementations. But unless gRPC-internals would use the feature (currently, there is no run-priority concept internally), then `priority` would have been added speculatively at this point.

L82-core-event-engine.md
drfloob: I've changed the definition to `intptr_t keys[2]`.

L82-core-event-engine.md
drfloob: @markdroth, please correct me if wrong, but I believe the main justification against an interface was that we wanted to avoid dynamic allocation in the Listener and Endpoint APIs where possible. Do you have any other objections here?

L82-core-event-engine.md
drfloob: Agreed. This is an iomgr limitation at the moment, `grpc_resource_user`s do not make copies of or take ownership of their assigned names. I'll leave a TODO to fix that.

L82-core-event-engine.md
drfloob: > if I upgrade grpc and an option is added to this struct, how can calling code guarantee it's observed (this is the minimum bar to making a useful addition here)

I disagree. If gRPC ships with an implementation that gRPC internals all agree with, there's no question. If some custom EventEngine does not want to honor some set of RunOptions, I think that's entirely the implementer's prerogative. As I commented above with the `priority` example, if some impl does not distinguish between run priorities, I don't think that should be a show stopper, nor do I think gRPC internals would necessarily do something differently if that option were honored or not.

> How could we verify that RunOptions is understood by an implementation?

Can you give examples of when you think that would be meaningful?

> how [can] an EventEngine implementor know that the interface has expanded and some modifications are required?

Since this is a public API, we will file gRFCs for all EventEngine API changes. The mailing list will be notified as well.

L82-core-event-engine.md
drfloob: Good catch, thanks.

L82-core-event-engine.md
drfloob: The idea was that there can be synchronous and asynchronous failures. The value of observing a synchronous failure is questionable, though, and it's not consistent with the Read or Write APIs for example. We were waiting too see how a few implementations looked to determine whether we want to keep the return Status, but I tend to agree we'll probably want to return `void` here and signal failure through the callback alone. @nicolasnoble based on your work so far, does it seems reasonable to remove the return Status?

L82-core-event-engine.md
ctiller: Fair enough. I don't think it'd massively change direction to be fair, although it might add a step or two before we remove experimental, and I imagine first steps as experimental apis will necessarily have a default event engine - it's a sensible first step, and it'd be very hard to expunge iomgr without it.

My larger question is if it's a sensible end state, or if we end up with a more robust core of we omit the concept of a default event engine as part of that core.

I don't think we need to make this decision this week, but we should have a deeper discussion on this prior to grfc acceptance.

L82-core-event-engine.md
ctiller: ReadOptions I was wondering about the urgent flag we currently have... I have no deep insight into it's purpose, but at some point we felt like we needed it, and maybe we'll want such flexibility again (actually akin to priority for RunOptions, but I think it would be a mistake to tie the concepts!)

L82-core-event-engine.md
ctiller: Regarding API changes: if we make a semantic addition here, grfcs and mailing list posts are likely insufficient. On a larger project: it's likely whomever did the integration work that required the custom event engine has moved on, and other folks are only loosely aware of the state required. If we add required semantics to RunOptions it's likely that ignoring them will work kind of but introduce subtle bugs, leading to all sorts of difficult to diagnose from a distance behaviors.

Which means that our rule for RunOptions must be, if we include it without a versioning scheme, that anything added must be documented thoroughly that it may be ignored at will by any and all EventEngine implementations, and we must train our team thoroughly that any usage of RunOptions could be ignored by customer code and to aware of that when reviewing code and when diagnosing support tickets.

As to a change that might cause problems... Consider: adding a 'same_thread' constraint to Run to ensure that whatever function is passed to Run is additionally run within the same thread, presumably to catch some thread safety optimization.

L82-core-event-engine.md
drfloob: Discussed this externally on another PR, seems like there are no objections. I'll make this a pure virtual interface.

L82-core-event-engine.md
drfloob: Same as above, this will become an interface.

L82-core-event-engine.md
drfloob: Pinging @nicolasnoble for another opinion on this. I feel that if anything in `RunOptions` becomes essential to EventEngine functionality, that's a code smell. Backwards-incompatible changes should be first-class parts of the API, and merit gRPC version changes.

L82-core-event-engine.md
drfloob: Updated, PTAL

L82-core-event-engine.md
drfloob: Updated, PTAL

L82-core-event-engine.md
drfloob: That's a good question, Soheil! (and the same question for `GetLocalAddress` below). The style guide says:

> Prefer to return by value or, failing that, return by reference. Avoid returning a pointer unless it can be null.

Within an Endpoint's lifetime, when might an Endpoint's peer address (or local address) be null? I can't think of any valid scenarios, so I'm in favor of changing this to return a reference. @markdroth WDYT?

drfloob: I've made the change, thanks for pointing it out Soheil!

L82-core-event-engine.md
```diff
@@ -0,0 +1,538 @@
...
+                        SliceAllocator::AllocateCallback cb) = 0;
+};
+```
+
+The `SliceAllocatorFactory` allows `EventEnigine::Listener`s to create and
+provide `SliceAllocator`s to `EventEngine::Endpoint`s as incoming connections
+are accepted.
+
+```cpp
+class SliceAllocatorFactory {
```
piwicode, line 105: What is the benefit of having a `SiceAllocator` and a `SliceAllocatorFactory`?
Are there expected locality or lifecycle conciderations?
I would recommend to move the `peer_name` to Allocate and to remove the factory.

drfloob: As explained above, the SliceAllocator transparently enrolls EventEngines into the ResourceQuota system. The ResourceQuota system is a public feature, so we can't break that. Therefore something like the SliceAllocator needs to exist. The SliceAllocatorFactory mirrors the ResourceQuota itself (which is a parent to multiple independent resource users that can be reclaimed separately, incrementally if the quota has been exhausted), while the SliceAllocator hides the resource users within. I believe Craig has started some public documentation on the resource quota feature, it's fairly sparse at the moment.

The lifetimes are managed by smart pointers in the Connect and CreateListener APIs here. I believe our main implementations will let these objects remain valid until gRPC is shut down. Do you have lifetime questions that aren't answered by that?

The `peer_name` is an identifier for the Allocator (for the endpoint, essentially), not for a single allocation request, so it should remain in the constructor.

L82-core-event-engine.md
```diff
@@ -0,0 +1,538 @@
...
+    /// Called when the listener has accepted a new client connection.
+    using AcceptCallback = std::function<void(std::unique_ptr<Endpoint>)>;
+    virtual ~Listener() = default;
+    /// Bind an address/port to this Listener.
+    ///
+    /// It is expected that multiple addresses/ports can be bound to this
+    /// Listener before Listener::Start has been called. Returns either the
+    /// bound port or an appropriate error status.
+    virtual absl::StatusOr<int> Bind(const ResolvedAddress& addr) = 0;
+    virtual absl::Status Start() = 0;
```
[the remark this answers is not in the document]
drfloob: > What is going to happen if Bind is called after Start?

The expectation is in the Bind method description: "Returns either the bound port or an appropriate error status.". I imagine on most systems, in most EventEngine implementations, that would constitute an error.

> What is the semantic of destroying a Listener?

Can you clarify your question? Are you asking "what does it mean to destroy a listener"?

Regarding the last point, that's an interesting suggestion! I'm curious what @markdroth and @nicolasnoble think about the feasibility.

drfloob (commented):
Thank you for the feedback, Pierre!

L82-core-event-engine.md
drfloob: It is a bit clunky. We do, however, need some solution for I/O to occur "behind the scenes", in a way that applications can control if they wish. Either that, or applications would pay the price of instantiating whatever mechanism gRPC provides for the background I/O - likely some default EventEngine in the current design, with its threads, memory usage, etc..

For example, last I checked, the Global XDS client channel has no surface API that would allow custom EventEngines to be provided. I think there are a handful such situations where some application-provided global default would be helpful.

L82-core-event-engine.md
```diff
@@ -0,0 +1,538 @@
...
+```cpp
+  /// An Endpoint represents one end of a connection between a gRPC client and
+  /// server. Endpoints are created when connections are established, and
+  /// Endpoint operations are gRPC's primary means of communication.
+  ///
+  /// Endpoints must use the provided SliceAllocator for all data buffer memory
+  /// allocations. gRPC allows applications to set memory constraints per
+  /// Channel or Server, and the implementation depends on all dynamic memory
+  /// allocation being handled by the quota system.
+  class Endpoint {
```
ctiller, line 165: We need some mechanism for Endpoint to participate in ResourceQuota reclamation.
There are many possible designs here, we've discussed some internally, but that functionality is a requirement.

L82-core-event-engine.md
```diff
@@ -0,0 +1,538 @@
...
+gRPC public APIs contain methods that revolve around the concept of a system
+file descriptor (fd) for network operations. This is not a cross-platform
+concept, and many of the platforms that gRPC runs on do not natively support
+fds. For that reason, we did not want to include any mention of fds in the
+EventEngine API.
+
+Since we're committed to not making any breaking changes with this work, we have
+made a compromise to continue to support those fd-specific methods in the
+default EventEngine implementation alone. In other words, if your application
+provides a custom EventEngine, the fd-specific APIs will be no-ops.
```
markdroth, line 484: These functions should probably fail instead of just being silent no-ops.  I might even be open to them crashing the whole process (which may be the only option for APIs that return `void`).

L82-core-event-engine.md
```diff
@@ -0,0 +1,538 @@
...
+
+The ability to provide custom EventEngines at runtime will initially be
+available in the C++ and C-core APIs only. Adding support for custom
+EventEngines written in wrapped languages, or providing EventEngines at runtime
+from wrapped languages, is not a focus of this work.  Challenges may include the
+management of dll-barrier crossing issues, and potential performance losses.
+
+### Public file descriptor operations are not supported for custom EventEngines
+
+gRPC public APIs contain methods that revolve around the concept of a system
```
markdroth, line 475: We know that we have these methods in C++ and C-core.  Have we also audited wrapped languages to see if any of these APIs are used there?  If not, we should.

markdroth (commented):
Overall, this is looking very good!

L82-core-event-engine.md
markdroth: With regard to read cancellation, I think the current plan is that the only time a read would need to be cancelled would be if the endpoint is shutting down.  I don't think we have any cases where we need to cancel a single read without shutting down the endpoint.  But if you have such a case in mind, we can discuss.

ctiller: I don't... I think I had an API symmetry kick last week.

That said: I think if we resolve RunOptions as wanting to keep it, we should probably add ReadOptions, and similarly if we decide RunOptions is not the right thing we should not add ReadOptions.

L82-core-event-engine.md
ctiller: I'd put forward a vote to just remove RunOptions from the API then: extending it seems too difficult to actually rely on, and any changes ought to merit a version change.

L82-core-event-engine.md
```diff
@@ -0,0 +1,538 @@
...
+    /// Cancel an asynchronous lookup operation.
+    virtual void TryCancelLookup(LookupTaskHandle handle) = 0;
+  };
+```
+
+### Task Execution
+
+```cpp
+  /// Retrieves an instance of a DNSResolver.
+  virtual absl::StatusOr<std::unique_ptr<DNSResolver>> GetDNSResolver() = 0;
```
drfloob, line 341: @nicolasnoble: Chatting with @piwicode, it makes sense to do away with the StatusOr wrapper. If the EventEngine can't provide a resolver, that's probably a critical problem.

ctiller: I know of one or two deployments where it's debatable whether having a dns resolver gives value. These are usually very size constrained and want to do some bespoke connection management, and if there's an option to omit it, they'd probably take it.

These users also tend to be big fans of `channel_from_fd`.

L82-core-event-engine.md
```diff
@@ -0,0 +1,538 @@
...
+
+/// A set of parameters used to configure an endpoint, either when initiating a
+/// new connection on the client side or when listening for incoming connections
+/// on the server side. An EndpointConfig contains a set of zero or more
+/// Settings. Each setting has a unique name, which can be used to fetch that
+/// Setting via the Get() method. Each Setting has a value, which can be an
+/// integer, string, or void pointer. Each EE impl should define the set of
+/// Settings that it supports being passed into it, along with the corresponding
+/// type.
+class EndpointConfig {
```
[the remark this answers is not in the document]
drfloob: They all share the same `grpc_event_engine::experimental` namespace, which provides logical grouping. I don't know of any good reasons we'd want to weaken encapsulation by allowing privileged access from inner classes to EventEngine internals (should it become an abstract class instead of pure interface). There's also the added verbosity of the resulting code if the classes are nested (an extra `EventEngine::` peppered around).

Do you have arguments as to why we should nest all related classes within the EventEngine class?

L82-core-event-engine.md
```diff
@@ -0,0 +1,478 @@
...
+
+```cpp
+  /// Called when a new connection is established.
+  ///
+  /// If the connection attempt was not successful, implementations should pass
+  /// the appropriate statuses to this callback. For example, callbacks might
+  /// expect to receive DEADLINE_EXCEEDED statuses when appropriate, or
+  /// CANCELLED statuses on EventEngine shutdown.
+  using OnConnectCallback =
+      std::function<void(absl::StatusOr<std::unique_ptr<Endpoint>>)>;
```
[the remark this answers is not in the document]
veblush: I think we can accept this although there are some details to be determined in new abseil promise doc. New proposal doesn't have restriction on which classes are okay to use in the public API so you can use any Abseil public classes.

L82-core-event-engine.md
```diff
@@ -0,0 +1,539 @@
...
+    /// When data is available on the connection, that data is moved into the
+    /// \a buffer, and the \a on_read callback is called. The caller must ensure
+    /// that the callback has access to the buffer when executed later.
+    /// Ownership of the buffer is not transferred. Valid slices *may* be placed
+    /// into the buffer even if the callback is invoked with a non-OK Status.
+    ///
+    /// For failed read operations, implementations should pass the appropriate
+    /// statuses to \a on_read. For example, callbacks might expect to receive
+    /// CANCELLED on endpoint shutdown.
+    virtual void Read(Callback on_read, SliceBuffer* buffer) = 0;
```
drfloob, line 176: We need to make the contract clear here that there can be _at most one_ outstanding read on each `Endpoint` at any given time. Calls to `Read` while a read is already in flight (i.e. the `on_read` callback has not been executed yet for some previous call to `Read`) should result in an error. We could change the API to `virtual absl::Status Read(...)` and return a friendly error synchronously. We could immediately call their callback with some error status we define here. We could state that the EventEngine implementation should abort, given that something is using the API improperly and needs to be fixed. @ctiller @markdroth @nicolasnoble thoughts? I'm leaning towards the 2nd option, executing the callback immediately with an error status; I think I'd prefer the first option if we had other uses of returning an immediate error.

L82-core-event-engine.md
```diff
@@ -0,0 +1,539 @@
...
+    /// until \a on_writable is called. The \a data in SliceBuffer will remain
+    /// valid after calling \a Write, but its state is otherwise undefined.  All
+    /// bytes in \a data must have been written before calling \a on_writable
+    /// unless an error has occurred.
+
+    ///
+    /// For failed write operations, implementations should pass the appropriate
+    /// statuses to \a on_writable. For example, callbacks might expect to
+    /// receive CANCELLED on endpoint shutdown.
+    virtual void Write(Callback on_writable, SliceBuffer* data) = 0;
```
drfloob, line 190: The same issue as with `Read` above, there should be at most one single outstanding `Write` for each Endpoint at any given time.

L82-core-event-engine.md
drfloob: Discussing with @markdroth, @ctiller, and @nicolasnoble, it seems best to abort in this situation. Clients that are using the API incorrectly are unlikely to recover gracefully. I've added that language to the documentation.

L82-core-event-engine.md
drfloob: Same as for `Read`, EventEngines should abort. I've updated the RFC.

L82-core-event-engine.md
drfloob: After meeting about it, we concluded it makes sense to remove `RunOptions` and revisit the need for them before we promote this API from experimental status. In the meantime, we don't have any current use for the expansion, and we will otherwise be able to modify the API if we need to. I've removed `RunOptions` from the proposal.

L82-core-event-engine.md
drfloob: Closing the loop: we removed the `*Options` structs as YAGNI. The API is experimental and can be modified for a bit. We may introduce these structs later if we have cases for them.

L82-core-event-engine.md
drfloob: Fixed upstream.

tanger7719 (approved):

PR #246, opened 2021-06-18
A43: gRPC Authorization API

ashithasantosh:
This proposal discusses how gRPC authorization solution will be made available to OSS gRPC users. It covers details on gRPC SDK authorization policy (user facing policy language) and shows how to enable authorization controls in C++, Go and Java.

### DIFF
--- a/A43-sdk-authorization.md
+++ b/A43-sdk-authorization.md
@@ -1,5 +1,5 @@
 A43: SDK authorization
-
+----
 * Author(s): [Ashitha Santhosh](https://github.com/ashithasantosh)
 * Approver: [Doug Fawley](https://github.com/dfawley),
             [Eric Anderson](https://github.com/ejona86),
@@ -8,36 +8,34 @@ A43: SDK authorization
             [Yihua Zhang](https://github.com/yihuazhang)
 * Status: Draft {Draft, In Review, Ready for Implementation, Implemented}
 * Implemented in: <language, ...>
-* Last updated: 2021-07-19
-* Discussion at: (filled after thread exists)
+* Last updated: 2021-08-30
+* Discussion at: https://groups.google.com/g/grpc-io/c/m4Lns0tO-fE
 
 ## Abstract
 
-gRPC is currently implementing an open-source standard authorization solution
-based on Envoy RBAC. This proposal discusses how this framework would be made 
-available to non-xDS OSS gRPC users to perform per-RPC authorization checks.
+gRPC is currently implementing an open-source standard authorization solution.
+This proposal discusses how this framework would be made available to OSS gRPC
+users to perform per-RPC authorization checks.
 
 ## Background
 
-Authentication verifies the authority of the request. Authorization ensures the 
-authenticated authority has access to the requested resource.
+Authentication verifies the identity of the requestor. Authorization ensures that
+the requestor has sufficient access to make a particular request.
 
 gRPC offers built-in authentication mechanisms like TLS, ALTS and even allow
 users to plug in their own authentication systems. However, gRPC doesn't have any
 standard authorization solution available. If a user wanted to perform per RPC
 authorization check, they have to implement their own solution in the application.
-With standard gRPC authorization solution (under implementation), users can simply
-provide a policy configuration for their service, then the server verifies
-whether the client is authorized to access the requested service, and for which
-RPC methods. This differs from client-side authorization (also known as hostname
-verification) where client verifies server's identity.
+The standard gRPC authorization mechanism described in this proposal allows users
+to simply provide an authorization policy for their server. Then, for each RPC,
+the server verifies whether the client is authorized to issue the RPC.
 
 We have two different use-cases for authorization-
 1. gRPC Proxyless Service Mesh (or xDS authz) - Here Istio or Traffic Director
    control plane sends Envoy RBAC policy configs to xDS enabled gRPC servers for
    authorization
 2. SDK authorization (or non-xDS authz) - Authorization is made available to
-   non-xDS OSS gRPC users like network switches.
+   OSS gRPC users like network switches.
 
 This document discusses details of SDK authorization use-case. xDS authorization
 will be convered in a different proposal.
@@ -68,22 +66,26 @@ Following is the JSON schema of SDK Authorization Policy Version 1.0
   "title": "AuthorizationPolicy",
   "definitions": {
   	"rule": {
-  	  "description": "Specification of rules. An empty rule is always matched (i.e., both source and request are empty)",
+  	  "description": "Specification of rules. An empty rule is always matched"
+  	    "(i.e., both source and request are empty)",
   	  "type" : "object",
   	  "properties": {
   	  	"name": {
-  	  	  "description": "The name of an authorization rule. It is mainly for monitoring and error message generation.",
+  	  	  "description": "The name of an authorization rule. It is mainly for"
+  	  	    "monitoring and error message generation.",
 		  "type": "string"
 		},
 		"source": {
-		  "description": "Specifies attributes of a peer. Fields in the source are ANDed together, once we support"
-		    "multiple fields in the future. If not set, no checks will be performed against the source.",
+		  "description": "Specifies attributes of a peer. Fields in the source are"
+		    "ANDed together, once we support multiple fields in the future. If not"
+		    "set, no checks will be performed against the source.",
 			"type": "object",
 		  "properties": {
 			"principals": {
-			  "description": "A list of peer identities to match for authorization. The principals are one"
-				"of, i.e., it matches if one of the principals matches."
-				"The field supports Exact, Prefix, Suffix, and Presence matches."
+			  "description": "A list of peer identities to match for authorization."
+			   "The principals are one of, i.e., it matches if one of the principals"
+			   "matches."
+			   "The field supports Exact, Prefix, Suffix, and Presence matches."
 				  "- Exact match: \"abc\" will match on value \"abc\"."
 				  "- Prefix match: \"abc*\" will match on value \"abc\" and \"abcd\"."
 				  "- Suffix match: \"*abc\" will match on value \"abc\" and \"xabc\"."
@@ -96,15 +98,17 @@ Following is the JSON schema of SDK Authorization Policy Version 1.0
 		  }
 		},
 	    "request": {
-	      "description": "Specifies attributes of a request. Fields in the request are ANDed together. If not set, no"
-			"checks will be performed against the request.",
+	      "description": "Specifies attributes of a request. Fields in the request"
+	       "are ANDed together. If not set, no checks will be performed against"
+	       "the request.",
 		  "type": "object",
 		  "properties": {
 		    "paths": {
-			  "description": "A list of paths to match for authorization. This is the fully qualified name"
-			    "in the form of \"/package.service/method\". The paths are ORed together, i.e., it"
-				"matches if one of the paths matches."
-				"This field supports Exact, Prefix, Suffix, and Presence matches."
+			  "description": "A list of paths to match for authorization. This is the"
+			   "fully qualified name in the form of \"/package.service/method\". The"
+			   "paths are ORed together, i.e., it matches if one of the paths"
+			   "matches."
+			   "This field supports Exact, Prefix, Suffix, and Presence matches."
 				  "- Exact match: \"abc\" will match on value \"abc\"."
 				  "- Prefix match: \"abc*\" will match on value \"abc\" and \"abcd\"."
 				  "- Suffix match: \"*abc\" will match on value \"abc\" and \"xabc\"."
@@ -115,15 +119,15 @@ Following is the JSON schema of SDK Authorization Policy Version 1.0
 			  }
 			},
 			"headers": {
-			  "description": "A list of HTTP header key/value pairs to match against, for potentially"
-			    "advanced use cases. The headers are ANDed together, i.e., it matches only if *all*"
-			    "the headers match.",
+			  "description": "A list of HTTP header key/value pairs to match against,"
+			   "for potentially advanced use cases. The headers are ANDed together,"
+			   "i.e., it matches only if *all* the headers match.",
 			  "type": "array",
 			  "items": {
 				"type": "object",
 				"properties": {
 				  "key": {
-					"description": "The name of the HTTP header to match. The following "
+					"description": "The name of the HTTP header to match. The following"
 					  "headers are *not* supported: \"hop-by-hop\" headers (e.g.,"
 					  "those listed in \"Connection\" header), the \"Host\" header,"
 					  "HTTP/2 pseudo headers (\":\"-prefixed) and headers prefixed"
@@ -131,10 +135,10 @@ Following is the JSON schema of SDK Authorization Policy Version 1.0
 					"type": "string"
 				  },
 				  "values": {
-				    "description": "A list of header values to match. The header values"
-				      "are ORed together, i.e., it matches if one of the values matches."
-					  "Multi-valued headers are considered a single value with commas"
-					  "added between values."
+				    "description": "A list of header values to match. The header"
+				     "values are ORed together, i.e., it matches if one of the values"
+				     "matches. Multi-valued headers are considered a single value with"
+				     "commas added between values."
 					  "This field supports Exact, Prefix, Suffix, and Presence match."
 						"- Exact match: \"abc\" will match on value \"abc\"."
 						"- Prefix match: \"abc*\" will match on value \"abc\" and \"abcd\"."
@@ -155,25 +159,29 @@ Following is the JSON schema of SDK Authorization Policy Version 1.0
 	  "required": ["name"]
 	}
   },
-  "description": "AuthorizationPolicy defines which principals are permitted to access which resource. Resources are RPC methods"
-    "scoped by services.",
+  "description": "AuthorizationPolicy defines which principals are permitted to"
+   "access which resource. Resources are RPC methods scoped by services.",
   "type": "object",
   "properties": {
 	"name": {
-	  "description": "The name of an authorization policy. It is mainly for monitoring and error message generation.",
+	  "description": "The name of an authorization policy. It is mainly for"
+	   "monitoring and error message generation.",
 	  "type": "string",
 	},
 	"deny_rules": {
-      "description": "List of deny rules to match. If a request matches any of the deny rules, then it will be denied. If none of"
-		"the deny rules matches or there are no deny rules, the allow rules will be evaluated.",
+      "description": "List of deny rules to match. If a request matches any of the"
+       "deny rules, then it will be denied. If none of the deny rules matches or"
+       "there are no deny rules, the allow rules will be evaluated.",
 	  "type": "array",
       "items": {
 	    "$ref": "#/definitions/rule",
 	  },
 	},
 	"allow_rules": {
-      "description": "List of allow rules to match. The allow rules will only be evaluated after the deny rules. If a request"
-	    "matches any of the allow rules, then it will allowed. If none of the allow rules matches, it will be denied.",
+      "description": "List of allow rules to match. The allow rules will only be"
+       "evaluated after the deny rules. If a request matches any of the allow"
+       "rules, then it will allowed. If none of the allow rules matches, it will"
+       "be denied.",
 	  "type": "array",
 	  "items": {
 		"$ref": "#/definitions/rule",
@@ -322,35 +330,52 @@ different languages.
 
 #### C++
 
-1. Application enables SDK authorization in gRPC Server.
-
-- Policy is statically initialized
+1. C++ authorization provider API
 
 ```C++
-grpc::Status status;
-std::shared_ptr<AuthorizationPolicyProviderInterface> provider = 
-	StaticDataAuthorizationPolicyProvider::Create(authz_policy, &status);
-ServerBuilder builder;
-builder.SetAuthorizationPolicyProvider(provider);
-std::unique_ptr<grpc::Server> server = builder.BuildAndStart();
-```
+// Wrapper around C-core grpc_authorization_policy_provider. Internally, it
+// handles creating and updating authorization engine objects, using SDK
+// authorization policy.
+class AuthorizationPolicyProviderInterface {
+ public:
+  virtual ~AuthorizationPolicyProviderInterface() = default;
+  virtual grpc_authorization_policy_provider* c_provider() = 0;
+};
 
-- Policy is reloaded dynamically from file path
+// Implementation obtains authorization policy from static string. This provider
+// will always return the same authorization engines.
+class StaticDataAuthorizationPolicyProvider
+      : public AuthorizationPolicyProviderInterface {
+ public:
+  static std::unique_ptr<StaticDataAuthorizationPolicyProvider>
+     Create(const std::string& authz_policy, grpc::Status* status);
 
-```C++
-grpc::Status status;
-std::shared_ptr<AuthorizationPolicyProviderInterface> provider = 
-	FileWatcherAuthorizationPolicyProvider::Create(
-		authz_policy_path, /*refresh_interval_sec=*/3600, &status);
-ServerBuilder builder;
-builder.SetAuthorizationPolicyProvider(provider);
-std::unique_ptr<grpc::Server> server = builder.BuildAndStart();
+  ~StaticDataAuthorizationPolicyProvider override();
+ private:
+  grpc_authorization_policy_provider* provider_;
+};
+
+// Implementation obtains authorization policy by watching for changes in
+// filesystem. This provider will return up-to-date authorization engines.
+class FileWatcherAuthorizationPolicyProvider final
+    : public AuthorizationPolicyProviderInterface {
+ public:
+  static std::unique_ptr<FileWatcherAuthorizationPolicyProvider>
+    Create(const std::string& authz_policy_path,
+           unsigned int refresh_interval_sec,
+           grpc::Status* status);
+
+  ~FileWatcherAuthorizationPolicyProvider override();
+ private:
+  grpc_authorization_policy_provider* provider_;
+};
 ```
 
 2. C-core APIs
 
 ```C++
-/** Channel args for grpc_authorization_policy_provider. If present, enables gRPC authorization check. */
+/** Channel args for grpc_authorization_policy_provider. If present, enables gRPC
+ *  authorization check. */
 #define GRPC_ARG_AUTHORIZATION_POLICY_PROVIDER "grpc.authorization_policy_provider"
 
 /**
@@ -395,182 +420,34 @@ GRPCAPI void grpc_authorization_policy_provider_release(grpc_authorization_polic
 
 ```
 
-3. C-core Authorization Providers
-
-```C++
-struct grpc_authorization_policy_provider
-    : public DualRefCounted<grpc_authorization_policy_provider> {
-  virtual AuthorizationEngines engines() const= 0;
-};
-
-// Provider class will get SDK Authorization policy from string during
-// initialization. This policy will be translated to Envoy RBAC policies and
-// used to initialize allow and deny AuthorizationEngine objects. This provider
-// will return the same authorization engines everytime.
-class StaticDataAuthorizationPolicyProvider
-    : public grpc_authorization_policy_provider {
- public:
-  static absl::StatusOr<RefCountedPtr<grpc_authorization_policy_provider>>
-  	Create(absl::string_view authz_policy);
-
- private:
-  RefCountedPtr<AuthorizationEngine> allow_engine_;
-  RefCountedPtr<AuthorizationEngine> deny_engine_;
-};
-
-// Provider class that will get SDK Authorization Policy by watching for changes in
-// filesystem. When we read the file contents for the first time, we will translate
-// the SDK Authorization policy from file to Envoy RBAC policy and use the
-// the Envoy RBAC policy to initialize allow and deny grpc_authorization_engine
-// objects. This class contains a refresh thread, which will periodically (every
-// refresh_interval_sec) load file contents in specified path, and upon
-// modification invoke translator code, and replace existing engines with new
-// engines initialized with new policy.
-class FileWatcherAuthorizationPolicyProvider final
-        : public grpc_authorization_policy_provider {
-  public:
-    static absl::StatusOr<RefCountedPtr<grpc_authorization_policy_provider>>
-        Create(const std::string& authz_policy_path,
-            unsigned int refresh_interval_sec){
-    // We will perform an initial read, upon failure return error back to
-    // application. This is required to ensure that the file watcher provider
-    // always has a valid policy to make authorization decisions against. Once we
-    // have a valid policy, and if the reload fails (due to invalid policy or I/O
-    // error), we will continue using the latest valid policy to make authorization
-    // decisions.
-    }
-
-  private:
-    RefCountedPtr<AuthorizationEngine>> allow_engine_;
-    RefCountedPtr<AuthorizationEngine>> deny_engine_;
-};
-
-```
-
-4. C++ authorization providers
-
-```C++
-// C++ wrapper around C-core grpc_authorization_policy_provider type.
-class AuthorizationPolicyProviderInterface {
- public:
-  virtual ~AuthorizationPolicyProviderInterface() = default;
-  virtual grpc_authorization_policy_provider* c_provider() = 0;
-};
-
-class StaticDataAuthorizationPolicyProvider
-      : public AuthorizationPolicyProviderInterface {
- public:
-  static std::unique_ptr<StaticDataAuthorizationPolicyProvider>
-     Create(const std::string& authz_policy, grpc::Status* status);
-
-  ~StaticDataAuthorizationPolicyProvider override() {
-    grpc_autz_provider_release(provider_);
-  }
- private:
-  grpc_authorization_policy_provider* provider_;
-};
-
-// Similarly for FileWatcherAuthorizationPolicyProvider.
-class FileWatcherAuthorizationPolicyProvider final
-    : public AuthorizationPolicyProviderInterface {
- public:
-  static std::unique_ptr<FileWatcherAuthorizationPolicyProvider>
-    Create(const std::string& authz_policy_path,
-           unsigned int refresh_interval_sec,
-           grpc::Status* status);
-};
-
-
-```
-
-5. SDK policy translator
-
-```C++
-struct RbacPolicies {
-  Rbac deny_policy;
-  Rbac allow_policy;
-};
-
-// Translates SDK authorization policy to Envoy RBAC policies. Returns
-// error on failure.
-// authz_policy: Authorization Policy string in JSON format.
-absl::StatusOr<RbacPolicies> GenerateRbacPolicies(
-    absl::string_view authz_policy);
-```
-
-6. RBAC Engine
-
-```C++
-// Interface for gRPC authorization engines
-class AuthorizationEngine : public RefCounted<AuthorizationEngine> {
- public:
-  struct Decision {
-    enum class Type {
-      kAllow,
-      kDeny,
-    };
-    Type type;
-    std::string matching_policy_name;
-  };
-
-  virtual Decision Evaluate(const EvaluateArgs& args) const = 0;
-};
-
-// GrpcAuthorizationEngine can be either an Allow engine or Deny engine. This
-// engine makes authorization decisions to Allow or Deny incoming RPC request
-// based on permission and principal configs in the provided RBAC policy and the
-// engine type. This engine ignores condition field in RBAC config. It is the
-// caller's responsibility to provide RBAC policies that are compatible with
-// this engine.
-class GrpcAuthorizationEngine : public AuthorizationEngine {
- public:
-  // Builds GrpcAuthorizationEngine without any policies.
-  explicit GrpcAuthorizationEngine(Rbac::Action action) : action_(action) {}
-  // Builds GrpcAuthorizationEngine with allow/deny RBAC policy.
-  explicit GrpcAuthorizationEngine(Rbac policy);
-
-  // Evaluates incoming request against RBAC policy and makes a decision to
-  // whether allow/deny this request.
-  Decision Evaluate(const EvaluateArgs& args) const override;
-
-  //...
-};
-
-```
-
-#### Go
-
-1. Application enables SDK authorization in gRPC Server.
+3. Application enables SDK authorization in gRPC Server.
 
 - Policy is statically initialized
 
-```Go
-creds := credentials.NewServerTLSFromFile(certFile, keyFile)
-i, err := authz.NewStatic(authzPolicy)
-serverOpts := []grpc.ServerOption{
-  grpc.Creds(creds),
-  grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
-  grpc.ChainStreamInterceptor(i.StreamInterceptor),
-}
-s := grpc.NewServer(serverOpts...)
+```C++
+grpc::Status status;
+std::shared_ptr<AuthorizationPolicyProviderInterface> provider = 
+	StaticDataAuthorizationPolicyProvider::Create(authz_policy, &status);
+ServerBuilder builder;
+builder.SetAuthorizationPolicyProvider(provider);
+std::unique_ptr<grpc::Server> server = builder.BuildAndStart();
 ```
 
 - Policy is reloaded dynamically from file path
 
-```Go
-creds := credentials.NewServerTLSFromFile(certFile, keyFile)
-i, err := authz.NewFileWatcher(authzPolicyFile, 1 * time.Hour)
-serverOpts := []grpc.ServerOption{
-  grpc.Creds(creds),
-  grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
-  grpc.ChainStreamInterceptor(i.StreamInterceptor),
-}
-s := grpc.NewServer(serverOpts...)
-// In the end, free up the resources used for policy refresh.
-defer i.Close()
+```C++
+grpc::Status status;
+std::shared_ptr<AuthorizationPolicyProviderInterface> provider = 
+	FileWatcherAuthorizationPolicyProvider::Create(
+		authz_policy_path, /*refresh_interval_sec=*/3600, &status);
+ServerBuilder builder;
+builder.SetAuthorizationPolicyProvider(provider);
+std::unique_ptr<grpc::Server> server = builder.BuildAndStart();
 ```
 
-2. Authorization Server Interceptors
+#### Go
+
+1. Authorization Server Interceptors
 
 - Static initialization
 
@@ -585,22 +462,25 @@ type StaticInterceptors struct {
 // NewStatic returns a new StaticInterceptors from a static authorization policy
 // JSON string.
 func NewStatic(authzPolicy string) (*StaticInterceptors, error) {
-  // ...
+	// ...
 }
 
+// UnaryInterceptor intercepts incoming Unary RPC requests.
+// Only authorized requests are allowed to pass. Otherwise, an unauthorized
+// error is returned to the client.
 func (i *StaticInterceptors) UnaryInterceptor(
 	ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler)
 		(resp interface{}, err error) {
-  // First invoke denyEngine.Evaluate. If the evaluation result is DENY, close the
-  // call with UNAUTHENTICATED error, otherwise invoke allowEngine.Evaluate. If the
-  // evaluation result from is ALLOW, invoke handler to complete the RPC. If no matches
-  // are found, deny the request by default, close the call with UNAUTHENTICATED error.
+	// ...
 }
 
+// StreamInterceptor intercepts incoming Stream RPC requests.
+// Only authorized requests are allowed to pass. Otherwise, an unauthorized
+// error is returned to the client.
 func (i *StaticInterceptors) StreamInterceptor(
 	srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler)
 		(err error) {
-  // Similar to UnaryInterceptor implementation.
+	// ...
 }
 ```
 
@@ -619,13 +499,9 @@ type FileWatcherInterceptors struct {
 
 // NewFileWatcher returns a new FileWatcherInterceptors from a policy file
 // that contains JSON string of authorization policy and a refresh duration to
-// specify the amount of time between policy refreshes.
-func NewFileWatcher(file string, duration time.Duration) (*FileWatcherInterceptors, error) {
-  // Read the policy file and create new internal Interceptors. Will return error on
-  // failure.
-  // Creates a background go routine for policy refresh.
-  // ...
-}
+// specify the amount of time between policy refreshes. Creates a background go
+// routine for policy refresh.
+func NewFileWatcher(file string, duration time.Duration) (*FileWatcherInterceptors, error);
 
 // run is a long running goroutine which watches for changes in file path, and
 // updates the internalInterceptors upon modification.
@@ -659,96 +535,39 @@ func (i *FileWatcherInterceptors) Close() {
 }
 ```
 
-3. SDK policy translator
-
-```Go
-package authz
-
-// Translates a gRPC authorization policy in JSON string to two Envoy RBAC policies (deny and
-// allow policy). If the policy cannot be parsed or is invalid, an error will be returned.
-func translatePolicy(policyStr string) (*v3rbacpb.RBAC, *v3rbacpb.RBAC, error) {}
-```
-
-4. RBAC Engine
-
-```Go
-package authz
-
-type RPCData struct {
-  MD metadata.MD
-  PeerInfo *peer.Peer
-  FullMethod string
-  // ...
-}
-
-type engine struct {
-  policies map[string]*policyMatcher
-  action action
-}
-
-type ChainEngine struct {
-  chainedEngines []*engine
-}
-
-// NewChainEngine returns a chain of RBAC Engines, used to make authorization decisions on
-// incoming RPCs. Returns a non-nil error for invalid policies.
-func NewChainEngine(policy []*v3rbacpb.RBAC) (*ChainEngine, error) {}
-
-// IsAuthorized determines if an incoming RPC is authorized based on chain of RBAC engines.
-func (cre *ChainEngine) IsAuthorized(ctx context.Context) error {}
-```
-
-#### Java
-
-1. Application enables SDK authorization in gRPC Server.
+2. Application enables SDK authorization in gRPC Server.
 
 - Policy is statically initialized
 
-```Java
-AuthorizationServerInterceptor authzServerInterceptor;
-try {
-  authzServerInterceptor = AuthorizationServerInterceptor.create(authzPolicy);
-} catch (IllegalArgumentException iae) {
-  // ...
+```Go
+creds := credentials.NewServerTLSFromFile(certFile, keyFile)
+i, err := authz.NewStatic(authzPolicy)
+serverOpts := []grpc.ServerOption{
+  grpc.Creds(creds),
+  grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
+  grpc.ChainStreamInterceptor(i.StreamInterceptor),
 }
-Server server =
-    Grpc.newServerBuilderForPort(port, serverCreds)
-        .addService(service)
-        .intercept(authzServerInterceptor)
-        .build()
-        .start();
+s := grpc.NewServer(serverOpts...)
 ```
 
 - Policy is reloaded dynamically from file path
 
-```Java
-ScheduledExecutorService scheduledExecutor =
-    Executors.newSingleThreadScheduledExecutor(
-        new ThreadFactoryBuilder()
-            .setNameFormat(...)
-            .setDaemon(true)
-            .build());
-FileAuthorizationServerInterceptor authzServerInterceptor;
-try {
-  authzServerInterceptor = FileAuthorizationServerInterceptor.create(authzPolicyFile);
-} catch (IllegalArgumentException iae) {
-  // ...
-} catch (IOException ioe) {
-  // ...
+```Go
+creds := credentials.NewServerTLSFromFile(certFile, keyFile)
+i, err := authz.NewFileWatcher(authzPolicyFile, 1 * time.Hour)
+serverOpts := []grpc.ServerOption{
+  grpc.Creds(creds),
+  grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
+  grpc.ChainStreamInterceptor(i.StreamInterceptor),
 }
-Closeable closeable = authzServerInterceptor.scheduleRefreshes(
-    1, TimeUnit.HOURS, scheduledExecutor);
-Server server =
-    Grpc.newServerBuilderForPort(port, serverCreds)
-        .addService(service)
-        .intercept(authzServerInterceptor)
-        .build()
-        .start();
+s := grpc.NewServer(serverOpts...)
 // In the end, free up the resources used for policy refresh.
-closeable.close();
+defer i.Close()
 ```
 
-3. Authorization Server Interceptors
+#### Java
+
+1. Authorization Server Interceptors
 
 ```Java
 package io.grpc.authz;
@@ -767,11 +586,7 @@ public final class AuthorizationServerInterceptor implements ServerInterceptor {
   public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
       ServerCall<ReqT, RespT> call, Metadata headers,
       ServerCallHandler<ReqT, RespT> next) {
-    // First invoke denyRbacEngine.evaluate(call, headers). If the evaluation result
-    // is DENY, close the call with UNAUTHENTICATED error, otherwise invoke
-    // allowRbacEngine.evaluate(call, headers). If the evaluation result from the
-    // allow engine is ALLOW, invoke the next call handler. If no matches are found,
-    // deny the request by default, close the call with UNAUTHENTICATED error.
+    // ...
   }
 
   // Static method that creates an AuthorizationServerInterceptor.
@@ -835,55 +650,52 @@ public final class FileAuthorizationServerInterceptor implements ServerIntercept
 }
 ```
 
-4. SDK policy translator
+2. Application enables SDK authorization in gRPC Server.
+
+- Policy is statically initialized
 
 ```Java
-package io.grpc.authz;
-
-class RbacPolicies {
-  private final RBAC denyRbacPolicy;
-  private final RBAC allowRbacPolicy;
-
-  private RbacPolicies(Builder builder);
-  RBAC getDenyRbacPolicy();
-  RBAC getAllowRbacPolicy();
-
-  static final class Builder {
-    // ...
-  }
+AuthorizationServerInterceptor authzServerInterceptor;
+try {
+  authzServerInterceptor = AuthorizationServerInterceptor.create(authzPolicy);
+} catch (IllegalArgumentException iae) {
+  // ...
 }
-
-class AuthorizationPolicyTranslator {
-  // Translate a gRPC authorization policy in JSON string to two RBAC policies, a
-  // deny RBAC policy followed by an allow RBAC policy. If the policy cannot be
-  // parsed or is invalid, an IllegalArgumentException will be thrown.
-  public static RbacPolicies translate(String authorizationPolicy)
-      throws IllegalArgumentException {
-  }
-}
+Server server =
+    Grpc.newServerBuilderForPort(port, serverCreds)
+        .addService(service)
+        .intercept(authzServerInterceptor)
+        .build()
+        .start();
 ```
 
-5. RBAC Engine
+- Policy is reloaded dynamically from file path
 
 ```Java
-package io.grpc.xds.internal.rbac;
-
-public class AuthorizationDecision {
-  public enum Decision {
-    ALLOW,
-    DENY,
-  }
-  private Decision decision;
-  private String matchingPolicyName;
+ScheduledExecutorService scheduledExecutor =
+    Executors.newSingleThreadScheduledExecutor(
+        new ThreadFactoryBuilder()
+            .setNameFormat(...)
+            .setDaemon(true)
+            .build());
+FileAuthorizationServerInterceptor authzServerInterceptor;
+try {
+  authzServerInterceptor = FileAuthorizationServerInterceptor.create(authzPolicyFile);
+} catch (IllegalArgumentException iae) {
+  // ...
+} catch (IOException ioe) {
+  // ...
 }
-
-public class RbacEngine {
-  public RbacEngine(RBAC policy);
-  // Given a RPC call and headers, evaluate the policy.
-  public AuthorizationDecision evaluate(
-      ServerCall<ReqT, RespT> call, Metadata headers) {
-  }
-}
+Closeable closeable = authzServerInterceptor.scheduleRefreshes(
+    1, TimeUnit.HOURS, scheduledExecutor);
+Server server =
+    Grpc.newServerBuilderForPort(port, serverCreds)
+        .addService(service)
+        .intercept(authzServerInterceptor)
+        .build()
+        .start();
+// In the end, free up the resources used for policy refresh.
+closeable.close();
 ```
 
 ## Rationale
--- a/A43-sdk-authorization.md
+++ b/A43-sdk-authorization.md
@@ -682,7 +682,12 @@ policy to provider/interceptor. The provider then forwards the JSON policy to Po
 translator. The translator converts JSON policy to Envoy RBAC protos (Allow and/or
 Deny policy). Translator errors out on I/O errors or if the policy does not
 represent JSON schema it currently supports. Ultimately the generated RBAC policies
-are used to create Envoy RBAC engine(s).
+are used to create Envoy RBAC authorization engine(s). Then, for each incoming RPC
+request, we will invoke the Engines to get the authorization decision.
+
+SDK authorization can be enabled in xDS enabled servers, which means both
+authorization paths (SDK and xDS authorization) can co-exist. For a request to
+be allowed, both paths must allow the request.
 
 As mentioned previously, authorization APIs take policy in JSON format, instead of
 protobuf. This is done to avoid the dependency on protobuf. We have users in OSS
--- a/A43-sdk-authorization.md
+++ b/A43-sdk-authorization.md
@@ -199,12 +199,12 @@ Each rule has the following semantics -
      against the source fields and apply that to any action.
 2. Each source could contain a list of principals. The principals are ORed together,
    i.e. it matches if one of them matches.
-   Sequence of steps to evaluate each principal from config
-   - If TLS is not used, matching fails.
-   - A match is found if principal matches URI SANs or DNS SANs or Subject field from
-     x509 client certificate.
-   - We also get a match if there is no client certificate, and principal is an empty
-     string.
+   Sequence of steps to evaluate each principal from config -
+   1. If TLS is not used, matching fails.
+   2. A match is found if principal matches any one of URI SANs or DNS SANs or Subject
+      field from x509 client certificate.
+   3. We also get a match if there is no client certificate, and principal is an empty
+      string.
 3. Each request could contain a list of URL paths (i.e. fully qualified RPC methods)
    and list of http headers to match. Refer JSON schema above to understand matching
    semantics.
--- a/A43-sdk-authorization.md
+++ b/A43-sdk-authorization.md
@@ -428,7 +428,7 @@ type StaticInterceptors struct {
 // NewStatic returns a new StaticInterceptors from a static authorization policy
 // JSON string.
 func NewStatic(authzPolicy string) (*StaticInterceptors, error) {
-	// ...
+  // ...
 }
 
 // UnaryInterceptor intercepts incoming Unary RPC requests.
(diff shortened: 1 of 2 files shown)


A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
+A43: SDK authorization
+
+Author(s): [Ashitha Santhosh](https://github.com/ashithasantosh)
+Approver: 
+Status: Draft {Draft, In Review, Ready for Implementation, Implemented}
+Implemented in: <language, ...>
```
sanjaypujare, line 6: Some md formatting issue. Doesn't render correctly after this point.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+service-level and method-level access control for a service. Engines process 
+incoming RPC request attributes against policy configs and make a decision on 
+whether to allow or deny the request. The decision depends on the type of policy 
+(if the policy is an allowlist or denylist) and whether a matching policy was found.
+
+[RBAC filter]: https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/rbac/v3/rbac.proto
+
+The RBAC engine implementation is used by xDS API (i.e. Proxyless Service Mesh, 
+when Istio or Traffic Director sends Envoy RBAC policy to xDS enabled gRPC servers 
+for authorization) and SDK API (which makes authorization available to OSS gRPC users).
```
sanjaypujare, line 29: Specifically non-xDS OSS gRPC users (since xDS API users are also OSS gRPC users)

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+[RBAC filter]: https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/rbac/v3/rbac.proto
+
+The RBAC engine implementation is used by xDS API (i.e. Proxyless Service Mesh, 
+when Istio or Traffic Director sends Envoy RBAC policy to xDS enabled gRPC servers 
+for authorization) and SDK API (which makes authorization available to OSS gRPC users). 
+This document shares details on the latter.
+
+##Related Proposals:
+
+* [A41: xds RBAC Support][A41]
```
sanjaypujare, line 34: Link to the PR?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+
+SDK authorization policy guarantees stability and backward compatibility. So if
+a user writes a config, it will always be supported by gRPC unless the policy
+undergoes major versioning change like adding required fields. The policy is 
+extendable, we may add new fields in the future as new requirements are introduced, 
+we won’t be removing fields or updating semantics of existing fields.
+
+Following is the JSON schema of SDK Authorization Policy Version 1.0
+
+```json
```
sanjaypujare, line 51: Can you reduce indentation to 2 spaces so it's easier to view the text?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+RBAC proto and provided to Engines for evaluation. Note that SDK authorization policy 
+is a subset of Envoy RBAC, and it does not support all the fields that are present in 
+Envoy RBAC.
+
+### API
+
+gRPC will support both static initialization and dynamically reloading the policy
+from filesystem. In static initialization, the policy will be provided as a JSON
+string. In dynamic file reloading, the application will specify the file path that
+contains the authorization policy in JSON format.
```
sanjaypujare, line 245: What happens when the Dynamic file loader reads invalid file contents? Do you keep the previous policy in force? Or do you retry the read in that case? Also how do you prevent contention with the writer of the policy?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+try {
+  authzServerInterceptor = AuthorizationServerInterceptor.create(authzPolicy);
+} catch (IllegalArgumentException iae) {
+  // ...
+}
+Server server =
+    ServerBuilder.forPort(port)
+        .addService(service)
+        .useTransportSecurity(certChainFile, privateKeyFile)
+        .intercept(authzServerInterceptor);
```
sanjaypujare, line 317: 1) What happens if you provide multiple AuthorizationServerInterceptor interceptors this way? How will you catch such a situation?
2) Is it possible to have a simpler interface similar to C++ so the user doesn't know it's an interceptor e.g.
```
AuthorizationPolicyProvider authzPolicyProvider = AuthorizationPolicyProvider.create(authzPolicy);
Server server =    ServerBuilder.forPort(port)
        .addService(service)
        .useTransportSecurity(certChainFile, privateKeyFile)
        .authzPolicyProvider(authzPolicyProvider);
        .build().start();
```

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+
+As mentioned previously, internally gRPC Authorization engine consumes Envoy RBAC 
+proto. The user provided JSON SDK authorization policy will be translated to Envoy 
+RBAC proto and provided to Engines for evaluation. Note that SDK authorization policy 
+is a subset of Envoy RBAC, and it does not support all the fields that are present in 
+Envoy RBAC.
+
+### API
+
+gRPC will support both static initialization and dynamically reloading the policy
```
sanjaypujare, line 242: It should be mentioned that applying more than one policy (of any kind) is not allowed (and why - because the semantics are not define and/or implementation). And how this will be detected and prevented.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+
+```Java
+AuthorizationServerInterceptor authzServerInterceptor;
+try {
+  authzServerInterceptor = AuthorizationServerInterceptor.create(authzPolicy);
+} catch (IllegalArgumentException iae) {
+  // ...
+}
+Server server =
+    ServerBuilder.forPort(port)
```
sanjaypujare, line 314: How about (also) showing  `io.grpc.Grpc.newServerBuilderForPort(int port, ServerCredentials creds)` to illustrate since it uses the new `ServerCredentials`?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+        .addService(service)
+        .useTransportSecurity(certChainFile, privateKeyFile)
+        .intercept(authzServerInterceptor)
+        .build()
+        .start();
+// In the end, free up the resources used for policy refresh.
+closeable.close();
+```
```
sanjaypujare, line 350: Can this also be simplified to eliminate a lot of boilerplate?
```
FileAuthorizationPolicyProvider fileAuthzProvider 
     = FileAuthorizationPolicyProvider.create(
               authzPolicyFile, scheduledExecutor, 1, TimeUnit.HOURS);

Server server =    ServerBuilder.forPort(port)
        .addService(service)
        .useTransportSecurity(certChainFile, privateKeyFile)
        .authzPolicyProvider(fileAuthzProvider);
        .build()
        .start();
```

1) The same `authzPolicyProvider()` method is used and it does the right setup based on whether received provider is static or dynamicFile based
2) The closeable is also internally managed by the server where is will call close on the internally created Closeable

The only downside I see is this cannot be shared among multiple Servers.

Also consider using `io.grpc.Grpc.newServerBuilderForPort(int port, ServerCredentials creds)` in the example

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+			"properties": {
+				"name": {
+					"description": "The name of an authorization rule. It is mainly for monitoring and error message generation.",
+					"type": "string"
+				},
+				"source": {
+					"description": "Specifies attributes of a peer. Fields in the source are ANDed together, once we support" 
+					  "multiple fields in the future. If not set, no checks will be performed against the source.",
+					"type": "object",
+					"properties": {
```
sanjaypujare, line 67: Only `principals` supported? What about source IP address? Why can't we support that? If mTLS is not set then principals cannot be used. Should we mention that?

A43-sdk-authorization.md
sanjaypujare, line 242: mTLS is a pre-requisite for using authz because it only supports `principals` in the policy. It should be mentioned here and also checked in the implementation?

sanjaypujare, line 242: There are a couple of interop issues with xDS.

1) If server is an xDS server and receives authz policies from the xDS control plane, how will those policies and the SDK supplied authz policies interact? If this is an error, how will it be caught?

2) If the server also uses XdsServerCredentials and it tells the Server to use only TLS or plaintext, what happens to SDK supplied authz policies?

sanjaypujare (changes requested):
Made a few comments and would like to see them addressed

A43-sdk-authorization.md
yihuazhang: +1

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+Last updated: 2021-06-17
+Discussion at: (filled after thread exists)
+
+##Abstract
+
+gRPC is currently implementing an open-source standard authorization solution
+based on Envoy RBAC. This proposal discusses how this framework would be made 
+available to OSS gRPC users to perform per-RPC authorization checks.
+
+##Background
```
yihuazhang, line 16: Sorry for giving the confusing advice during our 1:1. But the current content in `Background` is closer to `Implementation details`. In `Background`, I suggest discussing 1) the difference between authentication vs authorization, 2) client-side authz (i.e., hostname verification) vs server-side authz, and 3) mention xDS authz.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+
+gRPC authorization support which is currently under implementation, internally 
+implements RBAC Engines based on Envoy RBAC policies. RBAC policy provides
```
yihuazhang, line 19: Yeah, `RBACK engine` itself is indeed implementation details, so not sure if we want to discuss it here.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+			"items": {
+				"$ref": "#/definitions/rule",
+			},
+		}
+	},
+	"required": ["name", "allow_rules"]
+}
+```
+
+gRPC SDK authorization policy has a list of allow rules and a list of deny rules.
```
yihuazhang, line 175: I suggest having another section before/after this section that explains each field in more depth:
- `Principle`:  when TLS is used, the peer principal comes from the first URI SAN (i.e., SPIFFE identity) or DNS SAN of the peer x.509 certificate. Also, when insecure channel is used, there should be no `peer`, and the request should be rejected. 
- What if both `request` and `source` are empty in `Rule`? It should be treated as MATCH, and depending on whether it is an allow or deny policy, we should allow or reject the request.

A43-sdk-authorization.md
yihuazhang: They are all great questions!

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+								"- Prefix match: \"abc*\" will match on value \"abc\" and \"abcd\"."
+								"- Suffix match: \"*abc\" will match on value \"abc\" and \"xabc\"."
+								"- Presence match: \"*\" will match when the value is not empty.",
+							"type": "array",
+							"items": {
+								"type": "string"
+							}
+						},
+						"headers": {
+							"description": "A list of HTTP header key/value pairs to match against, for potentially"
```
yihuazhang, line 103: Is it correct that for HTTP headers, we should match them against the hard-coded values defined in [here](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)?

A43-sdk-authorization.md
yihuazhang: I think if the server is xDS server or uses XdsServerCredentials, it should **only** try the authz policy delivered by xDS control plane (if provided) because there is no guarantee that xDS authz and SDK authz will not diverge in semantics in future. @ejona86, @markdroth, @dfawley, @sanjaypujare, WDYT?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+        .addService(service)
+        .useTransportSecurity(certChainFile, privateKeyFile)
+        .intercept(authzServerInterceptor)
+        .build()
+        .start();
+// In the end, free up the resources used for policy refresh.
+closeable.close();
+```
+
+##Rationale
```
yihuazhang, line 352: Maybe we can also add the rationale for using JSON-format authz policy, instead of proto-format.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,364 @@
...
+
+We decided to create a new policy language "SDK authorization policy" instead of 
+consuming Envoy RBAC directly due to following reasons
+- Envoy RBAC is a complex language, and we preferred using a simple human readable 
+  policy language
+- With our own language, we can provide a stable API, even when Envoy undergoes 
+  versioning updates.
+
+##Implementation
```
yihuazhang, line 361: I would suggest adding more details about the high-level implementation idea. That is, briefly explaining 
- policy provider
- policy translator
- RBAC engine (implemented as a filter in C-core and an interceptor in Java and Go).

yihuazhang:
Thanks @ashithasantosh for the gRFC. I also left some comments. PLMK if you have any questions.

A43-sdk-authorization.md
ashithasantosh: Sure. I reduced the indentation. It looks much cleaner now. Let me know if you would like me to reduce the line length further.

A43-sdk-authorization.md
ashithasantosh: There are two cases to consider here -
1. If initial file reload fails - we won't create the policy provider, and simply return error to application
2. if file reload fails on refresh - then we skip that update, and use previously valid policy to make authorization decisions.

About the contention, if say the writer is updating the file at the same time that the thread reads from it, then it probably depends on when the file completed update. It is possible that when policy is updated, we won't immediately start using it. It may get taken up in the next refresh interval.

A43-sdk-authorization.md
ashithasantosh: I included details on how we would match the principal. This also covers your suggestion, that if TLS is not used, then principals should fail to match.

In the SDK design document, we only have principals in source. I am guessing the fields were decided based on the current requirements (from Brixia), and source IP wasn't required. So to keep it simple, it wasn't included in MVP. 
Wdyt about adding source ip later, once we have MVP implemented?

A43-sdk-authorization.md
ashithasantosh: So we do check for authenticated connection when we are matching the principal values
https://github.com/grpc/grpc/blob/b9a643a81736d9efce5b0b6fce1cf6b1f66e1320/src/core/lib/security/authorization/matchers.cc#L177
I am not sure about mTLS being a pre-requisite for authz. Say we have allow_all kind policy (source and request empty), then it wouldn't check for whether connection is authenticated. It is only done when we use principal field.

A43-sdk-authorization.md
ashithasantosh: No worries :) Updated the background.

A43-sdk-authorization.md
ashithasantosh: I added more details. Please check if it looks good to you.

A43-sdk-authorization.md
ashithasantosh: Added text.

A43-sdk-authorization.md
ashithasantosh: Sure. I created a section for high level implementation.

A43-sdk-authorization.md
ashithasantosh: Maybe we could have SDK authorization in xDS enabled servers. In which case, ex. C++, both the filters (xDS and non-xDS) may be executed. Since they are separate filters, we have the flexibility to arrange SDK filter to be higher in stack relative to xDS.

A43-sdk-authorization.md
ashithasantosh: Yes! Since internally in the RBAC engine we would be fetching the header from gRPC Metadata. So if ":method" is unavailable it would be hard coded to POST, and server would then deny requests for all other method types. xDS RBAC support proposal goes through the caveats in detail.

Its worth noting that we don't support all the headers in SDK API.  The following headers are *not*
supported: "hop-by-hop" headers (e.g., those listed in "Connection" header), HTTP/2 pseudo headers (":"-prefixed), the "Host" header, and headers prefixed with "grpc-". If Authorization policy has any of these headers, then the policy will be marked invalid during translation.

A43-sdk-authorization.md
sanjaypujare: > we have allow_all kind policy

Sure, I didn't think of that but what's the value of such a policy?

A43-sdk-authorization.md
sanjaypujare: > Wdyt about adding source ip later, once we have MVP implemented?

Sounds good to me if that was the plan. But if the engine is going to be shared with the xDS RBAC implementation and that one does support source IP address, I wonder what's the additional effort to add here.

A43-sdk-authorization.md
sanjaypujare: > It is possible that when policy is updated, we won't immediately start using it. It may get taken up in the next refresh interval.

That's okay, I wasn't worried about missing the update. I was wondering if you are going to have extra code to guard against IO errors due to contention.

A43-sdk-authorization.md
ashithasantosh: I am not aware of any plan on adding source_ip to SDK policy. Maybe @markdroth @ejona86 @dfawley would know. 

About the shared engine implementation. There are several fields in RBAC that aren't supported by SDK authorization policy. It does not affect the SDK path, as we would be translating the SDK policy to Envoy RBAC (which would only contain the SDK fields), and the engine evaluation just works with that policy. So there is nothing extra that we have to do in the engine, to support SDK authorization.

Translator code
https://github.com/grpc/grpc/blob/master/src/core/lib/security/authorization/rbac_translator.h#L34

A43-sdk-authorization.md
ashithasantosh: We could have allow_rules as below
allow_rules : [{"name": "allow-all"}]
Since source and request are empty, this would translate to an ALLOW RBAC policy with ANY permission and ANY principal.
This means, the server would allow all incoming requests. Without checking for TLS within engine.

However, we do assert for grpc_auth_context in C-core filter for ex. Which would exist only if connection is authenticated.

sanjaypujare: > However, we do assert for grpc_auth_context in C-core filter for ex. Which would exist only if connection is authenticated.

So the only use of allow_all policy is to make sure the connection is mTLS aka authenticated. Otherwise not setting a policy is like allow_all. Is that right?

ashithasantosh: Yes, that's my understanding.
For Java and Go, it depends on how the server is configured.
I am not sure what would happen if connection is not mTLS and we use SDK authz. I don't think the implementation checks for it. Will have to look into it.

@markdroth @ejona86 Should we explicitly mention that mTLS is a pre-requisite for using SDK authorization, and ensure that all three stacks check for this in their implementation?

A43-sdk-authorization.md
ashithasantosh: For the MVP, we are not doing anything extra to guard against IO contention.

A43-sdk-authorization.md
ashithasantosh: That is a very good question. With the current approach, we could provide multiple AuthorizationServerInterceptors and each RPC will be validated against these interceptors.
We can avoid that and limit the SDK interceptor to just one instance using the wrapper approach u mentioned. And it looks much cleaner. 

Is there a reason why we chose to use interceptors directly in the Java API design? @ejona86 @sanjaypujare

A43-sdk-authorization.md
sanjaypujare: > ... So there is nothing extra that we have to do in the engine, to support SDK authorization.

Exactly. So my point is if the shared engine supports (source) IP address in the Envoy RBAC policy, the SDK authz feature can take advantage of that with minimal efforts for adding source IP support unless my assumptions are wrong...

A43-sdk-authorization.md
sanjaypujare: > ...In which case, ex. C++, both the filters (xDS and non-xDS) may be executed....

I am okay with this provided the semantics are well defined and everyone agrees with them.

Another idea I had was something similar to "fallback credentials" of XdsCredentials. We can define our SDK API in such a way so that it is possible for the user to say that use the xDS provided authz policy if present otherwise use this policy (the one I am supplying thru the API).

ashithasantosh: That's right! So we should accept only a single SDK authorization policy per gRPC Server.
Since we do deny by default in SDK path, if we have multiple SDK policies, say different policies for different services in a server, we would only evaluate against the first policy in chain if no match were found.

Only C++, currently ensures this behavior, where we accept only a single policy.
In Java and Go, since we expose the interceptors, users could create multiple sdk interceptors (therefore multiple policies) per server.  Should we be updating the design to check and prevent this, or just state explicitly that only one policy per server?

sanjaypujare: > ...Should we be updating the design to check and prevent this, or just state explicitly that only one policy per server?

Is it easy/possible to detect and prevent? What does it take?

A43-sdk-authorization.md
yihuazhang: LGTM. Thanks.

A43-sdk-authorization.md
yihuazhang: I think we should have mTLS as a pre-requisite for authz. It does not make sense to run the authz check against unauthenticated principles. Going back to your example, if we want to allow everything, why bother configuring authz in the first place?

yihuazhang: I am still not convinced we should allow the use of SDK authorization in xDS servers because it will give users an impression that SDK authz will be always compatible with xDS authz for which we do not have the guarantee. If we want to support the idea of "fallback authz" for xDS servers, the authz check should also follow the xDS authz semantics.

yihuazhang: Let me take back the above comment. There could be use cases where a service owner may want to allow some RPC methods for ALL principles while restricting others to only authenticated principles. In Envoy RBAC, it also supports both authenticated and unauthenticated principles: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#envoy-v3-api-msg-config-rbac-v3-principal-authenticated

yihuazhang (commented):
LGTM overall. 

- I would still like to see if we can get an agreement on whether we should allow SDK authz in xDS use cases. 
- There are some security issues supporting "path" field. We need to see how it gets resolved before making a decision to support it in SDK.

A43-sdk-authorization.md
ashithasantosh: Based on the email discussion, we won't be updating the current design.

A43-sdk-authorization.md
ashithasantosh: Updated the snippet to show new ServerCredentials. Would you like me to show both code snippets?

A43-sdk-authorization.md
ashithasantosh: Based on our email discussion, we won't be updating the current designs.  It is the user's responsibility to inject compatible policies when using multiple interceptors. I included a paragraph in proposal recommending single policy per grpc server.

A43-sdk-authorization.md
ashithasantosh: Based on email discussion, we have decided to keep the current design.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,424 @@
...
+   - If only source is empty, we evaluate against the request fields and apply that to
+     any user (See example below). Similarly, if only request is empty, we evaluate
+     against the source fields and apply that to any action.
+   - This also applies to sub fields, like if the principals in source is empty, the
+     behavior would be similar to an empty source.
+2. Each source could contain a list of principals. The principals are ORed together,
+   i.e. it matches if one of them matches.
+   Sequence of steps to evaluate each principal from config -
+   1. If TLS is not used, matching fails.
+   2. A match is found if principal matches any one of URI SANs or DNS SANs or Subject
```
ejona86, line 206: This is different behavior than RBAC. In RBAC, URI is used if there is such a SAN, then DNS, and Subject is only checked if there were no URI or DNS SANs.

A43-sdk-authorization.md
ejona86: I don't think there's any need to restrict to mTLS. If there is a principal in the rule, then that requires mTLS to match, but the rest of things still work without mTLS. Since other sources could be added in the future, there's nothing preventing non-mTLS usages becoming more useful in the future. And I agree having a way to let a few methods through via plaintext while requiring TLS for the rest is useful.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,424 @@
...
+
+### User facing Authorization Policy
+
+gRPC SDK authorization policy is the user facing policy language that allows
+service owners and/or security admins to enable per-RPC authorization checks.
+
+SDK authorization policy guarantees stability and backward compatibility. So if
+a user writes a config, it will always be supported by gRPC unless the policy
+undergoes major versioning change like adding required fields. The policy is 
+extendable, we may add new fields in the future as new requirements are introduced,
```
ejona86, line 55: Hmm... if a new field is added and a rule uses it on an older version of gRPC, then the restriction will be ignored. Seems that could cause trouble for allow rules. Do we want to do anything to address that?

A43-sdk-authorization.md
ejona86: I don't see any problem here. xDS will run its policies, and SDK will run its policies. This seems no different than the "multiple SDK interceptors" case which we recently agreed was fine. We aren't encouraging it, but if someone does it things are okay.

Note Java/Go have no technical means to prevent mixing this authz with xDS.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,424 @@
...
+Following code snippets show how to enable authorization in gRPC servers in 
+different languages.
+
+#### C++
+
+1. Static Initialization
+
+```C++
+std::shared_ptr<AuthorizationPolicyProviderInterface> provider = 
+	StaticDataAuthorizationPolicyProvider::Create(authz_policy);
```
ejona86, line 290: What happens if the policy is invalid?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,424 @@
...
+ServerBuilder builder;
+builder.SetAuthorizationPolicyProvider(provider);
+std::unique_ptr<grpc::Server> server = builder.BuildAndStart();
+```
+
+2. Dynamic file reloading
+
+```C++
+std::shared_ptr<AuthorizationPolicyProviderInterface> provider = 
+	FileWatcherAuthorizationPolicyProvider::Create(authz_policy_path, /*refresh_interval_sec=*/3600);
```
ejona86, line 300: Does this do I/O like in Go/Java to guarantee there is a policy? What if loading from file fails?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,424 @@
...
+  grpc.Creds(creds),
+  grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
+  grpc.ChainStreamInterceptor(i.StreamInterceptor),
+}
+s := grpc.NewServer(serverOpts...)
+// In the end, free up the resources used for policy refresh.
+defer i.Close()
+```
+
+#### Java
```
ejona86, line 336: Examples are nice, but could we include the API definitions? We don't need to show their implementation, but the API surface that will be added would be useful.

A43-sdk-authorization.md
ashithasantosh: Yes! The design decisions are similar to Java/Go.
If the file reload fails the first time, we return the error status back to application and no provider will be created.
If the file reload fails, after the provider is initialized with valid policy, we just skip that update and use the last valid policy to make authorization decisions.

A43-sdk-authorization.md
ashithasantosh: Oh sorry! I forgot to add the updates to API when I copied this from design document. Updated the snippet to match implementation.
If the policy is invalid, it returns null provider, and we can view the error status in status parameter.

A43-sdk-authorization.md
ashithasantosh: Sorry for the confusing wording. I was trying to say the same thing. Updated text. PTAL.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,424 @@
...
+1. Each rule has a name, a source and a request to match against. For a rule to match,
+   both source and request must match.
+   - If both source and request are empty, and the policy is an allow list, we will
+     accept all RPC requests. On the other hand, if the policy were a denylist, we
+     would deny all RPC requests.
+   - If only source is empty, we evaluate against the request fields and apply that to
+     any user (See example below). Similarly, if only request is empty, we evaluate
+     against the source fields and apply that to any action.
+   - This also applies to sub fields, like if the principals in source is empty, the
+     behavior would be similar to an empty source.
```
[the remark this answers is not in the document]
ashithasantosh: Makes sense We also want to be able to accept unset principals, which would apply to any authenticated user. Updated the text. Will make sure to update translator implementations in C++ and Go soon.

A43-sdk-authorization.md
ashithasantosh: Sounds good. So we will check for mTLS only when using principals field.
In which case, there won't be any difference in having SDK filter initialized with allow_all, or having no SDK filter, right?

A43-sdk-authorization.md
ashithasantosh: If a new field is added and user creates a config which includes this new field, and we run it against an older version of gRPC, that field will be ignored in the evaluation. Required or optional, this is undesired behavior. But I thought this should be user's responsibility to make sure they run against the right version. So if they want to use new fields, they would have to upgrade to the required version. I couldn't think of a way to address this. Do you have any suggestions?
Also you mentioned allow rules specifically, but we would have the same concern for deny rules as well?

This paragraph mainly talks about the life of a policy. So as long as we didn't introduce any "required" fields (we won't be deleting any fields), the same policy could be used across different gRPC versions.

A43-sdk-authorization.md
ashithasantosh: Sounds good. Included API definitions.

A43-sdk-authorization.md
ejona86: Deny would also have trouble, but they would fail things that shouldn't be failed. That is annoying, but isn't a security problem.

The only idea I have to resolve this problem other than looking for unknown fields is have a "version" field that starts at 1. When we add new fields that must be understood, then any config containing those fields would require version to be 2. But if these are hand-edited it doesn't help too much.

@yihuazhang, have any thoughts?

yihuazhang: Thanks for bringing this up, and I think there are two separate problems here:

1. Detect the problem of a "new" SDK policy being evaluated by an "old" gRPC library.
2. Action to be taken when 1) happens - either ignore the new field and continue with policy evaluation or disallow the use of SDK. 

For 1), one option I can think of is to ship an SDK JSON schema within each gRPC release, that should be considered as a source of truth for the specific gRPC release, and use it to validate a user provided SDK policy. It does not take any user provided inputs (e.g., version number) during validation, and the approach (i.e., JSON schema validation) seems pretty well adopted. 

For 2), from the security perspective, I prefer to disallow the use of SDK. By silently ignoring the new field, it may give the service owner false impression that SDK is functioning as expected. More importantly, it may allow the requests that should be rejected. Another option is to log an error when the problem happens, but it may not help much if the service owner does not consume the log in a correct way (e.g., never check the log).

yihuazhang: For 1), another option is to code the translator parsing logic to represent the SDK schema it currently supports, in which case there will be no need to ship the schema with its gRPC release. This is what gRPC Go currently does. @dfawley F.Y.I.

ashithasantosh: Currently both C++ and Go RBAC translator ignores fields other than the ones mentioned in JSON schema. It could be new/unknown fields. So I will look into the JSON library and figure out on how we can error out and mark the policy as invalid if there are fields other than what's expected in the current version. Will update the proposal with the details.

ashithasantosh: The current idea is that translator will mark the policy as invalid, if the policy contains fields that are unknown to the current JSON schema. Updated the proposal accordingly.
So far I checked C++ and Go, we can support this behavior.

Though this means, we won't have backward compatibility. We won't be able to use a policy with older gRPC versions.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+## Background
+
+Authentication verifies the authority of the request. Authorization ensures the 
+authenticated authority has access to the requested resource.
+
+gRPC offers built-in authentication mechanisms like TLS, ALTS and even allow
+users to plug in their own authentication systems. However, gRPC doesn't have any
+standard authorization solution available. If a user wanted to perform per RPC
+authorization check, they have to implement their own solution in the application.
+With standard gRPC authorization solution (under implementation), users can simply
```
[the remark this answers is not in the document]
ashithasantosh: Done.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+provide a policy configuration for their service, then the server verifies
+whether the client is authorized to access the requested service, and for which
+RPC methods. This differs from client-side authorization (also known as hostname
```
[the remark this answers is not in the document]
ashithasantosh: Sure. Removed line.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+
+## Abstract
+
+gRPC is currently implementing an open-source standard authorization solution
+based on Envoy RBAC. This proposal discusses how this framework would be made 
+available to non-xDS OSS gRPC users to perform per-RPC authorization checks.
+
+## Background
+
+Authentication verifies the authority of the request. Authorization ensures the
```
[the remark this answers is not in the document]
ashithasantosh: Updated the line. Since authentication is not a strict requirement for authorization, I reworded it as -
Authorization ensures that that requestor has sufficient access to make a particular request.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+### User facing Authorization Policy
+
+gRPC SDK authorization policy is the user facing policy language that allows
+service owners and/or security admins to enable per-RPC authorization checks.
+
+SDK authorization policy is extendable, we may add new fields in the future as
+new requirements are introduced, we won’t be removing fields or updating
+semantics of existing fields. However, this means if user wants to use policy
+with new fields, they need to use the right gRPC version. Using policy with new
+fields, with an older version of gRPC, will result in policy being marked invalid.
```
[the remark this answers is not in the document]
ashithasantosh: Yes, C++ work is pending. This is in my TODO list (go/grpc-authorization-track-sheet)
Go work is in review.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+* Author(s): [Ashitha Santhosh](https://github.com/ashithasantosh)
+* Approver: [Doug Fawley](https://github.com/dfawley),
+            [Eric Anderson](https://github.com/ejona86),
+            [Mark Roth](https://github.com/markdroth),
+            [Sanjay Pujare](https://github.com/sanjaypujare),
+            [Yihua Zhang](https://github.com/yihuazhang)
+* Status: Draft {Draft, In Review, Ready for Implementation, Implemented}
+* Implemented in: <language, ...>
+* Last updated: 2021-07-19
+* Discussion at: (filled after thread exists)
```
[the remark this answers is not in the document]
ashithasantosh: Done.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+
+## Abstract
+
+gRPC is currently implementing an open-source standard authorization solution
+based on Envoy RBAC. This proposal discusses how this framework would be made
```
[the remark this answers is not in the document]
ashithasantosh: Sounds good. Updated the proposal.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+   authorization
+2. SDK authorization (or non-xDS authz) - Authorization is made available to
+   non-xDS OSS gRPC users like network switches.
+
+This document discusses details of SDK authorization use-case. xDS authorization
+will be convered in a different proposal.
+
+## Related Proposals:
+
+* [A41](https://github.com/grpc/proposal/pull/237): xds RBAC Support
```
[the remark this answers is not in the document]
ashithasantosh: Sure. Moved to rationale.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+				]
+			}
+		}
+	]
+}
+```
+
+### High level Implementation
+
+gRPC Authorization internally implements RBAC Engines based on Envoy RBAC policies.
```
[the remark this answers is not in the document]
ashithasantosh: Updated the proposal. Please take a look and let me know if this looks good. Thank you!

A43-sdk-authorization.md
```diff
@@ -0,0 +1,905 @@
...
+contains the authorization policy in JSON format.
+
+We recommend users to use a single SDK authorization policy per gRPC server. If
+there are multiple policies, then there is a possibility that all the policies
+may not be evaluated against. For ex. if we have two policies for two different
+services say service A and service B. RPC to service B may get rejected, without
+even evaluating against service B policy, because it is evaluated after service A
+policy. On getting no match, service A policy could deny by default.
+
+Following code snippets show how to enable authorization in gRPC servers in
```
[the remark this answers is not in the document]
ashithasantosh: Sounds good. Removed the internal implementation details.
A pending item on me is to include the callback feature to the APIs. Will update the proposal soon (aiming this week).

A43-sdk-authorization.md
ashithasantosh: Done.

A43-sdk-authorization.md
ashithasantosh: Thank you for the discussion. I agree we should check for TLS only with principal fields. For the other fields, TLS will not be a pre-requisite.

ashithasantosh: Thank you for the discussion. I think we have agreement that SDK authorization can be enabled in xDS enabled servers, allowing both paths to co-exist. I added a statement in the rationale section.

A43-sdk-authorization.md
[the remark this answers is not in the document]
ashithasantosh: A policy that allows any authenticated user would look like -
{
    "name": "authz",
    "allow_rules": [
        {
            "name": "allow_authenticated",
            "source": {
                "principals": []
            }
        }
    ]
}

A43-sdk-authorization.md
```diff
@@ -0,0 +1,713 @@
...
+
+gRPC SDK authorization policy is the user facing policy language that allows
+service owners and/or security admins to enable per-RPC authorization checks.
+
+SDK authorization policy is extendable, we may add new fields in the future as
+new requirements are introduced, we won’t be removing fields or updating
+semantics of existing fields. However, this means if user wants to use policy
+with new fields, they need to use the right gRPC version. Using policy with new
+fields, with an older version of gRPC, will result in policy being marked invalid.
+This is essential, because otherwise we may end up allowing an RPC request without
```
sanjaypujare, line 45: If a policy's deny rule (but not allow rules) uses new fields and you mark the policy invalid you will end up allowing an RPC that could otherwise have been denied. In such a case your statement doesn't apply and you are not mitigating a security risk. Isn't that right?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,713 @@
...
+{
+  "title": "AuthorizationPolicy",
+  "definitions": {
+  	"rule": {
+  		"description": "Specification of rules. An empty rule is always matched"
+  		  "(i.e., both source and request are empty)",
+  		"type" : "object",
+  		"properties": {
+  			"name": {
+  				"description": "The name of an authorization rule. It is mainly for"
```
sanjaypujare, line 60: Does the `name` need to be unique in a given set of allow/deny rules ?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,713 @@
...
+  		}
+  	}
+  },
+  "required": ["name", "allow_rules"]
+}
+```
+
+#### Details
+
+gRPC SDK authorization policy has a list of *allow rules* and a list of *deny rules*.
```
sanjaypujare, line 192: The description here assumes a single authz policy attached to a server. Even if your JSON schema doesn't allow multiple policies what happens if you attach multiple policies like this (in Java):

```
authzServerInterceptor1 = FileAuthorizationServerInterceptor.create(authzPolicyFile1);
authzServerInterceptor2 = FileAuthorizationServerInterceptor.create(authzPolicyFile2);

Server server =
    Grpc.newServerBuilderForPort(port, serverCreds)
        .addService(service)
        .intercept(authzServerInterceptor1)
        .intercept(authzServerInterceptor2)
        .build()
        .start();
```

Is this allowed or will this be reported as an error? If not an error how are multiple policies handled?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,713 @@
...
+The following sequence is followed to make an authorization decision -
+1. Check for a match in *deny rules*, if a match is found, the request will be denied.
+   If no match is found in deny rules, or there are no deny rules execute next step.
+2. Check for a match in *allow rules*, if a match is found, the request will be allowed.
+   Note that allow rules is a required field here, if not present, policy is invalid.
```
sanjaypujare, line 197: `allow_rules` is a required field but its value can be an empty array. Is that also an invalid policy?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,713 @@
...
+3. Each *request* could contain a list of URL *paths* (i.e. fully qualified RPC
+   methods) and list of http *headers* to match. Refer JSON schema above to
+   understand matching semantics.
+   - If the sub-fields are empty, like if the *paths* is empty (assuming *headers*
+     is unset), the behavior would be similar to an empty *request* since no other
+     *request* fields are set.
+
+#### Example
+
+In the following policy example
```
sanjaypujare, line 232: How about including a "header" example in the snippet?

A43-sdk-authorization.md
```diff
@@ -0,0 +1,713 @@
...
+   - If both *source* and *request* are empty, the *rule* always matches. It is a
+     wildcard that can be used in a deny list to deny all RPCs or in an allow list to
+     accept all RPCs except those matching in the deny list.
+   - If only *source* is empty, we evaluate against the *request* fields and apply
+     that to any user (See example below). Similarly, if only *request* is empty, we
+     evaluate against the *source* fields and apply that to any action.
+2. Each *source* could contain a list of *principals*. The *principals* are ORed
+   together, i.e. it matches if one of them matches.
+   Sequence of steps to evaluate each principal from config -
+   1. If TLS is not used, matching fails.
```
sanjaypujare, line 212: You mean "if mTLS is not used..." . Mere TLS doesn't have the client identity available to the server

A43-sdk-authorization.md
```diff
@@ -0,0 +1,713 @@
...
+Traffic Director control plane sends Envoy RBAC policy configs to xDS enabled gRPC
+servers for authorization. ([A41](https://github.com/grpc/proposal/pull/237): xDS RBAC Support)
+
+Overall SDK authorization flow is as follows. User supplies gRPC SDK authorization
+policy to provider/interceptor. The provider then forwards the JSON policy to Policy
+translator. The translator converts JSON policy to Envoy RBAC protos (Allow and/or
+Deny policy). Translator errors out on I/O errors or if the policy does not
+represent JSON schema it currently supports. Ultimately the generated RBAC policies
+are used to create Envoy RBAC authorization engine(s). Then, for each incoming RPC
+request, we will invoke the Engines to get the authorization decision.
```
sanjaypujare, line 699: Any mention of optimization/caching such as:

- when a policy contains only principals (and no paths/headers) it's result is the same for all RPCs on a connection since the principals for a connection don't change
- if you cache the authorization result for a given set of headers+paths+principals then the cache can be used instead of evaluating the policy for every single RPC. 

Are they just implementation details or are they not being considered for now?

A43-sdk-authorization.md
ashithasantosh: We will be returning error status back to application notifying them of unknown/new fields. Previously, if we were using an older version, and deny rules had some new fields, then we would have just ignored that field, and evaluated based on the other fields (which are supported by that version), so we could have ended up allowing a request that should have been denied. Here instead of silently ignoring those new fields, we are returning an error back to user. In case of FW, the errors will be logged (except first reload, where we return error). Then the user can choose to fix the version or update the policy.

A43-sdk-authorization.md
ashithasantosh: 1. `{ name:"...", allow_rules:[] }` is invalid due to empty allow rules
2. `{ name:"...", allow_rules:[{name:"allow_all"}] }` is valid
3. `{ name:"...", deny_rules:[...] }` is invalid due to missing allow_rules

In the question, you are referring to the first case right? that should be invalid. But I realized the C++ code does not consider that case, and ends up creating an allow authorization engine with empty policies. So during evaluation, since there is no match, allow engine will always return deny decision. Thank you for bringing this up. Will update the code to return invalid for first case.

A43-sdk-authorization.md
ashithasantosh: It can be mTLS or TLS. We could have a policy to allow authenticated users to perform some action.

A43-sdk-authorization.md
ashithasantosh: Sure thing. Done.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,713 @@
...
+Authentication verifies the identity of the requestor. Authorization ensures that
+the requestor has sufficient access to make a particular request.
+
+gRPC offers built-in authentication mechanisms like TLS, ALTS and even allow
+users to plug in their own authentication systems. However, gRPC doesn't have any
+standard authorization solution available. If a user wanted to perform per RPC
+authorization check, they have to implement their own solution in the application.
+The standard gRPC authorization mechanism described in this proposal allows users
+to simply provide an authorization policy for their server. Then, for each RPC,
+the server verifies whether the client is authorized to issue the RPC.
```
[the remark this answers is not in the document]
ashithasantosh: Changed the text to "server verifies whether the client is authorized to make the RPC request."

Since the statement starts off as `server verifies`, I think this works. Let me know if you think we should still change it. And I can work on the wording:)

A43-sdk-authorization.md
sanjaypujare: But TLS (which is not mTLS) does not provide client authentication. So with TLS there cannot be "authenticated clients". On line 214 you say "if principal matches URI SANs from the client certificate". There is no client certificate in case of a TLS connection which is what I was trying to say.

ashithasantosh: That is a good point. 
Referring to xDS RBAC [proposal](https://github.com/grpc/proposal/blob/master/A41-xds-rbac.md): 
" If principal_name is unset, then Authenticated is said to match if the connection uses TLS; a client certificate is not necessary. " 
Also looking at envoy code, AFAIU its not constrained to mTLS
https://github.com/envoyproxy/envoy/blob/53f95570a7e9d7fa5414a4bc145852524fa017ed/source/extensions/filters/common/rbac/matchers.cc#L206
https://github.com/envoyproxy/envoy/blob/53f95570a7e9d7fa5414a4bc145852524fa017ed/envoy/ssl/connection.h#L79

Referring to my text in the proposal-
"if principal matches URI SANs from the client certificate"
If the peer certificate is not present, it will just skip this matching and move on to the next step
https://github.com/grpc/grpc/blob/master/src/core/lib/security/authorization/matchers.cc#L187

But I do realize that the term authenticated user for TLS seems incorrect, since the client never presents certificates. Let me look into Envoy code and documentation.

Adding @ejona86 for insights.

A43-sdk-authorization.md
ashithasantosh: Currently we do not have any such optimization/caching. For each incoming RPC, we evaluate against the SDK policy. I think several things need to be considered when doing such optimizations, especially with file reloading support. Being aware of the memory usage. Developing an efficient caching structure that can scale as the policy grows. Having some sort of monitoring to actually see the optimization benefits etc.
Thanks for bringing this up. We can definitely pick this up later.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,721 @@
...
+string. In dynamic file reloading, the application will specify the file path that
+contains the authorization policy in JSON format.
+
+Valid user provided SDK authorization policy creates authorization engine(s).
+In the case of file watcher, we internally create thread(C-core)/ goroutine(Go)/
+scheduled service(Java) which will be used to read the policy file periodically,
+and update the authorization engines. During the first file read, if the policy
+is invalid or there are I/O errrors, we will return error back to application and
+SDK authorization won't be enabled. If the error occurs on a later reload, then
+that particular reload will be skipped, and we will continue to use the latest
```
yihuazhang, line 302: nit. Should we also notify the application somehow that the reload fails (maybe via log)?

yihuazhang (approved):
Thanks @ashithasantosh for the great effort in addressing reviewers' comments. I left some nits, but LGTM overall.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,721 @@
...
+This proposal discusses how this framework would be made available to OSS gRPC
+users (like network switches) to perform per-RPC authorization checks.
+
+## Background
+
+Authentication verifies the identity of the requestor. Authorization ensures that
+the requestor has sufficient access to make a particular request.
+
+gRPC offers built-in authentication mechanisms like TLS, ALTS and even allow
+users to plug in their own authentication systems. However, gRPC doesn't have any
```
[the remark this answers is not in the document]
ashithasantosh: yes, referring to user defined metadata credentials.
https://grpc.io/docs/guides/auth/#extending-grpc-to-support-other-authentication-mechanisms

A43-sdk-authorization.md
ashithasantosh: Yes, we do have logging on failed reload. Added to the proposal.

A43-sdk-authorization.md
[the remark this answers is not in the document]
ashithasantosh: So the error will be returned when we try to create provider instance. 
And if we don't check for the error, we could end up creating server without SDK authorization.

That's why I emphasized on "invalid" policy here and the later sections cover what happens if the policy is invalid.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,721 @@
...
+   1. If TLS is not used, matching fails.
+   2. If there is no client certificate, we get a match if principal is an empty
+      string.
+   3. If we have a client certificate, we check against certificate contents. A
+      match is found if principal matches URI SANs from the certificate. If
+      there are no URI SANs in certificate, or no match was found, we check
+      against DNS SANs. Similarly, if certificate has no DNS SANs or match wasn't
+      found, we check against Subject field from certificate.
+   
+   Consider the case where the *principals* list is empty, then we only check for
```
ejona86, line 222: In this case does principals have to be set to an empty list or is simply have the source object be present?

If an empty list is necessary, then that makes it incompatible with proto.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,721 @@
...
+   step 1 above, i.e. rule matches for any authenticated user.
```
ejona86, line 223: If only step 1 is checked, then that means all users accessing the service over TLS are permitted. For only authenticated users, it looks like you should use `*`.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,721 @@
...
+// capability.
+public final class FileAuthorizationServerInterceptor implements ServerInterceptor {
+  private volatile AuthorizationServerInterceptor internalAuthzServerInterceptor;
+  private final String policyFile;
+  private FileTime lastModifiedTime;
+
+  // Constructor
+  private FileAuthorizationServerInterceptor(File policyFile) {
+    // Read policy from policyFile and create an internalAuthzServerInterceptor. An
+    // IOException or IllegalArgumentException will be thrown if the policy file
```
ejona86, line 586: I think for this case we'll just have IOException. If the policy is invalid we'll just wrap the exception in an IOException. IllegalArgumentException would be more appropriate if the file does not exist (since the file name is the only argument here), but FileNotFoundException is generally used for that case and is an IOException.

ejona86 (commented):
This looks very close. I think @markdroth's approval is the main thing lacking. Once he approves (or undoes "request changes"), I can provide final approval.

A43-sdk-authorization.md
ashithasantosh: Here, if the `principals: []`, then rule matches for any authenticated (TLS/mTLS) user, similar to `principals:["*"]`
Though the end result is the same, there are some differences.
In the "*" case, after the check for authenticated connection, since we first match against URI SANs, it will match against the first string and return successful match.

If it were an empty list (i.e. `principals: []`), then we can emulate the Envoy behavior of unset principal.
https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/rbac/v3/rbac.proto#config-rbac-v3-principal-authenticated
So, it will only check for authenticated connection, and return the match result accordingly.

For `source: {}` or missing source, principals would be set to ANY match.

Empty principals is implemented in Go, and C++ is in review.
https://github.com/grpc/grpc-go/blob/master/authz/rbac_translator.go#L165

Let me know your opinion. I can then accordingly update the proposal and related code.

A43-sdk-authorization.md
ashithasantosh: Sounds good. Updated the proposal.

A43-sdk-authorization.md
ejona86: I see the current Envoy behavior as subtle and not something to emulate (we have to to a degree, but we don't have to be happy about it). I think we want this to remain proto-compatible, which means we can't special-case present-but-empty array.

You can get the same behavior via `principals: ["*", ""]`. So it seems these special semantics aren't necessary.

ashithasantosh: Sure. Updated the proposal. Will make updates to the implementations.

A43-sdk-authorization.md
ashithasantosh: Done.

A43-sdk-authorization.md
ashithasantosh: So missing source, `source: {}`, `principals: []` will all result in ANY rule type for Principal.

ashithasantosh: @ejona86 I missed one of the reasons we chose empty principals. In TLS, where we do not have peer certificate, principals:["*"] would not match (won't match empty string since "*" is presence match). We may need to add special case in engine implementations in all 3 stacks to handle this.

ejona86: I thought I already mentioned that, and the solution:

> You can get the same behavior via `principals: ["*", ""]`. So it seems these special semantics aren't necessary.

ejona86: > But I felt it looked a bit hacky, since this representation is a result of internal implementation.

How so? `"*"` is defined as everything but empty string, so add an entry for empty string as well.

ashithasantosh: Sure. Sounds good.

A43-sdk-authorization.md
```diff
@@ -276,40 +283,26 @@ In the following policy example
...
 from filesystem. In static initialization, the policy will be provided as a JSON
 string. In dynamic file reloading, the application will specify the file path that
 contains the authorization policy in JSON format.
 
+Valid user provided SDK authorization policy creates authorization engine(s).
+In the case of file watcher, we internally create thread(C-core)/ goroutine(Go)/
+scheduled service(Java) which will be used to read the policy file periodically,
+and update the authorization engines. During the first file read, if the policy
+is invalid or there are I/O errrors, we will return error back to application and
+SDK authorization won't be enabled. If the error occurs on a later reload, then
```
markdroth, line 298: Saying that SDK authz won't be enabled implies that the server will start with no authz policy protection at all, which is not what we want.  This should probably just say something like "so that the gRPC server will not start".

A43-sdk-authorization.md
```diff
@@ -322,35 +315,52 @@ different languages.
 
 #### C++
 
-1. Application enables SDK authorization in gRPC Server.
-
-- Policy is statically initialized
+1. C++ authorization provider API
```
markdroth, line 318: Instead of using numbering here, please use subsection headers.  For example:

```
##### C++ Authorization Provider API
```

```
##### C-core Authorization Provider API
```

```
##### Example: Using a Static Authz Provider in C++
```

And so on.

Same for all languages.

A43-sdk-authorization.md
```diff
@@ -827,77 +615,101 @@ public final class FileAuthorizationServerInterceptor implements ServerIntercept
...
+RBAC policy provides service-level and method-level access control for a service.
+Engines process incoming RPC request attributes against policy configs and make a
+decision on whether to allow or deny the request. The decision depends on the type
+of policy (if the policy is an allowlist or denylist) and whether a matching policy
+was found.
+
+This engine implementation is shared by SDK-based authorization and xDS-based
+authorization. In gRPC Proxyless Service Mesh (or xDS authorization), Istio or
+Traffic Director control plane sends Envoy RBAC policy configs to xDS enabled gRPC
+servers for authorization. ([A41](https://github.com/grpc/proposal/pull/237): xDS RBAC Support)
```
markdroth, line 695: This link should point to `A41-xds-rbac.md` instead.

A43-sdk-authorization.md
```diff
@@ -827,77 +615,101 @@ public final class FileAuthorizationServerInterceptor implements ServerIntercept
...
+support all the fields that are present in Envoy RBAC. 
+
+RBAC policy provides service-level and method-level access control for a service.
+Engines process incoming RPC request attributes against policy configs and make a
+decision on whether to allow or deny the request. The decision depends on the type
+of policy (if the policy is an allowlist or denylist) and whether a matching policy
+was found.
+
+This engine implementation is shared by SDK-based authorization and xDS-based
+authorization. In gRPC Proxyless Service Mesh (or xDS authorization), Istio or
```
markdroth, line 693: I don't think it makes sense for this document to reference things like "Proxyless Service Mesh" or TD or Istio.  Instead, it should stick to discussing open-source protocols like xDS.  So this can say something like, "The xDS server sends an RBAC policy config to the xDS-enabled gRPC server for authorization."

markdroth (commented):
This looks really good overall!  I have just a few minor comments remaining, none of them really substantive.

A43-sdk-authorization.md
ashithasantosh: Sure. Updated text. Thank you!

A43-sdk-authorization.md
ashithasantosh: Done.

A43-sdk-authorization.md
ashithasantosh: Done.

A43-sdk-authorization.md
ashithasantosh: I removed the part "and SDK authorization won't be enabled". PTAL

A43-sdk-authorization.md
```diff
@@ -690,9 +684,8 @@ of policy (if the policy is an allowlist or denylist) and whether a matching pol
 was found.
 
 This engine implementation is shared by SDK-based authorization and xDS-based
-authorization. In gRPC Proxyless Service Mesh (or xDS authorization), Istio or
-Traffic Director control plane sends Envoy RBAC policy configs to xDS enabled gRPC
-servers for authorization. ([A41](https://github.com/grpc/proposal/pull/237): xDS RBAC Support)
+authorization. In xDS authorization, the xDS server sends an RBAC policy config to
+xDS-enabled gRPC server for authorization.([A41](https://github.com/grpc/proposal/blob/master/A41-xds-rbac.md): xDS RBAC Support)
```
markdroth, line 688: This should be a relative link, just `A41-xds-rbac.md`, not `https://github.com/grpc/proposal/blob/master/A41-xds-rbac.md`.

A43-sdk-authorization.md
```diff
@@ -294,10 +294,10 @@ Valid user provided SDK authorization policy creates authorization engine(s).
 In the case of file watcher, we internally create thread(C-core)/ goroutine(Go)/
 scheduled service(Java) which will be used to read the policy file periodically,
 and update the authorization engines. During the first file read, if the policy
-is invalid or there are I/O errrors, we will return error back to application and
-SDK authorization won't be enabled. If the error occurs on a later reload, then
-that particular reload will be skipped and error will be logged, and we will
-continue to use the latest valid policy to make authorization decisions.
+is invalid or there are I/O errrors, we will return error back to application.
```
markdroth, line 297: I think this should explicitly say something like "we will return an error back to the application, which will generally result in the gRPC server not starting."  That way, we make it clear what the impact of this API is.

A43-sdk-authorization.md
ashithasantosh: Done.

A43-sdk-authorization.md
ashithasantosh: Sure. Updated text. Thank you!

markdroth (approved):
This looks great!

ejona86 (approved):
Go ahead and update the Status to (one of) Ready for Implementation/Implemented and I'll merge.

ejona86:
The approver is also generally just one person. They make sure that there's consensus between interested parties. We had decided earlier I'd be the approver (due to my involvement with RBAC).

@dfawley, should we wait on your review? I'd feel better if someone had at least looked over the Go section. My understanding is that none of this should be surprising and based on previous discussions with you/some-other-Go-person.

ashithasantosh:
Thank you all for the detailed review and the interesting questions! :)

A43-sdk-authorization.md
```diff
@@ -0,0 +1,707 @@
...
+
+```Go
+creds := credentials.NewServerTLSFromFile(certFile, keyFile)
+i, err := authz.NewStatic(authzPolicy)
+serverOpts := []grpc.ServerOption{
+  grpc.Creds(creds),
+  grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
+  grpc.ChainStreamInterceptor(i.StreamInterceptor),
+}
+s := grpc.NewServer(serverOpts...)
```
dfawley, line 520: IMO it's clearer to remove the slice and do:
`s := grpc.NewServer(grpc.Creds(...), grpc.ChainUnaryInterceptor(...), ...)`

Also below.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,707 @@
...
+creds := credentials.NewServerTLSFromFile(certFile, keyFile)
+i, err := authz.NewFileWatcher(authzPolicyFile, 1 * time.Hour)
+serverOpts := []grpc.ServerOption{
+  grpc.Creds(creds),
+  grpc.ChainUnaryInterceptor(i.UnaryInterceptor),
+  grpc.ChainStreamInterceptor(i.StreamInterceptor),
+}
+s := grpc.NewServer(serverOpts...)
+// In the end, free up the resources used for policy refresh.
+defer i.Close()
```
dfawley, line 535: This should be done immediately upon `NewFileWatcher`s return.  (After checking `err`)

A43-sdk-authorization.md
```diff
@@ -0,0 +1,707 @@
...
+func (i *FileWatcherInterceptor) Close() {
+  i.cancel()
+}
+```
+
+##### Example: Using a Static Authz Interceptor in gRPC-Go Server
+
+```Go
+creds := credentials.NewServerTLSFromFile(certFile, keyFile)
+i, err := authz.NewStatic(authzPolicy)
```
dfawley, line 514: Maybe add a comment after this line: `// Ensure err is nil.`

A43-sdk-authorization.md
```diff
@@ -0,0 +1,707 @@
...
+
+As mentioned previously, authorization APIs take policy in JSON format, instead of
+protobuf. This is done to avoid the dependency on protobuf. We have users in OSS
+that use gRPC without protobuf. Another reason is to have a consistent API across
+languages.
+
+## Implementation
+
+The implementation order will be C++, Go, Java and then wrapped languages. The 
+implementation will be done by ashithasantosh.
```
dfawley, line 707: Is this just "Done in all languages" now?

dfawley:
> @dfawley, should we wait on your review?

Apologies.  I had done a review pass on this long ago but never sent my pending comments.  Sending now.

dfawley (changes requested):
I will say that I find the name of this feature using the term "SDK" to be unusual, since we have never used this term before, despite having many previous proposals with cross-language APIs in them.  But if @markdroth and @ejona86 are fine with it then it's not a big sticking point.

A43-sdk-authorization.md
ashithasantosh: Its done in C++ and Go. Java is currently under implementation.

A43-sdk-authorization.md
ashithasantosh: Good point. Done.

A43-sdk-authorization.md
ashithasantosh: Done

A43-sdk-authorization.md
```diff
@@ -0,0 +1,707 @@
...
+}
+
+// NewFileWatcher returns a new FileWatcherInterceptor from a policy file
+// that contains JSON string of authorization policy and a refresh duration to
+// specify the amount of time between policy refreshes.
+func NewFileWatcher(file string, duration time.Duration) (*FileWatcherInterceptor, error);
+
+// run is a long running goroutine which watches for changes in file path, and
+// updates the internalInterceptors upon modification.
+func (i *FileWatcherInterceptor) run(ctx context.Context) {
```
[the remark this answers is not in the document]
ashithasantosh: Done

A43-sdk-authorization.md
```diff
@@ -0,0 +1,707 @@
...
+#### Go
+
+##### Authorization Server Interceptors
+
+```Go
+package authz
+
+// StaticInterceptor contains engines used to make authorization decisions.
+type StaticInterceptor struct {
+  engines ChainEngine
```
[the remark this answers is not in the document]
ashithasantosh: Removed fields that are not exported.

ashithasantosh:
> I will say that I find the name of this feature using the term "SDK" to be unusual, since we have never used this term before, despite having many previous proposals with cross-language APIs in them. But if @markdroth and @ejona86 are fine with it then it's not a big sticking point.

I used the term from gRPC SDK Authorization Policy document containing the policy details. I think we wanted to differentiate from xDS authorization. That's why using SDK authorization instead of gRPC authorization here.

A43-sdk-authorization.md
```diff
@@ -0,0 +1,707 @@
...
+		(err error) {
+  // ...
+}
+
+type FileWatcherInterceptor struct {
+  internalInterceptors StaticInterceptors
+  policyFile           string
+  policyContents       string
+  refreshDuration      time.Duration
+  cancel               context.CancelFunc
```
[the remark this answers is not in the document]
ashithasantosh: Done

A43-sdk-authorization.md
dfawley: _After_ checking `err`.  You have this above now.

dfawley (commented):
> I used the term from gRPC SDK Authorization Policy document containing the policy details. I think we wanted to differentiate from xDS authorization. That's why using SDK authorization instead of gRPC authorization here.

We also have concepts like "gRPC retry" vs. "xDS retry", so I'm not sure why we think people would be likely to confuse "gRPC authz" with "xDS authz".  Meanwhile, "SDK" is not a term we've ever used before.  🤷

ashithasantosh:
> > I used the term from gRPC SDK Authorization Policy document containing the policy details. I think we wanted to differentiate from xDS authorization. That's why using SDK authorization instead of gRPC authorization here.
> 
> We also have concepts like "gRPC retry" vs. "xDS retry", so I'm not sure why we think people would be likely to confuse "gRPC authz" with "xDS authz". Meanwhile, "SDK" is not a term we've ever used before. 🤷

Sure. We can replace SDK with gRPC in the title and everywhere else in the proposal, and will also do a cleanup in the implementations. Let me add @markdroth @ejona86 to this discussion to see if they have any preferences wrt naming.

ejona86:
I always found the term SDK applied to authz to either be meaningless or misleading, depending on what it was addressing. So I'd be happy to see it go away. I didn't argue about it because we had bigger fish to fry. If there's a strong need for a differentiator (which, as with Doug, is unclear to me), then "API" seems clearly superior to SDK.

markdroth:
I am not necessarily super attached to the term "SDK" here, but I also don't really object to it.  It doesn't bother me enough to warrant the amount of work I suspect we'd need to do to remove the term from everywhere we've been using it.  But if Doug or Eric feels strongly and Ashitha is willing to do the work, I don't object to changing it.

ashithasantosh:
@dfawley @markdroth @ejona86
Thank you for the discussion!
I will then go ahead and remove SDK term from everywhere and replace them with gRPC. Ex. the title of this proposal will be `A43: gRPC authorization`.

sanjaypujare:
> @dfawley @markdroth @ejona86 Thank you for the discussion! I will then go ahead and remove SDK term from everywhere and replace them with gRPC when necessary. Ex. the title of this proposal will be `A43: gRPC authorization`.

IIRC the reason we decided to call it Authorization SDK is to differentiate from the xDS based authorization. So "gRPC Authorization API" might be (more) appropriate? In any case the file name still has `-sdk-authorization.md` (nit)

ashithasantosh:
> IIRC the reason we decided to call it Authorization SDK is to differentiate from the xDS based authorization. So "gRPC Authorization API" might be (more) appropriate? In any case the file name still has `-sdk-authorization.md` (nit)

Sure. Updated the title and also the file name. Thank you!:)

dfawley (approved):

PR #247, opened 2021-06-29
A36: replace destination_ports with source_prefix_ranges in the example

sanjaypujare:
Since destination_ports is not an array (a singleton) and source_prefix_ranges is an array.

### DIFF
--- a/A36-xds-for-servers.md
+++ b/A36-xds-for-servers.md
@@ -217,7 +217,7 @@ its present fields, which avoids repeating fields and converts the matchers into
 disjunctive normal form. That is, the matcher:
 
 ```
-destination_ports: 80, 8080
+source_ports: 80, 8080
 prefix_ranges: 192.168.0.0/16, 10.0.0.0/8
 source_type: EXTERNAL
 ```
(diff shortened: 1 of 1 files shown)


sanjaypujare:
@ejona86 hope you agree with this change...

ejona86:
Matching on source_ports is really weird; I can't actually think of a use for it even in Envoy. I want to take a look to see if there is a more natural alternative.

sanjaypujare:
> Matching on source_ports is really weird; I can't actually think of a use for it even in Envoy. I want to take a look to see if there is a more natural alternative.

Agreed but the current example is syntactically wrong since `destination_port` is not an array but `source_ports` is. May be `source_prefix_ranges` ?

ejona86:
Yeah, let's do source_prefix_ranges. How about:

```
prefix_ranges: 192.168.0.0/24, 10.1.0.0/16
source_prefix_ranges: 192.168.1.0/24, 10.2.0.0/16
source_type: EXTERNAL
```

sanjaypujare:
Done

ejona86 (approved):
Thank you.

PR #250, opened 2021-07-20
A39 update: Add support for terminal filters.

markdroth:
This matches Envoy's handling, removes special-case handling that we had for the router filter, and allows us to catch more problems at config validation time.

C-core implementation in grpc/grpc#26742.

### DIFF
--- a/A39-xds-http-filters.md
+++ b/A39-xds-http-filters.md
@@ -4,7 +4,7 @@ A39: xDS HTTP Filter Support
 * Approver: ejona86, dfawley
 * Status: Approved
 * Implemented in: C-core
-* Last updated: 2021-02-12
+* Last updated: 2021-07-20
 * Discussion at: https://groups.google.com/g/grpc-io/c/M-l8k2v5snY
 
 ## Abstract
@@ -27,7 +27,7 @@ that will enable us to implement functionality that, in Envoy, is
 provided by its out-of-the-box filters (e.g., fault injection or
 authorization), which we also want to be provided out-of-the-box in gRPC.
 
-### Related Proposals: 
+### Related Proposals:
 
 - [A27: xDS-Based Global Load Balancing](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)
 - [A30: xDS v3 Support](https://github.com/grpc/proposal/blob/master/A30-xds-v3.md)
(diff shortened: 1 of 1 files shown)


dfawley (approved):

A39-xds-http-filters.md
```diff
@@ -233,9 +236,9 @@ gRPC will support the [router
 filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter),
 which is used for the proto message type
 [`envoy.extensions.filters.http.router.v3.Router`](https://github.com/envoyproxy/envoy/blob/18db4c90e3295fb2c39bfc7b2ce641cfd6c3fbed/api/envoy/extensions/filters/http/router/v3/router.proto#L23).
-In practice, in order for gRPC to function properly, this filter is
-required and must be last in the filter list, although that is not
-enforced at config validation time.
+This filter is a terminal filter, so it must be the last one in the
```
ejona86, line 239: This isn't strictly true given the above. "A" terminal filter must be last, but it doesn't have to be this one. I think this needs to be reworded if this is the only supported terminal filter.

A39-xds-http-filters.md
```diff
@@ -78,12 +78,14 @@ message type for its top-level configuration, but not all filters
...
 use the same message type for their override configuration as for their
 top-level configuration.  Therefore, every filter will be registered
-one or more message types.
+by one or more message types.
 
 Each filter implementation will provide the following:
 - Method(s) for validating the config protos when parsing an xDS response.
 - An indication of whether it is supported on the gRPC client and/or gRPC
   server.
+- An indication of whether the filter is a terminal filter (i.e., it
```
ejona86, line 87: It seems we could delete this and change the later text to read
> In addition,
 validation will fail if a terminal filter is not the last filter in
 the chain or if a non-terminal filter is the last filter in the chain.
 The only currently-supported terminal filter is
  `envoy.extensions.filters.http.router.v3.Router`.
 Any validation error will cause the Listener resource to be NACKed.

That allows you to use the registry if you want without implying that we support other terminal filters.

markdroth: Updated wording to clarify this.

A39-xds-http-filters.md
markdroth: Done.

ejona86 (approved):

PR #251, opened 2021-07-21
A36 update: Require HCM to be the terminal network filter

ejona86:
This is to match Envoy's handling. See also
https://github.com/grpc/proposal/pull/250 for the http filter change.

### DIFF
--- a/A36-xds-for-servers.md
+++ b/A36-xds-for-servers.md
@@ -5,7 +5,7 @@ A36: xDS-Enabled Servers
 * Approver: markdroth
 * Status: Ready for Implementation
 * Implemented in: <language, ...>
-* Last updated: 2021-02-25
+* Last updated: 2021-07-21
 * Discussion at: https://groups.google.com/g/grpc-io/c/CDjGypQi1J0
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


markdroth (approved):

dfawley (approved):
cc @zasweq

ejona86:
CC @yashykt, @YifeiZhuang

YifeiZhuang:
> CC @yashykt, @YifeiZhuang

@ejona86  thank you.
To make sure I understand things correctly and completely, we do will do these in java, right?
1. add `isTerminalFilter()` API in `Filter` interface and their implementations.
2. add `isTerminal == isLast` validation is `XdsClient` upon receiving listener update response, for both client and server httpFilter configs.
3. remove `LameFilter` entirely.
4. change to only allow one HCM in each server listener filterChain in XdsClient parsing validation logic.

ejona86:
> add isTerminalFilter() API in Filter interface and their implementations.

I don't think we'll add the API isTerminalFilter(). From what I saw, it is easiest to just hard-code the check for Router in validation of XdsClient. But yes, the other parts are what I was thinking.

Note that the HCM validation logic is already hard-coded; it seems with this change we don't even need a loop any more, but whatever ¯\_(ツ)_/¯

YifeiZhuang:
> I don't think we'll add the API isTerminalFilter(). From what I saw, it is easiest to just hard-code the check for Router in validation of XdsClient. But yes, the other parts are what I was thinking.

Got it. Just it might be a conflict for the coming soon server validation because java share the same validation code. But this is trivial. As long as we know that "Router is not the last" is invalid  ¯_(ツ)_/¯

PR #252, opened 2021-07-29
L82: Proposed additions core-cpp public headers to support TLS Key export in GRPC C++

Vignesh2208:
Relevant Github Issue: https://github.com/grpc/grpc/issues/24944
Relevant PR: https://github.com/grpc/grpc/pull/26812


markdroth:
The number "L82" is already taken by #245.  Please use "L83" for this.

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,93 @@
...
+### Related Proposals:
+
+N/A
+
+## Proposal
+
+The proposal extends the existing `grpc::experimental::TlsCredentialsOptions` class with methods to set configuration for supporting Tls Key export at run-time. The proposed changes
+include:
+
+* A `grpc_tls_key_log_format` enum is added to `include/grpc/grpc_security_constants.h`
```
ctiller, line 27: Let's omit this for now.

Instead, add a channel arg (string) for the path, and if we need to also specify the format we can add a secondary parameter in the future.

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,93 @@
...
+...
+ServerBuilder builder;
+builder.AddListeningPort(server_address,
+      grpc::experimental::TlsCredentials(options));
+```
+
+
+## Rationale
+
+Alternatives using ENVIRONMENT Variables to set TLS Key export configurations were considered. These allow enabling/disabling key export without having to rebuild the binary. However the disadvantages here include:
```
ctiller, line 79: I'm still an advocate for requiring both a channel arg and an environment variable... having two locks on such a potentially damaging feature seems useful.

ctiller (commented):
Thanks for writing this up!

L83-core-cpp-tls-key-export-feature.md
Vignesh2208: Sure. Should we get rid of the struct grpc::experimental::TlsKeyLoggerConfig as well ? Currently it has two members: the file path string and file log format. Getting rid of the file log format would leave it with just one member. I could get rid of the struct but it would make it harder to add any new extension in the future.

L83-core-cpp-tls-key-export-feature.md
Vignesh2208: You mean like an environment variable override i.e, log keys only when both the env variable and the above configs are set ?

ctiller:
@yihuazhang @ZhenLian @markdroth I don't think I'm sufficiently up to date with the TLS creds model to be an approver here, can someone else take that role?

yihuazhang:
@ZhenLian, since the new feature will be added to advanced TLS, do you mind taking a look? PLMK if you are busy, I can also take the role.

ZhenLian:
@yihuazhang Yeah this is in my list, but feel free to take another look if possible. 
Sorry about the delay - I was busy with perf&intern stuff, etc in the past few weeks.

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,93 @@
...
+* Author(s): Vignesh Babu (Vignesh2208)
+* Approver: ctiller
+* Status: In Review
+* Implemented in: core, cpp
+* Last updated: 07/29/2021
+* Discussion at: https://github.com/grpc/grpc/issues/24944
+
+## Abstract
+
+This proposal discusses additions to the core/cpp public headers to support exporing of TLS Keys on the client and server side to aid in decryption of packet captures using tools such as Wireshark.
```
yihuazhang, line 12: How about renaming it to `TLS session keys` to make it more clear?

L83-core-cpp-tls-key-export-feature.md
yihuazhang: One benefit I can think of using channel args is that it is agnostic to credential types and thus can support both SSL and TLS credentials. However, I do not think that benefit outweighs all three reasons you mentioned here, especially there is no motivation to support the feature in the old SSL stack.

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,93 @@
...
+
+The proposal extends the existing `grpc::experimental::TlsCredentialsOptions` class with methods to set configuration for supporting Tls Key export at run-time. The proposed changes
+include:
+
+* A `grpc_tls_key_log_format` enum is added to `include/grpc/grpc_security_constants.h`
+  ```
+  /** Supported TLS Keylogging formats **/
+  typedef enum {
+    // Default key logging formal is NSS compatible.
+    TLS_KEY_LOG_FORMAT_NSS = 0,
```
yihuazhang, line 32: Can you please add `GRPC_` prefix to all these constants?

L83-core-cpp-tls-key-export-feature.md
yihuazhang: Craig, I agree it is a potentially damaging feature if misused. But in the current design, in order to enable it, a user needs to explicitly create a `TlsKeyLoggerConfig` instance and set it in `TlsServerCredentialsOptions`, which I think is sufficient to confirm user's intention to enable the feature. I am concerned that introducing extra knobs may result in usability problems.

yihuazhang (commented):
LGTM overall. Left some minor comments.

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,93 @@
+Title
```
markdroth, line 1: This should say "L83: TLS Session Key Export in gRPC C++".

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,93 @@
...
+
+Currently unlike the GRPC Go implementation, GRPC C++ does not allow export of TLS keys at run time. This is needed to decrypt and analyse packet captures for debugging purposes. The keys when exported in files in a specific format, can be read by debugging tools like wireshark and they can automatically decrypt packet captures.
+
+### Related Proposals:
+
+N/A
+
+## Proposal
+
+The proposal extends the existing `grpc::experimental::TlsCredentialsOptions` class with methods to set configuration for supporting Tls Key export at run-time. The proposed changes
```
markdroth, line 24: The fact that we're using `experimental` here gives me pause, since experimental APIs don't actually require gRFCs at all.  TlsCreds as a whole is still experimental, and there will be a gRFC for it as a whole when we're ready to de-experimentalize it (see #205 for some work-in-progress).

I suspect this means that we should just add this API to the TlsCreds gRFC that @ZhenLian is working on, rather than having this separate gRFC.  And not having the gRFC merged should not block the implementation, since the API is still currently experimental -- it just means it may be subject to change later.

L83-core-cpp-tls-key-export-feature.md
markdroth: I think we definitely want this feature to be configured via TlsCreds, not via channel args.  That's how we're doing all credential configuration, and I think it would be quite strange to do this one feature differently.

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,93 @@
...
+  /** Supported TLS Keylogging formats **/
+  typedef enum {
+    // Default key logging formal is NSS compatible.
+    TLS_KEY_LOG_FORMAT_NSS = 0,
+    // More formats can be added here for future extensions if desired.
+    TLS_KEY_LOG_FORMAT_NONE
+  } grpc_tls_key_log_format;
+  ```
+
+* A `grpc::experimental::TlsKeyLoggerConfig` struct is defined in `include/grpcpp/security/tls_credentials_options.h` to support configuration for Key export. It currently has two members:
```
markdroth, line 38: Please document both the C-core API and the C++ API for the TlsCreds options.

markdroth (changes requested):
Thanks for writing this up!

Given that TlsCreds as a whole is still experimental and that @ZhenLian is still working on a gRFC for it, I think it would probably make sense to just proceed with the implementation as experimental and merge this content into that in-progress gRFC for TlsCreds as a whole.

L83-core-cpp-tls-key-export-feature.md
Vignesh2208: Yes @ZhenLian also thought it would be a good idea to merge them. I'll do that after this is approved.

L83-core-cpp-tls-key-export-feature.md
Vignesh2208: Done. I have added new sections to document them.

L83-core-cpp-tls-key-export-feature.md
Vignesh2208: There is a GRPC_TLS_KEY_LOGGING_ENABLED environment variable to override right now. I think we can remove this at a later point if there are usability issues.

L83-core-cpp-tls-key-export-feature.md
Vignesh2208: Done.

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * An opaque type pointing to a Tls Key logger helper object which provides
+ * methods to log TLS session keys to files.
+ */
+typedef struct gprc_tls_key_logger grpc_tls_key_logger;
+
+/**
+ * An opaque type pointing to a Tls key logging configuration.
+ */
+typedef struct grpc_tls_key_log_config grpc_tls_key_log_config;
```
ZhenLian, line 101: Since `grpc_tls_key_log_config` is an opaque type, how can the users specify different configurations through it?
I thought we were going to have something like this:
```
typedef struct tls_key_log_format {
  ...
}

typedef struct grpc_tls_key_log_config {
  const char* log_file_path;
  tls_key_log_format format;
}
void grpc_tls_credentials_options_set_tls_key_log_config(
    grpc_tls_credentials_options* options, grpc_tls_key_log_config* config);
```
Then we don't need to expose `gprc_tls_key_logger` at all.

ZhenLian (changes requested):

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,202 @@
...
+GRPCAPI void grpc_tls_key_logger_destroy(grpc_tls_key_logger* key_logger);
+
+/**
+ * Initializes a Tls Key logger registry. Tls Key logger objects can only be
+ * created after the registry is initialized.
+ */
+GRPCAPI void grpc_tls_key_logger_registry_init();
+```
+
+# Example Usage
```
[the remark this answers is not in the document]
Vignesh2208: Done.

L83-core-cpp-tls-key-export-feature.md
```diff
@@ -0,0 +1,202 @@
...
+* Author(s): Vignesh Babu (Vignesh2208)
+* Approver: ctiller
+* Status: In Review
+* Implemented in: core, cpp
+* Last updated: 09/28/2021
+* Discussion at: https://github.com/grpc/grpc/issues/24944
+
+# Abstract
+
+This proposal discusses additions to the core/cpp public headers to support exporing of TLS session keys on the client and server side to aid in decryption of packet captures using tools such as Wireshark.
```
ZhenLian, line 12: nit: exporing/exporting

ZhenLian (approved):
This is looking pretty good!

ZhenLian:
@markdroth Mark, can I know in general, when do we expect a gRFC to be merged? Do we merge it when the proposed API is implemented, or we merge it when the proposed API become stable?
I know here we are waiting for https://github.com/grpc/proposal/pull/205 to get in, but given the recent work to unify the security API, the newly proposed change is unlikely to get implemented throughout the end of this year. That might further defer gRFC(s) like this.
There are still some rooms for improvements around `Provider` and `Verifier`, but the `TlsCredentials` and `TlsCredentialsOptions` themselves are pretty "stable" now. I wonder if we can come up with a way to merge the gRFCs step by step? Maybe I can write the gRFC for the `TlsCredentials` and `TlsCredentialsOptions`, and then we can gradually merge some small features, like TLS key exporting, or TLS version selection(which is a feature blocking for a really long time, since last year I think, @matthewstevenson88)? For gRFCs about `Provider` or `Verifier`, it's OK to put them in `Open` state, since they are still not finalized yet.
This also helps to keep each gRFC relatively short and staying on one single security topic, which might be easier for future readers to read. What do you think about this idea?
@yihuazhang FYI.

PR #253, opened 2021-08-02
A44: gRPC xDS Retry Support

dapengzhang0:
@dfawley @ejona86 @markdroth Please review.


A44-xds-retry.md
```diff
@@ -0,0 +1,135 @@
...
+### Related Proposals
+
+- [A6: gRPC Retry Design](https://github.com/grpc/proposal/blob/master/A6-client-retries.md)
+- [A31: gRPC xDS Timeout Support and Config Selector Design](https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md)
+
+## Proposal
+
+### Features to support
+
+Due to the differences between gRPC and Envoy, incomplete implementation status of the original [retry gRFC](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) and unresolved [compatibility issues](https://github.com/envoyproxy/envoy/issues/17412) with dynamic cluster selection, we propose initially to support only some of the retry features in Envoy.
```
markdroth, line 34: I think this can just say "gRPC will support the following xDS retry features."  The details of why we aren't going to support individual features can be covered below.

markdroth (changes requested):
This looks good overall!

My only substantive comment is the one about explaining why we are deferring or not supporting each individual feature.

A44-xds-retry.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Background
+
+gRPC library provides a [built-in retry](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) mechanism based on gRPC retry policy in service config. Envoy also provides a [proxy-level retry](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_routing.html?highlight=hedge#retry-semantics) based on either route configuration or request headers for either REST or gRPC requests. The Envoy retry for gRPC requests is triggered either based on the failure indicated by the initial metadata in the response, in the same way as gRPC built-in retry, or based on a per-try-timeout. For proxyless gRPC (aka gRPC-xDS), it is proposed to support various features and configurations in route action such as timeout and retry using a per-RPC ConfigSelector (see [gRPC xDS Config Selector Design](https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md)). It is natural to fit Envoy retry config into gRPC retry policy with proper extension.
```
ejona86, line 15: Remove `?highlight=hedge` from URL.

A44-xds-retry.md
```diff
@@ -0,0 +1,135 @@
...
+Due to the differences between gRPC and Envoy, incomplete implementation status of the original [retry gRFC](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) and unresolved [compatibility issues](https://github.com/envoyproxy/envoy/issues/17412) with dynamic cluster selection, we propose initially to support only some of the retry features in Envoy.
+
+#### Supported Retry Features
+
+1. Retry on a given set of gRPC status codes (x-envoy-retry-grpc-on values) specified by the retry_on config.
+1. Backoff retry on retriable error status by a time given by the retry_back_off config.
+1. Limit the number of attempts that can be retried for a given RPC by the num_retries config.
+
+
+#### Deferred or Unsupported Retry Features
```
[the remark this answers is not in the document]
dapengzhang0: Great suggestion. Thank you so much for the detailed explanation for each field!

> The retry_host_predicate and host_selection_retry_max_attempts fields, because this mechanism is different than the one described for hedging in gRFC A6, so we will evaluate this later, possibly as part of implementing hedging.

I wonder why compare these fields with hedging and possibly implement as part of hedging. Is Envoy implementing this as part of Envoy's hedging?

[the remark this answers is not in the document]
markdroth: In gRFC A6, I believe the intent was that the feature of avoiding an endpoint that you had already sent to was primarily intended for the hedging case, because it really doesn't make sense to have two outstanding attempts of the same request to the same endpoint.  I don't think there was any particular reason why we couldn't have used the same logic for retries, but it wasn't considered as important there, which is why A6 discusses this only in the context of hedging.

If/when we actually implement something like this, I think we probably *should* do it for both retries and hedging.  But since it is more important for hedging than retries, implementing hedging will probably be the impetus for working on this feature.

A44-xds-retry.md
```diff
@@ -24,50 +24,50 @@ There are some key differences between gRPC’s and Envoy’s retry policy:
...
+- Limit the number of attempts that can be retried for a given RPC by the `num_retries` config.
 
-1. Retry on a given set of gRPC status codes (x-envoy-retry-grpc-on values) specified by the retry_on config.
-1. Backoff retry on retriable error status by a time given by the retry_back_off config.
-1. Limit the number of attempts that can be retried for a given RPC by the num_retries config.
+Features that we will not support right now but may add in the future:
 
+- Hedging support, because not all gRPC languages have implemented hedging yet, and we didn't want to wait for that before making retries configurable via xDS.
+- The `per_try_timeout` field, because we can't safely support this now without hedging support, lest we break users when we add hedging support. We will add this later, at the same time as we add hedging support. See the ["Rationale" section](#Supporting-per_try_timeout) below for more details.
+- Any `retry_on` conditions from values of [x-envoy-retry-on](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#x-envoy-retry-on).
```
markdroth, line 44: I think this line should be deleted.  This is already listed below under "features that we do not plan to support", and I think that's more accurate -- this is not something that we'll be adding later.

A44-xds-retry.md
```diff
@@ -24,50 +24,50 @@ There are some key differences between gRPC’s and Envoy’s retry policy:
...
+- The `retry_host_predicate`, `host_selection_retry_max_attempts` and `retry_priority` fields, because this mechanism is different than the one described for hedging in [gRFC A6](A6-client-retries.md#hedging-policy), so we will evaluate this later, possibly as part of implementing hedging.
```
markdroth, line 45: I don't think we'll ever support `retry_priority`, because gRPC does not support HTTP/2 priorities.  So that belongs in the next section.

markdroth (changes requested):
This looks good overall.  Just a few minor comments remaining.

A44-xds-retry.md
dapengzhang0: Moved `retry_priority` to the next section.

markdroth (approved):
Looks great!

markdroth:
(We just need to wait until Aug 16 to merge.)

PR #255, opened 2021-08-11
A36 update: Document expected behavior when observing RouteConfiguration

ejona86:
The text permitted certain carve-outs if RouteConfiguration was not
being observed, but it didn't describe the behavior if
RouteConfiguration _was_ observed.

-----

Split out of #237, as it is grown a bit.

CC @markdroth, @dfawley, @yashykt, @YifeiZhuang, @zasweq

### DIFF
--- a/A36-xds-for-servers.md
+++ b/A36-xds-for-servers.md
@@ -217,9 +217,9 @@ period of 10 minutes for long-lived RPCs, such that clients will reconnect and
 have the updated configuration apply. This applies equally to an update in a
 RouteConfiguration provided inline via the `route_config` field as it is part of
 the Listener, but it does not apply to an updated RouteConfiguration provided by
-reference via `rds` field. Infrastructure for "not serving" mode may be used for
-the graceful shutdown, but only if it can be used without causing spurious RPC
-or connection failures. Applying updates to a Listener should be delayed until
+reference via `rds` field. Draining must not cause the server to spuriously fail
+RPCs or connections, so the listening port must not be closed as part of the
+process. Applying updates to a Listener should be delayed until
 dependent resources have been attempted to be loaded (e.g., via RDS). The
 existing resource loading timeout in XdsClient prevents the update from being
 delayed indefinitely and the duplicate resource update detection in XdsClient
--- a/A36-xds-for-servers.md
+++ b/A36-xds-for-servers.md
@@ -5,7 +5,7 @@ A36: xDS-Enabled Servers
 * Approver: markdroth
 * Status: Ready for Implementation
 * Implemented in: <language, ...>
-* Last updated: 2021-07-21
+* Last updated: 2021-08-11
 * Discussion at: https://groups.google.com/g/grpc-io/c/CDjGypQi1J0
 
 ## Abstract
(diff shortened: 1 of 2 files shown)


A36-xds-for-servers.md
```diff
@@ -192,8 +190,48 @@ required. This is to allow implementations that only support L4 xDS features to
...
+a non-existent resource, then all RPCs processed by that HttpConnectionManager
+will fail with UNAVAILABLE.
+
+[Like in Envoy][envoy lds], updates to a Listener cause all older connections on
+that Listener to be gracefully shut down (i.e., "drained") with a default grace
+period of 10 minutes for long-lived RPCs, such that clients will reconnect and
+have the updated configuration apply. This applies equally to an update in a
+RouteConfiguration provided inline via the `route_config` field as it is part of
+the Listener, but it does not apply to an updated RouteConfiguration provided by
+reference via `rds` field. Infrastructure for "not serving" mode may be used for
```
markdroth, line 220: I don't understand what this sentance means:

"""
Infrastructure for "not serving" mode may be used for the graceful shutdown, but only if it can be used without causing spurious RPC or connection failures.
"""

If we get an LDS update, then presumably we should switch over to the new config for incoming connections ~immediately (modulo a short delay to get an RDS resource, if any, but that will be shorter than the 10m drain grace time), so there should be no period of time during which we are not listening to report as "not serving".

A36-xds-for-servers.md
```diff
@@ -192,8 +190,48 @@ required. This is to allow implementations that only support L4 xDS features to
...
+dependent resources have been attempted to be loaded (e.g., via RDS). The
+existing resource loading timeout in XdsClient prevents the update from being
+delayed indefinitely and the duplicate resource update detection in XdsClient
+prevents replacing the Listener when nothing changes. The grace period should be
+adjustable when building the XdsServer and should be described as the "drain
+grace time."
+
 [Filter.typed_config]: https://github.com/envoyproxy/envoy/blob/928a62b7a12c4d87ce215a7c4ebd376f69c2e080/api/envoy/config/listener/v3/listener_components.proto#L40
 [TypedStruct.type_url]: https://github.com/cncf/udpa/blob/cc1b757b3eddccaaaf0743cbb107742bb7e3ee4f/udpa/type/v1/typed_struct.proto#L38
+[A28-validation]: A28-xds-traffic-splitting-and-routing.md#response-validation
```
markdroth, line 232: Would it make sense to add a comment in A28 that points to this gRFC, so that people reading that one know it's been updated by this one?  (I'm thinking something like the "updated by:" field in RFCs.)

markdroth (commented):
Thanks for doing this!

A36-xds-for-servers.md
```diff
@@ -192,8 +190,48 @@ required. This is to allow implementations that only support L4 xDS features to
...
+as RDS does not distinguish between client-side and server-side. That is
+predomenently defined in [gRFC A28][A28-validation], although note that
+configuration for all VirtualHosts have been validated on client-side since
+sharing the XdsClient was introduced, yet was not documented in a gRFC. The
+validation must be updated to allow [Route.non_forwarding_action][] as a valid
+`action`. The VirtualHost is selected on a per-RPC basis using the RPC's
+requested `:authority`. Routes are matched the same as on client-side.
+`Route.non_forwarding_action` is expected for all Routes used on server-side and
+`Route.route` continues to be expected for all Routes used on client-side; a
+Route with an inappropriate `action` causes RPCs matching that route to fail
```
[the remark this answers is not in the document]
ejona86: Updated as discussed in TL meeting. We will not use TRANSIENT_FAILURE and instead fail RPCs with UNAVAILABLE the same on client-side and server-side. I've also documented that filters should be executed in such a case, but we're not worried about C doing otherwise on client-side for the while.

markdroth (approved):
Looks great!

PR #256, opened 2021-08-25
G2: gRPC over HTTP/3

JamesNK:
A specification for gRPC over HTTP/3.

There is a lot of overlap with the content in https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md. For example, request and response headers are unchanged between HTTP/2 and HTTP/3. Rather than duplicate content, the proposal focuses on the differences between HTTP/2 and HTTP/3.

Preview: https://github.com/JamesNK/proposal/blob/jamesnk/http3/G2-http3-protocol.md

### DIFF
--- a/G2-http3-protocol.md
+++ b/G2-http3-protocol.md
@@ -98,7 +98,7 @@ n/a|INADEQUATE_SECURITY| PERMISSION_DENIED … with additional detail indicating
 `GOAWAY` and `PING` frames exist in HTTP/3 and serve the same purpose as in HTTP/2.
 One notable difference is the `GOAWAY` frame in HTTP/2 reports the last
 successfully processed stream ID. In HTTP/3 the `GOAWAY` frame ID value must be greater
-that the last successfully processed stream ID.
+than the last successfully processed stream ID.
 
 ### Exceeding deadlines
 
--- a/G2-http3-protocol.md
+++ b/G2-http3-protocol.md
@@ -1,8 +1,8 @@
 G2: gRPC over HTTP/3
 ----
 * Author(s): [James Newton-King](https://github.com/jamesnk)
-* Approver: 
-* Status: Draft
+* Approver: ejona86
+* Status: In review
 * Implemented in: grpc-dotnet
 * Last updated: 2021-08-25
 * Discussion at: 
@@ -24,10 +24,10 @@ offered by HTTP/3:
 * Improved experience when there is connection packet loss.
 * Client supports transitioning between networks.
 
-[PROTOCOL-HTTP2.md](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) covers gRPC's HTTP semantics, along with some content
-that is specific to HTTP/2. Either this document should be generalized to include
-HTTP/2 and HTTP/3 information, or a new gRPC over HTTP/3 protocol document should
-be created.
+[PROTOCOL-HTTP2.md](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)
+covers gRPC's HTTP semantics, along with some content that is specific to HTTP/2.
+Either this document should be generalized to include HTTP/2 and HTTP/3 information,
+or a new gRPC over HTTP/3 protocol document should be created.
 
 ### Related Proposals:
 
@@ -38,14 +38,16 @@ n/a
 ### HTTP semantics
 
 The shape of a gRPC HTTP requests and responses are unchanged in HTTP/3. The
-content in [PROTOCOL-HTTP2.md](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) that covers requests and responses can be referred
-to directly, without any duplication.
+content in [PROTOCOL-HTTP2.md](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)
+that covers requests and responses can be referred to directly, without any
+duplication.
 
 Notably, unlike gRPC-Web, the content-type of `application/grpc` is unchanged.
 
 ### Transport mapping
 
-The gRPC over HTTP/2 specification [discusses HTTP2 transport mapping](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#http2-transport-mapping). The content discussed is mostly applicable to HTTP/3.
+The gRPC over HTTP/2 specification [discusses HTTP2 transport mapping](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#http2-transport-mapping).
+The content discussed is mostly applicable to HTTP/3.
 
 #### Stream Identification
 
@@ -57,7 +59,8 @@ The relationship between `DATA` frames and length prefixed messages are unchange
 
 #### Errors
 
-HTTP/3 [has different error codes](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-8.1) from HTTP/2. HTTP/3 errors are sent via QUIC frames instead of an `RST_STREAM` HTTP frame.
+HTTP/3 [has different error codes](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-8.1)
+from HTTP/2. HTTP/3 errors are sent via QUIC frames instead of an `RST_STREAM` HTTP frame.
 
 HTTP/3 errors are used in three situations:
 
--- a/G2-http3-protocol.md
+++ b/G2-http3-protocol.md
@@ -122,6 +122,23 @@ appropriate for a deadline exceeded, because both stream directions should be ab
 However, `STOP_SENDING` should be sent along with `RESET_STREAM` if the deadline is exceeded
 while the request side is in-progress.
 
+### HTTP/3 negotation
+
+A client and server both need to support HTTP/3 for it to be used. There are two common
+scenarios for establishing an HTTP/3 connection:
+
+* gRPC client uses TLS and is configured in advance to only use HTTP/3. The client starts a
+  QUIC+HTTP/3 connection to server. If the server doesn't support HTTP/3 then the connection
+  fails.
+* gRPC client uses TLS and is configured to use HTTP/2 or greater. In this case an initial
+  HTTP/2 call is made, server returns response with an `alt-svc` header that tells the client
+  that HTTP/3 is available. HTTP/2 connection is replaced with QUIC+HTTP/3 and future calls
+  use HTTP/3. If the server doesn't support HTTP/3 then the client continues to use HTTP/2.
+
+HTTP/3 requires TLS and will never be used by an insecure channel.
+
+API for configuring channel HTTP negotation behavior is up to implementations.
+
 ## Rationale
 
 HTTP/3 is an upcoming Internet Protocol. There needs to be a standardized agreement for how
--- a/G2-http3-protocol.md
+++ b/G2-http3-protocol.md
@@ -5,7 +5,7 @@ G2: gRPC over HTTP/3
 * Status: In review
 * Implemented in: grpc-dotnet
 * Last updated: 2021-08-25
-* Discussion at: 
+* Discussion at: https://groups.google.com/g/grpc-io/c/b1dWReBGyX4
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


G2-http3-protocol.md
```diff
@@ -0,0 +1,142 @@
...
+
+`GOAWAY` and `PING` frames exist in HTTP/3 and serve the same purpose as in HTTP/2.
+One notable difference is the `GOAWAY` frame in HTTP/2 reports the last
+successfully processed stream ID. In HTTP/3 the `GOAWAY` frame ID value must be greater
+that the last successfully processed stream ID.
+
+### Exceeding deadlines
+
+When an RPC has exceeded its deadline in HTTP/2, the server will complete the response
+with a gRPC status code of DEADLINE_EXCEEDED, then send a `RST_STREAM` frame to the
```
ejona86, line 102: Actually, we generally use a `RST_STREAM(CANCEL)` for server→client direction when the deadline expires as it is cheap and avoids flow control so any buffered data get thrown away. We generally expect that doesn't matter much because `RST_STREAM(CANCEL)` for client→server is racing with it. When our client implementations see a CANCELLED status they double-check whether the deadline has been exceeded, and if so replace the status and trailers with DEADLINE_EXCEEDED.

What you describe here is a fine server behavior, because, again, what the server does here shouldn't matter except for a race and it will resolve fine in a race. But it does keep the stream alive longer.

G2-http3-protocol.md
```diff
@@ -0,0 +1,142 @@
...
+
+The `RST_STREAM` frame doesn't exist in HTTP/3. QUIC supports stopping the reading and sending parts of a QUIC
+stream independently. Two QUIC frames have replaced `RST_STREAM` functionality:
+
+* [RESET_STREAM](https://www.rfc-editor.org/rfc/rfc9000.html#name-reset_stream-frames) - Abruptly terminate the sending part of a stream.
+* [STOP_SENDING](https://www.rfc-editor.org/rfc/rfc9000.html#name-stop_sending-frames) - Requests that a peer cease transmission on a stream.
+
+`RESET_STREAM` is not appropriate because the response with the DEADLINE_EXCEEDED status
+code should reach the client. Because QUIC frames can be received out-of-order,
+sending this QUIC frame after completing the response can cause the client
```
ejona86, line 114: Why would this be sent after completing the response? If the server has completed the response, the RPC is over and the deadline no longer matters.

G2-http3-protocol.md
```diff
@@ -0,0 +1,142 @@
...
+
+`STOP_SENDING` should be sent to the client by a gRPC server after it completes the response.
+It accurately communicates to the client that it should stop sending RPC messages.
+The application error code should be H3_NO_ERROR(0x0100).
+
+## Rationale
+
+HTTP/3 is a [highly requested feature](https://github.com/grpc/grpc/issues/19126) for gRPC.
+
+HTTP/3's benefits include:
```
ejona86, line 125: These bullets are just repeating the background. Remove them?

Rationale is really about "why did you make decision X instead of option Y." The highly requested issue and the mention of standardization is fine here, but normally I'd put it in background. (Background provides context and can explain why we are doing this feature.)

G2-http3-protocol.md
JamesNK: I agree. I'm explaining here why RESET_STREAM is the wrong thing to do.

I'm going into detail here because on the surface RESET_STREAM is the first frame someone might try to use as a replacement for HTTP/2's RST_STREAM.

ejona86: I was saying the construction made no sense to me. There is never any point that you'd "send this QUIC frame after completing the response." "Don't do that, but still use RESET_STREAM." "If a deadline occurs, use RESET_STREAM" and yes, that "can cause the client to abort the request" but the client has probably _already_ aborted the request.

JamesNK: FYI while you were replying I edit my comment to clarify what I meant by completing a response 😄

ejona86: I don't understand why RESET_STREAM is the wrong thing to use. It seems _at least_ RESET_STREAM should be used, and if the client hasn't half-closed yet then STOP_STREAM as well. That is [even mentioned in the spec](https://www.rfc-editor.org/rfc/rfc9000.html#section-3.5-8):

> An endpoint that wishes to terminate both directions of a bidirectional stream can terminate one direction by sending a RESET_STREAM frame, and it can encourage prompt termination in the opposite direction by sending a STOP_SENDING frame.

If you have to do the `grpc-status: DEADLINE_EXCEEDED` approach, then really we'd also want to allow a truncated message if the status code is not OK. But I don't see what HTTP/2 vs HTTP/3 (or even HTTP/1.1) has to do with that; the reasoning for that approach seems the same in each.

JamesNK: I thought with an exceeded deadline situation that it was best practice was for a server to gracefully complete the response by sending trailers with a grpc-status of DEADLINE_EXCEEDED, and then send RST_STREAM. That means if there is a race condition where the response is received before the client timer exceeded the deadline then the correct status (DEADLINE_EXCEEDED) is returned to the user.

If the server sends RESET_STREAM (NO_ERROR) then the client gRPC library will throw an error, and if the client's deadline timer hasn't triggered then they'll probably get an INTERNAL_ERROR status.

ejona86: > I thought with an exceeded deadline situation that it was best practice was for a server to gracefully complete the response by sending trailers with a grpc-status of DEADLINE_EXCEEDED, and then send RST_STREAM

grpc-status closure, followed by RST_STREAM(OK) if client hasn't half-closed, is a thing for normal errors. But we don't use it for when the deadline expires. Instead we use RST_STREAM(CANCEL). C/Go/[Java then have code on the client-side](https://github.com/grpc/grpc-java/blob/4fa612ae3d8c3d38e84184f633fc4fbf02b93a35/core/src/main/java/io/grpc/internal/ClientCallImpl.java#L691-L703) to handle the processing race. And yes, I'd believe that isn't actually documented anywhere. Although truncating the response is mentioned as part of RST_STREAM(CANCEL).

For HTTP/3, the equivalent seems to be RESET_STREAM(H3_REQUEST_CANCELLED) (followed by normal handling if client hasn't half-closed). Not that I know what I'm talking about with H3. I've read like... 5 paragraphs of the spec at this point.

ejona86: I guess I should ask: when the deadline expires and you want to send `grpc-status`, how are you avoiding truncating a message in grpc-dotnet? In Java and Go, the grpc framing is pretty near the application and the network thread just sees raw HTTP body bytes so has no way to synchronize to a message boundary.

JamesNK: I don't recall off the top of my head. I'd need to check.

> grpc-status closure, followed by RST_STREAM(OK) if client hasn't half-closed, is a thing for normal errors. But we don't use it for when the deadline expires. Instead we use RST_STREAM(CANCEL). C/Go/Java then have code on the client-side to handle the processing race.

Hmmm, I didn't know that! Looks like grpc-dotnet has been doing deadlines quite differently.

No one has brought up this deadline race as an issue when using the grpc-dotnet client with other implementations. I would have expected to see them if the server is sending back the response you described. Perhaps I added some protection like that. It's been a couple of years since I looked closely at deadline handling logic.

> For HTTP/3, the equivalent seems to be RESET_STREAM(H3_REQUEST_CANCELLED) (followed by normal handling if client hasn't half-closed).

Yes, that matches your description of HTTP/2 deadline behavior.

ejona86: Created https://github.com/grpc/grpc/pull/27216 to improve HTTP/2 documentation. I think what I'm adding there will make the HTTP/3 behaviors to document here more obvious.

biels (approved):

G2-http3-protocol.md
JamesNK: Fixed

G2-http3-protocol.md
JamesNK: Fixed.

Also, I made improvements to the grpc-dotnet client so that RST_STREAM(CANCEL) is mapped to DEADLINE_EXCEEDED if received when the client's deadline has been exceeded.

G2-http3-protocol.md
```diff
@@ -0,0 +1,137 @@
...
+
+HTTP/3 is an upcoming Internet Protocol. There needs to be a standardized agreement for how
+gRPC over HTTP/3 works to maintain interoperability in the gRPC eco-system.
+
+## Implementation
+
+A trial implementation in grpc-dotnet is underway. .NET 6 is adding preview support for
+HTTP/3 to the .NET server and client. grpc-dotnet leverages that underlying HTTP/3 support.
+
+## Open issues (if applicable)
```
sanjaypujare, line 135: QUIC always uses security (TLS) which means gRPC-over-http3 always requires certs/keys IIUC. Does it impact the gRPC API or the usage? To be addressed in this gRFC?

JamesNK: HTTP/3 would never be negotated over an insecure connection.

Scenarios would be:
1. gRPC client uses prenegotated H2C over insecure connection
2. gRPC client uses TLS and requests only HTTP/2. Ignores `alt-svc` header returned from the server.
3. gRPC client uses TLS and requests HTTP/2 or greater. In this case an initial HTTP/2 call is made, server returns response with an `alt-svc` header that tells the client that HTTP/3 is available. HTTP/2 connection is replaced with QUIC+HTTP/3 and future calls use HTTP/3.
4. gRPC client uses TLS and is configured in advance to only do HTTP/3. QUIC+HTTP/3 connection is made to server.

larspaulsen:
Hi James,

Do you happen to know how gRPC Servers or Client e.g. SDK kind of pluming/wrapper code get notified of transport errors for any of the following types of streaming gRPC connections (service interface call/functions), other than implementing their own heartbeat timeout mechanism?:

1. Server Streaming.
2. Client Streaming.
3. Bi-directional Streaming.

?

Thank you,

-lars

G2-http3-protocol.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Rationale
+
+HTTP/3 is an upcoming Internet Protocol. There needs to be a standardized agreement for how
+gRPC over HTTP/3 works to maintain interoperability in the gRPC eco-system.
+
+## Implementation
+
+A trial implementation in grpc-dotnet is underway. .NET 6 is adding preview support for
+HTTP/3 to the .NET server and client. grpc-dotnet leverages that underlying HTTP/3 support.
```
clayg, line 133: @lidizheng what about doing some python asyncio prototyping!?  https://github.com/aiortc/aioquic

lidizheng: I understand that building a MVP gRPC Python HTTP/3 on top of the `aioquic` library probably isn't that hard. Currently, gRPC Python uses gRPC Core as the transport layer, it's not that trivial to move existing code onto another transport library. I hope if we want to invest resources, we have a working architecture and aiming for long-term support.

ejona86:
@JamesNK, is there anything in this PR you'd like to change before I merge it? It looks like there's only one typo to address from my perspective.

G2-http3-protocol.md
```diff
@@ -0,0 +1,137 @@
...
+n/a|STREAM_CLOSED|No mapping as there is no open stream to propagate to. Implementations should log.
+n/a|COMPRESSION_ERROR|INTERNAL
+n/a|INADEQUATE_SECURITY| PERMISSION_DENIED … with additional detail indicating that permission was denied as protocol is not secure enough for call.
+
+#### Connection management
+
+`GOAWAY` and `PING` frames exist in HTTP/3 and serve the same purpose as in HTTP/2.
+One notable difference is the `GOAWAY` frame in HTTP/2 reports the last
+successfully processed stream ID. In HTTP/3 the `GOAWAY` frame ID value must be greater
+that the last successfully processed stream ID.
```
JamesNK, line 101:
```suggestion
than the last successfully processed stream ID.
```

JamesNK:
> @JamesNK, is there anything in this PR you'd like to change before I merge it? It looks like there's only one typo to address from my perspective.

Fixed typo. I also added a section on HTTP/3 negotiation in reply to @sanjaypujare's comment.

ejona86 (approved):

PR #257, opened 2021-08-27
L84: Add grpc_call_failed_before_recv_message() to C-core API

ananda1066:
Proposal to move `grpc_call_failed_before_recv_message` from subsurface to the C-core surface API.

### DIFF
--- a/L81-cpp-call-failed-before-recv-message.md
+++ b/L81-cpp-call-failed-before-recv-message.md
@@ -5,7 +5,7 @@ Title
 * Status: Draft
 * Implemented in: C++
 * Last updated: 2021-08-27
-* Discussion at: TBD
+* Discussion at: https://groups.google.com/g/grpc-io/c/UNBHGr6feeY
 
 ## Abstract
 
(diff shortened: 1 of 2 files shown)


ctiller (dismissed):

L81-cpp-call-failed-before-recv-message.md
```diff
@@ -0,0 +1,38 @@
+Title
```
markdroth, line 1: Please rename this file to L84-core-call-failed-before-recv-message.md and change this line to `L84: Add grpc_call_failed_before_recv_message() to C-core API`.

markdroth (commented):
I'll let Craig review, but this LGTM overall.

Note that the number L81 was already taken.  (You generally have to look at pending PRs in addition to the files that already exist in the repo.)

L81-cpp-call-failed-before-recv-message.md
ananda1066: Ah done, thanks!

ctiller (approved):

PR #259, opened 2021-09-03
L85: Native cross-process communication on Android using BinderChannel

sifmelcara:
Preview: https://github.com/sifmelcara/proposal/blob/core-binder-transport/L85-core-binder-transport.md

### DIFF
--- a/L85-core-binder-transport.md
+++ b/L85-core-binder-transport.md
@@ -6,13 +6,13 @@
 *   Approver: TBD
 *   Status: Draft
 *   Implemented in: gRPC-core, C++
-*   Last updated: August 31, 2021
+*   Last updated: September 30, 2021
 *   Discussion at: https://groups.google.com/g/grpc-io/c/n8Qhlfwd8tE
 
 ## Abstract
 
-Implements [BinderChannel protocol](proposal L73) in gRPC-core transport layer,
-and provide C++ interface for Android native code.
+Implements [BinderChannel protocol](proposal L73) in gRPC-core, and provide C++
+interface for Android native code.
 
 ## Background
 
@@ -44,28 +44,24 @@ After the connection is established, the rest can be done in pure C++.
 
 ### Client side API
 
-Since the channel creation process requires interactions with Android runtime,
-we will have a special API for creating binder transport channels.
+Since we need more than a simple URI string to create the channel, we will have
+separate APIs for creating binder transport channels.
 
 ```cpp
-void CreateBinderChannel(
+std::shared_ptr<grpc::Channel> CreateBinderChannel(
     void* jni_env, jobject context, const std::string& package_name,
     const std::string& class_name,
-    std::shared_ptr<grpc::binder::SecurityPolicy> security_policy,
-    std::function<void(void*, std::shared_ptr<grpc::Channel>)>
-        on_channel_created);
+    std::shared_ptr<grpc::binder::SecurityPolicy> security_policy);
 
-void CreateCustomBinderChannel(
+std::shared_ptr<grpc::Channel> CreateCustomBinderChannel(
     void* jni_env, jobject context, const std::string& package_name,
     const std::string& class_name,
     std::shared_ptr<grpc::binder::SecurityPolicy> security_policy,
-    const grpc::ChannelArguments& args,
-    std::function<void(void*, std::shared_ptr<grpc::Channel>)>
-        on_channel_created);
+    const grpc::ChannelArguments& args);
 
 class SecurityPolicy {
  public:
-  // return true if the uid is authorized to connect
+  // returns true if the UID is authorized to connect
   virtual bool IsAuthorized(int uid) = 0;
 };
 ```
@@ -74,7 +70,8 @@ class SecurityPolicy {
 Java code to establish the connection.
 
 `context`: The service will only be considered required by this client for as
-long as the `context` exists (See [bindService] for details).
+long as the `context` exists (See [bindService] for details). This object should
+be a `android.content.Context`.
 
 `package_name` and `class_name`: These 2 arguments will be used to create a
 [ComponentName], which identifies a specific service to connect to.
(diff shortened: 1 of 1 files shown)


sifmelcara:
cc @markdroth 
cc @ctiller 
cc @markb74 
cc @ejona86 
cc @TaWeiTu
cc @napathome

L85-core-binder-transport.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Server side API
+
+Service implementation will be an Android [bound service] in Java, and it should return
+a binder object created by our transport implementation to the client.
+
+We propose to introduce a new API that returns credentials for binder server:
+
+```Cpp
+std::shared_ptr<ServerCredentials> BinderServerCredentials(std::shared_ptr<grpc::binder::SecurityPolicy> security_policy);
```
TaWeiTu, line 101: We should include the C API `grpc_server_add_binder_port` into this RFC per discussions made in https://github.com/grpc/grpc/pull/27036.

sifmelcara: Mark and Craig: Can you confirm if we only need C API for this specific API are we need C API for every C++ API? What's the reason to introduce C API?

L85-core-binder-transport.md
```diff
@@ -0,0 +1,181 @@
...
+    void* jni_env, jobject context, const std::string& package_name,
+    const std::string& class_name,
+    std::shared_ptr<grpc::binder::SecurityPolicy> security_policy,
+    std::function<void(void*, std::shared_ptr<grpc::Channel>)>
+        on_channel_created);
+
+class SecurityPolicy {
+ public:
+  // return true if the uid is authorized to connect
+  virtual bool IsAuthorized(void* jni_env, int uid) = 0;
```
beshaya, line 60: To my knowledge, there is no way to get access to PackageManager in NDK, so it would be easier on implementing teams if this function included the package name and signing certificate(s) to the c++ client; otherwise clients trying to implement security will need to include extra java code to call into PackageManager.

Alternatively, including a utility function to get the package name + signing certificate from c++ code could be used.

L85-core-binder-transport.md
```diff
@@ -0,0 +1,181 @@
...
+object in `Parcel`, we will print an error message and fail the RPC call.
+
+### Error handling
+
+We will mark the API as experimental for now and experiment it in real world use
+cases. After that we will integrate proper error handling into our interface.
+
+### Android API level requirement
+
+For Java interoperability, we will require the device to have API level 32
```
beshaya, line 166: What features require API level 32? It seems unfortunate that this would rule out use of BinderChannel on all existing Android devices until receiving an OS update. Is it possible to target a lower API level?

L85-core-binder-transport.md
```diff
@@ -0,0 +1,181 @@
...
+we want to prevent user from using the (not connected) channel immediately and
+wait for RPC call to success and causing deadlock.
+
+### Parcelable objects
+
+The protocol and Java implementation allows passing Parcelable objects. We
+currently don't have a plan to support it in C++. When we see a parcelable
+object in `Parcel`, we will print an error message and fail the RPC call.
+
+### Error handling
```
beshaya, line 159: Could you also make a note about what happens to the BinderChannel if the remote process is killed? Will the channel automatically try to reconnect the way a TCP or UDP channel does, or does the calling code need to manage reconnection?

sifmelcara: Added. Our Java class will receive notification when the connection is broken, and our code should be able to handle the connection recovery (and check authorization again if the server app is updated) for user.

Ref: https://developer.android.com/reference/android/content/ServiceConnection

L85-core-binder-transport.md
sifmelcara: [A new API for disabling header](https://cs.android.com/android/platform/superproject/+/master:frameworks/native/libs/binder/ndk/include_ndk/android/binder_ibinder.h;l=234?q=AIBinder_Class_disableInterfaceTokenHeader&ss=android%2Fplatform%2Fsuperproject)

> Is it possible to target a lower API level?

For C++ Server <--> C++ Client scenario it is possible to target API level 31

L85-core-binder-transport.md
sifmelcara: Our plan is to provide many possible implementations of SecurityPolicy that makes sense, so client unlikely need to customize this

Yeah we can include a C++ utility function to get package name + signing certificate so that user can check certificate in pure C++. (This is a little bit off topic, and the utility function not necessary to be part of gRPC I guess)

L85-core-binder-transport.md
```diff
@@ -0,0 +1,191 @@
...
+we establish the connection between server and client, the rest can be done in
+pure C++.
+
+### Client side API
+
+Since the channel creation process requires interactions with Android runtime,
+we will have a special API for creating binder transport channels.
+
+```cpp
+void CreateBinderChannel(
```
TaWeiTu, line 50: We should also let users provide `ChannelArguments` for custom channel creation. For example, add another API `CreateCustomBinderChannel()` which takes a `ChannelArguments` and forwards the calls to `CreateBinderChannel()` to `CreateCustomBinderChannel()` with the default `ChannelArguments`.

sifmelcara:
Hi Craig would you be able to take a look or advice us how to proceed? Thank you!

L85-core-binder-transport.md
```diff
@@ -0,0 +1,206 @@
...
+connection (since APIs for setting up connections are only available in Java).
+After the connection is established, the rest can be done in pure C++.
+
+### Client side API
+
+Since we need more than a simple URI string to create the channel, we will have
+separate APIs for creating binder transport channels.
+
+```cpp
+std::shared_ptr<grpc::Channel> CreateBinderChannel(
```
markdroth, line 51: Do you actually need a special channel creation function?  Could you instead pass `jni_env` and `context` via channel args, `package_name` and `class_name` via the URI, and `security_policy` via a new credential type?

sifmelcara: `jni_env` and `context` is mandatory for establishing the connection and they are pointers to special JVM structure. I'm not sure about the convention and assumed that `channel_args` is mainly for specifying configurations and might not be suitable for this use case. `jni_env` and `context` are non-optional and they don't feel like "channel arguments" since they are basically only used for establishing the connection and is not related to `grpc::Channel` logic.

For `package_name` and `class_name` it is possible to encode them into URI but it is simpler to not include them, since 1. `package_name` and `class_name` are also only used single-time when establishing the connection (in Java). 2. `package_name` and `class_name` is not the only way to establish the connection. For example, in the future we might support creating channel from a pointer to binder, in that case it is similar to `grpc_insecure_channel_create_from_fd` and URI does not make too much sense for that case.

For `security_policy` I think it cannot be easily implemented with gRPC's credential design (I might be wrong). gRPC credential is about writing/reading some kind of token/signature in wire data. However our security policies are using system calls to verify if the connection is allowed and has nothing to do with wire data or headers.

So in the end I decided that it is easier (for users and for us) to introduce a separate interfaces instead of squeezing everything into existing `CreateChannel` interface.

PR #260, opened 2021-09-03
A46: xDS NACK Semantics Improvement

markdroth:


ejona86 (approved):
Beautiful. That was well documented.

A46-xds-nack-semantics-improvement.md
```diff
@@ -0,0 +1,119 @@
...
+  to `ROUND_ROBIN` instead of considering the resource invalid.
+  However, this is not semantically correct behavior; it could cause
+  unexpected load on servers and a large number of unnecessary connections
+  to backends.  Also, while it would address the specific case that
+  triggered this design, it would not address the general case of seeing
+  an unsupported value in a supported field.
+
+## Implementation
+
+[A description of the steps in the implementation, who will do them, and when.  If a particular language is going to get the implementation first, this section should list the proposed order.]
```
dfawley, line 115: TBD? Remove?

dfawley (approved):
A++++++++++++++++++ would read again.

murgatroid99:
Should a NACK sent in this situation indicate which resources are rejected vs accepted in a machine-readable way (i.e. other than in the error message)?

A46-xds-nack-semantics-improvement.md
```diff
@@ -0,0 +1,119 @@
...
+
+Note that gRPC's original behavior was intended to make sure that the
+control plane had a clear picture of what configuration the client is
+using.  However, it turns out that Envoy currently has cases where it
+will apply some valid resources from a response that it NACKs, which
+means that the control plane *already* did not have that kind of clear
+picture.  To ensure that the control plane will have that kind of clear
+picture, there will be subsequent work to extend the xDS protocol to
+allow per-resource NACKing instead of per-response NACKing, although
+that will be a broader project.
```
srini100, line 95: You could add that the CSDS service in a gRPC client can be used to get the current state of of the resources as a workaround to this control plane limitation.

markdroth:
@murgatroid99 
> Should a NACK sent in this situation indicate which resources are rejected vs accepted in a machine-readable way (i.e. other than in the error message)?

There's no way to do that today.  That's what I was referring to in the doc by saying that there will be some future work to change the xDS protocol to make that possible.

A46-xds-nack-semantics-improvement.md
markdroth: Good idea.  Done.

A46-xds-nack-semantics-improvement.md
markdroth: I'll update this section once I have an implementation PR for C-core out for review.

srini100 (approved):
Looks great! Thanks for taking care of this.

howardjohn (approved):

A46-xds-nack-semantics-improvement.md
```diff
@@ -0,0 +1,122 @@
...
+using.  However, it turns out that Envoy currently has cases where it
+will apply some valid resources from a response that it NACKs, which
+means that the control plane *already* did not have that kind of clear
+picture.  To ensure that the control plane will have that kind of clear
+picture, there will be subsequent work to extend the xDS protocol to
+allow per-resource NACKing instead of per-response NACKing, although
+that will be a broader project.  Note that the CSDS service in the gRPC
+xDS client (see [gRFC A40](A40-csds-support.md)) can be used to get the
+current state of the resources in the client, which is likely to be more
+accurate than the view fromthe xDS server.
```
dapengzhang0, line 98: from the

markdroth: Thanks, fixed.

A46-xds-nack-semantics-improvement.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Related Proposals: 
+* [gRFC A27: xDS-Based Global Load Balancing](A27-xds-global-load-balancing.md)
+* [gRFC A42: xDS Ring Hash LB Policy](A42-xds-ring-hash-lb-policy.md)
+
+## Proposal
+
+We will change gRPC's behavior such that when a response is NACKed, gRPC
+will still use all valid resources from the response; it will ignore
+only the invalid resources.
```
dapengzhang0, line 72: If the list of resources is a prioritized list or a filter chain, and one of the resources is invalid, do we consider the whole priority list or filter chain valid?

markdroth: In the context of this document, a "resource" means only a complete xDS-level config resource -- i.e., Listener, RouteConfiguration, Cluster, or ClusterLoadAssignment.  The existing rules for validating those resources tell us whether an individual resource is valid, so those rules would cover things like how to interpret a priority list or filter chain.

dapengzhang0: Got it. This doc is only for "top-level" xDS resources.

A46-xds-nack-semantics-improvement.md
```diff
@@ -0,0 +1,122 @@
...
+This document proposes a behavior change to address this problem.
+
+### Related Proposals: 
+* [gRFC A27: xDS-Based Global Load Balancing](A27-xds-global-load-balancing.md)
+* [gRFC A42: xDS Ring Hash LB Policy](A42-xds-ring-hash-lb-policy.md)
+
+## Proposal
+
+We will change gRPC's behavior such that when a response is NACKed, gRPC
+will still use all valid resources from the response; it will ignore
```
menghanl, line 71: How do we update CSDS?

The invalid resources can be updated to the most recent version, and NACK

How the valid resources?
- Do we also update them to the most recent version, and NACK
- Or keep the previous version and ACK
- Or the most recent version and ACK???

markdroth: In CSDS, we report status on a per-resource basis, so it's fine for each resource to be in a different state.  The invalid resources will show status NACKED, and the valid resources can show ACKED with the new version.

A46-xds-nack-semantics-improvement.md
```diff
@@ -0,0 +1,122 @@
...
+LB policy, as per [gRFC A27](A27-xds-global-load-balancing.md),
+but newer clients now support the `RING_HASH` policy, as per [gRFC
+A42](A42-xds-ring-hash-lb-policy.md).  If a Cluster resource specifies
+an unsupported LB policy, clients will consider the resource invalid,
+which will cause them to NACK the response.  So if an xDS server cannot
+be sure that all of its clients have been upgraded to a version that
+supports the `RING_HASH` policy, then it cannot safely send a Cluster
+resource configuring that policy, because that change would cause older
+clients to stop functioning.
+
```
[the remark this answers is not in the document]
markdroth: We still don't actually have a design for version negotiation, so that's not a solution for the immediate problem.  And even once we do have such a design, I suspect that that's not going to be the best way to address this kind of use-case, because it would require the control plane to have a bunch of logic to determine how to configure older clients that don't support the current configuration.

I think having the control plane generate different versions of the config will make sense for cases where we are deprecating one set of fields and replacing them with another set, where the semantic meaning of the underlying config can be expressed equally well via both sets of fields.  But I think it will be too complex to do in cases where the client simply does not support the features that the configuration is intending to configure, because I don't think the control plane can really know what to do in that case.

A46-xds-nack-semantics-improvement.md
dapengzhang0: What if an invalid resource is so broken that is doesn't have a resource name to update CSDS, e.g. the ANY type message of that particular resource can not be parsed?

dapengzhang0: > What if an invalid resource is so broken that it doesn't have a resource name to update CSDS

Oh, I think in that case we don't update CSDS for that resource even without the change.

markdroth: Yes, that's correct.  If we can't even find the resource's name, we have no way to report the problem via CSDS, but that's not affected by this proposal.

PR #264, opened 2021-09-09
A36 Core/C++ API update : Use a struct for serving status

yashykt:
A36 Core/C++ API update: Use a struct for serving status to allow more fields to be added in the future


markdroth (approved):
Thanks for remembering to update this!

ejona86 (approved):

PR #266, opened 2021-10-06
A39 update: Require that the HTTP filter list be non-empty.

markdroth:
Envoy doesn't actually seem to have a validation check for this, but I suspect that's just an oversight on their part, because I don't see any way that the HCM can work without a terminal filter (and there's only one such filter, which is the router filter).

### DIFF
--- a/A39-xds-http-filters.md
+++ b/A39-xds-http-filters.md
@@ -141,6 +141,8 @@ filters.  In each entry, gRPC will look at the following fields:
 
 When gRPC receives the `HttpConnectionManager` proto in the LDS
 response, it will validate the list of filters as follows:
+- There must be at least one filter in the list, or the Listener
+  resource will be NACKed.
 - Every filter in the list must have a unique instance name.  Any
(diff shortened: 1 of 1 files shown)


ejona86:
CC @YifeiZhuang 

Looks like Java is already prohibiting empty http filter list:
https://github.com/grpc/grpc-java/blob/fc57cad4ec7774eefd9ae4990f57805da84f611f/xds/src/main/java/io/grpc/xds/ClientXdsClient.java#L757

I'll note that gRFC A36 xDS for Servers already has sufficient language (and the Java implementation [looks good](https://github.com/grpc/grpc-java/blob/fc57cad4ec7774eefd9ae4990f57805da84f611f/xds/src/main/java/io/grpc/xds/ClientXdsClient.java#L333)):
> Additionally, the FilterChain is only valid if filters contains exactly one entry for the envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager filter, HttpConnectionManager hereafter, as the last entry in the list.

yashykt (approved):

PR #268, opened 2021-10-20
A47: xDS Federation

markdroth:


A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+into them is one of the servers that is already configured in the
+bootstrap file (i.e., it matches server name, channel creds, and server
+features).
+
+### Temporary environment variable protection
+
+While this feature is in development (until it has passed interop
+testing and is deemed production-ready), parsing of the new bootstrap
+config fields will be disabled unless the
+`GRPC_EXPERIMENTAL_XDS_FEDERATION` environment variable is set to true.
```
menghanl, line 659: Should we use `GRPC_XDS_EXPERIMENTAL_FEDERATION`, just to be consistent with the other env vars?  (swap `XDS` and `EXPERIMENTAL`)

e.g. RBAC is `GRPC_XDS_EXPERIMENTAL_RBAC` https://github.com/grpc/proposal/blob/master/A41-xds-rbac.md#temporary-environment-variable-protection

A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+  glob collections.  gRPC will eventually need this for scalability
+  reasons for [LEDS](https://github.com/envoyproxy/envoy/issues/10373),
+  but it's not necessary today, so we are decoupling it in order to
+  reduce the scope of this design.
+- Support for the `alt` directive.  We do not currently have a use-case
+  for which this is required in gRPC.
+- xDS redirects.  We do not currently have a use-case for which this is
+  required in gRPC.
+- Support for dynamic parameters.  This will be necessary in the
+  not-too-distance future, but it will be covered in a separate design.
```
sanjaypujare, line 53: "not-too-distant"

A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+// Authorities map.
+"authorities": {
+  "xds.example.com": {
+  }
+}
+```
+
+If a gRPC server is created listening on `0.0.0.0:8080`:
+- The server Listener resource name will be
+  `xdstp://xds.example.com/envoy.config.listener.v3.Listener/grpc/server/0.0.0.0:8080`.
```
sanjaypujare, line 300: Should this string have the second `:` percent-encoded?

A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+"client_default_listener_resource_name_template": <string>,
+```
+
+There will be two changes to the existing
+`server_listener_resource_name_template` described in [gRFC A36][A36]:
+- If the template starts with `xdstp:`, it will be interpreted as a
+  new-style name, in which case the authority of the URI will be used to
+  select the relevant configuration in the `authorities` map.
+- When replacing the `%s` token, if the template starts with `xdstp:`,
+  the replacement string will be %-encoded in the resulting resource
```
sanjaypujare, line 192: The doc uses both %-encoding or percent-encoding. Will be good to stick to just one term.

A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+### Temporary environment variable protection
+
+While this feature is in development (until it has passed interop
+testing and is deemed production-ready), parsing of the new bootstrap
+config fields will be disabled unless the
+`GRPC_EXPERIMENTAL_XDS_FEDERATION` environment variable is set to true.
+
+## Rationale
+
+For security reasons, we are focusing on statically configuring a clien
```
sanjaypujare, line 663: s/clien/client/

A47-xds-federation.md
markdroth: Fixed.

A47-xds-federation.md
markdroth: Changed to use "percent-encoding" everywhere.

A47-xds-federation.md
markdroth: No.  A colon does not need to be percent-encoded in the path part of the URI.

A47-xds-federation.md
markdroth: Fixed.

A47-xds-federation.md
markdroth: Let's try to use the `GRPC_EXPERIMENTAL_` prefix going forward.  That way, we can have the same convention for both xDS and non-xDS functionality.

A47-xds-federation.md
sanjaypujare: I was going by the text on lines 434-438: "... then it will percent-encode the listening address before replacing the %s token in the template." I thought it meant unconditionally percent-encode but you are now saying encode only if needed e.g. if `%s` is in the query parameter part.

A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+- The target URI has no authority, so gRPC will look at the bootstrap
+  config's `client_default_listener_resource_name_template` field, which
+  is not set, so it defaults to `%s`.  So the Listener resource name will be
+  `server.example.com`, just like it would be prior to this design.
+- The Listener resource name does not start with `xdstp:`, so we do not
+  use the `authorities` map at all.
+- We use the top-level `xds_server` list, which is the same one that we
+  would have used prior to this design.
+
+If a gRPC client channel is created for `xds://xds.example.com/server.example.com`:
```
sanjaypujare, line 231: Having "example.com" as the top-level-domain in both the authority and the path part has no significance other than it being a typical use-case, is that right? So this example could very well be `xds://xds.authority.com/server.example.com` . If true, wouldn't it be better to not use "example.com" in both places to avoid implying some sort of connection?

A47-xds-federation.md
markdroth: I don't think colon would need to be percent-encoded anywhere in the URI.  @ejona86 can correct me if I'm wrong.

A47-xds-federation.md
markdroth: Done.

A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+up-front and depend on idleness behavior, or they can lazily create the
+channel when it's actually needed.  Similarly, if a channel stops being
+used, implementations can either depend on idleness behavior to tear
+down the connections, or they can ref-count the channel and shut it down
+when it's no longer needed.  (The expected tear-down after being idle is
+O(5-15m).)
+
+##### Resource Cache
+
+The resource cache in the `XdsClient` is currently a two-level map, keyed
```
menghanl, line 464: What's the reason for the extra level? With the new style resource names, shouldn't every resource have a unique name?
We can keep the level for resource type, but mainly for the convenience, it won't be necessary anymore.

A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+first by resource type and then by resource name.  This will need to be
+changed to use a three-level map, keyed first by authority, then by
+resource type, and then by resource name.
+
+The top-level map in the cache is keyed by authority.  Resources without
+any authority will use the empty string as the key.  The value of this
```
menghanl, line 470: > Resources without any authority

The old-style resources don't have authority. And they use the top level server (same as before).

Is empty string a valid key for the bootstrap authority map?
If so, empty-string-authority can be different from no-authority.

A47-xds-federation.md
```diff
@@ -0,0 +1,682 @@
...
+  do not do any lookup in the `authorities` map.
+- We default to the top-level `xds_server` list, which is the same one
+  that we would have used prior to this design.
+
+#### Example: New-Style Names on gRPC Client
+
+The following new fields are added to the bootstrap config:
+
+```
+"client_default_listener_resource_name_template": "xdstp://xds.authority.com/envoy.config.listener.v3.Listener/%s",
```
menghanl, line 250: Do we check the type (`envoy.config.listener.v3.Listener` in this example) to make sure it's valid?
What if the users set this to `envoy.config.listener.v3.Cluster`? (So it's an LDS watch for a CDS resource?)
Do we fail? Or do we sent it as a LDS request, assuming the users know what they are doing?

---

Similarly, what if in an LDS response, the resource name's type is not Listener?
Do we NACK it?
Or we accept it and call the watcher?

markdroth: Good questions.  Let me answer the more general issue first, then I'll come back to the bootstrap file.

We should check the type on input, from both the `XdsClient` API and on the wire, and reject new-style resource names whose types do not match the type of resource we're told to look at:
- If a caller of the `XdsClient` API starts a watch on a resource of type (e.g.) Listener with a new-style name that indicates a different type, we should reject it.  In C-core, we're handling this by basically immediately reporting an error to the watcher and then never sending another update to it.  We don't actually subscribe to the resource on the wire.
- If we receive a resource on the wire in a response for resource type (e.g.) Listener that contains a new-name that indicates a different type, we NACK.  But that shouldn't actually happen in practice, since we should never subscribe to such a name in the first place.

Now, coming back to the question of the bootstrap file, I can think of arguments both for and against doing this validation.  Here are some of the factors to consider:
- The expansion of this template will be done outside of the `XdsClient` (on the client side, it will be done in the `xds` resolver; on the server side, it will be done in the xDS-specific server code), and the caller will then start a watch for the resulting resource name on the `XdsClient`.  So technically, we don't really need to do this validation at bootstrap-reading time, because the watcher will simply fail when we start a watch on the resulting resource name.  (Counter-argument: Surfacing the error earlier might be more friendly to users, if it turns out that this is a common mistake.)
- There may be cases in the future where it will be legitimate to have a different resource type here.  For example, we might someday want to be able to support pointing to an individual entry within a ListenerCollection resource (see xRFC TP1 for info on [list collection resources](https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md#list) and the [`#entry=<name>` fragment syntax for xdstp URIs](https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md#uri-based-xds-resource-names)).  (Counter-argument: If/when we ever decide to support this, we could change the validation we do at bootstrap parsing time.  But having older clients flat-out reject this might cause more of a problem for backward-compatibility.)
- In principal, it should be possible for the `%s` token to include the resource type, in which case we can't actually do this validation in the first place.  For example, it should be possible for the template to be `xdstp:%s` and for the client to create a channel for `xds:envoy.config.listener.v3.Listener/server.example.com`.  Or, there could be an even more esoteric case, like the template being `xdstp:envoy.config.listener.v3.%s` and the channel being created for `xds:Listener/server.example.com`.  (Counter-argument: I can't think of a reason for anyone to actually do this... although there could potentially be such a case in the future, if/when we support ListenerCollection resources.)

There's enough of a grey area here that my preference would be to say that we should *not* explicitly check this at bootstrap parsing time, but I'm open to being convinced otherwise if you or other reviewers feel otherwise.  WDYT?

A47-xds-federation.md
markdroth: Two reasons:
1. Memory efficiency.  It seems wasteful to replicate the same authority string many times in the cache.  (Note that this same argument applies to resource type.)
2. We need a place to store some per-authority data anyway, such as the server (or, eventually, list of servers) to use for that authority, and a ref to the server channel that we're currently using for it.

Also, note that the resource type level is still needed for old-style names, because a control plane could still use the same old-style resource name for multiple resource types.

A47-xds-federation.md
markdroth: Good catch!  Yes, this could be a problem in the case where the top-level server list differs from the server list in the `""` entry in the `authorities` map.  In that case, we should use the top-level server list for old-style resource names, but we should use the server list in the `""` entry for new-style names with no authority.  (So the difference here is actually old-style names vs. new-style names; for a new-style name, there is no difference between empty-string authority and no authority.)

I've updated the text here to clarify.

@donnadionne We didn't handle this case properly in grpc/grpc#27570.  Can you please put together a PR to fix this?  We can probably handle it by adding a prefix to the authority that we use in the map key, so that old-style resources use authority "old:" and new-style resources use authority "xdstp:<authority_name>" (where "<authority_name>" may be empty).

markdroth (commented):
Thanks for the comments, @menghanl -- these are useful discussion points to clarify!

A47-xds-federation.md
menghanl: So this is more for efficiency, not correctness, is that right?

I was thinking about making all the authorities share one cache, so all the resources are kept in one place (in one two-level map, not with the extra level of authority).
It simplifies CSDS a bit.
And one goal of the new-style resource name is to make it easier for caching, so I thought this should be a reasonable choice.

One complexity I got is for the SotW APIs (LDS and CDS), those needs per authority tracking to know which resource is removed. But for RDS/EDS, the implementation before federation should just work.

A47-xds-federation.md
menghanl: Not checking at bootstrap parsing time SGTM.

One note is that, even if we don't check at bootstrap parsing time, but we check at `client.Watch` time, it won't support any new types. So this decision doesn't affect backward-compability.

Not checking at bootstrap parsing time also has the advantage that it will fail only if the invalid template string is used. If it's set for an authority that's never picked, it won't cause any errors. (Or maybe this is a disadvantage?)

A47-xds-federation.md
markdroth: We still need a place to store the list of servers (and their associated configuration) to use for each authority, and we need a place to store a ref to the current channel being used for each authority.  If you don't store that information here, where will you store it?  Just have another map for it?

If you have another way of dealing with the above, then yes, it's "just" for efficiency, not correctness.  However, given the price of memory these days, I think the efficiency argument is actually a significant one.  If you just store the literal resource names as strings, then you're going to be duplicating the authority and resource type strings in memory many, many times.

> It simplifies CSDS a bit.

For us, using a multi-level map didn't make a big difference for CSDS.  We basically needed two nested loops instead of just one, but it wasn't really that much more complicated.

> And one goal of the new-style resource name is to make it easier for caching, so I thought this should be a reasonable choice.

To be clear, the goal of the new-style naming is actually just to provide a global naming scheme so that clients can get different resources from different servers without having any naming conflicts.  It's not really intended to make implementations simpler; in fact, as evidenced by this discussion, implementations will likely get a bit more complex. :)

> One complexity I got is for the SotW APIs (LDS and CDS), those needs per-server tracking to know which resource is removed.

Right.  I think that's another argument for having a two-level map: it avoids the need to iterate over every resource for every LDS or CDS response.  If you have a large number of authorities, you should ideally be able to look only at the resources for the authorities that are currently using the server that you're talking to.

In C-core, we are currently iterating through the list of authorities and skipping the ones whose server channel is not the channel that we received the response on.  And if we wind up in a situation where there are a very large number of authorities, we could optimize that further by keeping a list of authorities to look at in the object that holds the server channel.

A47-xds-federation.md
markdroth: Good point.  IMHO, the fact that any given template could affect only a subset of usage and might never actually be used in a given client is another good reason not to do this validation at bootstrap parsing time, since that does provide more options for forward-compatibility.

A47-xds-federation.md
```diff
@@ -0,0 +1,684 @@
...
+// "xdstp://xds.authority.com/envoy.config.listener.v3.Listener/grpc/server/%s"
+"server_listener_resource_name_template": <string>,
+```
+
+#### Example: Bootstrap Config Contains No New Fields
+
+If the bootstrap config does not contain any of the new fields described
+above, then here's how the resource names will be determined.
+
+If a gRPC client channel is created for `xds:server.example.com`:
```
dapengzhang0, line 221: Should we use `xds:///server.example.com`?  In java `xds:server.example.com` is not a URI String

A47-xds-federation.md
ejona86: Unconditionally percent-encode would imply the letter 'a' should become %61. `:` is in `pchar` so it can go most places in a URI. That said, it is a bit unclear what should be percent encoded. Should we say to percent encode anything that isn't `unreserved`, `sub-delims`, `:`, `@` ("if it can't be a pchar, without being percent encoded")? That would mean the permitted literal characters are: `ALPHA / DIGIT / "-" / "." / "_" / "~" / "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "=" / ":" / "@"`

sanjaypujare: I meant unconditionally percent-encode reserved characters. In any case it looks okay to not percent-encode ":" in this case based on https://datatracker.ietf.org/doc/html/rfc3986#section-2.2 which has this "If a reserved character is found in a URI component and no delimiting role is known for that character, then it must be interpreted as representing the data octet corresponding to that character's encoding in US-ASCII."

A47-xds-federation.md
markdroth: As per RFC-3986, `xds:server.example.com` is a valid URI with scheme `xds` and path `server.example.com`.

I'm aware that Java doesn't currently handle this format.  @ejona86 and I have discussed this in the past, and he agrees that it needs to be fixed, but I don't think that work has been done yet.

A47-xds-federation.md
markdroth: I made one more small change here: now that the authority key includes whether the resources are old-style or new-style, there's no more need to encode that difference in the lower level of the map.

markdroth: Donna fixed this in C-core in grpc/grpc#28106.

A47-xds-federation.md
```diff
@@ -0,0 +1,693 @@
...
+check the cache to see if the resource is already present, in which case
+it will be returned immediately.  Otherwise, it will send a message
+subscribing to that resource, using the appropriate xDS server from the
+authority.
+
+#### LRS Server Representation
+
+We currently represent the LRS server name in several places:
+- In the `CdsUpdate` struct returned by the `XdsClient` CDS watcher.
+- In the LB policy configs for the `xds_cluster_resolver` and
```
menghanl, line 558: The field (same in both balancers), is called `lrsLoadReportingServerName`.
Do we keep using that field? Or rename it? Or add a new field?

A47-xds-federation.md
```diff
@@ -0,0 +1,693 @@
...
+- In the `XdsClient` load reporting APIs.
+
+Currently, the name is represented as just a string, and it basically
+has only two values: unset, meaning that load reporting is not enabled,
+or the empty string, meaning to use the same server as we're using for
+ADS.  The latter meaning worked fine when there was only one possible
+server for us to talk to, but in the presence of federation, there can
+be multiple servers, and we need a way to indicate which one we mean.
+
+In the general case, we actually need more than just the server name for
```
menghanl, line 569: Will we report loads to a server that's not already an `authority` in the bootstrap config?
(We only support `self` today, and there's no xDS `ConfigSource` to specify a complete authority-config or `XdsServer`).

I'm thinking if we can just specify the `authority` name here, and the xdsclient can look up for the real config in the bootstrap.

A47-xds-federation.md
markdroth: Good question.  This made me realize that we should really spell out the exact LB policy config changes, and that made me notice a few problems.  The latest commit has a concrete proposal.

A47-xds-federation.md
markdroth: The `XdsServer` message here is basically our equivalent of the xDS `ConfigSource` proto.  If we did in the future decide to accept arbitrary `ConfigSource`s, it could specify a server other than one already listed in the bootstrap file, in which case we'd populate the `XdsServer` message from the `ConfigSource` (the [`target_uri` field](https://github.com/envoyproxy/envoy/blob/9d25410796882dabc4952ca26cab608930693e84/api/envoy/config/core/v3/grpc_service.proto#L235) would map to our `server_uri` field, the [`channel_credentials` field](https://github.com/envoyproxy/envoy/blob/9d25410796882dabc4952ca26cab608930693e84/api/envoy/config/core/v3/grpc_service.proto#L237) would map to our `channel_creds` field, and we'd seed our `server_features` field from the [`transport_api_version` field](https://github.com/envoyproxy/envoy/blob/9d25410796882dabc4952ca26cab608930693e84/api/envoy/config/core/v3/config_source.proto#L82)).  Even though we're not going to allow that today, we are trying to set the stage for doing so later.

I don't think we want to use the authority here, for several reasons:
- The authority as a concept is really an xDS-specific thing, and it makes sense only in the context of new-style xDS resource names.  LRS is not the same as xDS (it's a companion protocol to xDS but not identical to it), and it does not use resource names to specify the LRS server itself.
- If we did use an xDS authority name to specify the LRS server, then that would cause problems later if we wanted to support an LRS server that is different from the xDS server.
- In the short term, there will be many configs for which the LRS server is also the xDS server but the only xDS server configured in the bootstrap file is the global one, not one using an authority.  If we used an authority to specify the name, then we would have no way to indicate to use the global xDS server from the boostrap file.

A47-xds-federation.md
markdroth: I've sent grpc/grpc-proto#101 with the service config changes, as well as a couple of other missing updates.

markdroth: CC @dapengzhang0 as FYI for Java impl.

A47-xds-federation.md
ejona86: > the fact that any given template could affect only a subset of usage and might never actually be used in a given client is another good reason not to do this validation at bootstrap parsing time, since that does provide more options for forward-compatibility.

Delayed validation has no relation to forward-compatibility. Since it _may_ be used by a client, you can't break the older clients with new configuration. So it seems safe to validate up-front. This also has the benefit that someone testing the bootstrap doesn't need to test all client code paths in hopes of finding every last delayed validation. In short: if it would break at runtime and we need to validate it anyway, then validating it as early as possible is beneficial.

ejona86 (approved):
LGTMing as I'll be OOO for two weeks and I feel comfortable with it going in as long as you've read my comments.

A47-xds-federation.md
```diff
@@ -0,0 +1,730 @@
...
+// with an "xds:" URI with no authority.
+//
+// If starts with "xdstp:", will be interpreted as a new-style name,
+// in which case the authority of the URI will be used to select the
+// relevant configuration in the "authorities" map.
+//
+// The token "%s", if present in this string, will be replaced with
+// the service authority (i.e., the path part of the target URI
+// used to create the gRPC channel).  If the template starts with
+// "xdstp:", the replaced string will be percent-encoded.  The replacement
```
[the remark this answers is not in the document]
markdroth: Done.

A47-xds-federation.md
```diff
@@ -0,0 +1,730 @@
...
+```
+target_hostname = StripLeadingSlash(target_uri.path)
+if target_uri.authority is set:
+  authority_config = bootstrap.LookupAuthority(target_uri.authority)
+  if authority_config not found:
+    handle as invalid target URI
+  else:
+    template = authority_config.client_listener_resource_name_template
+    if template not set:
+      template = ("xdstp://" + target_uri.authority +
```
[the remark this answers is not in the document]
markdroth: Good point.  Fixed.

Sent grpc/grpc#28641 to fix C-core.

@dapengzhang0 @menghanl Please take note for Go and Java.

A47-xds-federation.md
```diff
@@ -0,0 +1,730 @@
...
+  a list of servers, ordered by priority.  (In other words, the
+  authority entry in the cache will change which server channel it is
+  using based on which server is currently reachable.)
+
+The second-level map will be the resource type, exactly the same as the
+top-level map is today.
+
+The third-level map will be the resource name.  For old-style resource
+names, this will be the full resource name, exactly as it is today.  For
+new-style names, it will be the `id` part of the `xdstp:` URI (i.e., the
```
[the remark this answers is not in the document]
markdroth: Changed wording to indicate that only these parts are required, but implementations may use the full resource name if they want.

A47-xds-federation.md
```diff
@@ -0,0 +1,730 @@
...
+If a new-style name has duplicate values for the same context param, the
+behavior is undefined.  The easiest thing is probably to just use the
+last value seen, since this allows the implementation to just populate
+the map by iterating over the query params.
+
+##### Watcher Changes
+
+When a caller starts a watch on an old-style resource name (one that
+does not start with `xdstp:`), the `XdsClient` will behave essentially as
+it does today, using the empty string as the authority name in the
```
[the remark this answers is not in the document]
markdroth: Good catch.  Forgot to update this in an earlier change.  Fixed.

A47-xds-federation.md
```diff
@@ -0,0 +1,730 @@
...
+resource cache.
+
+When a caller starts a watch on a new-style resource name (one that
+starts with `xdstp:`), the `XdsClient` will parse the URI.  It will use
+the authority of the URI as the primary key in the resource cache (using
+the empty string if the URI has no authority).  It will canonify any
+context parameters that are part of the resource name (i.e., removing
+duplicate keys and sorting keys in lexicographic order).  It will then
```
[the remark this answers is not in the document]
markdroth: The important part here is which server the subscription request gets sent to.  I've updated the wording to clarify this.

A47-xds-federation.md
markdroth: > > The authority as a concept is really an xDS-specific thing, and it makes sense only in the context of new-style xDS resource names. LRS is not the same as xDS (it's a companion protocol to xDS but not identical to it), and it does not use resource names to specify the LRS server itself.
> 
> Right now LRS is only "return to sender", so we aren't specifying the LRS server as much as the CDS source.

Even if we were phrasing it as "the CDS source", we still wouldn't want to specify it as an authority, because authority as a concept is completely irrelevant to LRS.

To say this another way, I think that in the fullness of time, we will want to move the LRS functionality out of the `XdsClient` and into a separate interface.  At that point, it will not make sense to specify the LRS server as an authority, because the authority will be a concept relevant to the `XdsClient`, not to the LRS interface.

> 
> > If we did use an xDS authority name to specify the LRS server, then that would cause problems later if we wanted to support an LRS server that is different from the xDS server.
> 
> What problems does it cause? We'd just have to add more code to support it and get rid of our current simple-string approach. But doing that in the future is just as easy as doing it now, except that we don't have to do it now.

We need to change the representation now to support federation anyway.  I don't think it makes sense to change to something that I'm confident will not work in the long run.  It seems better to change to something that I think will continue to work in the future, so that we minimize the chances of having to go change it again later.

> 
> In truth, it isn't clear this is the best future approach either. To avoid security problems, we have the restriction in the "XdsClient Load Reporting APIs" section that the config used here is already known from the bootstrap file. 

I think that the current security restrictions about requiring the LRS server to be one of the xDS servers in the bootstrap file is probably a short-term restriction.  I think that in the long term, we are going to have to find a way to solve the problem of being able to trust the control plane, because if we don't do that, we're going to start facing increasingly serious security problems that we can't work around easily.  These are problems that are larger in scope than simply how to specify the name of another control plane to talk to.


> It seems far easier if this is just use a "cookie", and that cookie could be the authority for SelfConfigSource. In the future we can generate an id (random, sequential, or hash) for xds-provided ConfigSources and store them in a map. That is obviously secure, trivial to plumb, and easier to implement in load reporting (vs a linear scan of all servers in bootstrap with a very specific equals check).

I think that approach seems like an option only because we happen to have xDS and LRS both handled via the `XdsClient` interface today.  I don't think it will make sense in the long run.

> 
> > In the short term, there will be many configs for which the LRS server is also the xDS server but the only xDS server configured in the bootstrap file is the global one, not one using an authority. If we used an authority to specify the name, then we would have no way to indicate to use the global xDS server from the boostrap file.
> 
> There's lots of disallowed charactes to use `&%/oldstyle`. Or if that bothers you too much, just use a different field name for old vs new. This seems a very small problem to instead argue for doing this much more invasive change within this design.

I think this would actually be more special-casey and more complicated than the current design.  It bakes in the assumption that you have to know whether it's an old or new name, which in principle should have no relevance to LRS, because that's a purely xDS-relevant concept.

I think that expressing the LRS server name directly is actually quite clean and general-purpose.  The only special case we have is that the `XdsClient` currently requires that the name actually be one of the ones in the bootstrap file, and that special case is well-contained and will be easy to remove in the future.

A47-xds-federation.md
```diff
@@ -0,0 +1,730 @@
...
+  }
+}
+```
+
+New-style resource names used as cache keys must be
+[normalized](https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md#normalization).
+In particular, this means sorting the context params in normal lexicographic
+order.
+
+If a new-style name has duplicate values for the same context param, the
```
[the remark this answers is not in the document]
markdroth: Today, just via the bootstrap file.  But this section is talking about the behavior inside of the `XdsClient` -- i.e., the behavior that needs to be provided by the xDS transport protocol layer.  In the future, we may change the code in the xDS resolver to get context parameters from some other source (e.g., we may decide to start supporting context params specified via the "xds:" target URI).  But the `XdsClient` layer should handle them the same way, no matter where they come from.

A47-xds-federation.md
markdroth: I don't think this is completely divorced from forward compatibility.  I think there could be cases where we would want to use a new feature for a particular new authority but not break any existing authorities.  In that case, validating this at bootstrap parsing time would make that impossible, whereas defering the check to run-time would provide the flexibility to say that old clients won't work with the new authority but will continue to work with existing authorities.

That having been said, although I still lean toward not doing this validation at bootstrap-parsing time, I can still see both sides here.  Does anyone else have any strong opinions on this?

ejona86: > I think there could be cases where we would want to use a new feature for a particular new authority but not break any existing authorities.

I see. I buy that to a degree. I do wonder "what would that feature look like" given how we've been maintaining compatibility to date by adding new fields, so that simply shouldn't be a problem for the majority case. But I still agree there may be something there for some case. This specific case, however, is a new field so that argument doesn't really hold.

Looking again at this case, I'm noticing now we'd have to parse the URL before the `%s` replacement, which seems bad. So I'm now agreeing we shouldn't validate this early. Yes, we could replace empty string or some such, but it won't be reliable.

A47-xds-federation.md
ejona86: > Even if we were phrasing it as "the CDS source", we still wouldn't want to specify it as an authority, because authority as a concept is completely irrelevant to LRS.

Well, to the perspective of the receiver in the service config, it'd just be a "cookie" or "server id". It'd be treated as opaque.

> We need to change the representation now to support federation anyway.

It is a string today. I was suggesting leaving it as a string. The only difference is lrs would need to pass that string into xds client when reporting.

> To say this another way, I think that in the fullness of time, we will want to move the LRS functionality out of the XdsClient and into a separate interface.

When we do that, there's a serious security issue to resolve. We _can't_ treat this as a general-purpose `ConfigSource`. Somehow xds client and the lrs client would need to coordinate (maybe through a third class, LrsApprovedServerContainer) to make sure the server config is trusted. Since they have to directly coordinate, we can pass an id or "handle" instead of the full configuration.

I'd also accept using "server1", "server2", etc, and not using the authority. Any string generation works; I can easily live without using the authority string, although I do think it'd be the easiest for implementation and debugging. We can change what the string is at any time and only the component that generated the string should care (i.e., xds client).

> I think that the current security restrictions about requiring the LRS server to be one of the xDS servers in the bootstrap file is probably a short-term restriction. I think that in the long term, we are going to have to find a way to solve the problem of being able to trust the control plane,

That's not at all what I was talking about. The design here and my proposal both work fine if we trust the control plane. The point is that "ConfigSources must be trusted, and service config isn't trusted", so we can give such sources a name when we get them (from bootstrap or control plane) and use that name. We don't need to provide the full config in service config because LRS _isn't allowed_ to use any arbitrary config; at any given time there's a fixed number of ConfigSources it can use.

> I think that approach seems like an option only because we happen to have xDS and LRS both handled via the XdsClient interface today

In all futures, xds client and lrs client have to coordinate for security. Unless we wanted to start signing config snippets in service config, which still requires coordination for the key.

>> There's lots of disallowed charactes to use `&%/oldstyle`
> I think this would actually be more special-casey and more complicated than the current design. It bakes in the assumption that you have to know whether it's an old or new name

Again, only xds client has to care (or a new class that mediates between xds client and lrs client). I'm not wild about a separate field, but it certainly is "less complex" than XdsServer solution. The special string seems very minor, especially since such a name may already exist (`old:` mentioned elsewhere).

> The only special case we have is that the XdsClient currently requires that the name actually be one of the ones in the bootstrap file, and that special case is well-contained and will be easy to remove in the future.

That special case can never be removed. It can be enhanced to be "in the bootstrap file or from control plane", but we must always have the validation check in place.

> in principle should have no relevance to LRS, because that's a purely xDS-relevant concept.

Well, right now security is only designed for xDS and ConfigSource must be trusted. But if we consider a gRPC-LB and RLS approach, LRS could become safe if CallCredentials were stripped... Mostly? Arbitrary ChannelCredentials configuration makes me nervous with its power. Ummm... is it safe to send load reports to an attacker's server, even if the credentials don't get leaked? The Node proto, cluster names, localities, or ORCA metrics may be poor to send to an attacker?

At the very least, LRS would need special xDS vs non-xDS logic. I'm sympathetic to this idea, but it seems premature to claim that XdsServer is well suited to the task.

ejona86: Just to be clear after the book: I can live with the XdsServer message in this design. It seems unnecessary, but it's not _that_ bad and all this is internal. If it is a misstep it wouldn't be hard to undo.

dfawley (approved):

A47-xds-federation.md
markdroth: Eric, you raise a good point about the issue of trusting the service config, as opposed to trusting the xDS control plane.  I think we have some disagreements about whether or not the service config should be considered untrusted in the general case, and it will require deeper discussion to come to consensus on that in the long term.  But as you say, I don't think this is a blocker for the current design.

Let's continue to discuss this as we move forward.

A47-xds-federation.md
markdroth: That's a good point.  Sounds like we've converged on this.

A47-xds-federation.md
```diff
@@ -0,0 +1,734 @@
+A47: xDS Federation
+----
+* Author(s): Mark D. Roth (@markdroth)
+* Approver: @ejona86, @dfawley
+* Status: {Draft, In Review, Ready for Implementation, Implemented}
+* Implemented in: <language, ...>
```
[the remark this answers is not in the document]
markdroth: Updated.

PR #269, opened 2021-10-21
A48: xDS Least Request LB Policy

erikjoh:


erikjoh:
@srini100 As discussed in the meeting. Not sure who to ping for help with figuring out some of the TODOs in here.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,184 @@
+A48: xDS Least Request LB Policy
+----
+* Author(s): erikjoh & [fabric](https://github.com/orgs/spotify/teams/fabric)
+* Approver: a11r
```
srini100, line 4: Please change the approver to markdroth

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,184 @@
...
+The subchannel state handling implemented in the `round_robin` policy will be re-used for `least_request_experimental`.
+Implementations may break out this subchannel state handling and have `round_robin` and `least_request_experimental`
+inherit the logic instead. This would be similar to the "base balancer" already implemented in grpc-go
+[here](https://github.com/grpc/grpc-go/blob/03268c8ed29e801944a2265a82f240f7c0e1b1c3/balancer/base/balancer.go).
+
+TODO: Possibly flesh this out with more explicit logic pending proposal PR review.
+
+##### Aggregated Connectivity State
+
+The `least_request_experimental` policy will use the same heuristic for determining the aggregated
```
ejona86, line 104: I don't think we want to go with that approach. I think we'd want to keep round_robin-style. We had to change things up for ring-hash because connection establishment in the degenerate case is serial, which could be _slooow_. For RR and it seems this least-request policy, connecting to backends is done in parallel, so I don't see a need to do anything special.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,184 @@
...
+
+##### Picker Behavior
+
+The picker should be given a list of all subchannels with the `READY` state.
+With this list, the picker should pick a subchannel based on the following pseudo-code:
+
+```
+candidate = null;
+for (int i = 0; i < choiceCount; ++i) {
+    sampled = subchannels[random_integer(0, subchannels.length)]
```
ejona86, line 118: super-nit: seems this pseudo-code language uses semicolons, but the semicolon is missing here.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,184 @@
...
+    if (candidate == null) {
+        candidate = sampled;
+        continue;
+    }
+
+    if (sampled.active_requests < candidate.active_requests) {
+        candidate = sampled;
+    }
```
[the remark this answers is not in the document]
ejona86: The current loop seems fine. In real life, I'd sort of expect us to do:
```
candidate = subchannels[random_integer(0, subchannels.length)];
for (int i = 1; i < choiceCount; ++i) {
    sampled = subchannels[random_integer(0, subchannels.length)];
    if (sampled.active_requests < candidate.active_requests) {
        candidate = sampled;
    }
}
return candidate;
```

It just needs to communicate.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,184 @@
...
+The `least_request_experimental` policy will associate one outstanding request counter for each subchannel.
+These counters should additionally share the lifecycle with its corresponding subchannel.
+The counter for a subchannel should be atomically incremented by one once it has been picked by the picker.
+In the `PickResult` for each language-specific implementation (
+[Java](https://github.com/grpc/grpc-java/blob/1f90e0e28d5628195cb1f861b73e45ed003f2973/api/src/main/java/io/grpc/LoadBalancer.java#L561-L571),
+[C++](https://github.com/grpc/grpc/blob/4567af504ed53cc8398e53bf1efd23f753d43bb8/src/core/ext/filters/client_channel/lb_policy.h#L178-L201),
+[Go](https://github.com/grpc/grpc-go/blob/01ed64857e3146000ec99cdea4f2932204f17cdd/balancer/balancer.go#L256-L269)
+), the picker should add a
+callback for atomically decrementing the subchannel counter once the RPC finishes (regardless of `Status` code).
+In some cases, these callbacks may not be called.
```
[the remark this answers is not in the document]
ejona86: We've come across this problem before with circuit breaker. Maybe we can just link to what we did there. I'll note that the implementation in Java was racy, and we just accepted that the counters could be exceeded briefly. So we may need to accept that here as well the counters may be slightly old, although there are certainly more risks here by accepting the raciness.

`streamCreated()` may be fine from within a LB, but we used `newClientStreamTracer()` previously. You can track how we maintained [`inFlights` in the cluster policy](https://github.com/grpc/grpc-java/blob/59c6b49fd4051b8d9d5cf958a762007a2f60280c/xds/src/main/java/io/grpc/xds/ClusterImplLoadBalancer.java#L179).

ejona86 (commented):
This looks pretty good from my perspective. Only a few details to work out. Seems easy to implement and would be valuable.

A48-xds-least-request-lb-policy.md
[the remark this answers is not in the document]
erikjoh: @ejona86 I think it's probably fine that counters are slightly outdated here, since they should in most cases still be fairly close to the real value and thereby still relevant for least request.
What concerns me more is whether there is a case when `newClientStreamTracer()` is called but `streamClosed()` is not. If that is possible you would have a zombie outstanding request that never gets removed from the counter.
But since you did this approach for `inFlights` in the cluster policy this is either not possible or extremely unlikely.
For the least request POC in grpc-java we made [here](https://github.com/erikjoh/grpc-java/blob/8ccc41b7cdbd082d9cfb5aef631709b7c181f3da/core/src/main/java/io/grpc/util/LeastRequestLoadBalancer.java#L390-L424), we copied that exact approach and it seems to work fine so far at least.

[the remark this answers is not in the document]
ejona86: > think it's probably fine that counters are slightly outdated here

It very well might be, but do understand the policy devolves in this case. Consider the case when the number of samples is close to the number of addresses and there's bursty QPS. Basically it would be finding the address with the "true (observed) minimal" requests, which will be the same for every RPC in a burst. The bursts are _probably_ not a concern in practice, but it depends heavily on the size of the race window.

> What concerns me more is whether there is a case when newClientStreamTracer() is called but streamClosed() is not.

That shouldn't happen. The tracer instance has a similar lifetime to the stream, and streams had better be closed. If there's a leak here I'd consider that a bug and we'd need to fix it.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,184 @@
...
+
+The `least_request_experimental` policy will have the following config:
+
+```proto
+message LeastRequestLoadBalancingConfig {
+  uint32 choice_count = 1;  // Optional, defaults to 2.
+}
+```
+
+The `choice_count` determines the number of randomly sampled subchannels which should be compared
```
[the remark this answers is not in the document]
erikjoh: Even though Envoy has no limitation in place I think it makes sense to have one in gRPC!
If others agree on 10 being a reasonable max choice count we can use that I think.
Capping it using `number of endpoints` may also be a good idea but it possibly could reduce performance in smaller scale scenarios. For example with a `choice_count=3` and only two endpoints `a` and `b` where `b` performs much worse than `a` (i.e. much more outstanding requests), the probability of picking the bad one `b` goes from `25%` to `12.5%` if we still allow three choices instead of two (since sampling is independently random).

A48-xds-least-request-lb-policy.md
erikjoh: Interesting point! In that case I'll change this to be consistent with RR instead.

A48-xds-least-request-lb-policy.md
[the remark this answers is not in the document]
erikjoh: > The bursts are probably not a concern in practice, but it depends heavily on the size of the race window.

I *think* Envoy accepts this raciness in favor of a safer but slower picker (will try to check later as it may be an interesting point to mention in the proposal text).
This could also be an additional reason to limit `choice_count`, since a higher value means a bigger race window.

A48-xds-least-request-lb-policy.md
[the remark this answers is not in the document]
erikjoh: I thought about this a bit more and decided to go ahead and update with max 10 now already.
Happy to update again if it makes sense to change the max value or also take `number of endpoints` into account.

A48-xds-least-request-lb-policy.md
erikjoh: Updated now. Possibly listing the rules isn't needed here so could remove that if it doesn't make sense.

A48-xds-least-request-lb-policy.md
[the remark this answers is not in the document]
erikjoh: Updated the text now and added a full section in the `Rationale` section further below.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,225 @@
...
+```
+candidate = null;
+for (int i = 0; i < choiceCount; ++i) {
+    sampled = subchannels[random_integer(0, subchannels.length)];
+    if (candidate == null) {
+        candidate = sampled;
+        continue;
+    }
+
+    if (sampled.active_requests < candidate.active_requests) {
```
markdroth, line 135: Where is the policy going to store the count of active requests for each subchannel?

In Java and Go, this will probably be fairly straightforward; the policy can implement its own subchannel wrapper to store this field.  But C-core has a slightly more complicated model for handling subchannels, which will make this a bit more challenging.

In Java and Go, the LB policy is effectively the single owner of the subchannel.  Whenever the LB policy is given an updated list of addresses, it compares that list to the previous list and keeps the subchannels it already had for any addresses that are still present on the new list.  This avoids unnecessary connection churn caused by deleting the existing subchannel and creating a new subchannel for the same address.  And if those subchannels are wrapped by the LB policy, that wrapper will be a stable object for each address that it can use to store the active request count.

In C-core, however, the subchannels are ref-counted and can have multiple owners, and the `SubchannelInterface` object returned to the LB policy is just a thin wrapper that holds a ref to the real subchannel.  The channel maintains a pool of subchannels, and when an LB policy asks the channel to "create" a subchannel for a given address, if the pool already contains a subchannel for that address, then the channel returns a `SubchannelInterface` object that holds a new ref to that existing subchannel instead of creating a new subchannel.  So when an LB policy is given an updated list of addresses, it doesn't bother to compare with the old list; it just unconditionally "creates" a new subchannel for every address in the list and unrefs the subchannels in the old list.  If there are addresses in common, the new ref will have been created before the old ref is released, so there will not wind up being any connection churn.

What this means is that every time the `least_request` policy gets an updated address list, it will wind up replacing the `SubchannelInterface` object for a given address, even though the underlying subchannel is not changing, which means that the subchannel wrapper is not going to be a stable place to keep the active request count.  So we probably need to come up with something different there.

ejona86 (approved):

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,225 @@
...
+
+##### Outstanding request counters
+
+The `least_request_experimental` policy will associate one outstanding request counter for each subchannel.
+These counters should additionally share the lifecycle with its corresponding subchannel.
+The counter for a subchannel should be atomically incremented by one after it has been successfully
+picked by the picker. Due to language-specifics, the increment may occur either before or during stream creation.
+In the `PickResult` for each language-specific implementation (
+[Java](https://github.com/grpc/grpc-java/blob/1f90e0e28d5628195cb1f861b73e45ed003f2973/api/src/main/java/io/grpc/LoadBalancer.java#L561-L571),
+[C++](https://github.com/grpc/grpc/blob/4567af504ed53cc8398e53bf1efd23f753d43bb8/src/core/ext/filters/client_channel/lb_policy.h#L178-L201),
```
markdroth, line 153: Note that I'm soon going to merge grpc/grpc#27947, which will change the API in C-core.  The new API handles an edge case that would have caused a problem for this design, in which the current API was not getting notified when a pick result was abandoned.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Rationale
+
+### Support for Weighted Least Request
+
+While also supporting the "*all weights not equal*" (Weighted Least Request) case
+[from Envoy](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#weighted-least-request)
+would enable more advanced use cases, there are some missing pieces in gRPC that
+are required for that to be technically feasible.
+For example, gRPC ignores the per-endpoint `load_balancing_weight` field as a result
```
markdroth, line 181: We actually added support for this field in C-core as part of implementing `ring_hash`.  I don't think it would be hard to support in other languages either.

What would be harder would be actually changing `ROUND_ROBIN` to support these weights.  There is another open gRFC proposal for that in #202, although it's been silent for a while, so I'm not sure if the author is still working on it.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,225 @@
...
+
+One possible downside of the Envoy implementation, is that the random selection will allow a subchannel to
+be sampled multiple times and thereby be unavoidably picked with some probability.
+This means that even a frozen endpoint would never fully be excluded by the picker.
+There exists a
+[closed PR to Envoy](https://github.com/envoyproxy/envoy/pull/11006)
+that attempted to solve this.
+In that PR, there were some questions around whether there would be any significant improvement for >5 endpoints.
+
+### Outstanding request counter raciness
```
markdroth, line 198: I think that the new API in C-core that I mentioned above would actually allow you to eliminate most of this raciness -- or at least to err on the side of considering an address to have every request you assign to it, even if some of those picks get abandoned.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,225 @@
...
+### Related Proposals: 
+
+This proposal builds on earlier work described in the following gRFCs:
+* [gRFC A27: xDS-based Global Load Balancing](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)
+* [gRFC A42: xDS Ring Hash LB Policy](https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md)
+
+## Proposal
+
+Support `LEAST_REQUEST` LB policy and configuration of it using xDS.
+Specifically, gRPC will only support the "*all weights equal*" least request algorithm as
```
markdroth, line 36: How will the policy react if it gets more than one instance of the same address in the address list?  That's not likely to happen in the xDS case, but it can happen with DNS or other custom resolvers.

In this case, will the policy de-dup the addresses and assume just one instance, or will it treat each instance of the address as a separate address to be chosen from?

markdroth (commented):
This looks pretty good overall!  I've got a few comments and questions, just to make sure we've thought through all of the edge cases.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,225 @@
...
+* [Go](https://github.com/grpc/grpc-go/blob/03268c8ed29e801944a2265a82f240f7c0e1b1c3/balancer/base/balancer.go)
+
+##### Aggregated Connectivity State
+
+The `least_request_experimental` policy will use the same heuristic for determining the aggregated
+connectivity state as the `round_robin` policy.
+The general rules in order of precedence are:
+
+1. If there is at least one subchannel with the state `READY`, the aggregated connectivity state is `READY`.
+2. If there is at least one subchannel with the state `CONNECTING` or `IDLE`,
```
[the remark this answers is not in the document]
ejona86: This is the behavior of RR, since when a subchannel enters IDLE the policy immediately requests a connection. IDLE will only ever be a very transient state.

A48-xds-least-request-lb-policy.md
erikjoh: Nice!
Then I would say it makes sense to wait until `subchannel_call_tracker` is available before implementing this.
Would be quite bad for the lb policy if outstanding requests are leaked.
We concluded in an earlier thread here that the Java `ClientStreamTracer` should be safe to use in this design.
Possibly we need to make sure the grpc-go callback also is safe for this design as well.

I'll can update the proposal here and reference that PR as a dependency.

A48-xds-least-request-lb-policy.md
erikjoh: Hmmm, interesting edge-case!
First of all, I assume you mean ip+port by "address" in this case (so e.g. "10.1.2.3:443")?
If a dns/custom resolver today receives duplicate addresses I presume it's being done to more heavily weigh traffic to some address when using round_robin.
If least_request were to de-dup in such a case, it would probably defeat the purpose behind having duplicate addresses in the first place.
And while the traffic weigh would not work nearly the same between RR and LR in this case, there would still be a somewhat heavier preference towards addresses that are duplicated in LR.
So if round_robin today doesn't do any de-dup of addresses and has a 1:1 mapping between addresses and subchannels I think least_request should probably follow that approach.

Does that make sense?
If so, then I can add this clarification in here somewhere :smile:

A48-xds-least-request-lb-policy.md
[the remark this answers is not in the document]
erikjoh: @markdroth This list is coming from this comment [here](https://github.com/grpc/grpc/blob/79d684529d9db60aa2c5e82deb43e297a6818cfb/src/core/ext/filters/client_channel/lb_policy/round_robin/round_robin.cc#L306-L319).
The difference is the `or IDLE` that I added since subchannels immediately enter `CONNECTING` in this case as @ejona86 points out.
In grpc-java, the implementation even explicitly checks for `IDLE` [here](https://github.com/grpc/grpc-java/blob/fe9026ed8ae5025e92d452e077bfa0b329c123a0/core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java#L186-L191).
Looking at the different implementations, it seems like there may be some slight inconsistencies which makes it tricky to define the current round robin behavior that least request should mimic.

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,225 @@
...
+   the aggregated connectivity state is `CONNECTING`.
+3. If all subchannels have the state `TRANSIENT_FAILURE`, the aggregated connectivity state is `TRANSIENT_FAILURE`.
```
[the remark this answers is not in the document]
erikjoh: So my interpretation of this is that the aggregated connectivity state can only transition `TRANSIENT_FAILURE -> READY` and not `TRANSIENT_FAILURE -> CONNECTING`.
How about adding something like this?

```suggestion
3. If all subchannels have the state `TRANSIENT_FAILURE`, the aggregated connectivity state is `TRANSIENT_FAILURE`.

Also note that since subchannels in this case only can transition from the state `TRANSIENT_FAILURE` to the state `READY` ,
the aggregated state also only may transition from the state `TRANSIENT_FAILURE` to the state `READY`.
```

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+The `choice_count` determines the number of randomly sampled subchannels which should be compared
+for each pick by the picker.
+Envoy currently supports any `choice_count` value greater than two.
+In the `least_request_experimental` policy however, only an effective `choice_count` in the
+range `[2, 10]` will be supported. If a `LeastRequestLoadBalancingConfig` with a `choice_count > 10`
+is received, the `least_request_experimental` policy will set `choice_count = 10`.
+
+##### Subchannel state handling
```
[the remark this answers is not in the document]
erikjoh: Hmm, it was tricky to find the equivalent logic for the different languages and also it's hard for me to tell if there are any slight discrepancies between them.
The [java implementation](https://github.com/grpc/grpc-java/blob/fe9026ed8ae5025e92d452e077bfa0b329c123a0/core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java#L138-L156) has been the easiest to follow for me so if I just where to explain the "Subchannel state handling" there it would be something like:

```
1. New subchannels begin with the `IDLE` state and are immediately requested to connect.
2. When a subchannel enters either `TRANSIENT_FAILURE` or `IDLE` states, a re-resolve is triggered.
3. When a subchannel enters the `IDLE` state, it's immediately requested to connect.
4. When a subchannel enters the `TRANSIENT_FAILURE` state, it can only transition directly to the `READY` state.
5. Removed subchannels cause a disconnect and release of the underlying resources.
```

Possibly it's better to just describe it like this here instead of trying to find/link to implementations in each language.

A48-xds-least-request-lb-policy.md
erikjoh: I'm not super familiar with the C-core implementation and would need to spend some more time to give a more proper answer to this.
I would (probably naively) think that doing a similar approach to what `XdsClusterImplLb` does for circuit breaking would work reasonably well here as well with some small adjustments.
So the least request implementation would have a `SubchannelCallCounterMap` similar to [CircuitBreakerCallCounterMap](https://github.com/grpc/grpc/blob/79d684529d9db60aa2c5e82deb43e297a6818cfb/src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_impl.cc#L47-L107) that somehow uses the real subchannel as key to find the corresponding counter.
One downside with this however, is that the lb policy might need to do a diff between re-resolves to clean up counters for removed addresses.
As for churn, it might not be too bad since I guess an address that gets re-introduced wouldn't/shouldn't have any outstanding requests anyway (even though it already might have an open connection since before).

markdroth: Yeah, that's definitely one possibility.  The entries in the map could be handled via simple ref-counts held by wrapping the `SubchannelInterface` objects (and probably also by the per-call completion callbacks that decrement the counters), so they'd clean themselves up automatically when the address stops being used.

But this brings up another question: Do we want these counters to be global?  In the case of circuit breaking, the counters are global, because they're per-cluster limits, and no matter how many gRPC channels are talking to the same xDS cluster, we want to apply the same limits across all of them.  But outside of the xDS space, LB policies are generally per-channel with no state shared across channels, and it would be somewhat surprising if they shared state globally.  It seems like the goal here is to build the `least_request` policy in a way that works for both the xDS and non-xDS use-cases, so we should carefully think through what behavior we want here.

(Note that the `xds_cluster_impl` policy *is* xDS-specific, so this isn't an issue there.  And the `ring_hash` policy, which is *not* xDS-specific, doesn't have this issue, since the same hash algorithm will be used in every channel even without any global state.  So I think this is the first time we're really having to figure out how to bridge this particular gap.)

In the xDS case, we probably would want to have global counters here, because if a given cluster uses the `least_request` policy, we ideally want to consider all traffic going to the cluster (i.e., we would use the same argument that we used to choose global state for circuit breaking).  If we don't do that, then this will be a case where our behavior will differ from Envoy's, so at minimum, we should document that.

In the non-xDS case, it would probably be a bit surprising if two channels started counting each others' traffic.  But on the other hand, it probably wouldn't actually hurt anything, and it might actually wind up yielding better balancing overall.  However, it is a notable difference from how existing policies work.

It's also worth noting that in Java and Go, each channel will have its own set of subchannels, but C-core by default shares subchannels between channels with the same target and channel args.  If we use global counters, that seems a little more intuitive in the C-core model, whereas it seems a little strange in the Java and Go model; if we don't use global counters, then the opposite is true.  I don't think this will cause any actual problem either way; it's just a little counter-intuitive to share a connection but not see all of the call traffic on the connection (or vice-versa).

In terms of implementation, note that if we do want global counters, then we'd probably need a global map similar to the one we use for circuit breaking, even in Java and Go.  We'll probably need to give some thought to exactly what keys we use for the global map; we probably want it to be more than just the address, because we probably don't want to share counters in the case where there are two channels talking to two different services via the same proxy.

@ejona86 or @dfawley, any thoughts on this?

A48-xds-least-request-lb-policy.md
markdroth: I've just merged that PR, so you can just refer to it as the current API now. :)

@dfawley tells me that the go callback is invoked even in the case where the pick is abandoned, so you should be able to do what you need in every language.

A48-xds-least-request-lb-policy.md
erikjoh: As you point out, there are several cases, especially in java/go, where pure global counters probably aren't a good fit.
For example, talking to different services through a proxy or using different subchannels/connections that target the same address.
One other aspect is that xDS-specific information like cluster name isn't even available in this LB layer, so using something like that for the key seems unfeasible.
Instead of defining it as global or non-global, one option could maybe be to just define the counters as being associated to each unique subchannel:

```
Similarly to how underlying subchannels each have a connection state attribute (READY,IDLE,etc.),
each underlying subchannel will be associated with a single counter for outstanding requests.
This may be implemented either using custom subchannel attributes or a global map that maps the subchannel
memory address to it's corresponding counter.
```

This means that in C-core, counters would be shared equivalently to how subchannels themselves are shared, while in Java/Go, the counters would never be shared since subchannels are never shared.

While this approach wouldn't optimize well for certain edge-cases, it's a fairly simple approach to reason about or debug if there ever are any LB issues.
With the circuit breaking and fully global counters, it's much easier in this regard since hitting the circuit breaker limits gives feedback in the form of an RPC error whereas LBs will be completely silent in all cases.

A48-xds-least-request-lb-policy.md
erikjoh: Interesting!
Looks like java at least supports it as well but the weight attribute is marked as:
```
Internal attributes used for xDS implementation. Do not use.
```
which means per-endpoint weights still may need to be standardized in a gRFC before it can be used in an LB implementation like this.
Will update the text to reflect this.

A48-xds-least-request-lb-policy.md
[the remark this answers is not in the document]
markdroth: Linking to the full files is probably fine.  Honestly, I don't really care about linking to implementations at all, so you could just remove that if you want, although I don't object to the links being here.  But the important thing is to describe in text the required behavior of the `least_request` policy, independently of any existing implementations.

A48-xds-least-request-lb-policy.md
markdroth: Just FYI, in C-core, we added the attribute in a way that is not xDS-specific:

https://github.com/grpc/grpc/blob/c1089d2964528b52a25c18d748910658e4762d40/src/core/ext/filters/client_channel/server_address.h#L116

Note that I think we'd view this more as a convention than as a required part of the API: it's up to each resolver impl to decide to add this attribute to the addresses it produces, and it's up to each LB policy impl to decide whether to look at this attribute and what to do with it.  So it's something that can be used when there are use-cases for which it makes sense, but it's not required or assumed to be used by any given resolver or LB policy.

But I think within those constraints, we could support it in `least_request`, the same way that we do in `ring_hash`.

A48-xds-least-request-lb-policy.md
markdroth: I'd like feedback from Eric and Doug before making a call here.  Personally, I'm leaning in the direction of global counters, since there are some advantages and there isn't really any compelling disadvantage; the behavior may be a little surprising, but it seems like it wouldn't actually hurt anything in common cases.

> One other aspect is that xDS-specific information like cluster name isn't even available in this LB layer, so using something like that for the key seems unfeasible.

One possible way to address that might be by having the `least_request` policy define a per-address attribute indicating some additional value to include in the key for the global map.  We could then have the xDS code pass the cluster name in using that attribute.

ejona86: TL;DR: My suggestion is the last (non-parenthetical) paragraph. Unfortunately it took this mini-proof to arrive there; no summary for you!

I think subchannel sharing is a red herring. C core can do the same behavior of Go and Java by simply doing the address-level diffing that Go and Java do; it can "manually" reassociate the counters from the old subchannel to the new subchannel when it gets address updates.

I think that's a separate discussion from "what scope should the counters be." My initial gut reaction was also they should be associated with the subchannel, and if we choose to do anything we might break subchannel sharing (e.g., by including the cluster name in channel args). But I think that is neither here nor there. With per-subchannel counts 1) we'd still lose any RPC counts when the policy using the subchannel isn't least-request, and 2) there's nothing all that special about _the connection_ as if the connection receives GOAWAY we'd sum its counts with any RPCs on a new connection.

This policy is about spreading load and reducing load to slower backends or backends that may have unluckily received more costly requests. There's no need for perfect global information here, as this client doesn't know the number of requests sent by other clients. There are probably multiple valid scopes for the counters from a correctness perspective, and so it is just an optimization problem when choosing between them.

Note that the policy semi-assumes a reasonable number of concurrent RPCs, otherwise it devolves into random. It doesn't _require_ many concurrent RPCs, but there's no benefit to using it over random/round_robin for low concurrent RPCs.

If we take a least-request policy with a fixed set of addresses and duplicate the entire thing N times (duplicating the counters as well), then the behavior of those N policies should actually be fairly similar to the single unified policy. That only devolves if the number of concurrent RPCs is low. If each policy receives a different type of traffic (different methods, with different cost per method), then that still holds; there can be small interactions between the different types of traffic, but they should be in the noise with sufficient concurrent RPCs. (For lower concurrent RPCs, separate policies might be _better_, as that will distribute each _type_ of RPC more evenly.)

The behavior would change if the N policies have different addresses. Assuming our target is to emulate Envoy's per-cluster-per-address scope, this means that per-shared-subchannel counters is likely a bad idea, but per-shared-subchannel-per-cluster would be okay (if we assume XdsClient sharing and use of EDS, to guarantee the same addresses in each policy).

I don't see a need for global state in Java and Go. In C, I think we should just select the easiest-to-implement approach, whether that is to "manually" transfer counts during address updates or to use the cluster and subchannel as keys into a global map. Assuming similar addresses provided to each LB instance, both approaches should have similar enough behavior that I think the main dimension to optimize is effort. If we don't think that is a safe assumption (e.g., not using EDS), then it is probably best for C to do the "manual" diffing to transfer counts similar to Go and Java.

(If we do go with a cluster+subchannel global map in C, I'd hope we design it in an xds-agnostic way so the policy can still be used outside of xds at some point. I think there are ways to do that easily, so I'm not discussing it here.)

dfawley: I somewhat agree with Eric, though I would say that aggregating as many sources as possible will yield the best results.  I.e. if you had a global view, you could direct traffic most efficiently.  If you have a full view of the current client _machine_, you could direct traffic a bit less efficiently.  If you limit it to the current client _process_, you lose some efficiency.  And if you limit it to a single client _instance_, you lose even more efficiency.  This goes with what Eric said that "the policy semi-assumes a reasonable number of concurrent RPCs" -- with more sources of RPCs, we will have more concurrent RPCs.

So, on average, the policy should give the best results if implemented as globally as possible.  If it was going to be difficult to implement in Go, I would say that the cost is not worth the benefits.  But I don't believe a global map would be problematic at all to implement, so I would be fine with tracking it globally.  That said, I also don't think it's likely to make much of a difference either way, since _most of the time_, "the current process" will devolve into a single client anyway.

markdroth: Thanks, that was really useful discussion!

@ejona86, to me, the key point in what you said was actually this:
> If we take a least-request policy with a fixed set of addresses and duplicate the entire thing N times (duplicating the counters as well), then the behavior of those N policies should actually be fairly similar to the single unified policy.

I think you're right about the various edge cases with low number of concurrent requests or different mixes of differently weighted types of requests, but those things can happen even across client processes or machines or any other scope we could consider.  So I don't think we need to really worry about those, and I agree with your point that in most situations where this policy will be used, the behavior should be the same no matter what scope we choose here.

So, I think we should stick with a purely local-to-the-LB-policy view.  Even if it turns out that C-core will want to put this in a map keyed by subchannel ID rather than storing it in the subchannel wrappers, the fact that that map isn't global means that we don't need a bunch of additional synchronization, which seems like a win.

@erikjoh, I think we can change this gRFC to say that the counters can be stored in a wrapped subchannel.  In C-core, there will likely be some indirection there -- the counters themselves will probably be in a separate map, and the wrapped subchannel will hold a ref to the relevant entry in that map -- but either way, the wrapped subchannel is the means that the picker will use to determine which counter to increment for a given request, so I think that's the important thing to note here.

A48-xds-least-request-lb-policy.md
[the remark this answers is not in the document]
markdroth: It looks like you've updated the text to refer to the "logical" subchannel state, as if that was an already well-understood term, but I don't think that most readers are going to know what that means.  I think it would be better to just say something like this:

For the purposes of the evaluating the aggregated connectivity state rules above, a subchannel that enters TRANSIENT_FAILURE should be considered to stay in TRANSIENT_FAILURE until it reports READY. If a subchannel is bouncing back and forth between TRANSIENT_FAILURE and CONNECTING while it attempts to reestablish a connection, we will consider it to be in state TRANSIENT_FAILURE for the purposes of the above aggregation rules.

A48-xds-least-request-lb-policy.md
erikjoh: Added `Allow picking same candidate multiple times` section further down in the gRFC.

A48-xds-least-request-lb-policy.md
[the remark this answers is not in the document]
erikjoh: Thanks for the clarification!
Added a slightly modified version of your suggestion and removed the `"logic"` .

A48-xds-least-request-lb-policy.md
erikjoh: Very informative discussion indeed!
@markdroth I've added a few sentences about this in the outstanding counters section and also moved that section before the picker section since it seems like a more logical ordering here.
I was a bit careful with the wording since I don't think the Java implementation actually needs to wrap the subchannels since subchannels are already LB-local and also the support for custom subchannel attributes can be utilized to store the counters.

A48-xds-least-request-lb-policy.md
erikjoh: Even though raciness may be eliminated in that particular case I still think it's worth mentioning that some small amount of raciness is theoretically possible.
Since the picker is fully concurrent, any large burst of picks could potentially also cause unfair bursts to individual upstream endpoint(s) which may overload them.
The race window here should be very minor however, this probably isn't a problem in practice.

A48-xds-least-request-lb-policy.md
erikjoh: Been thinking about this a bit more and I'm currently contemplating whether to include the "all weights not equal" (weighted least request) in this proposal or not.
The main reasons I had for not including it where:
1) I didn't want to deal with design of sending per-endpoint weights to LBs in each language in a non-xds-specific way.
2) I have a hard time seeing how the control plane can calculate valuable per-endpoints weights in cases where endpoints are considered equal (at least in terms of cpu/mem performance). The control plane would probably need to have per-endpoint load information either through load reports or other means. And even in that case the feedback loop may be too slow for large scale dynamic setups with high/bursty rps. This is where ORCA may be beneficial instead.

If you are saying I wouldn't have to think about 1) in this proposal, it all boils down to whether the feature is worth the moderately extra effort to implement or not.
If you think it's worth the effort, I can add the extra picker logic for the "all weights not equal" case and state that it's "optional for a language to implement given a non-xds-specific solution for passing the weights from the resolver".

A48-xds-least-request-lb-policy.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Proposal
+
+Support `LEAST_REQUEST` LB policy and configuration of it using xDS.
+Specifically, gRPC will only support the "*all weights equal*" least request algorithm as
+[defined by Envoy](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/load_balancers#weighted-least-request).
+
+### Least Request LB Policy
+
+#### xDS API Fields
```
markdroth, line 41: This section does not mention the `slow_start_config` field.  Is that going to be supported?

A48-xds-least-request-lb-policy.md
markdroth: I don't have any strong feeling one way or the other about whether the "all requests not equal" algorithm is useful.  The only factor that I would consider here is forward-compatibility: if it's not supported now and then we wind up adding it later, will clients start behaving differently when they get upgraded?  If that change is something that might cause problems, then it might make sense to implement it now, just so that we don't run into that later.  OTOH, if that's purely an optimization and not likely to cause any problems, then it may be fine to add it later if/when someone actually needs it.

A48-xds-least-request-lb-policy.md
markdroth: That's a good point.  I was thinking about the fact that you will need to create an object that contains both the subchannel and the counter for a given address, and the picker will need to access that object so that it can increment the counter when it returns the subchannel.  But you're right that that object doesn't necessarily need to wrap the subchannel, since this is a leaf policy and won't be passing the subchannel down to a child policy and then unwrapping it when it comes back from the child's picker.

A48-xds-least-request-lb-policy.md
markdroth: Hmm, I'm now thinking that maybe we went the wrong way with this, for two reasons.

First, it seems like the idea here is to have a different counter for each instance of the address.  However, given the discussion we just finished about where to store the counters, it seems a little strange to store two different counters for the same subchannel.

Second, in principle, having the same address listed multiple times is just another way of handling weighting, so I'm thinking that we should just consider this another way of expressing per-endpoint weight.  In other words, the equally weighted address list [A, B, A] should be considered equivalent to the weighted address list [A=2, B=1].  That's basically what the ring_hash policy does, so it seems consistent to do the same thing here.  And then the way we handle this becomes just part of the question of whether we support the "all requests not equal" algorithm.

WDYT?

A48-xds-least-request-lb-policy.md
erikjoh: I think we would prefer to not include the "all weights not equal" algorithm in this proposal. Even though I think the implementation doesn't need to be too tricky, there are still some tricky edge cases which probably need to be discussed/addressed.
As for the consequences of adding this to the implementation later, it seems like the behavioral change wouldn't be too severe/controversial, especially since it only would affect the case where the control plane is sending per-endpoint weights to a proxyless gRPC client which previously ignored them.
This exact problem already exists for round_robin as well, since the ROUND_ROBIN policy in xDS should mean weighted round robin if per-endpoint weights are present. But until https://github.com/grpc/proposal/pull/202 (or similar) is implemented, proxyless gRPC will ignore any per-endpoint weights for a cluster with ROUND_ROBIN policy.

I might also be willing to open a follow-up gRFC for this addition to least request, but it's currently not a priority for us to have it supported.

A48-xds-least-request-lb-policy.md
erikjoh: > it seems a little strange to store two different counters for the same subchannel.

I can't speak for C-core or Go until checking the implementations more carefully, but in Java it looks like duplicate addresses in the resolution result ([here](https://github.com/grpc/grpc-java/blob/a2398ce5dbe2eacf841d7af9c9aa5e912e8db0aa/core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java#L73)) also causes duplicate/separate subchannels. In this case, I think it still makes sense to have separate counters on these subchannels as it would increase the selection probability (even though it's not proportional to the "weight").
This would also make the implementation simpler, since more logic from round robin may be used as-is.

If we end up adding the Weighted LR algorithm, I think your suggestion makes sense.
One edge case I can think of is the case where the resolver also sends address weights.
Then [A=2,B=1,A=1] would have to become [A=3, B=1] and so on. Possibly the resolver should be responsible for sending a proper resolution result so that we don't need implicit aggregation in the LB policy like this.

So my suggestion is to still have one counter per subchannel (or subchannel ref in C-core) and use exactly the same logic as round robin for deciding how duplicate addresses should translate into duplicate subchannels (or subchannel refs).

A48-xds-least-request-lb-policy.md
erikjoh: Good catch! Was looking at an old version of the proto which didn't have that field.
This feature requires the Weighted Least Request algorithm so if we don't support that initially we wouldn't support `slow_start_config` either.
Also, I think support for this might be a bit tricky for the "*all weights equal*" case since Envoy seems to switch to the "*all weights not equal*" algorithm during endpoint startup and then back to the "*all weights equal*" algorithm once all endpoints have fully started. This might get a bit complicated in gRPC as it means the picker behavior would need to change in between resolves.

A48-xds-least-request-lb-policy.md
markdroth: @ejona86 tells me that round_robin in Java *does* de-dup subchannels for duplicate addresses (he pointed me at [this](https://github.com/grpc/grpc-java/blob/a2398ce5dbe2eacf841d7af9c9aa5e912e8db0aa/core/src/main/java/io/grpc/util/RoundRobinLoadBalancer.java#L82)), so there's only one connection created, but it still treats them as doubly weighted in terms of how it distributes RPCs across the addresses.  I think it would be consistent to do the same thing here.

For now, I think we can simply ignore duplicate addresses, since we're explicitly not supporting differently weighted endpoints.  But if we do add that later, we will want to treat this case as differently weighted endpoints.  (If we don't de-dup now, then it could cause behavior changes if we want to add support for differently weighted endpoints later.)

Given the different ways that address lists can be generated (DNS and EDS being two good examples), I think it makes sense for LB policies to understand both a weight attribute and the fact that a given address may be present in the list multiple times.  This is what we do in the ring_hash policy in C-core.

A48-xds-least-request-lb-policy.md
markdroth: Sounds good.  But let's explicitly state that it won't be supported in the doc, so that there's no confusion.

A48-xds-least-request-lb-policy.md
markdroth: Okay, sounds good.  I'm fine with not including the "all weights not equal" algorithm for now.

markdroth (commented):
Sorry for the delay in responding!  This got caught in my backlog after the Thanksgiving holiday.

A48-xds-least-request-lb-policy.md
erikjoh: Ahh, I see now, the EAGs in grpc-java have the same hashCode if the underlying addresses are equal. Indeed that would de-dup then :sweat_smile: 

> I think it would be consistent to do the same thing here.

@markdroth How would you like to phrase this in the proposal?
It's still not entirely clear to me if round_robin currently is consistent across languages in terms of handling duplicate addresses.

Maybe change the section further down to something like:

```
##### Duplicate addresses in resolution result

The `least_request_experimental` policy will be consistent with the `round_robin` policy in terms of handling duplicate addresses in the resolution result.

If support for the weighted least request algorithm is later added to the `least_request_experimental` policy, addresses should be de-duplicated and weighted accordingly just as the `ring_hash` policy does.
```

markdroth: It looks like all three languages de-dup to avoid creating redundant connections to the same address, but the weighting behavior is a little different.  Java and C-core both use it as a weighting indication, so that an address that is present twice in the list will get twice as many RPCs per pass through the list as an address that is present only once.  Go apparently does not currently do that weighting (i.e., an address that is present twice in the list will get just as many RPCs as an address that is present once), but @dfawley tells me he agrees that that's a bug, and he's noted it in grpc/grpc-go#2580 as something that needs to be fixed.  So I think we're generally in agreement about what the intended behavior is here.

In terms of the wording here, how about something like the following:

"""
In gRPC, if an address is present more than once in an address list, that is intended to indicate that the address should be more heavily weighted.  However, it should not result in more than one connection to that address.

In keeping with this, the `least_request_experimental` policy will de-duplicate the addresses in the address list it receives so that it does not create more than one connection to a given address.  However, because the policy will not initially support the weighted least request algorithm, we will ignore the additional weight given to the address.  If at some point in the future we do add support for the weighted least request algorithm, then an address that is present multiple times in the list will be given appropriate weight in that algorithm.
"""

erikjoh: Sound good!
Updated now.

ejona86 (approved):

markdroth (approved):
This looks great!

dfawley (approved):

PR #272, opened 2021-11-01
L89: C-core: Remove grpc_insecure_channel_create and grpc_server_add_insecure_http2_port from Core Surface API

yihuazhang:


L87-core-remove-grpc-insecure-channel-creation-api
```diff
@@ -0,0 +1,126 @@
...
+   TODO(hork): add channel_args to this API to allow endpoints and transports
+   created in this function to participate in the resource quota feature.
+   Note that this API currently only supports inseure server credentials
+   Using other types of credentials will result in a failure. */
+GRPCAPI void grpc_server_add_channel_from_fd(grpc_server* server,
+                                             void* reserved, int fd,
+                                             grpc_server_credentials* creds);
+
+```
+Fifth, remove `grpc_insecure_channel_create` and `grpc_server_add_insecure_http2_port` from `grpc.h`. After this change takes effect, the only way of creating gRPC insecure channels will be first creating insecure channel or server credentials, and then invoking `grpc_channel_create` or `grpc_server_add_http2_port` at client or server side, respectively.
```
sanjaypujare, line 110: It might be useful to note that this will have negligible or zero impact because this is a core API and not an end user API so users are likely to be affected by this removal (assuming true).

morgwai: i strongly agree with @sanjaypujare : when I first saw the thread on the mailing-list, I almost made my pants dirty and calmed down only after I saw his above comment.

yihuazhang: Thanks! I updated the text to reflect it.

sanjaypujare: The new text does acknowledge impact on users who "directly interact with Core surface APIs (e.g., by calling insecure channel creation API's)". Assuming such users exist, won't they benefit from deprecation first and then removal of the APIs?

yihuazhang: I do not think we need to keep backward-compatibility for C-core API users. Here is a quote from @markdroth  - "The C-core API is not a public API, and we don't offer a backward-compatibility guarantee for it. We are required to publish a gRFC for C-core API changes, but we are not required to make non-breaking changes"

markdroth: That's correct, we do not promise backward compatibility for the C-core API.

L89-core-remove-grpc-insecure-channel-creation-api
```diff
@@ -0,0 +1,128 @@
+Remove grpc_insecure_channel_create and grpc_server_add_insecure_http2_port from Core Surface API
```
markdroth, line 1: This filename needs a .md suffix so that github will format it correctly.

L89-core-remove-grpc-insecure-channel-creation-api
```diff
@@ -0,0 +1,128 @@
...
+* Rename `grpc_server_add_secure_http2_port` to `grpc_server_add_http2_port`.
+* Rename `grpc_secure_channel_create` to `grpc_channel_create` and re-order its parameters.
+* Move `grpc_channel_credentials` and `grpc_server_credentials`related API's from `grpc_security.h` to `grpc.h`.
+* Add `grpc_channel_create_from_fd` and `grpc_server_add_channel_from_fd` to Core Surface API.
+
+## Background
+
+Currently, the gRPC insecure build does not include any of the functions in `grpc_security.h`, nor does it include any of the code in `src/core/lib/security`. This means that it does not support channel or call credentials, security connectors, or client/server auth filters. It also does not link in SSL libraries.
+
+We could stop supporting insecure builds, so that all builds of gRPC include the functions in `grpc_security.h` and the code in `src/core/lib/security`.  However, the gRPC build would not directly include the SSL libraries.  Instead, there would be separate build targets for each of the credential types (both insecure and secure). Only those credential types that need SSL would pull in the SSL library dependencies.
```
markdroth, line 22: Please change the wording here to make it clear that we are actually making these changes.  Instead of "could" or "would", please use "will".

We should also explain why we're doing this.  Key points:
- The current approach requires different API entry points for secure vs. insecure channels and servers, and it requires a lot of conditional compilation that complicates our build system.
- The main reason for supporting insecure builds is because there are some users that don't need security but really need to minimize binary size (e.g., embedded systems), so they need a way to build without depending on the SSL libraries.
- The current approach was decided upon early in our development, before we had the credential APIs, and it was never updated when the credential APIs were added.  However, the credential APIs provide a much cleaner way to allow building gRPC without the SSL dependency.

To solve this problem, we are going to move to an approach where insecure channels and servers are handled the same way as secure channels and servers.  It will no longer be possible to build gRPC without support for credentials or to create a channel or server without specifying credentials, but we will provide an insecure credentials type to be used in cases where security is not needed.  The insecure credentials type will not depend on SSL libraries and will be included in all gRPC builds.  Other credential types that do depend on SSL libraries will be available in secure builds only, just as they are today.

L89-core-remove-grpc-insecure-channel-creation-api
```diff
@@ -0,0 +1,128 @@
...
+
+```
+/** Add a HTTP2 over an encrypted link over tcp listener.
+   Returns bound port number on success, 0 on failure.
+   REQUIRES: server not started */
+GRPCAPI int grpc_server_add_http2_port(grpc_server* server, const char* addr,
+                                       grpc_server_credentials* creds);
+```
+
+Fourth, add `grpc_channel_create_from_fd` and `grpc_server_add_channel_from_fd` to `grpc.h`. Note that the current APIs only support
```
markdroth, line 81: I think these functions should stay in grpc_posix.h. They're not supported on non-POSIX systems.

L89-core-remove-grpc-insecure-channel-creation-api
```diff
@@ -0,0 +1,128 @@
...
+either insecure channel credentials (at client-side) or insecure server credentials (at server-side), and using other types of
+credentials will result in a failure. In future, the API's will be improved to support other types of credentials.
```
markdroth, line 83: Please remove the apostrophe in "API's".

L89-core-remove-grpc-insecure-channel-creation-api
```diff
@@ -0,0 +1,128 @@
...
+   descriptor corresponding to a connected socket. Events from the file
+   descriptor may come on any of the server completion queues (i.e completion
+   queues registered via the grpc_server_register_completion_queue API).
+   The 'reserved' pointer MUST be NULL.
+   TODO(hork): add channel_args to this API to allow endpoints and transports
+   created in this function to participate in the resource quota feature.
+   Note that this API currently only supports inseure server credentials
+   Using other types of credentials will result in a failure. */
+GRPCAPI void grpc_server_add_channel_from_fd(grpc_server* server,
+                                             void* reserved, int fd,
```
markdroth, line 106: Let's get rid of this `reserved` parameter while we're changing this. I don't think we need it.

L89-core-remove-grpc-insecure-channel-creation-api
```diff
@@ -0,0 +1,128 @@
...
+```
+Fifth, remove `grpc_insecure_channel_create` and `grpc_server_add_insecure_http2_port` from `grpc.h`. After this change takes effect, the only way of creating gRPC insecure channels will be first creating insecure channel or server credentials, and then invoking `grpc_channel_create` or `grpc_server_add_http2_port` at client or server side, respectively.
+
+## Rationale
+
+* By removing gRPC insecure build, gRPC build system will be simplified. We do not need to maintain two implementations of certain features - one for secure build and another for insecure one. `grpclb_channel.cc` and `grpclb_channel_secure.cc` is an example.
+
+* From security perspective, if there is a need to restrict the usage of gRPC insecure channels, the problem is simplified to some extent because we only need to control a single means of creating insecure channels, instead of two.
+
+* Most importantly, since we introduce a different target for each credential type, for the binaries that do not rely on SSL-related functionalities, they do not need to bringin SSL dependency and thus will have a smaller binary size.
```
markdroth, line 118: s/bringin/bring in/

markdroth (commented):
Thanks for doing this!

L89-core-remove-grpc-insecure-channel-creation-api
yihuazhang: Done.

L89-core-remove-grpc-insecure-channel-creation-api
yihuazhang: Done.

L89-core-remove-grpc-insecure-channel-creation-api
yihuazhang: Done.

L89-core-remove-grpc-insecure-channel-creation-api
yihuazhang: Thanks for adding more contexts. I added them to the new text.

L89-core-remove-grpc-insecure-channel-creation-api.md
```diff
@@ -12,19 +12,33 @@ Remove grpc_insecure_channel_create and grpc_server_add_insecure_http2_port from
 * Remove `grpc_insecure_channel_create` and `grpc_server_add_insecure_http2_port` from Core Surface API.
 * Rename `grpc_server_add_secure_http2_port` to `grpc_server_add_http2_port`.
 * Rename `grpc_secure_channel_create` to `grpc_channel_create` and re-order its parameters.
-* Move `grpc_channel_credentials` and `grpc_server_credentials`related API's from `grpc_security.h` to `grpc.h`.
+* Move `grpc_channel_credentials` and `grpc_server_credentials`related APIs from `grpc_security.h` to `grpc.h`.
```
markdroth, line 15: Looks like you need to add a space before "related".

L89-core-remove-grpc-insecure-channel-creation-api.md
```diff
@@ -12,19 +12,33 @@ Remove grpc_insecure_channel_create and grpc_server_add_insecure_http2_port from
...
+libraries will be available in secure builds only, just as they are today.
+Also, the gRPC build will not directly include the SSL libraries.
+Instead, there will be separate build targets for each of the credential types, and
+only those credential types that need SSL will pull in the SSL library dependencies.
 
 
 ## Proposal
 
-First, move `grpc_channel_credentials` and `grpc_server_credentials`related API's from `grpc_security.h` to `grpc.h`.
+First, move `grpc_channel_credentials` and `grpc_server_credentials`related APIs from `grpc_security.h` to `grpc.h`.
```
markdroth, line 41: Need a space before "related" here too.

markdroth (approved):
Just a couple of remaining formatting nits; otherwise, this looks great!

@jtattermusch and @ctiller may also want to review.

As per gRFC comment policy, we'll need to wait until November 15th to merge this.

L89-core-remove-grpc-insecure-channel-creation-api.md
```diff
@@ -0,0 +1,142 @@
...
+* Discussion at: https://groups.google.com/g/grpc-io/c/9B4-AHCx_ZM 
+
+## Abstract
+
+* Remove `grpc_insecure_channel_create` and `grpc_server_add_insecure_http2_port` from Core Surface API.
+* Rename `grpc_server_add_secure_http2_port` to `grpc_server_add_http2_port`.
+* Rename `grpc_secure_channel_create` to `grpc_channel_create` and re-order its parameters.
+* Move `grpc_channel_credentials` and `grpc_server_credentials` related APIs from `grpc_security.h` to `grpc.h`.
+* Add `grpc_channel_create_from_fd` and `grpc_server_add_channel_from_fd` to Core Surface API.
+
```
jtattermusch, line 17: Listing the API changes you're planning is definitely a must, but since a big portion of the design discussions were also about the impact on the build targets, it probably makes sense to add a few sentences that describe the impact on existing build targets (e.g. `grpc_unsecure` and `grpc` which play the most important role here).
- let's mention the planned impact on the existing public bazel targets
- let's also mention the impact on the non-bazel targets (which are mostly represented by our cmake targets `grpc` and `grpc_unsecure`.

From you implementation PR, it looks like the `//:grpc` and `//:grpc_unsecure` targets will be kept and `//:grpc_unsecure` won't have a dependency on boringssl.
The cmake targets: it looks like we will still have `grpc` (with unconditional dependency on boringssl/openssl) and `grpc_unsecure` (NO  dependency on boringssl/openssl, also unconditionally).

If the above is correct, the proposal LGTM to me after adding the summary to the abstract.

L89-core-remove-grpc-insecure-channel-creation-api.md
```diff
@@ -0,0 +1,142 @@
...
+   TODO(hork): add channel_args to this API to allow endpoints and transports
+   created in this function to participate in the resource quota feature. */
+GRPCAPI void grpc_server_add_channel_from_fd(grpc_server* server,
+                                             int fd,
+                                             grpc_server_credentials* creds);
+
+```
+Fifth, remove `grpc_insecure_channel_create` and `grpc_server_add_insecure_http2_port` from `grpc.h`. After this change takes effect, the only way of creating gRPC insecure channels will be first creating insecure channel or server credentials, and then invoking `grpc_channel_create` or `grpc_server_add_http2_port` at client or server side, respectively.
+
+## Rationale
```
jtattermusch, line 126: I'm curious if the issue https://github.com/grpc/grpc/issues/27733 (which I mentioned in https://github.com/grpc/grpc/pull/25586#issuecomment-944154028) will be resolved by the changes in the implementation of this proposal. If so, that's very positive and it should be mentioned here.

If not, perhaps we should mention it in the "Open issues" section, since in our own codebase, we're not always honoring the mutual exclusivity of  grpc and grpc_unsecure and we do have targets (e.g. some tests) that transitively depend on both `grpc` and `grpc_unsecure`.

jtattermusch (commented):
LGTM after addressing my comments.

L89-core-remove-grpc-insecure-channel-creation-api.md
yihuazhang: > From you implementation PR, it looks like the //:grpc and //:grpc_unsecure targets will be kept and //:grpc_unsecure won't have a dependency on boringssl.
The cmake targets: it looks like we will still have grpc (with unconditional dependency on boringssl/openssl) and grpc_unsecure (NO dependency on boringssl/openssl, also unconditionally).

That's correct, Jan. I also updated the text to make it more clear. Thanks for pointing it out.

L89-core-remove-grpc-insecure-channel-creation-api.md
yihuazhang: Ah, I do not think there is a need to keep `_nosec` end2end tests any more as we removed all code related to insecure channel creation (e.g., `grpc_insecure_channel_create`) API in this gRFC, that are exercised by `_nosec` tests.

jtattermusch: I think that's a good point. If we no longer need the `_nosec` tests after merging this PR (do other reviewers agree with this statement?), we should definitely get rid of those tests in a followup PR. There's actually quite a bit of complexity in how the _SEC vs _NOSEC end2end tests get configured and run, so this cleanup will be very welcome.

jtattermusch (approved):

L89-core-remove-grpc-insecure-channel-creation-api.md
markdroth: Yes, I agree, we no longer need the `_nosec` tests.  I'm tempted to say that we should just remove those as part of Yihua's existing PR, but the PR is already quite large and invasive as it is, so I guess let's do that as a separate step.

yihuazhang:
@asraa, @yanavlasov I noticed `InsecureChannelCredentials` is used in envoy code base (e.g., [here](https://github.com/envoyproxy/envoy/blob/main/source/common/grpc/google_grpc_creds_impl.cc#L117-L119)), and I am wondering if the change in this gRFC (i.e., fail RPC if call credentials are sent over insecure channels) will break envoy tests. Any information will be appreciated.

yanavlasov:
> @asraa, @yanavlasov I noticed `InsecureChannelCredentials` is used in envoy code base (e.g., [here](https://github.com/envoyproxy/envoy/blob/main/source/common/grpc/google_grpc_creds_impl.cc#L117-L119)), and I am wondering if the change in this gRFC (i.e., fail RPC if call credentials are sent over insecure channels) will break any envoy tests. Any information will be appreciated.

I do not know for sure. There are some tests that use credentials here https://github.com/envoyproxy/envoy/blob/ae56bb42102df22401c5c0df9d3f8341b18acdfc/test/common/grpc/grpc_client_integration_test.cc#L450

If there is a simple way to do it, we can do dry-run and see how many tests will break in Envoy. But we can also just fix them when we pull a new version of gRPC.

yihuazhang:
> But we can also just fix them when we pull a new version of gRPC.
SG. PLMK if you find tests broken when importing the new version of gRPC, I am happy to help investigating.

yihuazhang:
I believe we can merge this proposal now.

PR #273, opened 2021-11-02
L88-cpp: Add conversion operators between absl::Status and grpc::Status.

ctiller:


markdroth (approved):
SGTM.

This would be the first place in the public C++ API where we refer to an absl type.  @veblush, any concerns about that?

markdroth:
(Note that we need to wait 2 weeks before merging this, as per gRFC policy.)

PR #275, opened 2021-11-16
Update readme and template to reflect actual process being followed

dfawley:


README.md
```diff
@@ -57,22 +62,16 @@ the final APPROVER is still ``a11r``.
...
-- ``#Gnnnn`` - Protocol level changes.
+- ``An`` - Affects all languages.
+- ``Pnn`` - Affects processes, such as the proposal process itself.
+- ``Lnnn`` - Language specific changes to external APIs or platform support.
+- ``Gnnnn`` - Protocol level changes.
+
+## Updating proposals
+When updating a proposal, the PR description should be named as follows:
+
+    $CategoryName## update: <description of change>``
```
markdroth, line 73: Please remove trailing back-ticks.

README.md
```diff
@@ -57,22 +62,16 @@ the final APPROVER is still ``a11r``.
...
-- ``#Pnn`` - Affects processes, such as the proposal process itself.
-- ``#Lnnn`` - Language specific changes to external APIs or platform support.
-- ``#Gnnnn`` - Protocol level changes.
+- ``An`` - Affects all languages.
+- ``Pnn`` - Affects processes, such as the proposal process itself.
+- ``Lnnn`` - Language specific changes to external APIs or platform support.
+- ``Gnnnn`` - Protocol level changes.
+
+## Updating proposals
+When updating a proposal, the PR description should be named as follows:
```
markdroth, line 71: Maybe we should say something about when and why we would update an existing gRFC instead of creating a new one?  For example, maybe something like "Sometimes, after a gRFC is approved, small changes are needed to the design.  In these cases, it is often simpler to modify the existing gRFC rather than create a new one."

README.md
```diff
@@ -1,52 +1,57 @@
...
+as new commits to the PR. The OWNER is encouraged to solicit as much feedback on
+the proposal as possible during this period.
 1. If there is consensus as deemed by the APPROVER during the comment period,
-the APPROVER will mark the proposal as final and assign it a gRFC number.
-Once this is assigned (as part of the closure of discussion), the OWNER will
-update the state of the PR as final and submit the PR.
-Commits must not be squashed; the commit history serves as a log of changes
-made to the proposal.
+the APPROVER will approve the PR on GitHub. The PR will then be merged by either
+the OWNER, if possible, or the APPROVER otherwise.
```
markdroth, line 54: Should we maybe explicitly say something here like "A proposal being merged indicates that it has been approved", just to make the state clear?

GRFC-TEMPLATE.md
```diff
@@ -2,21 +2,19 @@ Title
 ----
 * Author(s): [Author Name, Co-Author Name ...]
 * Approver: a11r
-* Status: {Draft, In Review, Ready for Implementation, Implemented}
```
markdroth, line 5: I'm thinking that we should probably go through the existing gRFCs and remove the "Status:" field, since it's pretty useless and misleading.  But we can do that in a separate PR.

markdroth (commented):
This looks good!  My comments are fairly minor.

GRFC-TEMPLATE.md
dfawley: Agreed.  I was thinking of the same, but maybe we should wait a month or two, since there are PRs pending with Status fields in them as well.

README.md
dfawley: It seems pretty clear from the preceding sentence to me...  I added a following sentence to remove any doubt.  PTAL.

README.md
dfawley: Done.

README.md
dfawley: Done (with slightly different wording).

dfawley (commented):
Thanks for the comments.  PTAL

GRFC-TEMPLATE.md
```diff
@@ -2,21 +2,19 @@ Title
...
 * Implemented in: <language, ...>
 * Last updated: [Date]
-* Discussion at: <google group thread> (filled after thread exists)
```
ejona86, line 8: I do think we need at least a notification to the grpc-io list. Yes, discussions rarely happen there, but the notification still seems fair. (This would be a bit different if GitHub had a way to watch for new PRs without also subscribing to those PRs.)

Also, removing the notification involved here seems over-stepping the sort of change we should do in this PR if we aren't going to go through a fuller process. Like, it seems such a change like removing the grpc-io threads needs a grpc-io thread. And we are currently creating threads, so it seems weird to remove it if this PR is to reflect the actual process.

README.md
```diff
@@ -1,78 +1,89 @@
...
-it can be updated to a status of ``Implemented`` with the implementing
-languages listed. (Listing versions is not required.)
+- ``An`` - Affects all languages.
+- ``Pnn`` - Affects processes, such as the proposal process itself.
+- ``Lnnn`` - Language specific changes to external APIs or platform support.
+- ``Gnnnn`` - Protocol level changes.
+
+## Updating proposals
+
+Sometimes small changes are needed to approved proposals.  Rather than create
```
ejona86, line 80: I do think we should have some hints for what "small" is.

> Sometimes small changes are needed to approved proposals. For example, editorial fixes, clarifying text, and detailed fixes noticed during implementation or usage. Rather than create new proposals for such changes, it is often better to revise the existing one.

GRFC-TEMPLATE.md
dfawley: Should we include a back-reference to the grpc-io thread from the gRFC though?  If we add this as a step to follow in the README, one could easily go to the grpc-io forum and search for the proposal name to find discussion about it.

Also, I would be in favor of removing it entirely, since the system we are following today (not as documented, but what we are doing) theoretically encourages discussion in two places.  But I'm fine with leaving it as part of the process in this PR.

README.md
dfawley: Honestly I think anything that's not considered a "major" change should be a modification of a gRFC instead of a new one.  A major change would be like an entirely different API or a wholly different mechanism for solving the same problem.  "Minor" as described below, IMO, is for editorial fixes and clarifying text, which is allowed to be approved without the 10 day waiting period.

dfawley: This is the original `README.md` text, for reference:

1. Once it is approved for submission by the arbiter, it goes into the ``Final`` state. Only minor changes are allowed (what qualifies as minor is left to the APPROVER).
1. If a proposal needs to be revisited, it can be moved back to the ``Draft`` or ``In Review`` state. This can happen if issues are discovered during implementation. At which point, the review process as described above must be followed.

GRFC-TEMPLATE.md
ejona86: I agree adding the back-reference is annoying, but I do think the need to add the backreference is what _causes_ the email to be sent to grpc-io. Without it I think it would often be forgotten. (And it won't be as obvious to the approver it is missing.)

README.md
ejona86: Oh, so that's where you got that language... I think then we should harken back to the original language.

> Only minor updates typically occur after a proposal is approved (and merged?). Minor updates do not require a comment period. What qualifies as minor is left to the APPROVER. If an update is needed and is not minor, then the process's 10 day comment period must be observed. This can happen if issues are discovered during implementation.
>
> When updating a proposal, the PR description should be named as follows:
>    $CategoryName## update: <description of change>

This is potentially more restrictive, but it also doesn't change what updates can occur and how compared to the existing text.

dfawley: > Only minor updates typically occur after a proposal is approved

Sorry, but I don't like your wording.  "Only minor updates typically occur" -- to say "only X is typical" is unusual and unnecessary.  The goal of the original process here was:

1. "Minor" changes to a gRFC are allowed with no 10-day discussion period.
2. Bigger changes to a gRFC are allowed with the 10-day discussion period.

I believe I expressed this adequately with my wording.  I've made a small change here to move stuff / simplify.  LMK if you disagree.

GRFC-TEMPLATE.md
dfawley: Re-added.

README.md
ejona86: I disagree with your interpretation of the original text. I think the "This can happen if issues are discovered during implementation" part is important, as it provides a scope/context of the type of change we're talking about. We're not talking about redesigning old gRFCs 5 years after it was implemented because we no longer like that approach. Such things should be a new gRFC. Your proposed text _encourages_ changing existing gRFCs, with no real limit on what sort of changes we'd expect to make.

dfawley: If a gRFC has been implemented we probably aren't going to be materially modifying it in place except possibly to correct a problem discovered later or add a necessary, small tweak.  I take the "This can happen if" as an example, not to mean "This can _only_ happen if".  IMO we should leave it up to our judgement on a case-by-case basis whether it's better to update a gRFC or create a new one.  That's realistically already what we are doing, and I don't see this minor wording change (done to remove now-unnecessary clauses with the removal of the "Status" field) affecting anything.

@markdroth, thoughts?

ejona86: > I take the "This can happen if" as an example, not to mean "This can only happen if".

I agree. But the example does provide a _guide_. If someone wanted to do a huge change we could say, "that's a bigger change than we'd expect for an approved gRFC" and they wouldn't be surprised, because you can see how far of a departure that huge change is from the example. That is gone in the new language, and the new language _encourages_ making the change to the existing gRFC without any limit, implied or explicit.

markdroth: I can see both sides of this.  I agree with Doug that we should ultimately leave the decision to the approver.  But I also agree with Eric that we don't want to leave this completely open-ended; I think we do want to make it clear that non-minor changes to an existing gRFC are discouraged.

I will note that we do have one example where we made significant changes to a gRFC after it was already implemented, which was the modifications to A51 in https://github.com/grpc/proposal/pull/345.  In hindsight, this really should have been done as a separate gRFC; I was the one who suggested modifying the existing gRFC, since I thought it would be a fairly simple change, but it turned out to be more complex than I expected.  I think that having written guidance discouraging that might have prompted someone to object before that PR got merged.

In general, my feeling is that any modification to an existing gRFC should be minor enough that it doesn't require the 10-day comment period.  I think that in most cases, if we're considering needing the 10-day comment period, that's probably a signal that we should be writing a new gRFC rather than modifying an existing one.  But I don't want to be prescriptive about this -- I do think we want to allow the approver to use their judgement.

I suggest the following wording:

"""
Sometimes changes are needed to already-merged proposals, such as when we discover problems during implementation.  Rather than create new proposals for such changes, it is often better to revise the existing one.  However, this approach may not be appropriate for larger changes, especially after the design has already been implemented in most languages.

To make changes to an already-merged proposal, a PR may be approved and merged at the OWNER's and APPROVER's convenience.  Note that the APPROVER will decide whether this kind of post-merge change is appropriate.  The approver will also determine whether the 10-day comment period is necessary, based on how significant the change is.
"""

dfawley: Seems fine to me.  Updated.

markdroth (approved):
Looks great!  Let's make sure Eric is happy with it, and then get it merged.

GRFC-TEMPLATE.md
```diff
@@ -2,7 +2,6 @@ Title
 ----
 * Author(s): [Author Name, Co-Author Name ...]
 * Approver: a11r
-* Status: {Draft, In Review, Ready for Implementation, Implemented}
 * Implemented in: <language, ...>
 * Last updated: [Date]
```
markdroth, line 6: Nit: Can we change "[Date]" to "YYYY-MM-DD", to make it clear that we expect ISO format here?

GRFC-TEMPLATE.md
```diff
@@ -2,7 +2,6 @@ Title
 ----
 * Author(s): [Author Name, Co-Author Name ...]
 * Approver: a11r
```
markdroth, line 4: Let's change the default approver to be "@markdroth".

GRFC-TEMPLATE.md
ctiller: Couldn't we just write a bot that took care of pushing these emails?

GRFC-TEMPLATE.md
dfawley: Done but I left out the "@".  In case you were curious, here are all the current "Approver" field values:

```
     34 markdroth
     24 a11r
     14 wenbozhu
     14 @ejona86, @dfawley
      7 gnossen
      6 vjpai
      5 @markdroth
      5 ejona86
      4 murgatroid99
      4 ctiller
      3 psrini
      3 [Eric Anderson](https://github.com/ejona86)
      3 ejona86, dfawley
      3 @ejona86
      2 srini100
      2 nicolasnoble
      2 mehrdada
      2 Mark Roth (@markdroth)
      2 markdroth, vjpai
      2 lidizheng
      2 ericgribkoff
      1 wenbozhu
      1 voidzcy
      1 vjpai, yashykt
      1 soheilhy
      1 sanjaypujare
      1 roth@google.com
      1 roth
      1 mxyan
      1 mehrdada / apolcyn
      1 Mark Roth (@markdroth), Craig Tiller (@ctiller)
      1 Mark Roth
      1 [@markdroth](https://github.com/markdroth)
      1 markdroth, ejona86, dfawley
      1 markdroth, ejona86
      1 @markdroth , @ctiller
      1 ** @markdroth
      1 mark
      1 lidizheng, gnossen
      1 jtattermusch and markdroth
      1 jtattermusch
      1 Jan Tattermusch & Wenbo Zhu
      1 gnossen, rmariano
      1 Eric Anderson
      1 ejona
      1 [Doug Fawley](https://github.com/dfawley)
      1 dapengzhang0
      1 @ctiller
      1 chwarr, jiagtaoli2016, vjpai, markdroth,
      1 AspirinSJL
      1 apolcyn
      1 a11r, psrini
```

GRFC-TEMPLATE.md
dfawley: Done

GRFC-TEMPLATE.md
dfawley: I don't feel strongly here, but the vast majority of usages elide the "@".  The character has no actual function AFAICT, but I guess it's an indication that the following letters probably make up a github username.

ctiller (approved):

ejona86 (approved):

PR #276, opened 2021-11-22
A49: Dynamic Connection Scaling

dfawley:
Just reserving the proposal number for now.


murgatroid99:
If I understand this design right, the client needs to get a service config to enable this feature. In my experience, our users often encounter this limit when talking to GFE servers while using Google Cloud API client libraries, or similar. For those users, what do we expect to be the source of the service config?

dfawley:
Default service config (and optionally disabling name resolver service config) can also be injected by the application.  Client libraries already does this, I believe, for other SC features.  This is how we plan to enable it there.

murgatroid99:
If the expected use case is for users to set this option statically at channel creation time, why not make it just a channel argument?

ejona86:
The service provider should be in control of the setting, which means service config is the ideal location.

A49-dynamic-connection-scaling.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Subchannel Sharing
+
+Subchannels may be shared across channels.  The `maxConnectionsPerSubchannel`
+setting will be applied to a shared subchannel and that subchannel will use the
+highest maximum of all channels currently sharing it.  If channels sharing the
+subchannel are closed, then the remaining channels' settings will be combined to
+determine the new max.  If the max is lowered to less than the number of open
+connections, the excess connections will not be closed as a result.
+Down-scaling will only happen as connections are lost naturally.
```
murgatroid99, line 169: I think this part can be simplified. In Node, and I think also Core, subchannels are only shared among channels with the same target and channel arguments. Two such channels should never have different service configs except in brief windows when updates may propagate to separate resolvers at slightly different times. So, there probably isn't much value in tracking which setting value is associated with which channel, and instead we can just use the latest value the subchannel has seen from any channel.

murgatroid99:
OK, then do any of the libraries already have a mechanism for propagating service config updates to subchannels?

ejona86 (approved):

PR #279, opened 2021-11-29
Add extension flags to the shutdown transaction

jdcormie:


markb74 (commented):
I mildly prefer not sharing the "protocol extension flags" concept across setup and shutdown transactions, and instead just adding "shutdown flags" here. We'll definitely use protocol extension flags soonish (for improved flow control) so I'd prefer if that concept remained solely about protocol negotiation 

I'll defer to Eric's judgement here since adding extension flags was his suggestion in the original grfc review.

jdcormie:
Updated to make "shutdown flags" their own thing and leave "protocol extension flags" alone.

markb74 (approved):

PR #280, opened 2021-11-30
Update the C++ API for connection drain grace time

yashykt:


ejona86 (approved):

A36-xds-for-servers.md
```diff
@@ -432,11 +433,18 @@ class XdsServerServingStatusNotifierInterface {
 
 class XdsServerBuilder : public ::grpc::ServerBuilder {
  public:
+  // Sets the drain grace period in ms for older connections when updates to a
```
ejona86, line 436: s/older/existing/ may be more clear.

markdroth:
We decided to do this via a channel arg, so this isn't necessary anymore.

yashykt:
I think we should mention this API anyway.

markdroth (approved):

PR #281, opened 2021-12-01
A50: gRPC xDS Outlier Detection Support

murgatroid99:


A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Failure Percentage Algorithm
+
+ 1. If the number of addresses is less than `failure_percentage_minimum_hosts`, stop.
+ 2. For each address:
+    1. If the percentage of ejected addresses is greater than `max_ejection_percent`, stop.
+    2. If the address's total request volume is less than `failure_percentage_request_volume`, continue to the next address.
+    3. If the address's failure percentage is greater than `failure_percentage_threshold`, then choose a random integer in `[0, 100)`. If that number is less than `enforcing_failure_percentage`, eject that address.
+
+### xDS Integration
```
markdroth, line 160: In this section, we should be very specific about how each of the fields in the xDS outlier detection config will be handled by gRPC.  In particular:
- We should explicitly list all of the xDS fields that we will look at, what values are legal, and how we will handle any illegal values in each field -- e.g., will we NACK or ignore the value or what?
- We should be explicit about which fields in the xDS config we will *not* support, and how we will handle those fields.  In the case of this particular design, I suspect we can just say something like "any field not listed above will be ignored by gRPC".

We need to pay careful attention to forward-compatibility here, especially in cases where we might decide to NACK.  We should carefully think through what would happen if in the future we add support for new fields and make sure that the behavior that we start with will not cause any problems.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+
+If the `Cluster` message for an underlying cluster has the `outlier_detection` field set, the `xds_cluster_resolver` will generate an `outlier_detection` LB policy config from it for each priority as the top-level LB policy for that priority. In that case, the child policy config will be the config that would otherwise be at the top level for the priority. When `outlier_detection` is configured, the LB policy tree will look like this when using round robin, and the `outlier_detection` LB policy will be inserted similarly when using ring hash:
```
markdroth, line 162: If we insert the `outlier_detection` LB policy in the tree only when outlier detection is configured in xDS, then a config change that enables or disables outlier detection will cause churn in the LB policy tree.  This is more of an issue in Java and Go than it is in C-core and Node, because the former two implementations do not have a subchannel pool and will therefore have to reconnect to all of the endpoints when this reconfiguration occurs.  But even in C-core and Node, there may be cases where child policies will lose some state when this reconfiguration occurs, which may not be optimal.

We could avoid this by saying that we always insert the outlier detection LB policy, even when outlier detection is not configured, in which case the policy is a no-op.  But this would come at the cost of having that policy in the tree even when it's not needed.

@ejona86 and @dfawley, thoughts?

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+  repeated LoadBalancingConfig child_policy = 13;
+}
+```
+
+### Outlier Detection Behavior
+
+Internally, the `outlier_detection` LB policy will store a map with keys that are subchannel addresses, and values that are objects that contain the following:
+
+ - The call result counter object (see [Call Counter section](#call-counter))
+ - The latest ejection timestamp, or `null` if the address is currently not ejected
```
markdroth, line 112: Could just set the timestamp to a representation of infinite-future.  In C-core, I would expect that to be fairly idiomatic, and it would not require using `absl::optional<>`.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+ 2. For each address:
+    1. If the percentage of ejected addresses is greater than `max_ejection_percent`, stop.
+    2. If the address's total request volume is less than `failure_percentage_request_volume`, continue to the next address.
+    3. If the address's failure percentage is greater than `failure_percentage_threshold`, then choose a random integer in `[0, 100)`. If that number is less than `enforcing_failure_percentage`, eject that address.
+
+### xDS Integration
+
+If the `Cluster` message for an underlying cluster has the `outlier_detection` field set, the `xds_cluster_resolver` will generate an `outlier_detection` LB policy config from it for each priority as the top-level LB policy for that priority. In that case, the child policy config will be the config that would otherwise be at the top level for the priority. When `outlier_detection` is configured, the LB policy tree will look like this when using round robin, and the `outlier_detection` LB policy will be inserted similarly when using ring hash:
+
+![gRPC Client Architecture Diagram](A50_graphics/grpc_xds_client_architecture.png)
```
markdroth, line 164: In the diagram, the red arrow from the `xds_cluster_impl` policy in priority 0 starts from the wrong place.

A50-xds-outlier-detection.md
markdroth, line 162: For the last sentance in this paragraph, I suggest saying:

"""
When outlier detection is configured, the `outlier_detection` policy will be inserted above the `xds_cluster_impl` policy.  For example, when using round robin, the LB policy tree will look like this:
"""

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+
+[Link to SVG file](A50_graphics/grpc_xds_client_architecture.svg)
+
+### Temporary environment variable protection
+
+During initial development, the xDS part of this feature will be enabled by the `GRPC_XDS_EXPERIMENTAL_ENABLE_OUTLIER_DETECTION` environment variable. This
```
markdroth, line 170: I think we'll want the same env var to guard even the non-xDS part of this feature, so let's remove the `XDS_` from the env var name.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+   - If a priority switchover occurs in the middle of an interval, the addresses in both priorities will have some traffic. This situation will always be temporary, so the per-priority logic will be correct starting in the following interval.
+   - The xDS specification of `failure_percentage_minimum_hosts` says that it counts all hosts in a cluster without considering traffic. As a result, looking at only individual priorities will result in gRPC not applying the failure percentage algorithm in some situations when Envoy would.
+ - With the introduction of aggregate cluster support in gRFC A37, there is not always an LB policy corresponding to the entirety of every underlying cluster. Instead, the aggregate and underlying cluster tree is flattened to a single priority list with all of the priorities for every cluster. So, there is no clear place to insert logic that applies to all of the addresses in a single underlying cluster. We considered adding a global map for outlier detection LB policies in the same cluster to share data, but it would add a lot of complexity and the returns would be minimal.
```
markdroth, line 198: Of the three bullets here, this bullet is the only one that actually explains why we are choosing to do this on a per-priority basis.  The other two are justifications why we think the impact of that decision will not be that bad, but they're not reasons for making that decision in the first place.

Given that, I suggest moving this bullet into the initial paragraph, then having a separate paragraph for each of the other two bullets.

markdroth, line 198: Please make "gRFC A37" a link to https://github.com/grpc/proposal/blob/master/A37-xds-aggregate-and-logical-dns-clusters.md.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+  - If the xDS LB policy tree ever changes to consistently have an LB policy that represents each underlying cluster, the xDS logic in this proposal can easily be modified to put the `outlier_detection` policy at that part of the tree and make it act over the whole cluster.
+
+### Non-xDS-specific LB policy
+
+We decided to implement outlier detection as a generic LB policy instead of making it xDS-specific, possibly as part of the `xds_cluster_impl` policy. The outlier detection logic has no real dependency on xDS, and making it generic in this way allows us to provide its value to more of our users.
+
+## Implementation
+
+Implementation will start in Java and Node, and will happen later in C++ and Go.
```
markdroth, line 207: We're actually planning to implement this in all languages this quarter.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
+A50: gRPC xDS Outlier Detection Support
+----
+* Author(s): murgatroid99
+* Approver: markdroth
+* Implemented in: C++, Java, Go, Node
```
markdroth, line 5: It's not actually implemented in any languages yet, so we should probably not list them yet.

markdroth (commented):
This looks really good!  Comments are mostly minor.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+
+  // The maximum time that an address is ejected for. If not specified, the default value (300000ms or 300s) or
+  // the base_ejection_time value is applied, whatever is larger.
+  google.protobuf.Duration max_ejection_time = 3;
+
+  // The maximum % of an address list that can be ejected due to outlier
+  // detection. Defaults to 10% but will eject at least one address regardless of the value.
+  google.protobuf.UInt32Value max_ejection_percent = 4;
+
+  // This factor is used to determine the ejection threshold for success rate
```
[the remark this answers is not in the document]
murgatroid99: One thing to consider here is that the current design accepts a configuration object that is empty other than the `child_policy`, which will result in all fields having their default values. Notably, that will effectively enable `success_rate` ejection and disable `failure_percentage`, because the default `enforcing_success_rate` is 100 and the default `enforcing_failure_percentage` is 0. Do we want to preserve that?

Also, as you wrote it, at least one of `success_rate_ejection` or `failure_percentage_ejection` must be set, but that object can contain all `null` fields, which I assume should be functionally equivalent to not setting the outer field at all. Was that your intention?

A50-xds-outlier-detection.md
murgatroid99: The algorithm does not just check the ejection time relative to the current time, it also has different behavior depending on whether an address currently is ejected or not. I found it useful to have a sentinel value to check for that, and having that be `null` clearly distinguishes it from a valid timestamp. Alternatively, we could have an additional boolean `ejected` field, and ignore this timestamp if it is false.

murgatroid99: In other words, I think that the behavior requires logic that is equivalent to interacting with an `optional` anyway, and I think that actually having an `optional` is a good way to express it.

markdroth: Okay, I'll buy that argument.

A50-xds-outlier-detection.md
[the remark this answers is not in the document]
markdroth: My general concern here is that since we are going to support this for both xDS and non-xDS use-cases, I think we should structure the LB policy config in a way that makes sense for the non-xDS use-case, since that's the one that users will interact directly with.  In the xDS use-case, we'll have machinery to convert from the xDS config to whatever form this LB policy config uses, so it will be fully automated, and we can handle more complexity there if we need to.

So given that, I think we should choose both a structure and a set of defaults that are most useful.  It does seem a little arbitrary to me to enable one ejection algorithm and disable another one by default.  It seems like it might be useful to instead have the user explicitly choose which algorithm(s) they want to use, and then assume reasonable defaults for the parameters for those algorithms.  So, for example, we would not enable success rate unless the message was present, but we would not require the `enforcing_success_rate` to be set since its default is 100.  But we could do the same thing for failure percentage by not enabling it unless the message was present, and then assuming a default of 100 for `enforcing_failure_percentage`.

I'm open to counter-arguments here if there's reason to believe that Envoy's defaults are actually the ones we want.  But I have enough experience with Envoy to not want to *assume* that those defaults are the right ones without thinking it through. :)

A50-xds-outlier-detection.md
murgatroid99: One concern I have is with the "Consecutive 5xx" ejection algorithm, which is enabled by default in the xDS configuration. If a user does not explicitly configure it, then the behavior will still change if we add support for it. I don't know if that's a significant problem, and it's also unlikely that we will ever support it.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
+A50: gRPC xDS Outlier Detection Support
+----
+* Author(s): murgatroid99
+* Approver: markdroth
+* Implemented in: C++, Java, Go, Node
+* Last updated: 2021-11-30
+* Discussion at: <google group thread> (filled after thread exists)
```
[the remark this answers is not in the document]
murgatroid99: Added.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,207 @@
...
+
+This proposal builds on earlier work described in the following gRFCs:
+* [gRFC A27: xDS-based Global Load Balancing](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)
+* [gRFC A30: xDS v3 Support](https://github.com/grpc/proposal/blob/master/A30-xds-v3.md)
+* [gRFC A37: xDS Aggregate and Logical DNS Clusters](https://github.com/grpc/proposal/blob/master/A37-xds-aggregate-and-logical-dns-clusters.md)
+* [gRFC A42: xDS Ring Hash LB Policy](https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md)
+
+## Proposal
+
+We will add a new `outlier_detection` load balancing policy that implements the outlier detection algorithm. This load balancing policy itself is not specific to xDS, but the xDS LB policy tree will use it.
```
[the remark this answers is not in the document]
murgatroid99: Done. I just used what you said.

A50-xds-outlier-detection.md
murgatroid99: Done. I just used what you said.

A50-xds-outlier-detection.md
murgatroid99: I rearranged this as you suggested.

A50-xds-outlier-detection.md
murgatroid99: Done.

A50-xds-outlier-detection.md
murgatroid99: Updated.

A50-xds-outlier-detection.md
murgatroid99: Done.

A50-xds-outlier-detection.md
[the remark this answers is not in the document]
murgatroid99: This seems reasonable to me. We could even allow both fields to be unset to make it easier to create a no-op outlier detection config like you mentioned in your other comment.

A50-xds-outlier-detection.md
markdroth: I think that particular case is probably okay.  As you say, we're unlikely to ever support that.  And even if we did, (a) there's no requirement that gRPC has the same defaults as Envoy, and (b) we could probably even get away with enabling this by default, since it takes effect only when the backends are failing requests anyway.  So we'd have some options here.

A50-xds-outlier-detection.md
```diff
@@ -159,16 +163,20 @@ The subchannel wrapper will track the latest state update from the underlying su
...
 
 ![gRPC Client Architecture Diagram](A50_graphics/grpc_xds_client_architecture.png)
 
 [Link to SVG file](A50_graphics/grpc_xds_client_architecture.svg)
 
+
+#### Validation
+
+The `google.protobuf.Duration` fields `interval`, `base_ejection_time`, and `max_ejection_time` must obey the restrictions in the [`google.protobuf.Duration` documentation](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration) and they must have non-negative values. The fields `max_ejection_percent`, `enforcing_success_rate`, `failure_percentage_threshold`, and `enforcing_failure_percentage` must have values less than or equal to `100`.
```
markdroth, line 175: Please be explicit that if any of these "must"s are violated, we will NACK the xDS resource.

murgatroid99: Done.

A50-xds-outlier-detection.md
murgatroid99: I went ahead and made this change.

A50-xds-outlier-detection.md
```diff
@@ -55,53 +55,69 @@ message OutlierDetectionLoadBalancingConfig {
...
+    // deviation of the mean success rate: mean - (stdev *
+    // success_rate_stdev_factor). This factor is divided by a thousand to get a
+    // double. That is, if the desired factor is 1.9, the runtime value should
+    // be 1900. Defaults to 1900.
+    google.protobuf.UInt32Value stdev_factor = 1;
+
+    // The % chance that an address will be actually ejected when an outlier status
+    // is detected through success rate statistics. This setting can be used to
+    // disable ejection or to ramp it up slowly. Defaults to 100.
+    google.protobuf.UInt32Value enforcing_success_rate = 2;
```
markdroth, line 74: Suggest calling this `enforcement_percentage`.

A50-xds-outlier-detection.md
```diff
@@ -55,53 +55,69 @@ message OutlierDetectionLoadBalancingConfig {
...
+  message FailurePercentageEjection {
+    // The failure percentage to use when determining failure percentage-based outlier detection. If
+    // the failure percentage of a given address is greater than or equal to this value, it will be
+    // ejected. Defaults to 85.
+    google.protobuf.UInt32Value threshold = 1;
+
+    // The % chance that an address will be actually ejected when an outlier status is detected through
+    // failure percentage statistics. This setting can be used to disable ejection or to ramp it up
+    // slowly. Defaults to 100.
+    google.protobuf.UInt32Value enforcing_failure_percentage = 2;
```
markdroth, line 102: Same here: Suggest calling this `enforcement_percentage`.

A50-xds-outlier-detection.md
```diff
@@ -126,10 +142,18 @@ The `outlier_detection` LB policy will provide a picker that delegates to the ch
 The `outlier_detection` LB policy will have a timer that triggers on a period determined by the `interval` config option, and does the following:
 
  1. For each address, swap the call counter's buckets in that address's map entry.
- 2. Run the [success rate](#success-rate-algorithm) and [falure percentage](#failure-percentage-algorithm) algorithms. To eject an address, it will set the address's ejection timestamp to the current time, increment the multiplier, and call `eject()` on each subchannel wrapper.
- 3. For each address in the map:
+ 2. If the `success_rate_ejection` configuration field is set, run the [success rate](#success-rate-algorithm) algorithm.
+ 3. If the `failure_percentage_ejection` configuration field is set, run the [falure percentage](#failure-percentage-algorithm) algorithm.
+ 4. Run the [success rate](#success-rate-algorithm) and [falure percentage](#failure-percentage-algorithm) algorithms.
```
markdroth, line 147: This should now be removed in favor of the previous two, right?

A50-xds-outlier-detection.md
markdroth: Okay.  I'd like feedback from Eric and Doug before finalizing, though.  The counter-argument to what I said above is that config changes to enable or disable outlier detection are probably quite rare, and it may not be worth imposing the extra overhead of the outlier detection policy when it's not being used.

markdroth (commented):
This looks really good!  Just one really noteworthy issue left to resolve, which will probably have to wait until Eric is back.

A50-xds-outlier-detection.md
murgatroid99: Yes, I took it out now.

A50-xds-outlier-detection.md
dfawley: SGTM; the overhead of the extra policy is pretty low, and if the reconnections are triggered, they will all happen at once, which seems pretty invasive.

dfawley (approved):

A50-xds-outlier-detection.md
ejona86: I'd prefer the tree to be as static as possible. That means config changes would be less invasive (less likely to trigger bug) and fewer code paths to consider for an RPC. The noop cost should be low, and could be lower than using it conditionally (because of branch prediction and JITs noticing a call site only calls one particular implementation).

As much as I like "not paying for things you aren't using," xDS _is_ using outlier detection, so I'd consider avoiding it premature optimization.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,258 @@
...
+
+The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, increment the corresponding counter in the map entry referenced by the subchannel wrapper that was picked.
+
+The `outlier_detection` LB policy will have a timer that triggers on a period determined by the `interval` config option, and does the following:
+
+ 1. For each address, swap the call counter's buckets in that address's map entry.
+ 2. If the `success_rate_ejection` configuration field is set, run the [success rate](#success-rate-algorithm) algorithm.
+ 3. If the `failure_percentage_ejection` configuration field is set, run the [falure percentage](#failure-percentage-algorithm) algorithm.
+ 4. For each address in the map:
+    - If the address is not ejected and the multiplier is greater than 0, decrement the multiplier.
```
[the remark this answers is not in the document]
murgatroid99: Done

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,258 @@
...
+    - If the address is ejected, and the current time is after `ejection_timestamp + min(base_ejection_time * multiplier, max(base_ejection_time, max_ejection_time))`, un-eject the address.
+
+### Ejection and Un-ejection
+
+To eject an address, set the current ejection timestamp to the current time, increment the ejection time multiplier, and call `eject()` on each subchannel wrapper in that address's subchannel wrapper list.
```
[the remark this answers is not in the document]
murgatroid99: Done.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,258 @@
...
+* [gRFC A42: xDS Ring Hash LB Policy](https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md)
+
+## Proposal
+
+xDS outlier detection supports configuring many different ejection algorithms. We plan to support only two of them, [success rate](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier#success-rate) and [failure percentage](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier#failure-percentage). See the "Rationale" section below for why we are not supporting the other algorithms.
+
+The outlier detection functionality will be provided via a new `outlier_detection` LB policy, which delegates to a child policy to perform the actual load balancing. The outlier_detection policy will monitor the results of calls to the endpoints that the child dispatches requests to, and it will use that data to eject endpoints based on the configured ejection algorithms and thresholds. The LB policy will eject endpoints by having their subchannels report `TRANSIENT_FAILURE` to the child policy.
```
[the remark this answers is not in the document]
murgatroid99: I clarified in the "Temporary environment variable protection" section that the environment variable protection also applies to registering the LB policy itself, as suggested by Mark in [this comment](https://github.com/grpc/proposal/pull/281#discussion_r791080154). As a result, I think we do not also need the "experimental" name in the LB policy.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,258 @@
...
+    1. If the percentage of ejected addresses is greater than `max_ejection_percent`, stop.
+    2. If the address's total request volume is less than `failure_percentage_ejection.request_volume`, continue to the next address.
+    3. If the address's failure percentage is greater than `failure_percentage_ejection.threshold`, then choose a random integer in `[0, 100)`. If that number is less than `failiure_percentage_ejection.enforcement_percentage`, eject that address.
+
+### xDS Integration
+
+If the `Cluster` message for an underlying cluster has the `outlier_detection` field set, the `xds_cluster_resolver` will generate an `outlier_detection` LB policy config from it for each priority as the top-level LB policy for that priority. In that `outlier_detection` field message, if the `enforcement_percentage` field is set to 0, the config `success_rate_ejection` field will be `null` and all `success_rate_*` fields will be ignored. If the `enforcing_failure_percent` field is set to 0 or `null`, the config `failure_percent_ejection` field will be `null` and all `failure_percent_*` fields will be ignored. Then the message fields will be mapped to config fields as follows:
```
markdroth, line 189: Since we've decided to always add the `outlier_detection` policy in the tree when we're using xDS, we should remove the words "If the Cluster message for an underying cluster has the outlier_detection field set", since that makes it sound conditional.

markdroth (commented):
Just one minor remaining wording nit.  Otherwise, looks great!

A50-xds-outlier-detection.md
murgatroid99: It is conditional, in the sense that if the field is not set, the procedure described here is not followed, and the no-op config is generated instead. That's described in the next paragraph, but that's after the list so I can understand if it doesn't seem connected.

markdroth: Ah, I see.  Yeah, it's a bit hard to follow that way for a first-time reader.  I think it would be clearer to say something like this:

"""
The `xds_cluster_resolver` LB policy will configure an `outlier_detection` LB policy instance for each priority as the top-level LB policy for that priority.  The configuration for that `outlier_detection` policy will be based on the `outlier_detection` field in the `Cluster` resource.

If the `outlier_detection` field is not set in the `Cluster` resource, a "no-op" `outlier_detection` config will be generated, with `interval` set to the maximum possible value and all other fields unset.

If the `outlier_detection` field *is* set in the `Cluster` resource, if the `enforcement_percentage` field is set to 0, the config `success_rate_ejection` field will be `null` and all `success_rate_*` fields will be ignored. If the `enforcing_failure_percent` field is set to 0 or `null`, the config `failure_percent_ejection` field will be `null` and all `failure_percent_*` fields will be ignored. Then the message fields will be mapped to config fields as follows:
"""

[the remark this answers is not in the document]
murgatroid99: Since `outlier_detection` is a field in the `Cluster` message for the underlying cluster, does that mean that this information should be included in the `DiscoveryMechanism` message in the `xds_cluster_resolver` LB policy config? Also, would it be reasonable to just add an `outlier_detection` field in that config with a message type that is almost identical to the actual `outlier_detection` LB policy config except for the `child_policy` field?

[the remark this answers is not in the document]
markdroth: Yes, you're correct, it will need to be inside of the `DiscoveryMechanism` message.

For the proto message, I see a few options:
1. Duplicate the `outlier_detection` LB policy config except for the `child_policy` field.  The down-side of this is that it would require essentially maintaining two copies going forward, although this message probably won't change very frequently, so that may not be a big deal.
2. Change the `outlier_detection` LB policy config proto such that all of the fields except `child_policy` are in a separate message, so that we can reuse that message in both places.  The down-side of this is that it adds another level of nesting in the common case that users actually have to deal with in the non-xDS case.
3. Keep the `outlier_detection` LB policy config as-is and re-use it completely, just adding a comment that the `child_policy` field will be ignored in the `xds_cluster_resolver` config.  This is a little ugly in terms of proto semantics, but since our code actually parses these configs in JSON form instead of proto form, that ugliess won't make any practical difference.

My inclination would be to go with option 3.

Note that in the implementation, we will want to have one piece of code to parse all of the common fields, rather than duplicating that functionality in both policies.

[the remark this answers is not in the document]
murgatroid99: I made an update addressing this comment and the other one. How does it look now?

A50-xds-outlier-detection.md
```diff
@@ -186,7 +186,74 @@ The subchannel wrapper will track the latest state update from the underlying su
...
+  // priorities.
+  repeated DiscoveryMechanism discovery_mechanisms = 1;
+
+  // Locality-picking policy.
+  // This policy's config is expected to be in the format used
+  // by the weighted_target policy.  Note that the config should include
+  // an empty value for the "targets" field; that empty value will be
+  // replaced by one that is dynamically generated based on the EDS data.
+  // Optional; defaults to "weighted_target".
+  repeated LoadBalancingConfig locality_picking_policy = 2;
```
markdroth, line 242: These two fields are actually no longer used; this was updated by [gRFC A42](https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md#change-child-policy-config-generation-in-xds_cluster_resolver-policy).

I think we should probably just bypass this kind of issue by listing only the new field to be added, not showing any of the additional existing fields.

As a follow-up, you can put together a PR for the grpc/grpc-proto repo to make the changes to the service config protos.

markdroth (commented):
Just one minor thing to be updated.  This looks really good!

A50-xds-outlier-detection.md
murgatroid99: Fixed. The follow-up PR is at grpc/grpc-proto#104.

markdroth (approved):
This looks great!

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,262 @@
...
+### Outlier Detection Behavior
+
+Internally, the `outlier_detection` LB policy will store a map with keys that are subchannel addresses, and values that are objects that contain the following:
+
+ - The call result counter object (see [Call Counter section](#call-counter))
+ - The latest ejection timestamp, or `null` if the address is currently not ejected
+ - The current ejection time multiplier, starting at 0
+ - A list of subchannel wrapper objects that correspond to this address
+
+When the `outlier_detection` LB policy receives an address update, it will create a map entry for each subchannel address in the list, and remove each map entry for a subchannel address not in the list, then pass the address list along to the subchannel.
```
ejona86, line 136: Is the final word "subchannel" supposed to be "subpolicy"?

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,262 @@
...
+ 1. For each address, swap the call counter's buckets in that address's map entry.
+ 2. If the `success_rate_ejection` configuration field is set, run the [success rate](#success-rate-algorithm) algorithm.
+ 3. If the `failure_percentage_ejection` configuration field is set, run the [falure percentage](#failure-percentage-algorithm) algorithm.
+ 4. For each address in the map:
+    - If the address is not ejected and the multiplier is greater than 0, decrease the multiplier by 1.
+    - If the address is ejected, and the current time is after `ejection_timestamp + min(base_ejection_time * multiplier, max(base_ejection_time, max_ejection_time))`, un-eject the address.
+
+### Ejection and Un-ejection
+
+To eject an address, set the current ejection timestamp to the current time, increase the ejection time multiplier by 1, and call `eject()` on each subchannel wrapper in that address's subchannel wrapper list.
```
ejona86, line 153: Implementation note: You probably don't want to use the current time _now_ when doing this. You probably want to grab the current time when you start the "loop over each address" process. Otherwise you can have aliasing issues where you are 1ms before the ejection window would expire. Grabbing the time at the start of the loop means the time is always _a bit older_ and the comparisons shouldn't offer any surprise.

Alternatively, you could subtract a fixed amount of time (e.g., 1% or 50% of `OutlierDetectionLoadBalancingConfig.interval`) when setting the value.

A50-xds-outlier-detection.md
```diff
@@ -0,0 +1,262 @@
...
+The `child_policy` config will be the `xds_cluster_impl` policy config that previously would have been at the top level for the priority. For example, when using round robin, the LB policy tree will look like this:
+
+![gRPC Client Architecture Diagram](A50_graphics/grpc_xds_client_architecture.png)
+
+[Link to SVG file](A50_graphics/grpc_xds_client_architecture.svg)
+
+
+#### Validation
+
+The `google.protobuf.Duration` fields `interval`, `base_ejection_time`, and `max_ejection_time` must obey the restrictions in the [`google.protobuf.Duration` documentation](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration) and they must have non-negative values. The fields `max_ejection_percent`, `enforcement_percentage`, `failure_percentage_threshold`, and `enforcement_percentage` must have values less than or equal to `100`. If any of these requirements is violated, the `Cluster` resource should be NACKed.
```
ejona86, line 221: That is from the xDS perspective. This is marked as a general-use policy. What is the behavior when parsing a service config?

ejona86 (approved):
I less-than-skimmed Rational; I don't think anything there would be surprising to me.

ejona86:
I failed to include: my stuff can go in as a followup PR without bothering me.

A50-xds-outlier-detection.md
murgatroid99: Fixed.

A50-xds-outlier-detection.md
murgatroid99: I changed it to use the timestamp of when the timer fires.

A50-xds-outlier-detection.md
murgatroid99: I created a separate validation section for the non-xDS case.

PR #282, opened 2021-12-10
L93: Add a new Node API to create credentials from a SecureContext

murgatroid99:


wenbozhu (approved):

ctiller:
There's already an L92 merged, so I guess this should be L93?

murgatroid99:
This PR had the L92 number before that one did, so that one should really be L93, but I guess it's too late now so I'll change this one.

PR #283, opened 2021-12-23
P5: JDK Version Support Policy

ejona86:


P5-jdk-version-support.md
```diff
@@ -0,0 +1,125 @@
...
+support dates are a good guideline.
+
+Java has been providing strong incentives to upgrade in the form of new features
+and performance. Developers want those benefits and so push their organizations
+to upgrade. It is in everyone's best interest to upgrade regularly, as otherwise
+Java comes dangerously close to ossifying as we have seen with the slow
+historical upgrades.
+
+Applications running on JDK versions not receiving Premier Support are much less
+likely to be actively developed and so are less like to need new features. Some
```
sanjaypujare, line 110: s/less like to/less likely to/

P5-jdk-version-support.md
```diff
@@ -0,0 +1,125 @@
...
+applications in a single, large application server is less common which allows
+upgrading pieces individually. The ecosystem has gotten more practiced at
+supporting new versions, less willing to be held back, and better at
+compartmentalizing legacy applications.
+
+gRPC is widely used and is lower in the dependency tree, so it has more
+conservative support. gRPC Java 1.0 supported Java 6, which was dropped
+following discussion in grpc/grpc-java#3961 because Java 6 was becoming
+unsupportable as the compiler was unmaintained and other libraries were dropping
+support. The discussion for Java 7 is happening at grpc/grpc-java#4671 and is
```
sanjaypujare, line 51: "The discussion for *dropping* Java 7"?

P5-jdk-version-support.md
```diff
@@ -0,0 +1,126 @@
...
+Android support is tracked separately and generally follows Google Play
+Services's supported API level, for security and as a reasonable guideline of
+what is worth supporting. Android support is not impacted by this gRFC as it has
+different options available to it (e.g., desugaring) and does not align cleanly
+to JDK versions.
+
+## Proposal
+
+gRPC Java may drop support for a JDK version when that version is no longer
+receiving [Premier Support from Oracle][Oracle Java support]. When doing so,
```
dapengzhang0, line 73: > When doing so,

There will be two actions in the process:

- choose a minor version between the most recent versions of grpc for supporting the old JDK 
- make a new release which drops the old JDK

Which happens first? (I think the first one)  When do we notify the community when we starts the process? Can the gap time between the two be unlimited?

dapengzhang0: > make a new release which drops the old JDK

Or we don't do this particularly?

ejona86: The smallest gap would be doing the two simultaneously from the external perspective. The release notes would say, "Java Y is now required. Non-Android users requiring Java Y-1, are encouraged to use v1.X".

When doing v1.X, I think we could say, "This is expected to be one of the last releases supporting Java Y and users requiring Java Y support are encouraged to use this release."

Right now defining the release after-the-fact is better because we haven't had a strong idea of when exactly we'd drop support. I had originally thought 1.36 would be a good candidate for Java 7 users, but it's now been 6 months and we still haven't dropped Java 7. With this policy planning is much easier so we could swap to the announce-ahead-of-time approach.

If you're asking if we could choose a year-old release, yes, we could try although it is easy to argue it isn't "recent." But I don't see much reason to do that. We want something recent because it'd already have the most features+bug fixes and less likely to require users to downgrade. But we may choose something slightly older because it has proved itself to have been a stable release. I think 6 months is the practical upper bound, with 3 months being more likely (which is just as a release would fall off the normal support window). Do you think that should be codified?

P5-jdk-version-support.md
```diff
@@ -0,0 +1,126 @@
...
+gRPC maintainers will publicize a recent minor version release (but not
+necessarily the most recent) to be the suggested release for users on the older
```
dapengzhang0, line 75: What's the concern if we _always_ choose the most recent minor version to publicize? Doing so we will never require users to downgrade. (For the upcoming release, even though we claim not to support the old JDK, we don't have to immediately migrate to build it with the new JDK or to use the new language features.)

ejona86: I think we will prefer a very recent version, but I think we'd also take into account something that seems to have been heavily used (and thus, we are more confident is stable). So we may use a slightly older version if it happened to have been heavily used with Google Client Libraries, Beam, or similar. But we'd still prefer the newest version that seemed to have been heavily used.

The downgrade problem isn't expected to be too severe for shorter periods because we've seen devs in these circumstances tend not to update frequently (which is also why it seems they don't really need features).

P5-jdk-version-support.md
dapengzhang0: > Do you think that should be codified?

No, it doesn't sound necessary.

dapengzhang0 (approved):

PR #285, opened 2022-01-14
A41: Clarify metadata match when invert is set

yashykt:


A41-xds-rbac.md
```diff
@@ -150,7 +150,7 @@ In RBAC `metadata` refers to the Envoy metadata which has no relation to gRPC
 metadata. Envoy metadata is generic state shared between filters, which has no
 gRPC equivalent. RBAC implementations in gRPC will treat Envoy metadata as an
 empty map. Since `ValueMatcher` can only match if a value is present (even
-`NullMatch`), the `metadata` matcher is guaranteed not to match.
+`NullMatch`), it is guaranteed not to match. If `invert` is set, it will match.
```
ejona86, line 153: I'm not wild about special-casing metadata matcher for describing invert behavior. I _know_ someone will complain to me saying, "but invert was only mentioned for 'metadata', so I thought it only applied there". If we need to call out invert behavior, make it separate because it applies to all parts. Include it near the "Because matching supports NOT, the matcher must still be processed even if a rule contains references to things that don't generally match" text?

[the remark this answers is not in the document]
yashykt: Yeah, I wanted to make it clear that `ValueMatcher` is the thing that won't match as compared to the overall `metadata` matcher which would take `invert` into account.

yashykt: Updated

ejona86:
Oh, you're talking about `MetadataMatcher.invert`. That is a new field that did not exist when this gRFC was made (looks like it was added a few days before the gRFC was finalized), and that I wasn't aware of.

Maybe we just document that it did not exist for the gRFC?

yashykt:
Updated to note that Java and Go do not support `MetadataMatcher.invert` as of yet.

menghanl:
Any updates here? Go is waiting for this to be merged before implementing the behavior.
(And if we do, maybe we should remove that Go/Java doesn't support the feature from the gRFC).

ejona86:
I'm not wild about this being a moving target. Again, I think we should just say the point in time this was done and imply nothing about new fields.

@menghanl, why is Go adding/waiting for invert?

menghanl:
We have an issue tracking this: https://github.com/grpc/grpc-go/issues/5150
And @zasweq was waiting for this gRFC to be merged.

ejona86:
That go issue sounds like, "hey it was added to the gRFC, let's add it to Go." I don't think it should be "added" to the gRFC. It came after the gRFC. The gRFC has nothing to do with it. We may say, "hey, by the way, there was stuff added to the proto after this that is not part of the gRFC" which helps understand the design. But that's different.

I think there have been some other discussions about when we'll add new fields, especially with matchers. This seems related. It should be discussed places other than here, I think.

dfawley:
What is this PR waiting for to be merged?  Or do we not want to update the gRFC?  If not, what documentation do we include this in?  Just the proto file's comments?

yashykt:
Based on @ejona86 's review, I guess this PR should be closed, since there is already a date on the gRFC (unless we are not comfortable with overloading the meaning of the last update for this purpose)

yashykt:
Also, I don't think https://github.com/grpc/grpc-go/issues/5150 needs to be blocked on this

dfawley:
> Also, I don't think https://github.com/grpc/grpc-go/issues/5150 needs to be blocked on this

I agree.  We will go ahead and implement this. What's the status in the other languages?

yashykt:
> I agree. We will go ahead and implement this. What's the status in the other languages?

Handled in Core

PR #289, opened 2022-02-19
A51: Custom Backend Metrics Support

YifeiZhuang:
Dumping design docs to gRFC per:
* server design doc: [go/grpc-custom-backend-metrics-design-docs](https://docs.google.com/document/d/1h3_yfcp7vUILJnZQSko8KR2Xy7H3J_OySCkVH8I4GBI/edit?resourcekey=0-wxlbrVxaif8Pf0xwBCT5Jw#heading=h.b5hiceid4qd1)
* client design doc: [go/grpc-lb-policy-backend-metric-api](http://go/grpc-lb-policy-backend-metric-api).


A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+In fact, many fine-tuned LB policies requires feedback signals from the
+backend to the data plane clients, e.g. weighted-round-robin LB policy.
+To fill the gap and complements custom LB policy, gRPC now brings APIs to allow users to inject custom 
+metrics at the server and consume the metrics from the client.
+The [ORCA](https://github.com/envoyproxy/envoy/issues/6614) proposal fits perfectly into the solution. 
+We propose to use ORCA service and message standards defined in UDPA protos.
+
+### Related Proposals:
+
+* [A36: xDS-Enabled Servers][A36]
```
markdroth, line 35: I don't think we need to reference either gRFC A36 or the upcoming xDS custom LB policy design here.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+Backend metric data will be encoded in the ORCA format for transmission on the wire. 
+Specifically, it will be transmitted as a binary-encoded [LoadReport](https://github.com/cncf/udpa/blob/04548b0d99d4e70b29310ebccc8e01f2deeed43a/udpa/data/orca/v1/orca_load_report.proto#L14) protobuf.
+
+### Per-Request Metrics Reporting
+
+#### Wire Format
+For metrics reported on a per-request basis, the ORCA LoadReport protobuf will be encoded in the 
+trailing metadata sent by the backend.  The metadata key will be `endpoint-load-metrics-bin`.
+
+Note that if the backend application adds this data to the trailing metadata of each RPC,
```
markdroth, line 66: This paragraph should probably move down to the "Rationale" section.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+that the client does not need, but this case should be fairly rare; 
+backends will generally not be written to populate data that they do not need.  
+Also, we want to avoid the complexity of introducing some kind of negotiation for the client to 
+tell the backend that it wants this data.
+
+#### The Client API
+LB policies will have the ability to access the load report data 
+as part of intercepting the trailing metadata sent from the server.
+Note that the API must be structured such that if multiple LB policies are interested in the data (e.g., both xDS and WRR),
+the protobuf deserialization needs to happen only once. The exact API for this may differ across languages,
```
markdroth, line 78: We should have an appendix that shows the APIs in the various languages.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+distinct from the callback it gets to intercept trailing metadata.  
+However, we think this is not a good idea, since it is likely to cause ordering problems
+(there is no guarantee of the order in which the callbacks would be seen by the LB policy) 
+and may incur additional synchronization overhead.
+
+#### The Server API 
+Using a server interceptor, it is convenient to attach the metrics as part of the
+trailing metadata of the server call and send back with the RPC response.
+
+Define a metric recorder that is unique per RPC. Server applications should obtain a reference and
```
markdroth, line 93: "We will define..."

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+
+#### The Server API 
+Using a server interceptor, it is convenient to attach the metrics as part of the
+trailing metadata of the server call and send back with the RPC response.
+
+Define a metric recorder that is unique per RPC. Server applications should obtain a reference and 
+use it to record metrics specifically for the current RPC call. Implementing this may leverage a 
+[context](https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html) or similar mechanism, 
+and a language should implement one if not already existed.
+The metric recoder looks like this:
```
markdroth, line 97: Instead of having Java-specific API info here, please describe the API in a language-agnostic way here, and move the Java API specification to an appendix (which we should also include for the other languages).

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+The data is dumped and translated to `OrcaLoadReport` at the end of each RPC by the interceptor, 
+and sent out as the trailing metadata.
+
+Note that reporting utilization metrics per-query is only useful in high QPS systems. 
+In such systems, the application that measures utilization data should consider sampling the data 
+to avoid performance degradation due to high rates of syscalls.
+
+
+### Out of Band Metrics Reporting
+For the out-of-band reporting, it requires a separate load reporting server running along with the
```
markdroth, line 132: It does not require a separate *server* (i.e., listening on a different port); it requires a separate RPC *service* (i.e., a handler for requests for a particular RPC method).

I don't think this needs to say anything about "running along with the backend server for as long as the server is alive", since that seems irrelevant and confusing.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+metrics data from the streaming server for each backend.
+The report request should specify metrics reporting interval, the lower bound of which should be capped 
+to a server configurable value, and there is no upper bound unless it is needed in the future.
+
+#### The Server API 
+Provide builtin implementation of `OpenRCAService` streaming service defined by ORCA in each supported language.
+Server applications are free to add such a load reporting server
+in a normal way that they register a service.
+
+The service holds the utilization metrics data in key/value pairs using a hashmap.
```
markdroth, line 146: I don't think we should over-specify the implementation here; the type of map to use is an implementation detail that this design doesn't really have to care about.  I think it's sufficient to say that it's a map.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+
+  // Records the CPU utilization metric measurement for the call. 
+  public CallMetricRecorder recordMemoryUtilizationMetric(double value);
+
+  // Returns the call metric recorder attached to the current call.
+  public static CallMetricRecorder getCurrent();
+}
+```
+
+All the methods should be thread safe.
```
markdroth, line 118: Thread-safety is an implementation detail; it does not need to be mentioned here (but it can be mentioned in the Java-specific appendix).

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+Recording one metrics multiple times overrides the previously provided values. 
+The methods can be called at any time during an RPC lifecycle,
+depending on the particular type of the cost or utilization measurement. For example, an RPC request size may
+be recorded at the beginning of the call, while a queue-size may be reported at any time during the call life-cycle.
+The data is dumped and translated to `OrcaLoadReport` at the end of each RPC by the interceptor, 
+and sent out as the trailing metadata.
+
+Note that reporting utilization metrics per-query is only useful in high QPS systems.
```
markdroth, line 126: This can probably be moved down to the "Rationale" section as well.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+
+### Out of Band Metrics Reporting
+For the out-of-band reporting, it requires a separate load reporting server running along with the 
+backend server for as long as the server is alive.
+
+At the client side, typically an LB policy, opens a stream on each
+established connection to the backend server to request the load report, and then periodically receiving
+metrics data from the streaming server for each backend.
+The report request should specify metrics reporting interval, the lower bound of which should be capped 
+to a server configurable value, and there is no upper bound unless it is needed in the future.
```
markdroth, line 139: I think we can omit the "and there is no upper bound unless it is needed in the future" part.  We're not including that, so let's not say anything about it.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+
+#### The Server API 
+Provide builtin implementation of `OpenRCAService` streaming service defined by ORCA in each supported language.
+Server applications are free to add such a load reporting server
+in a normal way that they register a service.
+
+The service holds the utilization metrics data in key/value pairs using a hashmap. 
+Expose set-style APIs for user applications to update the metrics data, pseudocode below:
```
markdroth, line 147: As above, let's describe the API in a language-agnostic way and move the Java-specifics to an appendix.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+
+// Update the memory utilization metrics data.
+void setMemoryUtilizationMetric(double value);
+
+// Clear the memory utilization metrics data.
+void deleteMemoryUtilizationMetric();
+}
+```
+
+All the methods above should be thread safe.
```
markdroth, line 184: As above, thread-safety may be something to mention for Java, but it's over-specifying the language-agnostic design.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+// Clear the memory utilization metrics data.
+void deleteMemoryUtilizationMetric();
+}
+```
+
+All the methods above should be thread safe. 
+Allowing updating each metric key individually is convenient for a server application to adopt where it 
+comes with multiple isolated components and each generates their own set of metrics.
+
+A minimum reporting interval (30s by default) could be configured when creating the service.
```
markdroth, line 188: s/could/can/

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+The minimum report interval is the lower bound of the OOB metrics report period. It means, in the 
+client OOB request message if the `report_interval` 
+goes below the service configured minimum interval then it is treated as the service minimum interval. 
+If `report_interval` is not specified (equal to 0), it is also treated as service minimum.
+There is no upper bound, the largest value you can specify is max integer.
+However, it might be useful to have a separate option called default report interval (default 1 minute)
```
markdroth, line 194: This paragraph can be removed (or, if you really want to keep it, it can be moved to the "Rationale" section to discuss possible future changes).

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+A minimum reporting interval (30s by default) could be configured when creating the service.
+The minimum report interval is the lower bound of the OOB metrics report period. It means, in the 
+client OOB request message if the `report_interval` 
+goes below the service configured minimum interval then it is treated as the service minimum interval. 
+If `report_interval` is not specified (equal to 0), it is also treated as service minimum.
+There is no upper bound, the largest value you can specify is max integer.
+However, it might be useful to have a separate option called default report interval (default 1 minute) 
+that will be used when `report_interval` is not specified (equal to 0), see the code comment above.
+
+Alternatively we can use callback-style APIs in a pull model, but this is suboptimal because
```
markdroth, line 197: This paragraph should be moved down to the "Rationale" section.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+it is not performance/scalability friendly.
+In a pull model, the time complexity of doing utilization measurement syscalls reaches O(M*N),
+M being the number of OOB clients and N being their frequency configuration.
+In a push model, the number of syscalls decouples with the number of OOB clients and their configurations.
+In fact, the server application is able to tune the performance by sampling the utilization data at a constant rate.
+
+Internally, the load report server will save each incoming client connection(`responseObserver`) 
+in a `RcaConnection` object or alike, with a custom frequency timer.
+
+```java
```
markdroth, line 207: I don't think we need Java-specific details here.  The language-agnostic description is sufficient.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+
+```
+
+Initially receiving a new request, or when the response timer fires, the server sends a response to 
+report metrics data. To generate a response, we copy the metrics data saved in the map, 
+but the response proto should only be produced once and be reused.  
+The report is always in a state-of-the-world fashion. The service should properly handle exceptions 
+when the client disconnects: it should terminate the stream, cancel the timer and remove the 
+`RcaConnection` immediately. Otherwise, the service should keep sending the report.
+Note that the frequency is final because the OOB request method is a server-streaming API.
```
markdroth, line 221: These 4 lines can be discussed in the "Rationale" section below.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+when the client disconnects: it should terminate the stream, cancel the timer and remove the 
+`RcaConnection` immediately. Otherwise, the service should keep sending the report.
+Note that the frequency is final because the OOB request method is a server-streaming API. 
+Clients would terminate the OOB stream and reconnect if they want to change the frequency. 
+Switching to a bidi stream will make it possible to change the reporting frequency in flight, 
+but it should not impact the public interface.
+
+If no one called to update the metrics data during a report interval, the data remains the same. 
+And the load reporting server will keep sending the metrics data regardless of whether the data has changed or not.
+There is a potential optimization, that is to skip sending the unchanged metrics data since last response.
```
markdroth, line 228: The rest of this paragraph can be moved to the "Rationale" section.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+and should not require heavy-weight operations like reconnecting to backends or recreating subchannels or LB policies.
+
+Each policy that has subscribed to the data will get a callback whenever a new message is sent to the client on the call.  
+As with per-request data, the protobuf deserialization will happen only once.  
+The data provided to the callback should be exposed to the LB policy using the same type as used for per-request data.
+
+The API for subscribing to out-of-band reports must be available to LB policies at any level of the routing hierarchy, 
+including both leaf LB policies (whose children are subchannels) and parent LB policies (whose children are other LB policies).  
+Note that there may be multiple LB policies in the routing hierarchy that are interested in subscribing.  
+For example, in the case where we are using the xDS policy for cluster routing and the WRR policy for intra-cluster routing,
```
markdroth, line 262: The wording here is a little out of date; it dates back to when all of the xDS functionality was in a single "xds" LB policy.  I suggest changing it to say something like this:

"""
For example, when using xDS, the `xds_cluster_impl` policy may need to see backend metric data to generate load reports for each cluster, and the WRR policy might need to see the same data for endpoint-picking within a locality.
"""

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+both policies may be interested in subscribing.  
+The WRR policy will be directly managing the subchannels and will therefore be able to subscribe directly against the subchannel API, 
+whereas the xDS policy will need to subscribe against the LB policy API of the WRR policy,
```
markdroth, line 265: I think we decided that this isn't the way we're actually going to do it.  Instead of having a way for parent policies to subscribe via the child policy (which would require adding new methods to the LB policy API itself), the parent policy will just subscribe to the subchannels as they are created before passing them back down to the child policy.

@ejona86 @dfawley Please confirm that this matches your recollection of where we left this.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+If the StreamCoreMetrics call returns any other status, the client will retry the call.  
+To avoid hammering a server that may be experiencing problems, the client will use exponential backoff between attempts.  
+When the client receives a message from the server on a given call, the backoff state is reset, 
+so the next attempt will occur immediately, but any subsequent attempt will be subject to exponential backoff.
+
+When the client subchannel is shutting down or when the backend sends a GOAWAY, 
+the client will cancel the StreamCoreMetrics call.  
+There is no need to wait for the final status from the server in this case.
+
+## Use Cases
```
markdroth, line 293: I think this section probably isn't needed.  We cover the use-cases sufficiently well above.

A51-xds-custom-backend-metrics.md
```diff
@@ -0,0 +1,382 @@
...
+because in OOB we need to identify which client the request cost reports is generated from and associate 
+it with the corresponding OOB stream in order to deliver the report accurately. 
+So generating the request cost report requires extra processing for each distinct client connection in the OOB scenario.
+This scenario is not supported until there is use case in the future.
+
+## Rationale
+
+## Implementation
+
+This will be implemented simultaneously in Java by @YifeiZhuang, in
```
markdroth, line 330: I don't think we need to say who's doing the implementation.  This line can probably just be removed.

markdroth (commented):
Thanks for doing this!  This looks like a great start.

A51-xds-custom-backend-metrics.md
YifeiZhuang: Thank you. Did we decide to do a wrapper `ForwardingLoadBalancerHelper `?  c.f. [java implementation](https://github.com/grpc/grpc-java/blob/881747a63df79a7baae252de86447c0997c76f69/xds/src/main/java/io/grpc/xds/OrcaOobUtil.java#L97-L147).

ejona86: Yeah, the text at the time of this comment doesn't seem likely in any language. I see the text changed for this already, and is now mirroring the Java approach. I think though that will probably need to go into the Java section and we leave this part here a bit more generic (just move the "We will provide a utility function... accepts subscriptions from any routing hierarchy." part to Java and leave the rest as-is?)

Last I heard, C was special-casing this in their subchannel API, so a parent can use the API on subchannel creation before returning a subchannel to a child. I thought Go was considering something similar, although that was partly because of the restrictions at the time; maybe Go has more options now.

A51-custom-backend-metrics.md
```diff
@@ -252,91 +206,119 @@ A client may change its requested interval at any time; this is expected to be a
...
+In fact, the server application is able to tune the performance by sampling the utilization data at a constant rate.
 
-2. Collect Utilization Metrics using Per-Query Reporting 
+It is by design that the OOB server sends the data regardless of whether it has been changed or 
+not since the last time it was sent. We considered a potential optimization, that is to skip 
+sending the unchanged metrics data since last response.
+To do that, we should not compare maps for de-duplication, instead, we can bump up a
+generation_id of the metrics data every time users make a change.
+Each client connection only needs to compare its previous generation_id to do deduplication.
+The optimization strategy can be implemented until there are concrete use cases in the future.
```
markdroth, line 282: s/util/in the future if/

A51-custom-backend-metrics.md
```diff
@@ -252,91 +206,119 @@ A client may change its requested interval at any time; this is expected to be a
...
 
-In a high QPS environment, reporting request cost using per-query is desired. 
-Compared with 1, the advantage of per-query is that the utilization metrics can be timely aligned with the requests. 
-It also communicates the metrics data rapidly and saves cost of the idle connection compared with OOB reporting.
+The reporting frequency in the OOB request is final because it is a server-streaming API.
+Clients would terminate the OOB stream and reconnect if they want to change the frequency.
+Switching to a bidi stream will make it possible to change the reporting frequency in flight,
+but it should not impact the public interface.
```
markdroth, line 287: s/it should not impact/this can be done in the future if needed without impacting/

A51-xds-custom-backend-metrics.md
markdroth: Yeah, the new text here is true of Java but not true in other languages -- C-core will not do this by wrapping the helper at all.

I suggest just removing the paragraph you added here (line 218 in the most recent commit).  I think there's probably enough context in the paragraphs above to make the context clear.

markdroth (commented):
This is looking really good!  Most of the remaining comments are fairly minor.

A51-custom-backend-metrics.md
```diff
@@ -80,170 +70,134 @@ but here are a couple of possible approaches:
...
-  StreamObserver<OrcaLoadReport> responseObserver;
-  Timer customFrequencyTimer;
-}
-
-```
-
+in an `RcaConnection` data structure or alike, with a custom frequency timer.
 Initially receiving a new request, or when the response timer fires, the server sends a response to 
 report metrics data. To generate a response, we copy the metrics data saved in the map, 
 but the response proto should only be produced once and be reused.
```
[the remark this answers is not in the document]
markdroth: (Actually, I decided to go ahead and implement this optimization in C-core.  But I still don't think we need to call it out in this design, because it is ultimately an implementation detail.)

markdroth (approved):
This looks great!

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,387 @@
...
+a child policy like round_robin, which does not know anything about out-of-band-reports and will not contain any
+code to handle them.
+
+When a subchannel first establishes a connection, if at least one LB policy is subscribing to out-of-band reporting, 
+the client will immediately start the `StreamCoreMetrics` call to the backend. 
+The client does not need to wait for this call to be established before it starts sending other calls on the subchannel.
+If the `StreamCoreMetrics` call fails with status `UNIMPLEMENTED`, the client will not retry the call, 
+and the subscribing LB polic(ies) will not receive any callbacks.
+However, the client will record a channel trace event indicating that this has happened. 
+It will also log a message at priority ERROR.
```
ejona86, line 200: What is this talking about? It seems likely inappropriate in Java.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,387 @@
...
+
+  // Records the memory utilization metric measurement for the call. 
+  public CallMetricRecorder recordMemoryUtilizationMetric(double value);
+
+  // Returns the call metric recorder attached to the current call.
+  public static CallMetricRecorder getCurrent();
+}
+```
+
+At the client side, Java will provide a utility function `OrcaClientStreamTracerFactory`.
```
ejona86, line 285: I think we need to mention how it will avoid double-parsing. The Factory may be duplicated multiple times, but it modifies `StreamInfo` when creating a `ClientStreamTracer` instance to become aware of other instances of the Factory. The first will parse the header and the following ones will internally register a callback with the first.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,387 @@
...
+
+#### Out-of-Band Reporting 
+At the server side, Java implemented the OOB service `OrcaServiceImpl`: users can optionally provide the minimum
+reporting interval configuration when creating the object. Then they can update utilization metrics
+data using the exposed APIs as follows. All the methods are thread safe.
+
+```java
+public class OrcaServiceImpl implements OpenRcaService {
+
+// Allow configuring minimum report interval. If not configured or badly configured non-positive,
```
ejona86, line 309: Indentation is wrong.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,387 @@
...
+
+The following pseudocode shows how a custom load balancing policy instantiates the helper wrapper, 
+installs the out-of-band metrics reporting mechanism, and then consumes the metrics reports, 
+which only requires users to register an `OrcaOobReportListener` that contains their own business logic:
+
+```java
+class CustomLoadBalancer extends LoadBalancer {
+  private final Helper helper;  // the original Helper
+  void handleNewServer(address) {
+  // Create OrcaHelperWrapper and a separate listener for each Subchannel
```
ejona86, line 358: Indentation is wrong.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,387 @@
...
+so users do not have to worry about client and server deployment coordination.
+
+The following pseudocode shows how a custom load balancing policy instantiates the helper wrapper, 
+installs the out-of-band metrics reporting mechanism, and then consumes the metrics reports, 
+which only requires users to register an `OrcaOobReportListener` that contains their own business logic:
+
+```java
+class CustomLoadBalancer extends LoadBalancer {
+  private final Helper helper;  // the original Helper
+  void handleNewServer(address) {
```
ejona86, line 357: It isn't clear to me what this function's purpose is. It creates stuff, but then throws is away.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,387 @@
...
+  Subchannel subchannel = orcaWrapper.asHelper().create(
+  CreateSubchannelArgs.newBuilder().setAddresses(address).build());
+  }
+}
+
+// Users implement a listener and pass it to the client library to receive
+// callbacks upon metric report updates from the backend.
+public interface OrcaOobReportListener {
+  //receive load report in the format of ORCA protocol.
+  void onLoadReport(OrcaLoadReport report);
```
ejona86, line 372: Looks like we have an issue is the current implementation that you can't distinguish which server this arrived from, yet the same listener is called for an entire LB policy. Seems there will need to be more glue here in some way.

A51-custom-backend-metrics.md
YifeiZhuang: Why the `ERROR` level message traced/[logged](https://github.com/grpc/grpc-java/blob/881747a63df79a7baae252de86447c0997c76f69/xds/src/main/java/io/grpc/xds/OrcaOobUtil.java#L508) in subchannelLogger is inappropriate?

A51-custom-backend-metrics.md
YifeiZhuang: Good catch. Yeah i agree that the `OrcaOobReportListener` passed to `OrcaReportingHelper` maps to a single helper. The helper is shared by all the OOB streams, so every server's metrics data is calling the same listener.
What about adding `Subchannel` to become `OrcaOobReportListener.onLoadReport(OrcaLoadReport report, Subchannel subchannel) `, maybe this can apply to per request listener as well.

A51-custom-backend-metrics.md
markdroth: This was intended to work the same way that we handle UNIMPLEMENTED for client-side health checking.  What does Java do in that case?  The original design for client-side health checking did say we should log at ERROR in this case, but maybe Java didn't implement that.

ejona86: 1. Why is it an ERROR? Seems it should be a WARNING since the implementation continues
2. WARNING logs are generally bad, because if something is only a _concern_ but not a _problem_ then logging all the time causes noise and hides legitimate problems. This is the same mentality used for golang not to have compiler warnings. There are some times it happens, but it should be exceedingly rare and certainly not as part of normal operation as is the case here
3. Java should not be logging by default. If we _can't_ propagate an error somewhere, then we may need to log, but that does demonstrate a design issue. We are a library and application authors do not want to see log noise (I'm aware of some cases we are already inappropriately logging, but several of those should be cleaned up and some can't easily.)

ejona86: Looks like Java [does do an error for health check](https://github.com/grpc/grpc-java/blob/a5829107c3dd771a43cb2ef9c6eb967aa13b8aa3/services/src/main/java/io/grpc/protobuf/services/HealthCheckingLoadBalancerFactory.java#L419-L421).

Per subchannel logging can spam things quite a lot, so essentially they "never should happen." With health check the expectation was there was explicit configuration in the service config. Basically, if UNIMPLEMENTED, then the service config is incorrect or something else is busted. For these metrics, I don't think we have a signal for whether this information is required or optional. Since there can also be per-RPC metrics, the LB policy might actually just be expected to receive the information from a different source.

If we have the error here, we are saying that using the util implies the server _must_ support the OOB service and we will not support it being optional.

A51-custom-backend-metrics.md
ejona86: Adding a `Subchannel` argument would certainly be closer, but there are some complications. First and easiest to resolve, we'll need to make sure to pass the correct Subchannel instance; different levels of the LB hierarchy see a subchannel at different wrapping levels. We'll need to remember the proper identity of the [subchannel when adding the listener](https://github.com/grpc/grpc-java/blob/a5829107c3dd771a43cb2ef9c6eb967aa13b8aa3/xds/src/main/java/io/grpc/xds/OrcaOobUtil.java#L278), and the proper identity of the subchannel is only known after the `if (augmented)` block, so we'd need to move that line that adds the listiner.

Second issue is that passing the subchannel through this callback API makes composition more difficult. It only works if the OrcaOobUtil is the top/last wrapper of the subchannel. If the subchannel is wrapped again after OrcaOobUtil, then using the subchannel easily becomes buggy. I'm thinking that it may be better to pass the listener inside CreateSubchannelArgs.

A51-custom-backend-metrics.md
markdroth: My expectation is that LB policies will need to be explicitly configured to request OOB backend metric data, since that's going to be a minority case, and that no one will configure that unless they expect the data to be available.  So I think the same argument would apply here as in the client-side health checking case.

markdroth:
I've sent YifeiZhuang/proposal#1 to add the C++ server-side OOB API to this gRFC.

A51-custom-backend-metrics.md
YifeiZhuang: > I'm thinking that it may be better to pass the listener inside CreateSubchannelArgs.

oh then it becomes similar to per request case where the user can track which listener is used for each subchannel/oob server. And they will be able to associate multiple listeners to a helper.

A51-custom-backend-metrics.md
YifeiZhuang: I am leaning towards a warning level log:
If users configured the client LB policy to have metrics but the server has not be properly configured yet, or during deployment, part of the backend has not been running the latest configuration.
This is an LB miss-configuration issue, even be transient. Creating a new connection may be successful. The main traffic might still going fine but the routing is not what the user is desired. Also, we are having ERROR for channel panic. So this miss configuration issue might be fair to get a WARNING: it also tells which server has which problem. Seems sufficient.

A51-custom-backend-metrics.md
markdroth: FWIW, in C-core, the API is to add a separate OOB watcher for each subchannel.

YifeiZhuang: @ejona Passing listener via CreateSubchannelArgs helps to solve the problem for leaf LB policy, it allows each subchannel have one listener. But seems not for non-leaf LB policy.

> different levels of the LB hierarchy see a subchannel at different wrapping levels. We'll need to remember the proper identity...

For an non-leaf LB hierarchy that may not use `CreateSubchannelArgs` directly, passing a [listener](https://github.com/grpc/grpc-java/blob/a5829107c3dd771a43cb2ef9c6eb967aa13b8aa3/xds/src/main/java/io/grpc/xds/OrcaOobUtil.java#L113) for each `helper` is still useful, e.g. when they put one callback for each child policy.

If we have a specific  CreateSubchannelArgs key for OOB listener, then we have to remove it after adding it to the listeners set of the OrcaReportingState, so that we do not adding them multiple times.

YifeiZhuang: We had a offline discussion that the current java `OrcaOobUtil` implementation needs to be changed, to allow to set multiple listeners to each subchannel and serve the needs of LB policies in different levels of LB hierarchy. We will probably leverage subchannel attributes to store the listeners and add them all at once to the OOB stream at the root of the LB hierarchy. This is because otherwise subchannel can be wrapped subchannel and cause problems. I will reflect the requirement in the gRFC.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,457 @@
...
+there will be an API that allows LB policies to subscribe to the out-of-band data. The client will
+maintain an open `OpenRCAService/StreamCoreMetrics` call whenever there is at least one LB policy 
+subscribing to the data. When a policy subscribes, it will indicate the interval at which it would like to get updates. 
+If multiple policies subscribe with different intervals, we will use the minimum of those intervals. 
+Note that the ORCA protocol involves the client telling the server the interval at which 
+it wants reports when the call starts, so if the client needs to change its requested interval, 
+it will need to cancel and restart the call.
+The API provided to LB policies needs to allow changing its requested interval at any time;
+this is expected to be a fairly rare operation
+(usually triggered by a config change for one of the subscribing LB policies),
```
ejona86, line 178: `s/,/./`

A51-custom-backend-metrics.md
```diff
@@ -160,23 +163,27 @@ Out-of-band data will be sent only if the client starts a streaming call for the
...
+The API provided to LB policies needs to allow changing its requested interval at any time;
+this is expected to be a fairly rare operation
 (usually triggered by a config change for one of the subscribing LB policies),
 It is important that a load balancing policy is able to re-config the frequency
 and should not require heavy-weight operations like reconnecting to backends or recreating subchannels or LB policies.
 
 Each policy that has subscribed to the data will get a callback whenever a new message is sent to the client on the call.
 As with per-request data, the protobuf deserialization will happen only once.
 The data provided to the callback should be exposed to the LB policy using the same type as used for per-request data.
+Currently, out-of-band reporting only supports utility metrics, details see the Rationale section,
```
ejona86, line 185: "details see the Rationale section" grammatically broken/weird in a couple ways. Would it be better as a parenthetical `(see Rationale section)`

A51-custom-backend-metrics.md
```diff
@@ -160,23 +163,27 @@ Out-of-band data will be sent only if the client starts a streaming call for the
...
+but we reserve the right to include costs and LP policy metrics filtering APIs in the future.
```
ejona86, line 186: "LP" seems it should be "LB", but I don't know what "LB policy metrics" is. It probably should just be "... and metrics filtering APIs...".

Although that's not at all what I was talking about with my "Since we'll be using the same type, should we have the client explicitly filter the costs from the object before calling the callbacks?" comment. I was meaning should the OOB util _implementation_ filter out costs if they happen to come from the server since we have no design of how they should be processed?

A51-custom-backend-metrics.md
```diff
@@ -284,8 +291,11 @@ public final class CallMetricRecorder {
 
 At the client side, Java will provide a utility function `OrcaClientStreamTracerFactory`. 
 Users can include it in the picker result in their custom LB policy’s 
-`LoadBalancer.SubchannelPicker` to let gRPC push notifications to the application. 
+`LoadBalancer.SubchannelPicker` to let gRPC push notifications to the application.
 It requires users to implement `OrcaPerRequestReportListener` in order to receive the metrics report callback.
+Note that java avoids trailers double-parsing: each `OrcaPerRequestReportListener` registered to each
```
ejona86, line 296: s/java/the implementation/ ? Or "the API"?

A51-custom-backend-metrics.md
```diff
@@ -306,38 +316,38 @@ data using the exposed APIs as follows. All the methods are thread safe.
...
-// report interval separately. If the default report interval is not configured or badly configured 
-// non-positive, the default is 1 min. 
-// public OpenRcaServiceImpl(long minInterval, TimeUnit timeUnit, long 
-// defaultInterval, TimeUnit)
-public OrcaServiceImpl(long minReportInterval, TimeUnit unit);
+  // Allow configuring minimum report interval. If not configured or badly configured non-positive, 
+  // the default is 30s.
+  // In the future, it might be useful to let users specify minimum report interval and default 
+  // report interval separately. If the default report interval is not configured or badly configured 
+  // non-positive, the default is 1 min.
```
ejona86, line 323: Is this 1 minute relating to the API that doesn't exist? It is strange having the API commented out as part of a different constructor. If you want to document the potential future API, just make it separate and give it its own documentation, but then comment it out.

A51-custom-backend-metrics.md
ejona86: > If we have the error here, we are saying that using the util implies the server must support the OOB service and we will not support it being optional.

I'm fine if we make this expectation explicit, in the gRFC and API. We will try to handle the case gracefully, but LBs will probably break if the server doesn't support it.

A51-custom-backend-metrics.md
YifeiZhuang: ugh...

> I was meaning should the OOB util implementation filter out costs if they happen to come from the server since we have no design of how they should be processed?

oh, i see, good point. OOB now would be nice to filter out if there are any cost from the server. but the server now doesn't have API support to add cost metrics on the OOB server, so it won't happen.

ejona86: It doesn't happen _today_ with _our implementation_. Someone could make their own implementation, or maybe in the future we design this stuff out and the server starts responding with costs. We shouldn't be assuming the server behavior outside that defined by the protocol, and the orca OOB protocol doesn't prevent costs from being sent.

A51-custom-backend-metrics.md
YifeiZhuang: > > If we have the error here, we are saying that using the util implies the server must support the OOB service and we will not support it being optional.
> 
> I'm fine if we make this expectation explicit, in the gRFC and API. We will try to handle the case gracefully, but LBs will probably break if the server doesn't support it.

I think warning is better based on previous conversation tho.

markdroth: I think this should be error, for consistency with what we do in the health checking case.  I think the two cases are exactly the same semantically.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,462 @@
...
+installs the out-of-band metrics reporting mechanism, and then consumes the metrics reports, 
+which only requires users to register an `OrcaOobReportListener` corresponding to the subchannel.
+
+```java
+class CustomLoadBalancer extends LoadBalancer { 
+  private final Helper helper;  // the original Helper
+  void handleNewServer(ResolvedAddresses address) {
+    // Create OrcaHelperWrapper and a separate listener for each Subchannel
+    OrcaListener listener = new OrcaOobReportListener();
+    OrcaHelperWrapper orcaWrapper = OrcaUtil.newOrcaHelperWrapper(helper);
```
ejona86, line 375: This wrapping should be done in the constructor. The LB shouldn't hang onto the original Helper.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,462 @@
...
+    Subchannel subchannel = orcaWrapper.asHelper().create(
+        CreateSubchannelArgs.newBuilder().setAddresses(address).build());
+    orcaWrapper.setListener(subchannel, listener,
+            OrcaReportingConfig.newBuilder().setInterval(40, SECONDS).build());
+    subchannel.updateAddresses(address);
```
ejona86, line 380: This does nothing, as the addresses were passed to create. Delete.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,462 @@
...
+The following pseudocode shows how a custom load balancing policy instantiates the helper wrapper, 
+installs the out-of-band metrics reporting mechanism, and then consumes the metrics reports, 
+which only requires users to register an `OrcaOobReportListener` corresponding to the subchannel.
+
+```java
+class CustomLoadBalancer extends LoadBalancer { 
+  private final Helper helper;  // the original Helper
+  void handleNewServer(ResolvedAddresses address) {
+    // Create OrcaHelperWrapper and a separate listener for each Subchannel
+    OrcaListener listener = new OrcaOobReportListener();
```
ejona86, line 374: The point of the new API is that the listener can know the subchannel. We should pass the subchannel into the listener's constructor in this example.

ejona86, line 374: `OrcaListener` doesn't seem to be a type and `OrcaOobReportListener` is an interface.

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,462 @@
...
+    OrcaHelperWrapper orcaWrapper = OrcaUtil.newOrcaHelperWrapper(helper);
+    Subchannel subchannel = orcaWrapper.asHelper().create(
+        CreateSubchannelArgs.newBuilder().setAddresses(address).build());
+    orcaWrapper.setListener(subchannel, listener,
```
ejona86, line 378: I really think this should be static, as I can't find any value in using the `orcaWrapper` and the static method is easier in the face of composition (if another helper needs to wrap the orca oob).

A51-custom-backend-metrics.md
```diff
@@ -0,0 +1,466 @@
...
+// Users implement a listener and pass it to the client library to receive
+// callbacks upon metric report updates from the backend.
+public interface OrcaOobReportListener {
+  //receive load report in the format of ORCA protocol.
+  void onLoadReport(OrcaLoadReport report);
+}
+
+```
+
+### Go
```
ejona86, line 397: I think we can just delete the Go section and have it be created when there is content to include.

ejona86 (commented):
This LGTM. OrcaMetrics class can be updated based on the recent API changes that were decided in Java. Annoying, but straight-forward.

Not giving GitHub-level approval as I am the Approver and it is still unclear if Doug needs to take a look.

A51-custom-backend-metrics.md
dfawley: SGTM.

Generally, should we replace all these sections with pointers to the implementation / API documentation once done?

dfawley (approved):

A51-custom-backend-metrics.md
YifeiZhuang: It looks either way exists, with pseudo-code + description pattern more popular among proposals.

ejona86: Generally that sort of information I add to the implementation section; I would have put this Java/C stuff in the Proposal section. But part of this strangeness is because our gRFCs have started including a lot more per-language information instead of saying, "these are the requirements; implement it how you desire" which applies equally to all languages.

https://github.com/grpc/proposal/blob/master/L74-java-channel-creds.md#implementation
https://github.com/grpc/proposal/blob/master/A8-client-side-keepalive.md#implementation (note: I added the C mapping because I had repeated difficulty figuring it out)
https://github.com/grpc/proposal/blob/master/A9-server-side-conn-mgt.md#implementation

dfawley: IMO we should scrub any language implementation stuff (except maybe high level design concepts) from gRFCs after the implementation is complete.  It's more likely to be inaccurate than the javadoc/godoc/etc, since: 1. the actual implementation (v1) may have needed to diverge from the initial gRFC, and 2. the implementation may change over time (new methods or fields are added, etc), but it is rare anyone will think to revisit the gRFC to update the documentation.

markdroth: I think it's fine to keep this here.  The APIs are part of the design here.  They will evolve over time, but so will the design itself.  This doc just shows the design at the time that we finish working on it; it's not meant to be a living doc that we always keep up to date.

PR #291, opened 2022-02-24
A6 update: .NET supports client retries

JamesNK:
See https://docs.microsoft.com/en-us/aspnet/core/grpc/retries?view=aspnetcore-6.0


JamesNK:
@jtattermusch

markdroth (approved):

PR #297, opened 2022-04-13
A50 update: address some conversations

murgatroid99:
This change addresses two recent conversations: one about optimizing the case in which outlier detection is configured to do nothing, and one about handling interval updates with existing timers.


ejona86 (approved):

A50-xds-outlier-detection.md
```diff
@@ -137,7 +137,7 @@ When the `outlier_detection` LB policy receives an address update, it will creat
 
 When the child policy asks for a subchannel, the `outlier_detection` will wrap the subchannel with a wrapper (see [Subchannel Wrapper section](#subchannel-wrapper)). Then, the subchannel wrapper will be added to the list in the map entry for its address, if that map entry exists. If there is no map entry, or if the subchannel is created with multiple addresses, the subchannel will be ignored for outlier detection. If that address is currently ejected, that subchannel wrapper's `eject` method will be called.
 
-The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, increment the corresponding counter in the map entry referenced by the subchannel wrapper that was picked.
+The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, increment the corresponding counter in the map entry referenced by the subchannel wrapper that was picked. If both the `success_rate_ejection` and `failure_percentage_ejection` fields are unset in the configuration, the picker should not do that counting.
```
menghanl, line 140: Do we also need to clear the counters in the map when a noop config is received?

(Or we only create the counters when we have a config that will use them)

murgatroid99: I guess the counters need to be cleared either when starting a timer without an existing timer, or when stopping timers because of a no-op config. If we want to optimize away the no-op case as much as possible, I think the first choice is better.

markdroth (approved):

menghanl (approved):

zasweq:
Just wanted to point out for implementors that I think that the last ejected time and timer start time are the same, and you only need one for both pieces of functionality.

PR #298, opened 2022-04-22
A52: gRPC xDS Custom Load Balancer Configuration

temawi:
A new proposal for refactoring how xDS LB policy configuration is processed and how custom LB policies can be configured.


A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,318 @@
...
+          }
+        },
+      }
+    }
+  }
+}
+```
+
+## Implementation
+Some initial refactoring to accommodate this desing has happened for Java. The Go library is yet to pick this up. In C-core the APIs that allow users to register custom LB policies are currently private. The plan is to make them public and once done this design can also be picked up for C-core.
```
markdroth, line 318: I don't think this needs to describe the current state of the implementation in each language.  If there are no implementation-specific details to describe, the "Implementation" section is not needed.

For C-core, I think this should say something like the following:

"""
In C-core, the LB policy API is not yet public, but it will be made public in the future when the EventEngine migration is far enough along to allow it.  In the interim, C-core will get the same changes to support custom xDS-based LB policies, so that when the C-core LB policy API later becomes public, this functionality will already be present.
"""

The other languages probably don't need to be mentioned at all.

markdroth (commented):
This looks really good!  Comments are mostly minor.

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,312 @@
+A52: gRPC xDS Custom Load Balancer Configuration
+----
+* Author(s): temawi
+* Approver: markdroth
+* Implemented in: Java, Go and C-core
+* Last updated: 2022-04-22
+* Discussion at:
```
markdroth, line 7: Please start an email thread on the grpc-io list and add a link to it here.  That will start the 10-day review period before we can merge the gRFC.

markdroth (commented):
This looks great!

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,312 @@
...
+  - The `hash_function` field will not be converted, but will need to be validated. Any function besides `XX_HASH` will cause the conversion to fail
+- `envoy.extensions.load_balancing_policies.round_robin.v3.RoundRobin`
+  - The gRPC policy name will be `round_robin`.
+  - Has no configuration - an empty JSON object is returned
+- `envoy.extensions.load_balancing_policies.round_robin.v3.WrrLocality`
+  - The gRPC policy name will be `xds_wrr_locality_experimental`
+  - Contains a `LoadBalancingPolicy` field that has a list of endpoint picking policies. This field will be processed by recursively calling the xDS LB policy registry, and the result will be used in the `child_policy` field in the configuration for the `xds_wrr_locality_experimental` policy (see the “xDS WRR Locality Load Balancer” sections for details on the config format).
+- `xds.type.v3.TypedStruct`
+  - This type represents a custom policy, deployed with the gRPC client by a user
+  - The gRPC policy name will be value of the `type_url` field in the `TypedStruct`, with the "`type.googleapis.com/`" prefix stripped out
```
ejona86, line 84: As we've discussed, the prefix here is weird. We probably want to follow protobuf here and use everything after the last `/`.

https://github.com/protocolbuffers/protobuf/blob/5f920d15071a160c65745568a85315cb6fd61600/src/google/protobuf/compiler/java/message.cc#L1687 (powers the [`is(Class)` function](https://github.com/protocolbuffers/protobuf/blob/5f920d15071a160c65745568a85315cb6fd61600/src/google/protobuf/compiler/java/message.cc#L1717))

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,312 @@
...
+To provide the `xds_wrr_locality` load balancer information about locality weights received from EDS, the cluster resolver will populate a new attribute in the resolved addresses. The new attribute will contain a map from a locality struct to a locality weight integer. Note that a change in this attribute would still allow associated subchannels to be reused - it will not affect their uniqueness.
+
+### Configuration Example
+
+This is a step-by-step example of how and example xDS configuration gets converted within a gRPC client. The example configuration is:
+
+```textproto
+{ // Cluster
+  load_balancing_policy: {
+    policies: [
```
ejona86, line 155: `[` for messages doesn't look like textproto format. The syntax highlighter is also complaining about this.

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,312 @@
...
+  "lb_policy_config": [
+    {
+      "xds_wrr_locality_experimental": {
+        "child_policy": [{
+          "myorg.MyCustomLeastRequestPolicy" : {
+            "choiceCount" : 2
+          },
+        }]
+      }
+  ]
```
ejona86, line 205: Missing a `}` before this.

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,312 @@
...
+The XdsClient will transfer the responsibility of converting the xDS configuration to the internal JSON format to a new *xDS LB policy registry*. This new registry will return an error if the xDS load balancing configuration cannot be converted, but it will not validate a converted configuration. It will be the responsibility of the XdsClient to validate that the converted configuration. It will do this by having the gRPC LB policy registry parse the configuration.
+
+The CDS resource is rejected and a NACK returned to the control plane in either of the following cases:
+
+- xDS LB policy registry cannot convert the configuration and returns an error 
+- gRPC LB policy registry cannot parse the converted configuration.
+
+### xDS LB Policy Registry
+
+This new registry will maintain a set of converters that are able to map from the xDS LoadBalancingPolicy message to the internal gRPC JSON format. The conversion process will be recursive so that e.g. the conversion of the new xDS WRR Locality policy (see below) will call into the registry to convert the child policies.
```
ejona86, line 56: @markdroth, we probably need to protect against excessive recursion in C? A malicious control plane could try to take over a client with a stack overflow. If this was in a single protobuf parsing step, then protobuf's parse depth checking would be enough. But since Any is concerned, it ends up not doing anything.

A52-xds-custom-lb-policies.md
markdroth: Textproto does actually support `[...]` for repeated fields these days.

A52-xds-custom-lb-policies.md
markdroth: Yeah, good point.  I assume we'd need to do this in all languages, not just C, since the xDS LB policy registry (whether implemented as a registry or just as a hard-coded parsing function) will do this recursion in all languages.

How about we just have a max depth of, say, 16?  Anything more than that would be NACKed.

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,312 @@
...
+  repeated LoadBalancingConfig xds_lb_policy = 2;
+
+}
+```
+
+To provide the `xds_wrr_locality` load balancer information about locality weights received from EDS, the cluster resolver will populate a new attribute in the resolved addresses. The new attribute will contain a map from a locality struct to a locality weight integer. Note that a change in this attribute would still allow associated subchannels to be reused - it will not affect their uniqueness.
+
+### Configuration Example
+
+This is a step-by-step example of how and example xDS configuration gets converted within a gRPC client. The example configuration is:
```
yashykt, line 150: s/and/an

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,312 @@
+A52: gRPC xDS Custom Load Balancer Configuration
+----
+* Author(s): temawi
+* Approver: markdroth
+* Implemented in: Java, Go and C-core
```
yashykt, line 5: Hasn't been implemented in Core yet :)

A52-xds-custom-lb-policies.md
ejona86: @markdroth, it supports `[...]` it for repeated messages? I've only seen it for repeated scalars. (In text_format.h, I think that is `SetUseShortRepeatedPrimitives(true)`.)

A52-xds-custom-lb-policies.md
temawi: I added a note about recursions deeper than 16 levels are considered invalid.

A52-xds-custom-lb-policies.md
temawi: Updated.

A52-xds-custom-lb-policies.md
yashykt: `Once a proposal is Final and if it has been implemented by a language, it can be updated to a status of Implemented with the implementing languages listed. (Listing versions is not required.)`

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,315 @@
...
+{
+  ...
+
+  "lb_policy_config": <array of LoadBalancingConfig JSON objects>
+  
+  ...
+}
+```
+
+The XdsClient will transfer the responsibility of converting the xDS configuration to the internal JSON format to a new *xDS LB policy registry*. This new registry will return an error if the xDS load balancing configuration cannot be converted, but it will not validate a converted configuration. It will be the responsibility of the XdsClient to validate that the converted configuration. It will do this by having the gRPC LB policy registry parse the configuration.
```
yashykt, line 47:
```suggestion
The XdsClient will transfer the responsibility of converting the xDS configuration to the internal JSON format to a new *xDS LB policy registry*. This new registry will return an error if the xDS load balancing configuration cannot be converted, but it will not validate a converted configuration. It will be the responsibility of the XdsClient to validate the converted configuration. It will do this by having the gRPC LB policy registry parse the configuration.
```

A52-xds-custom-lb-policies.md
temawi: Ok, I removed the languages for now.

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,315 @@
...
+// Converts an xDS cluster load balancing policy message to the internal
+// PolicyConfig JSON string.
+//
+// An error/exception will be returned if the message type
+// is not recognized, or if policy configuration is invalid in a way that
+// can't be determined by looking at the PolicyConfig output. 
+string convertToServiceConfig(envoy.config.cluster.v3.LoadBalancingPolicy loadBalancingPolicy)
+```
+
+This function will iterate over the list of `Policy` messages in `LoadBalancingPolicy`, attempting to convert each one to gRPC form, stopping at the first supported policy. The `Policy` message contains a `TypedExtensionConfig` message with the configuration information. `TypedExtensionConfig` in turn uses an Any typed `typed_config` field to store policy configuration of any type. This `typed_config` field is used to determine both the name of a policy and the configuration for it, depending on its type:
```
yashykt, line 72: So, for cases where we there are multiple policies in the list -
for example, [policyA, policyB], let's say policyB would run into an error if we tried to convert it to its JSON form, we would still accept this configuration since policyA was understood and we didn't feel the need to verify policyB.

The example can also be modified to say that policyA has troubles with JSON conversion but it is not supported, thereby being accepted by the client. In the future, the client gets updated to support policyA and it now sees a problem with the JSON conversion resulting in the new and updated client NACKing this configuration.

Is my understanding correct?

temawi: The first scenario you describe is correct. As long as the list of policies has one policy that is both supported by the client and parseable by the xds client, the configuration will be accepted.

In the second one, policyA would not be accepted since it is not supported by the client. The configuration as a whole would only be ACKed if there was another supported  policy in the list with a parseable config. If the client gets updated later to support policyA, it would still not be used if the config was invalid and the client would fall back to any other policy the config might have.

ejona86 (approved):

A52-xds-custom-lb-policies.md
```diff
@@ -0,0 +1,315 @@
...
+
+Regardless of the field that load balancing policy configuration is picked up from, internally `XdsClient` will represent the information in a uniform manner. The CDS update struct will be updated to carry LB policy configuration in JSON form. We move from an enum for the policy selection and a set of `oneof` configuration fields to having a [LoadBalancingConfig](https://github.com/grpc/grpc-proto/blob/ab96cf12ec7ce135e03d6ea91d96213fa4cb02af/grpc/service_config/service_config.proto#L560) JSON object.
+
+The load balancer config portion of the CDS update struct will change to take the following form:
+
+```c
+{
+  ...
+
+  "lb_policy_config": <array of LoadBalancingConfig JSON objects>
```
yashykt, line 41: Can you add a comment as to why this is an array?

yashykt (approved):

A52-xds-custom-lb-policies.md
temawi: It was just for compatibility with existing C config parsing libraries that expect to receive an array.

A52-xds-custom-lb-policies.md
```diff
@@ -34,7 +34,7 @@ Regardless of the field that load balancing policy configuration is picked up fr
 
 The load balancer config portion of the CDS update struct will change to take the following form:
 
-```json
+```c
```
markdroth, line 37: If this is supposed to be C, then I would format it like this:

```
struct XdsClusterResource {
  // ...

  // The gRPC LB policy config to be used as the child of the xds_cluster_impl LB policy.
  Json::Array lb_policy_config;

  // ...
}
```

markdroth (approved):
Looks great, modulo one small cosmetic comment.

A52-xds-custom-lb-policies.md
temawi: Ok I just switched to this and mentioned that this is what it would look like in C++/

PR #302, opened 2022-05-10
A53: Option for Ignoring xDS Resource Deletion

markdroth:


lepistone (commented):
Great to see this proposal, thank you!

Regarding the risk mentioned here that a faulty update isn't noticed promptly, I would like to be able from the point of view of a client to know when a deletion is ignored, so that I can alert.

I think it would be useful to have as part of the proposal a way for client systems to alert when this happens. The implementation doesn't really matter to me (a callback or an error-level log, I suppose) and it could even change by implementation.

Thank you!

markdroth:
@lepistone
> I think it would be useful to have as part of the proposal a way for client systems to alert when this happens. The implementation doesn't really matter to me (a callback or an error-level log, I suppose) and it could even change by implementation.

I'd prefer not to add a new API for this, but I'd be fine with generating an error log.

@ejona86 @dfawley Any concerns about an error log in this case?  I know Java has more restrictions around when it's okay to log than C++ does (not sure about Go).

ejona86:
I'm happy with logging here. I have some questions about when we will log and whether it is worth the effort to log when the resource is restored. But logging here seems useful and not generally noise.

A53-xds-ignore-resource-deletion.md
```diff
@@ -0,0 +1,92 @@
...
+basis.
+
+The new server feature will be the string `ignore_resource_deletion`; if
+present, this feature will be enabled.
+
+When this feature is enabled, the XdsClient will not invoke the
+watchers' `OnResourceDoesNotExist()` method when a resource is deleted,
+nor will it remove the existing resource value from its cache.
+
+Currently, this behavior will affect only LDS and CDS, because those are
```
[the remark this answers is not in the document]
markdroth: Good point.  Updated the text accordingly.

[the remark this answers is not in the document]
markdroth: Actually... on second thought, I'm not sure about this.  In C-core, we currently invoke the watcher's `OnError()` method when the resource timer fires, not the `OnResourceDoesNotExist()` method.  That's arguably wrong if this is the initial ADS stream after the client started up and it does not yet have a cached copy of the resource.  But I'm not sure that `OnResourceDoesNotExist()` is actually correct in the case where we were sent the resource on the old ADS stream but the timeout fired after the stream was closed and reestablished.  I think we probably shouldn't actually start the timer at all in that case, especially since it's possible that the server might not resend the resource as an optimization (see [spec](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#ack-nack-and-resource-type-instance-version)).

I need to think about this a bit more.

markdroth:
@ejona86 
> I have some questions about when we will log and whether it is worth the effort to log when the resource is restored.

Good question.  The most obvious way to do this would be to log every time we get an LDS or CDS response that does not include a resource that we already had, but that could cause a lot of duplicate logging.  If we're concerned about that duplicate logging, we could instead record a bit in the cache entry that says whether we've ignored a deletion for the entry, which would allow us to log only when the resource is initially deleted.  And that bit would also allow us to log when the resource comes back, if we decide that's something we want to do.

dfawley:
> @ejona86 @dfawley Any concerns about an error log in this case? I know Java has more restrictions around when it's okay to log than C++ does (not sure about Go).

According to [our guidelines](https://github.com/grpc/grpc-go/blob/master/Documentation/log_levels.md#error) for what level to log at, we would not use ERROR in Go, but WARNING instead.  Would that be okay?  (For us, ERROR is reserved for situations where grpc-go has a bug or the user did something wrong but the API can't return the error directly.  I.e. they should never happen in practice.)

A53-xds-ignore-resource-deletion.md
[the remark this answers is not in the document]
markdroth: So, having thought about this overnight, I believe that we should not actually start the timer in the first place upon ADS stream reconnects, because (a) starting the timer would make it impossible for servers to avoid resending the resource based on the `version_info` field, and (b) it doesn't really buy us anything to start the timer, because even if we could reliably tell that the resource did not exist after a stream restart, we cannot do so if the stream does not restart, so this still isn't really a reliable mechanism.  To be fair, the optimization based on the `version_info` field is not necessarily all that useful in the general case (the main case where it's useful is LDS/CDS wildcard subscriptions, which gRPC does not use), but there are some corner cases where it may come in handy, and the spec does allow it, so I think we need to handle this correctly for interoperability reasons.

So, given that, I think this gRFC does actually apply only for LDS and CDS in the SotW protocol variant, so I've restored the original wording here.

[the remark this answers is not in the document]
markdroth: I've sent grpc/grpc#29668 to fix this behavior in C-core.

markdroth:
As per today's discussion, I've added proposed logging behavior.  PTAL.

ejona86 (approved):

A53-xds-ignore-resource-deletion.md
```diff
@@ -0,0 +1,113 @@
...
+receives an LDS or CDS update that does not include the resource, if the
+bit is not set, it will log the ERROR message and set the bit; when the
+XdsClient receives an LDS or CDS update that includes a resource for
+which the cache entry bit is set, it will log the INFO message and unset the
+bit; when the XdsClient unsubscribes from a resource, if the cache entry bit
+is set, it will log the INFO message before deleting the cache entry.
+
+### Temporary environment variable protection
+
+This behavior is not configured via remote interaction, so it does not
```
ejona86, line 90: Note: we do have to handle backward compatibility with bootstrap files, so this is an incomplete argument.

ejona86:
FYI: We're thinking of exposing this as an opt-in argument to td-grpc-bootstrap to enable it before Traffic Director is fully handling the deletes.

A53-xds-ignore-resource-deletion.md
ejona86: No, old clients ignoring the field still doesn't mean it is safe. The only thing that would make it safe is if we don't think we need interop testing, unit testing is sufficient, and it will be implemented in a single PR. A client with this logic half-way implemented should not break if the bootstrap generator is updated.

markdroth: Yeah, I was making exactly those assumptions.  But you're right that this should be called out explicitly.

Done.

A53-xds-ignore-resource-deletion.md
```diff
@@ -0,0 +1,117 @@
...
+watchers' `OnResourceDoesNotExist()` method when a resource is deleted,
+nor will it remove the existing resource value from its cache.
+
+Currently, this behavior will affect only LDS and CDS, because those are
+the only two resource types for which the state-of-the-world xDS protocol
+variant allows the server to indicate a deletion to a client.  However,
+in the future, if/when gRPC adds support for the incremental xDS
+protocol variant, this behavior would affect all resource types.
+
+The XdsClient will generate an ERROR log when a resource deletion is
```
dfawley, line 75: I'd still lobby for WARNING here.  Our testing infrastructure actually fails tests if ERROR logs are encountered (unless they are predeclared), because we expect them to essentially not happen.  Also it seems weird to consider it an error that the client was configured to do a thing and its preconditions happened.  But if the client _isn't_ configured to do the thing and its preconditions happen, we don't log anything at all?

markdroth: As discussed, changed to use WARNING but indicate that ERROR can be used on implementations that don't support WARNING.

lepistone (approved):
I agree that this proposal as it stands effectively protects from a certain class of outages. Logging, while not the most convenient thing to alert on, can give some visibility as to what's going on. And the risk that an intentional update is not honored seem in check, since those updates would leave clients unable to make further progress. For the same reason, the fact that this makes proxyless gRPC diverge from the behaviour of Envoy seems an acceptable compromise.

Thank you, looking forward to see this implemented!

PR #303, opened 2022-05-12
A52 update: Mention gating by env flag

temawi:
cc: @markdroth, @dfawley

### DIFF
--- a/A52-xds-custom-lb-policies.md
+++ b/A52-xds-custom-lb-policies.md
@@ -312,5 +312,9 @@ The `weighted_target` configuration will be:
 }
 ```
 
+### Temporary environment variable protection
+During initial development, this feature will be enabled by the `GRPC_EXPERIMENTAL_ENABLE_CUSTOM_LB_CONFIG` environment variable. If unset or not true, the `load_balancing_policy` field in xDS Cluster proto will not be considered and load balancing policy configuration will be based on the `lb_policy` field. This environment variable protection will be removed once the feature has proven stable.
+
+
 ## Implementation
 In C-core, the LB policy API is not yet public, but it will be made public in the future when the EventEngine migration is far enough along to allow it. In the interim, C-core will get the same changes to support custom xDS-based LB policies, so that when the C-core LB policy API later becomes public, this functionality will already be present.
--- a/A52-xds-custom-lb-policies.md
+++ b/A52-xds-custom-lb-policies.md
@@ -313,7 +313,7 @@ The `weighted_target` configuration will be:
 ```
 
 ### Temporary environment variable protection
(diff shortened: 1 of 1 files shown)


temawi:
@yashykt FYI

A52-xds-custom-lb-policies.md
```diff
@@ -312,5 +312,9 @@ The `weighted_target` configuration will be:
 }
 ```
 
+During initial development, this feature will be enabled by the `GRPC_EXPERIMENTAL_ENABLE_CUSTOM_LB_CONFIG` environment variable. If unset or not true, the `load_balancing_policy` field in xDS Cluster proto will not be considered and load balancing policy configuration will be based on the `lb_policy` field. This environment variable protection will be removed once the feature has proven stable.
```
dfawley, line 316: `GRPC_EXPERIMENTAL_XDS_CUSTOM_LB_CONFIG`?

temawi: I did already implement it with the value I specified, but I can change if you feel it's worth it.

dfawley: It seems we have no standard here.

- `GRPC_XDS_EXPERIMENTAL_ENABLE_RING_HASH `
- `GRPC_XDS_EXPERIMENTAL_RBAC`
- `GRPC_EXPERIMENTAL_XDS_RLS_LB`
- `GRPC_EXPERIMENTAL_ENABLE_OUTLIER_DETECTION `

So I guess whatever is fine; good thing they're temporary. 🤷

ejona86 (approved):
We should keep an eye out for future xDS gRFCs to probably include XDS in the experimental variable, but these are meant to be water-under-the-bridge so it is fine here.

A52-xds-custom-lb-policies.md
markdroth: This should be `GRPC_EXPERIMENTAL_XDS_CUSTOM_LB_CONFIG`.

For the first several of these, we put `XDS` before `EXPERIMENTAL`, but the usage of this kind of env var is not really xDS-specific, so I think it makes sense to use `GRPC_EXPERIMENTAL_` as our default prefix for all such env vars.  And in this case, the feature is specifically xDS-based custom LB configs, so `XDS` should be in the name.

yashykt (approved):

A52-xds-custom-lb-policies.md
temawi: Updated to use GRPC_EXPERIMENTAL_XDS_CUSTOM_LB_CONFIG.

markdroth (approved):
Looks great.  Thanks!

PR #305, opened 2022-05-17
A50 update: Add details about environment variable protection

murgatroid99:
As discussed in the chat.


markdroth (approved):
Thanks for the update!

PR #311, opened 2022-06-12
BinderChannel: Clarify that only stream transactions are counted in flow control

sifmelcara:


sifmelcara:
reference: https://github.com/grpc/grpc-java/blob/a206cda1a8e4a84aeef073e298d76a2692a6a002/binder/src/main/java/io/grpc/binder/internal/BinderTransport.java#L417-L471

jdcormie (approved):

PR #314, opened 2022-07-22
A54: Restrict Control-Plane Status Codes

ejona86:


A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,136 @@
...
+ * `ALREADY_EXISTS`
+ * `FAILED_PRECONDITION`
+ * `ABORTED`
+ * `OUT_OF_RANGE`
+ * `DATA_LOSS`
+
+If an "inappropriate status" is detected, the status should be rewritten to use
+`INTERNAL` status code and include text within the status message making it
+clear a bug occurred. The original status code and message should be included
+within the new status message so the error information is not lost.
```
chwarr, line 84: Any concerns that copying over the original status code and message will leak information outside the expected boundary? (E.g., typically a top level exception handler for a RPC method will convert an unhandled exception into a generic INTERNAL error without including exception details--unless in debug mode--so that an attacker who causes an INTERNAL error doesn't easily get more information.)

ejona86: The status was received by the client (on the control plane), so it should be safe for the client to see (on the data plane). I'd worry that if we didn't include it we'd have trouble tracking down what is going on. Even with this status information the LB/NR isn't adding additional context, so we will probably need to do some guessing already as to where it came from.

I feel like the top-level exception handler is in a different situation as the exception isn't a grpc status.

chwarr: Makes sense. Thanks.

A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Background
+
+gRPC has a control plane consisting mainly of name resolvers and load balancers.
+Separately it has a data plane that processes individual RPCs. Error information
+from the control plane needs to pass to the data plane when errors are
+occurring, but the gRPC data plane is restricted from generating [status codes
+reserved for application use][statuscodes.md]: `INVALID_ARGUMENT`, `NOT_FOUND`,
+`ALREADY_EXISTS`, `FAILED_PRECONDITION`, `ABORTED`, `OUT_OF_RANGE`, and
+`DATA_LOSS`. Those status codes are reserved for each service.
```
markdroth, line 26: Suggest changing "for each service" to "for use by the application".

A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,136 @@
...
+[Spotify outage]: https://engineering.atspotify.com/2022/03/incident-report-spotify-outage-on-march-8/
+[issue 8950]: https://github.com/grpc/grpc-java/issues/8950
+
+## Proposal
+
+Add checks to each language's Channel that prevents the control plane from
+passing "inappropriate statuses" to the data plane. Specifically:
+
+ * When applying a failed Service Config parsing result from the name resolver
+ * When applying a failed Config Selector result from the name resolver (an
```
markdroth, line 61: This can reference [gRFC A31](https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md), where the config selector was specified.

A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,136 @@
...
+ * `ABORTED`
+ * `OUT_OF_RANGE`
+ * `DATA_LOSS`
+
+If an "inappropriate status" is detected, the status should be rewritten to use
+`INTERNAL` status code and include text within the status message making it
+clear a bug occurred. The original status code and message should be included
+within the new status message so the error information is not lost.
+
+Retry policy status code matching may only be checked after the control plane
```
markdroth, line 86: It might be worth explicitly stating that in the case of a pick failure with a disallowed status code, even though the call did not go out on the wire, it should not be subject to transparent retries  That's normally the case for a failed pick anyway, even if the pick failed with a status code that was allowed, so it shouldn't be surprising behavior, but it couldn't hurt to explicitly say so.

markdroth, line 86: It's probably also worth explicitly stating that the interactions with retry code affects only picker failures.  The other two categories of failures (resolver returning an error on the first attempt and config selector) both happen before the retry code, so they will never be retried, regardless of whether the status code is allowed or not.

A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,136 @@
...
+[statuscodes.md]: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+[Spotify outage]: https://engineering.atspotify.com/2022/03/incident-report-spotify-outage-on-march-8/
+[issue 8950]: https://github.com/grpc/grpc-java/issues/8950
+
+## Proposal
+
+Add checks to each language's Channel that prevents the control plane from
+passing "inappropriate statuses" to the data plane. Specifically:
+
+ * When applying a failed Service Config parsing result from the name resolver
```
markdroth, line 60: This isn't necessarily just a failed service config parse; it can also happen if the resolver encounters an error trying to obtain the service config in the first place.

A54-restrict-control-plane-status-codes.md
markdroth, line 86: This should probably reference [gRFC A6](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) for retries.

A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,136 @@
...
+ * `DATA_LOSS`
+
+If an "inappropriate status" is detected, the status should be rewritten to use
+`INTERNAL` status code and include text within the status message making it
+clear a bug occurred. The original status code and message should be included
+within the new status message so the error information is not lost.
+
+Retry policy status code matching may only be checked after the control plane
+status check and rewrite. However, if an inappropriate status is rewritten and
+the retry policy allows retrying `INTERNAL`, implementations are free to not
```
markdroth, line 88: Do you expect any of our implementations to actually do that?  Given the layering involved, it seems like we'd need to do extra work to *not* subject the failure to the normal configured retry policy, and I wasn't expecting that we had any reason to do that.

It might be clearer to just say something like "Note that a pick failure will (or "may", if there's a reason to be flexible) be subject to the configured retry policy, if any, regardless of whether the status code returned by the picker was an allowed code."

A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,136 @@
...
+
+This gRFC does not make it "okay" for a name resolver or load balancer to
+propagate status codes verbatum from the control plane. Such behavior would
+still be considered a bug and the `INTERNAL` error code will cause such bugs to
+fail noticably. The point of the gRFC is to reduce the damage to the system when
+these bugs are triggered. This does not improve reliability directly, but it
+does ease recovery.
+
+This gRFC purposefully does not limit interceptors, even interceptors that
+communicate with a control plane (e.g., RLQS). An interceptor like fault
```
markdroth, line 105: Most readers probably don't know what "RLQS" is, and we don't yet have a gRFC describing it.  How about instead just saying something like "an interceptor communicating with an external control plane for rate-limiting purposes"?

A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,136 @@
...
+
+Earlier in gRPC's life, there was relatively low risk of using a reserved status
+code in the data plane. There was a relatively fixed amount of control plane
+code, and much of it wasn't accessing gRPC-based services (e.g., DNS,
+pick_first, round_robin). At that point in time, a simple audit could reveal
+broken status code usages.
+
+Now, however, it is more common to implement custom name resolvers and load
+balancers and for those components to use gRPC themselves. This happened first
+with gRPC-LB, but also happens with xDS, Route Lookup Service (RLS), and others.
```
markdroth, line 36: We've never publically documented RLS, so it's probably best not to mention it explicitly here.  I think it's fine to say something like "also happens with xDS, and will happen with other control planes in the future".

markdroth (commented):
Looks very good overall!  Comments are mostly wording details.

A54-restrict-control-plane-status-codes.md
ejona86: I added it as a related gRFC. I didn't see any specific part that was useful to link to. The most precise language I found in it was "The retry policy will be implemented in-between the channel and the load balancing policy" which seems too imprecise to be helpful for this gRFC.

A54-restrict-control-plane-status-codes.md
ejona86: Our implementations probably won't, but I'm making it clear there are no guarantees here. We had explicitly stated in our discussions that we didn't really care the impact to retries. I need to say that the retry policy needs to be protected from these bugs, but I didn't want to imply that the retry policy must be consulted. There's really no guarantees there (even if all our implementations are likely to behave the same way).

I don't know what you're trying to achieve with the "Note that a pick failure..." language. That's the opposite of what I was trying to accomplish. If I was okay with that, I'd just delete this sentence without replacement.

A54-restrict-control-plane-status-codes.md
ejona86: That's a lot of words and specificity for a theoretical interceptor. I think I'd sooner remove the RQLS reference and provide no example. I thought that even if you didn't know RQLS it was helpful to those that do know it. I'm now linking to the .proto.

Fault injection now links to the gRFC as well.

A54-restrict-control-plane-status-codes.md
markdroth: I suggest making the word "Retry" at the start of this sentence a link to the gRFC, just so it's clear that that's the kind of retries we're talking about in this paragraph.

A54-restrict-control-plane-status-codes.md
markdroth: Yeah, deleting this sentence might make it more clear.

A54-restrict-control-plane-status-codes.md
markdroth: I guess I'm wondering what value we get out of allowing implementations to not retry.  It's not clear to me when we'd actually want to do that in practice, and I suspect the wording here is going to cause confusion over the actual behavior that users should expect.  It's not clear to me that the flexibility is worth that potential confusion.

My preference here would be to just say that the retry policy will be applied to the rewritten `INTERNAL`.

Alternatively, if we do want to allow implementations to not retry, how about replacing "implementations are free to not retry the RPC" with "implementations may retry the RPC"?  The word "may" implies that it's not mandatory, but the wording is positive instead of negative, which seems to more clearly point to the behavior we actually expect our implementations to exhibit.

ejona86: > I suspect the wording here is going to cause confusion over the actual behavior that users should expect

The point of the wording here was that there is no expected behavior.

> implementations may retry the RPC

That has its own confusions. I inverted it to be "may fail the RPC" which worked fine, but I decided to just remove it. The "may only" already allows such behavior (although not obviously so) and it doesn't matter much.

A54-restrict-control-plane-status-codes.md
```diff
@@ -0,0 +1,156 @@
...
+ * `NOT_FOUND`,
+ * `ALREADY_EXISTS`
+ * `FAILED_PRECONDITION`
+ * `ABORTED`
+ * `OUT_OF_RANGE`
+ * `DATA_LOSS`
+
+If an "inappropriate status" is detected, the status should be rewritten to use
+`INTERNAL` status code and include text within the status message making it
+clear a bug occurred. The original status code and message should be included
```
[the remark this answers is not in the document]
ejona86: I don't think so. We should only log when we can't report an error to the application, and we will be reporting this error to the application.

markdroth (approved):

dfawley (approved):

PR #315, opened 2022-07-26
A50 update: Improve handling of disabling the timer

murgatroid99:
As discussed out of band, the state should be reset more fully when disabling the timer when the OD LB policy is provided a config that completely disables ejection.

### DIFF
--- a/A50-xds-outlier-detection.md
+++ b/A50-xds-outlier-detection.md
@@ -149,7 +149,13 @@ The `outlier_detection` LB policy will have a timer that triggers on a period de
     - If the address is not ejected and the multiplier is greater than 0, decrease the multiplier by 1.
     - If the address is ejected, and the current time is after `ejection_timestamp + min(base_ejection_time * multiplier, max(base_ejection_time, max_ejection_time))`, un-eject the address.
 
-The `outlier_detection` LB policy will store the timestamp of the most recent timer start time. When a new config is provided, if the timer start timestamp is unset, set it to the current time and start the timer for the configured interval, then for each address, reset the call counters. If the timer start timestamp is set, instead cancel the existing timer and start the timer for the configured interval minus the difference between the current time and the previous start timestamp, or 0 if that would be negative. When the timer fires, set the timer start timestamp to the current time. If a config is provided with both the `success_rate_ejection` and `failure_percentage_ejection` fields unset, skip starting the timer and unset the timer start timestamp.
+The `outlier_detection` LB policy will store the timestamp of the most recent timer start time. When a new config is provided, if the timer start timestamp is unset, set it to the current time and start the timer for the configured interval, then for each address, reset the call counters. If the timer start timestamp is set, instead cancel the existing timer and start the timer for the configured interval minus the difference between the current time and the previous start timestamp, or 0 if that would be negative. When the timer fires, set the timer start timestamp to the current time.
+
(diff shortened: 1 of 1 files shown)


A50-xds-outlier-detection.md
```diff
@@ -149,7 +149,13 @@ The `outlier_detection` LB policy will have a timer that triggers on a period de
...
+The `outlier_detection` LB policy will store the timestamp of the most recent timer start time. When a new config is provided, if the timer start timestamp is unset, set it to the current time and start the timer for the configured interval, then for each address, reset the call counters. If the timer start timestamp is set, instead cancel the existing timer and start the timer for the configured interval minus the difference between the current time and the previous start timestamp, or 0 if that would be negative. When the timer fires, set the timer start timestamp to the current time.
+
+If a config is provided with both the `success_rate_ejection` and `failure_percentage_ejection` fields unset, skip starting the timer and do the following:
+
+ - Unset the timer start timestamp.
```
zasweq, line 156: Also stop outstanding timer? To not run interval timer algorithm again?

zasweq (changes requested):

A50-xds-outlier-detection.md
murgatroid99: I added that. It's sort of covered by the previous paragraph, but I think you're right that it's worth explicitly mentioning it here.

A50-xds-outlier-detection.md
```diff
@@ -149,7 +149,14 @@ The `outlier_detection` LB policy will have a timer that triggers on a period de
...
+The `outlier_detection` LB policy will store the timestamp of the most recent timer start time. When a new config is provided, if the timer start timestamp is unset, set it to the current time and start the timer for the configured interval, then for each address, reset the call counters. If the timer start timestamp is set, instead cancel the existing timer and start the timer for the configured interval minus the difference between the current time and the previous start timestamp, or 0 if that would be negative. When the timer fires, set the timer start timestamp to the current time.
```
zasweq, line 152: This sentence doesn't seem accurate. "If the timer start timestamp is set, instead cancel the existing timer and start the timer for the configured interval minus the difference between the current time and the previous start timestamp, or 0 if that would be negative. When the timer fires, set the timer start timestamp to the current time." The and implies that you do both if the timer start timestamp is set, but you only do the latter if it's not a no-op config.

A50-xds-outlier-detection.md
zasweq: Sweet thanks.

A50-xds-outlier-detection.md
zasweq: I guess this is covered by the line later, "skip starting the timer"

zasweq (approved):

markdroth (approved):

A50-xds-outlier-detection.md
murgatroid99: I modified this section to hopefully be clearer. Does it look better now?

zasweq: Yeah thanks I like the conditional, but the logic changed. Your new language only resets the call counters if not a no-op config. I think we want to keep it to clear call counters if timerStartTime is zero irrespective of no-op. Although thinking about it, it should be ok to only clear on a non no-op.

PR #320, opened 2022-08-19
A42 update: io.grpc.channel_id should be initialized with a random number

apolcyn:
Compliments some changes that have been made or are being made:

- https://github.com/grpc/grpc-java/pull/9453
- https://github.com/grpc/grpc-go/pull/5591
- https://github.com/grpc/grpc/pull/28986

### DIFF
--- a/A42-xds-ring-hash-lb-policy.md
+++ b/A42-xds-ring-hash-lb-policy.md
@@ -92,15 +92,15 @@ Here is how gRPC will handle each type of hash policy:
   the way that Envoy does.  However, we will support one special filter state
   [`key`](https://github.com/envoyproxy/envoy/blob/2443032526cf6e50d63d35770df9473dd0460fc0/api/envoy/config/route/v3/route_components.proto#L703)
   called `io.grpc.channel_id`, which will hash to the same value for all
-  requests on a given gRPC channel.  This can be used in similar situations
-  to where Envoy uses `connection_properties` to hash on the source IP address.
-  (Note that we do not recommend that applications create multiple gRPC
-  channels to the same virtual host, but if you do that, then the
-  behavior here will not be exactly the same as using `connection_properties`,
-  because each channel may use a different endpoint.). In order to facilitate
-  an even selection of backends across different channels running in different
-  processes or machines, the value of `io.grpc.channel_id` should be
-  initialized with a uniform random number.
+  requests on a given gRPC channel. In order to facilitate an even selection
+  of backends across different channels (which may or may not be in the same
+  process or machine), the value of `io.grpc.channel_id` should be initialized
+  with a uniform random number. This can be used in similar situations to where
+  Envoy uses `connection_properties` to hash on the source IP address. (Note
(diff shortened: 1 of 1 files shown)


A42-xds-ring-hash-lb-policy.md
```diff
@@ -97,7 +97,10 @@ Here is how gRPC will handle each type of hash policy:
   (Note that we do not recommend that applications create multiple gRPC
   channels to the same virtual host, but if you do that, then the
   behavior here will not be exactly the same as using `connection_properties`,
-  because each channel may use a different endpoint.)
+  because each channel may use a different endpoint.). In order to facilitate
```
dfawley, line 100: > (Note that we do not recommend that applications create multiple gRPC
> channels to the same virtual host, but if you do that, then the
> behavior here will not be exactly the same as using `connection_properties`,
> because each channel may use a different endpoint.)

Did we want that behavior?  Because we could initialize this ID globally to a process if desired.


Also: nit: no period after the parenthetical sentence.  But should this sentence come before it, maybe?

apolcyn: > Did we want that behavior? Because we could initialize this ID globally to a process if desired.

Changed wording to note that we do want even backend distribution across channels, which may or may not be in the same process or machine (i.e. we want even distribution across <i>all</i> channels).

We know of use cases that do use multiple channels within a process, and for those use cases, each channel should have affinity to its own backend (but hot spotting one backend is unnecessary for them).

> Also: nit: no period after the parenthetical sentence. But should this sentence come before it, maybe?

Moved things around to keep the parenthetical sentence at the end.

markdroth (approved):

ejona86 (approved):

dfawley (approved):

A42-xds-ring-hash-lb-policy.md
```diff
@@ -92,12 +92,15 @@ Here is how gRPC will handle each type of hash policy:
...
-  requests on a given gRPC channel.  This can be used in similar situations
-  to where Envoy uses `connection_properties` to hash on the source IP address.
-  (Note that we do not recommend that applications create multiple gRPC
-  channels to the same virtual host, but if you do that, then the
-  behavior here will not be exactly the same as using `connection_properties`,
-  because each channel may use a different endpoint.)
+  requests on a given gRPC channel. In order to facilitate an even selection
+  of backends across different channels (which may or may not be in the same
+  process or machine), the value of `io.grpc.channel_id` should be initialized
+  with a uniform random number. This can be used in similar situations to where
```
chwarr, line 98: Perhaps this instead?

```suggestion
  with a random number from a uniform distribution. This can be used in similar situations to where
```

"uniform random number" could be misread as "generate a single random number and use it for all the channel ids", but I think you're trying to say that each client id should be different (and if you want a uniform distribution across backends, the random number needs to be from a uniform distribution so that hashing doesn't get skewed).

apolcyn: I could see that. Changed to your wording.

PR #327, opened 2022-09-19
Added L102-cpp-version-macros

veblush:


markdroth (approved):
Looks great to me!

As per the gRFC process, we need to wait 10 business days to see if there are any comments before merging this.

chwarr (approved):

L102-cpp-version-macros.md
```diff
@@ -0,0 +1,46 @@
...
+libraries and applications using gRPC C++ to behave differently depending
+on the gRPC C++ version at compile-time.
+
+## Background
+
+It's a widely used code pattern to behave differenly based on the version
+information available at compile-time. Following is an example
+
+```
+#if GRPC_CPP_VERSION_MAJOR > 1 || (GRPC_CPP_VERSION_MAJOR == 1 && GRPC_CPP_VERSION_MINOR >= 50)
```
ctiller, line 22: This example seems really dangerous: there's no way we'd guarantee that a new feature from 1.50 would work the same in 2.0: the point of making that major version change is to eliminate such backwards compatibility guarantees.

I'd instead recommend one number that combines minor & patch version, and when we do a new major version, do that with a different symbol:

```
#if defined(GRPC_CPP_VERSION)
# if GRPC_CPP_VERSION > 5000 // minor version 50, patch 00
// Use 1.50 and up version of thing
# else
// Do 1.xx workaround
# endif
#elif defined(GRPC_CPP2_VERSION)
// use 2.0 api of the thing
#else
#error undetected grpc version 
#endif
```

veblush: This example doesn't show any feature available now will be available in the future. I think nothing can do that.  Also they can do the same thing with current proposal like

```
# ifdef GRPC_CPP_VERSION_MAJOR
  #if GRPC_CPP_VERSION_MAJOR == 2
  #elif GRPC_CPP_VERSION_MAJOR == 1
  ...
#endif
```

One thing I'd like to stick to here is to use the convention that developers are familiar with. I think there are two options we have; a traditional way to provide all version components as this proposal uses like gcc (`__GNUC_MINOR__`) and clang (`__clang_minor__`) and a protobuf way to provide a single number encompassing all version components like 15002 for 1.50.2. I think both would serve the purpose without need to learn something new. (also we can provide both but I'd like to follow this traditional one first) But other than that, it might be too much.

ctiller: The example you put in the text is absolutely doing something dangerous and presupposing that APIs will be the same in future versions. Abseil does something similar to what I'm proposing with their version/lts_version macros.

Minimally we need to make the examples we put in the RFCs safe and robust.

veblush: Yeah, let me update the example.

ctiller (approved):

PR #332, opened 2022-10-06
A27 update: note restrictions on locality priority and weight, and on addresses

markdroth:


dfawley (approved):

markdroth:
I've added a note here about addresses being unique, as per https://github.com/envoyproxy/envoy/issues/23445.

dfawley (approved):
Latest addition LGTM too.

ejona86 (approved):

PR #334, opened 2022-10-13
L104: C++ OpenCensus Plugin Public APIs

yashykt:


markdroth:
I don't think we want this gRFC.  We have not yet made these APIs externally visible, and I don't think we should start now, since OpenCensus is deprecated.

I realize that we need to expose these in the short term for another project, but we're going to keep them marked experimental, so I don't think we need a gRFC for that.

PR #335, opened 2022-10-26
A57: XdsClient Failure Mode Behavior

markdroth:
CC @murgatroid99 @apolcyn

### DIFF
--- a/A57-xds-client-failure-mode-behavior.md
+++ b/A57-xds-client-failure-mode-behavior.md
@@ -38,7 +38,7 @@ cause gRPC to incorrectly fail data plane RPCs).
 This document describes exactly how the XdsClient should behave in the
 face of these failure modes.
 
-### Related Proposals: 
+### Related Proposals:
 * [A27: xDS-Based Global Load Balancing][A27]
 
 [A27]: A27-xds-global-load-balancing.md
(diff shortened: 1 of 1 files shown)


A57-xds-client-failure-mode-behavior.md
```diff
@@ -0,0 +1,206 @@
...
+stream failure, which will make the client slower to restart the stream
+when connectivity is restored.  However, this is considered an
+acceptable trade-off to provide a better error message in the case of
+connectivity failure.
+
+### Handling Resources That Do Not Exist
+
+As per the xDS spec, the SotW protocol variants do not provide any
+explicit mechanism for the server to indicate that that the client has
+subscribed a resource that does not exist.  Therefore, when the client
```
ejona86, line 118: s/subscribed a/subscribed to a/

ejona86 (approved):
Just for the record, I disagree that "unfortunate side-effect of unnecessarily increasing the backoff delay" is negative; I think that could be fine. But it isn't that important and that language has little impact so I'm not concerned with it remaining.

A57-xds-client-failure-mode-behavior.md
markdroth: Fixed.

dfawley (approved):

PR #339, opened 2022-11-17
A6 update: make changes based on conversations

murgatroid99:
This PR includes a few different changes that we discussed while I was implementing the feature in Node. The main changes are:

 - Removing the underspecified statement that LB policies should use a list of previous backend choices when choosing a new backend.
 - Making the specifics of the memory management configuration options implementation choices.
 - Clarifying how LB policy pick results should be handled by retry logic, and how the client library detects that "The RPC reaches the server, but has never been seen by the server application logic".
 - Making it optional for client libraries to expose the `grpc-previous-rpc-attempts` value in response metadata.

I also wanted to update the graphics for the Transparent Retries section, but I don't know where the original drawings are.

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -3,8 +3,8 @@ gRPC Retry Design
 * Author(s): [Noah Eisen](https://github.com/ncteisen) and [Eric Gribkoff](https://github.com/ericgribkoff)
 * Approver: a11r
 * Status: Implemented
-* Implemented in: Java, .NET, Go except hedging, and C-Core except hedging
-* Last updated: 2022-02-14
+* Implemented in: Java, .NET, Node, Go except hedging, and C-Core except hedging
+* Last updated: 2022-11-17
 * Discussion at: https://groups.google.com/forum/#!topic/grpc-io/zzHIICbwTZE
 
 Table of Contents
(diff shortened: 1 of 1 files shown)


A6-client-retries.md
```diff
@@ -284,21 +282,24 @@ When an RPC is evicted from the buffer, pending hedged requests should be cancel
...
+In the second case, in which load balancing succeeds but the RPC never leaves the client, the client library can transparently retry until a success occurs, or the RPC's deadline passes.
 
-If the RPC reaches the gRPC server library, but has never been seen by the server application logic (the second case), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy. This extra caution is needed because this case involves extra load on the wire.
+If the RPC reaches the gRPC server library, but has never been seen by the server application logic (the third case), the client library will immediately retry it once. If this fails, then the RPC will be handled by the configured retry policy. This extra caution is needed because this case involves extra load on the wire. The client knows it is in this case if the stream ends with an `RST_STREAM` frame with the error code `REFUSED_STREAM` or if the HTTP/2 connection closes with a `GOAWAY` frame with the last stream identifier less than the stream's ID.
```
markdroth, line 302: The new sentence at the end of this paragraph is specific to the HTTP/2 transport, but there may be other transports, so I think we need to explain this in that context.  I suggest starting the sentence with "In the gRPC HTTP/2 transport, this occurs when".

markdroth (commented):
Thanks for doing this!

The diagrams for this design are stored here in SVG form.  I believe you can import that form into Google Drawings to modify it, then re-export as both SVG and PNG form to include in this PR.

A6-client-retries.md
murgatroid99: Done, with a slightly different wording.

murgatroid99:
I have not found a way to import an SVG into Google Drawings without using either a sketchy third-party site or a sketchy third-party program to turn it into an unfamiliar file format that Google Drawings apparently accepts for imports.

murgatroid99:
I will make a separate PR to suggest replacing and updating the diagrams with Mermaid diagrams.

PR #341, opened 2022-12-01
A52: Locality weight attribute for each address

temawi:
To account for an edge case where the same locality appears under more than one priority, each address should have the locality weight as an attribute instead of keeping them in a map.

@markdroth 
@ejona86 
@dfawley


markdroth (approved):
Thanks for updating this!

PR #343, opened 2023-01-28
A58: Weighted Round Robin LB Policy

yousukseung:


A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+  // Can be called concurrently.
+  function Pick();
+};
+```
+
+### Control Plane Behaviors
+
+`weighted_round_robin` control plane works identically to `round_robin` except the handling load reports and weights and the scheduler. The current `round_robin` implementation in each language should serve as full reference. This section outlines the differences and key behaviors.
+
+#### Aggregate Connectivity State
```
markdroth, line 88: Please combine this with the "Managing Subchannels" section and call it something like "Handling Subchannel Connectivity State Notifications".

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+In Go, `baseBalancer` handles the address update as in `round_robin`. When the address set changes, the `baseBuilder` calls the picker builder of `weighted_round_robin` if the aggregate connectivity state is `READY`. When the connectivity state of a subchannel changes, it triggers the same process as when the address set changes.
+
+In all languages, when a subchannel state is updated to `IDLE`, the control plane immediately requests a connection and treats it as if it is in `CONNECTING` for the aggregation purpose.
```
markdroth, line 111: Please change "control plane" to "LB policy".

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Subchannel Weights
+
+To enforce `blackout_period` and `weight_expiration_period`, it tracks two timestamps along with the weight. `last_updated` tracks the last time a non-zero report was received. `non_empty_since` tracks when the first non-zero load report was received, and it is reset when a subchannel comes back to `READY` or the weight expires so that it can be updated with the next non-zero load report. The blackout and expiration are enforced when the weight value is looked up. When the weight expires, `non_empty_since` is reset.
```
markdroth, line 115: "it tracks" -> "the LB policy tracks"

markdroth, line 115: I don't think we need the last sentence here ("When the weight expires, `non_empty_since` is reset."), since that's already covered two sentences earlier ("it is reset when a subchannel comes back to `READY` or the weight expires").

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+The weight of a backend is calculated as ***weight = qps / cpuutilization***.
+
+```
+function UpdateWeight(qps, cpu_utilization) {
+  var new_eight =
```
markdroth, line 121: This should be "new_weight".

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+     cpu_utilization == 0 ? 0 : qps / cpu_utilization;
+  if (new_eight == 0) return;
```
markdroth, line 123: This should also be "new_weight".

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+  last_updated_ = now();
+  weight_ = new_weight;
+}
+function OnSubchannelBecomesReady() {non_empty_since_ = infy;}
+function GetWeight() {
+   if (now() - last_updated_ >= weight_expiration_period_) {
+      non_empty_since_ = infy;
+      return 0;
+   }
+   if (now() - non_empty_since_ < blackout_period_) return 0;
```
markdroth, line 134: Need to also check `blackout_period_ > 0`.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+   return weight_;
+}
+
+```
+
+### Per-Call Load Reporting
+
+Necessary support for per-call load report handling in clients depends on [gRFC A51]. This section outlines how this features works with `weighted_round_robin`. xDS `cluster_impl` policy can serve as an example.
```
markdroth, line 142: I think this paragraph is not necessary.  We've already said above that this relies on gRFC A51.  And it's not useful to refer to existing code as a reference; the description here should be enough to explain how this works without that.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### C/C++
+Processing per-call load reports is supported with `SubchannelCallTrackerInterface`. `weighted_round_robin` implements this interface and attaches it to a subchannel `PickResult`. Later Finish() is called back with the load metrics and updates the weight.
+
+#### Java
+`OrcaPerRequestUtil.OrcaReportingTracerFactory` is available in `io.grpc.xds.orca` to process per-call backend load reports in `ClusterImplLoadBalancer`. This needs to be moved to a new package and shared with `weighted_round_robin`. Then the picker attaches this tracer to the subchannel selection and the listener takes the reference to the subchannel to which it updates the weight to in `onLoadReport()`.
+
+#### Go
+Per-call load report handling is fully implemented. Picker selection result `PickResult` has an optional callback `Done()` that is passed with the backend load metrics parsed from the response if available. With `weighted_round_robin`, the picker only sets Done() in `PickResult` when per-call load reporting is enabled, and the callback updates the weight.
```
markdroth, line 153: The first sentence is not necessary.  We've already said above that this depends on gRFC A51.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+`weighted_round_robin` defaults to per-call load reporting. The client does nothing when the per-call load report is expected but missing in the trailing metadata of the response. While `enable_oob_load_report` is set, per-call load reports are never parsed on the client side. The client does nothing when the per-call load report is present in the response when not expected.
+
+#### C/C++
+Processing per-call load reports is supported with `SubchannelCallTrackerInterface`. `weighted_round_robin` implements this interface and attaches it to a subchannel `PickResult`. Later Finish() is called back with the load metrics and updates the weight.
+
+#### Java
+`OrcaPerRequestUtil.OrcaReportingTracerFactory` is available in `io.grpc.xds.orca` to process per-call backend load reports in `ClusterImplLoadBalancer`. This needs to be moved to a new package and shared with `weighted_round_robin`. Then the picker attaches this tracer to the subchannel selection and the listener takes the reference to the subchannel to which it updates the weight to in `onLoadReport()`.
```
markdroth, line 150: There should be some way to express what needs to happen here that does not refer back to the xds_cluster_impl policy.  @ejona86

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### Go
+Per-call load report handling is fully implemented. Picker selection result `PickResult` has an optional callback `Done()` that is passed with the backend load metrics parsed from the response if available. With `weighted_round_robin`, the picker only sets Done() in `PickResult` when per-call load reporting is enabled, and the callback updates the weight.
+
+### OOB Load Reporting
+
+OOB load report handling in clients depends on [gRFC A51]. This section outlines how this feature works with `weighted_round_robin`.
```
markdroth, line 157: Same as above, I don't think this paragraph is needed.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+In Go, `baseBalancer` handles the address update as in `round_robin`. When the address set changes, the `baseBuilder` calls the picker builder of `weighted_round_robin` if the aggregate connectivity state is `READY`. When the connectivity state of a subchannel changes, it triggers the same process as when the address set changes.
+
+In all languages, when a subchannel state is updated to `IDLE`, the control plane immediately requests a connection and treats it as if it is in `CONNECTING` for the aggregation purpose.
+
+### Subchannel Weights
```
markdroth, line 113: Please add the following paragraph to this section:

"""
In Java and Go, the weight will be stored in a wrapped subchannel.  However, in C++, that approach won't work, because we create a new subchannel object for each address whenever we get a new address list, and we don't want to lose existing weight information when that happens.  So instead, we store the weights in a map keyed by address, and each subchannel list will take its own ref to the entry in the map for each subchannel in the list.
"""

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### Go
+Per-call load report handling is fully implemented. Picker selection result `PickResult` has an optional callback `Done()` that is passed with the backend load metrics parsed from the response if available. With `weighted_round_robin`, the picker only sets Done() in `PickResult` when per-call load reporting is enabled, and the callback updates the weight.
+
+### OOB Load Reporting
+
+OOB load report handling in clients depends on [gRFC A51]. This section outlines how this feature works with `weighted_round_robin`.
+
+In all languages, `report_interval` in `OrcaLoadReportRequest` is determined by the service config parameter `oob_reporting_period`. When it changes, a new stream request is sent. In Java/Go, the implementation needs to add support to enable and disable the OOB listener when `enable_oob_load_report` is updated. In C++ this happens automatically as a new subchannel list is created.
```
markdroth, line 159: The first sentence is a little misleading, because there is no guarantee that the `oob_reporting_period` in the WRR config will actually determine the `reporting_interval` sent on the ORCA stream.  It's possible that there could be another LB policy in the tree that is watching the same subchannel with a lower reporting interval, in which case that will be used instead.

Suggest the following wording instead:

"""
When OOB reporting is enabled, the LB policy gets data via an OOB watch on each subchannel with the requested reporting interval set to the value of the `oob_reporting_period` LB config field.  The watcher is started whenever the `enable_oob_load_report` LB config field becomes true and stopped whenever it becomes false, and it is restarted whenever the value of the `oob_reporting_period` field changes.  (In C++, these changes happen automatically, because we create a new subchannel for each resolver update, and we create a new OOB watcher if appropriate for each new subchannel object.)
"""

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+`weighted_round_robin` initiates OOB load reporting when the LB policy parameter `enable_oob_load_report` changes to true. When the backend returns `UNIMPLEMENTED` to the OOB load report request it proceeds with a log message as if it is not enabled and does not retry.
```
markdroth, line 161: The second sentence here is not needed, because that's already covered in gRFC A51.

Suggest just removing this paragraph in favor of the alternative wording I suggested above.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+`OrcaOobUtil` is available in `io.grpc.xds.orca` but currently not in use. This needs to be moved to a new package and shared with `weighted_round_robin`, which implements `ForwardingLoadBalancerHelper` that takes the helper from `OrcaOobUtil`. The listener takes the reference to the subchannel to which it updates the weight to in `onLoadReport()`. A new API is needed to set and unset the listener in `OrcaOobUtil` to toggle the OOB load report processing.
+
+#### Go
+
+`orca.OOBListener` is available for OOB load reporting. `weighted_round_robin` extends it and attach it to the subchannel to which it updates the weight in `onLoadReport()`. A new API is needed to support enabling and disabling OOB load reporting.
+
+
+### xDS Integration
+
+`weighted_round_robin` is enabled as the endpoint picking policy of `WrrLocality` of [gRFC A52]. The implementation is available in C++ and Java but not in Go yet.
```
markdroth, line 178: Please remove the second sentence.  This gRFC is about the design, not the implementation timeline.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+`weighted_round_robin` is added as a new LB policy per [gRFC A52] and is added to `XdsLbPolicyRegistry`. The name is prefixed with `ClientSide` to avoid confusion with the existing weighted round robin support via `ROUND_ROBIN`.
```
markdroth, line 180: This is probably a good place to mention the pending gRFC A34.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+  // …
+  policy_config_factories_.emplace(
+      ClientSideWeightedRounbRobinLbPolicyConfigFactory::Type(),
+      absl::make_unique<ClientSideWeightedRounbRobinLbPolicyConfigFactory>());
+}
+```
+
+In the xDS configuration for the client, `client_side_weighted_round_robin` is wrapped in `wrr_locality` as a child policy. Below is an example xDS configuration with `client_side_weighted_round_robin`. The config includes `round_robin` as a backup policy to fallback in case the client does not support `client_side_weighted_round_robin`. On the client the new xDS LB hierarchy in [gRFC A52] creates the gRPC LB policy `weighted_round_robin` as the endpoint-picking policy.
+
+```json
```
markdroth, line 210: This isn't JSON, it's textproto.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+{ // Cluster
+  name: "..."
+  type: EDS
+  load_balancing_policy: {
+    policies: [
+      typed_extension_config: {
+        name: "..."
+        typed_config: {
+          "@type": "type.googleapis.com/envoy.extensions.load_balancing_policies.wrr_locality.v3.WrrLocality"
```
markdroth, line 219: Suggest writing this in the normal textproto notation for `Any` fields:

```
typed_config {
  [type.googleapis.com/envoy.extensions.load_balancing_policies.wrr_locality.v3.WrrLocality] {
    endpoint_picking_policy: {
      ...
    }
  }
}
```

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+          endpoint_picking_policy: {
+            policies: [
+              typed_extension_config: {
+                name: "..."
+                typed_config: {
+                  "@type": "type.googleapis.com/envoy.extensions.load_balancing_policies.client_side_weighted_round_robin.v3.ClientSideWeightedRoundRobin"
```
markdroth, line 225: Same comment here with regard to textproto notation for `Any` fields.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+                    },
+                    // …
+                  ]
+                }
+              }
+              // Backup RR policy to fall back if WRR is unsupported.
+              typed_extension_config: {
+                name: "..."
+                typed_config: {
+                  "@type": "type.googleapis.com/envoy.extensions.load_balancing_policies.round_robin.v3.RoundRobin"
```
markdroth, line 240: Same here.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
+A58: `weighted_round_robin` LB policy
+----
+* Author(s): @yousukseung
+* Approver: @markdroth
+* Status: Draft
+* Implemented in: -
+* Last updated: 2023-01-24
+* Discussion at: -
```
markdroth, line 8: Please create a discussion thread for this and add the link here, as per the gRFC process.

markdroth (commented):
This looks like a good start!

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Background
+
+Today in gRPC weighted endpoint load balancing is possible with resolver supplied weights combined with the round robin policy on the client. This requires the control plane to centrally collect load metrics and compute weights for load balancing based on the backend load metrics. Client-side weighted round-robin enables load balancing between clients and backends without involving a centralized load collection. The necessary support to transmit load metrics from backends to clients is available such as [Open Request Cost Aggregation (ORCA)] and [gRFC A51]. This design proposes a new top-level load balancing policy `weighted_round_robin` that implements client-side load balancing.
+
+
+### Related Proposals:
+* [gRFC A51]
```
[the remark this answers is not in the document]
yousukseung: Fixed.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Background
+
+Today in gRPC weighted endpoint load balancing is possible with resolver supplied weights combined with the round robin policy on the client. This requires the control plane to centrally collect load metrics and compute weights for load balancing based on the backend load metrics. Client-side weighted round-robin enables load balancing between clients and backends without involving a centralized load collection. The necessary support to transmit load metrics from backends to clients is available such as [Open Request Cost Aggregation (ORCA)] and [gRFC A51]. This design proposes a new top-level load balancing policy `weighted_round_robin` that implements client-side load balancing.
+
+
+### Related Proposals:
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Background
+
+Today in gRPC weighted endpoint load balancing is possible with resolver supplied weights combined with the round robin policy on the client. This requires the control plane to centrally collect load metrics and compute weights for load balancing based on the backend load metrics. Client-side weighted round-robin enables load balancing between clients and backends without involving a centralized load collection. The necessary support to transmit load metrics from backends to clients is available such as [Open Request Cost Aggregation (ORCA)] and [gRFC A51]. This design proposes a new top-level load balancing policy `weighted_round_robin` that implements client-side load balancing.
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+* Author(s): @yousukseung
+* Approver: @markdroth
+* Status: Draft
+* Implemented in: -
+* Last updated: 2023-01-24
+* Discussion at: -
+
+## Abstract
+
+This document proposes a design for a new load balancing policy `weighted_round_robin`. This policy supports client-side weighted round robin load balancing. Unlike `grpclb` which follows the look-aside model, `weighted_round_robin` is client-side and requires direct load reporting from backends to clients. The policy supports either per-call or periodic out-of-band load reporting and depends on the [Open Request Cost Aggregation (ORCA)] support and [gRFC A51]. The client computes and tracks the weight of subchannels and performs weighted round-robin channel selection. The design also proposes to support `weighted_round_robin` as a custom load balancing policy in xDS as proposed in [gRFC A52].
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+* [Open Request Cost Aggregation (ORCA)]
+
+
+## Proposal
+
+Introduce a new LB policy `weighted_round_robin`. This policy implements client-side load balancing with direct load reports from backends. Either per-call or periodic out-of-band (OOB) load reporting is active at a time where per-call is the default. The policy is otherwise largely a superset of the existing policy `round_robin`.
+
+### LB Policy Config and Parameters
+
+Add the policy config with parameters as follows. This policy is enabled as the endpoint picking policy of `WrrLocality` in [gRFC A52], see [below](#xds-integration) for more.
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+```textproto
+message LoadBalancingConfig {
+   // …
+   WeightedRoundRobin weighted_round_robin_experimental = 18 [json_name = "weighted_round_robin_experimental "];
+}
+
+message WeightedRoundRobin {
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+`oob_reporting_period` is only used when OOB load reporting is used. Note the effective reporting period is determined as outlined in [gRFC A51] and may be different. The client requests the minimum of values from all interested LB policies, but the server enforces its own minimum reporting period. Defaults to 10 seconds.
+
+The client updates effective weights periodically every `weight_update_period`. This parameter defaults to 1 second. The lowest value allowed is 100ms. When set to a value lower than 100ms the period is set to 100ms. Weight updates triggered by changes in the connectivity state or backend addresses are not affected.
+
+### Earliest Deadline First (EDF) Scheduler
+
+`weighted_round_robin` introduces a scheduler that manages weighted subchannel selection. A new scheduler instance is created every time weights are updated.  A state object is passed when created so that the scheduler can start where it last left across multiple instances. The scheduler must be light and cheap to create and allow concurrent pick calls. The scheduler returns the index in the input weight array and the caller has to map the index to a subchannel.
```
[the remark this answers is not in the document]
yousukseung: Updated. Also added how EDF scheduling is applied, per Eric's comment.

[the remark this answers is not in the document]
yousukseung: Done, except I said "atomically updated" instead of saying an atomic. So far I avoided mentioning anything specific to the "static stride" as it is one way of implementing the EDF scheduling, and the state being one atomic is specific to the static stride. I guess we can state that i) the state is persisted and ii) updated after each pick calls (atomically). LMK.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+  // `next_seq_fn` returns a monotonically increasing
+  // number that wraps, must allow concurrent calls.
+  function EDFScheduler New(float[] weights, lambda () -> uint32 next_seq_fn);
+  // Returns the index of the subchannel selected.
+  // Can be called concurrently.
+  function Pick();
+};
+```
+
+### Control Plane Behaviors
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+The scheduler resides in the picker. When compared to `round_robin`, the scheduler conceptually replaces the monotonically increasing index counter of the picker. The picker in `weighted_round_robin` otherwise works the same externally as in `round_robin`.
+
+In C/C++, the scheduler is reference counted and its access is mutex protected. We may consider optimization in the future such as using RCU synchronization when available. Java/Go implementation accesses the scheduler with an atomic reference object.
+
+The picker starts a timer to swap the scheduler to update update weights every `weight_update_period`. This timer iterates over the subchannel list (wrapper) of the picker, loads the latest weight from each, builds a new scheduler and replaces it with the current one. Note this timer does not run in the control plane, but weights are atomic and reading weights does not require extra synchronization with the control plane. When the scheduler is swapped, ongoing pick calls that already took a reference of the outgoing scheduler can continue to refer to the old scheduler until the call returns.
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,251 @@
...
+
+The behavior remains the same as in `round_robin` in all languages.
+
+In C/C++, the control plane creates a new pending subchannel list upon getting a new address list and starts watching the connectivity state of its subchannels. Only subchannels in the `READY` state are included. The pending list is immediately promoted to the current list if it’s an initial update or the new list is empty (i.e. no subchannel). Otherwise the pending list is later promoted to the current list if the current list has none `READY`, or the pending list has at least one `READY`, or all subchannels in the pending list are `TRANSIENT_FAILURE`. The rest of the process is triggered upon the connectivity state of a subchannel changes in either the pending or current list. When the connectivity state of a subchannel changes and it is in the pending subchannel list, it checks whether to promote the pending list to current. If it’s in the current list it updates the aggregate connectivity state and creates a new picker. The kind of picker created depends on the aggregate state: an active Picker (w/ a scheduler) if `READY`, a QueuePicker if `CONNECTING`, or a TransientFailurePicker otherwise.
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Updated, PTAL.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Done.

ANA885874:
Thank you

A58-client-side-weighted-round-robin-lb-policy.md
ejona86: `ClusterImplLoadBalancer` isn't involved at all today.

yousukseung: Updated to only mention OrcaPerRequestUtil.OrcaReportingTracerFactory. Made similar changes to OOB and Go, PTAL.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -28,42 +29,56 @@ Introduce a new LB policy `weighted_round_robin`. This policy implements client-
...
-
 ### Earliest Deadline First (EDF) Scheduler
 
-`weighted_round_robin` introduces a scheduler that manages weighted subchannel selection. A new scheduler instance is created every time weights are updated.  A state object is passed when created so that the scheduler can start where it last left across multiple instances. The scheduler must be light and cheap to create and allow concurrent pick calls. The scheduler returns the index in the input weight array and the caller has to map the index to a subchannel.
+`weighted_round_robin` introduces a scheduler that manages weighted subchannel selection. A new scheduler will be created periodically to take new weight information into account. A state object is passed when created so that the scheduler can start where it last left across multiple instances. The scheduler must be light and cheap to create and allow concurrent pick calls. The scheduler returns the index in the input weight array and the caller has to map the index to a subchannel. The scheduler's state is stored in the LB policy and not the picker, and is atomically updated at the end of each pick by the scheduler.
```
markdroth, line 73: This doesn't say anything about how this state object is actually used.  What value is stored in the state object?  How is it used and updated by the scheduler?

Also, have we resolved whether we want all languages to use this approach, or are we okay with having different languages use different approaches?  I thought we said that our internal implementation does not use this atomic state approach in Java or Go.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -205,39 +202,31 @@ XdsLbPolicyRegistry::XdsLbPolicyRegistry() {
...
-                  fields: [
-                    {
-                      key: "enable_oob_load_report"
-                      value: false
-                    },
-                    // …
-                  ]
+        typed_config {
+          [type.googleapis.com/envoy.extensions.load_balancing_policies.wrr_locality.v3.WrrLocality] {
+          endpoint_picking_policy {
```
markdroth, line 217: This (and everything inside of it) should be indented one more level.

markdroth (commented):
This looks really good!  I think the only significant remaining issue is the one about the scheduler algorithm; I think we need more details about exactly what the requirements are for implementations and what things are permitted to be different across implementations.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Re-formatted.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Updated after chatting with Ivy and Erik.

I removed anything specific to the static stride and the global state object (next_seq lambda). Java will for now use a priority queue, randomize the starting index with each scheduler, and hold a lock in pick() (Java stubby implementation).

markdroth (approved):
Looks great!

ejona86 (approved):
More than a few comments, but most are very easy. The one about the environment variable is probably the most important of the comments.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+
+  // Load reporting interval to request from the server.  Note that the
+  // server may not provide reports as frequently as the client requests.
+  // Used only when enable_oob_load_report is true.  Default is 10 seconds.
+  google.protobuf.Duration oob_reporting_period = 2;
+
+  // A given endpoint must report load metrics continuously for at least
+  // this long before the endpoint weight will be used.  This avoids
+  // churn when the set of endpoint addresses changes.  Takes effect
+  // both immediately after we establish a connection to an endpoint and
```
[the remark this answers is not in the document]
yousukseung: I think we cannot tell if the remote end remained online, or rarely if it's the same VM since we disconnected. But the only bad case I can think of is where the set of back end keeps changing fast while they remain healthy, but this sounds unexpected to me.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+  // How often endpoint weights are recalculated.  Default is 1 second.
+  google.protobuf.Duration weight_update_period = 5;
+}
+```
+
+### Earliest Deadline First (EDF) Scheduler
+
+`weighted_round_robin` introduces a scheduler that manages weighted subchannel selection. A new scheduler will be created periodically or when the set of weights change to take new weight information into account. The scheduler must be cheap to create and allow concurrent pick calls to proceed with no or minimal contention. The scheduler returns the index in the input weight array and the caller has to map the index to a subchannel.
+
+The scheduler implements the [earliest deadline first][EDF] (EDF) scheduling. In its implementation, each subchannel is treated as a job whose deadline is inversely proportional to its weight. To avoid synchronization, each subchannel is given a random credit in the range [0, deadline] when first inserted. The subchannel with the next earliest deadline is selected each time and then inserted back with its deadline.
```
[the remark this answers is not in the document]
yousukseung: Updated.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+
+In C/C++, the scheduler is reference counted and its access is mutex protected. We may consider optimization in the future such as using RCU synchronization when available. Java/Go implementation accesses the scheduler with an atomic reference object.
+
+The picker starts a timer to swap the scheduler to update update weights every `weight_update_period`. This timer iterates over the subchannel list (wrapper) of the picker, loads the latest weight from each, builds a new scheduler and replaces it with the current one. Note that synchronization must be used to access the weights, since they are going to be set based on backend metric reports in either RPC reports or OOB responses but then accessed in this timer. When the scheduler is swapped, ongoing pick calls that already took a reference of the outgoing scheduler can continue to refer to the old scheduler until the call returns.
+
+Note only subchannels in `READY` state are passed to the scheduler. Subchannels that are included with a zero weight are still picked but with the average weight. When less than two subchannels have load info, it degrades to `round_robin`.
```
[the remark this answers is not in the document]
yousukseung: Re-worded. We don't swap in this case but WRR just works with uniform weight. so this part is just an FYI.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+
+`weighted_round_robin` uses the earliest deadline first (EDF) scheduling with a monotonically increasing counter as the state. This counter is persisted across multiple scheduler instances for continuity between weight updates.
```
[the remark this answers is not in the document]
yousukseung: Oops, I think this sentence was accidentally left behind, removed.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+
+When `oob_reporting_period` changes, the client implementation starts a new OOB load reporting stream as outlined in [gRFC A51][A51].
+
+Changes to `weight_update_period`, `blackout_period`, and `weight_expiration_period` take effective with a new picker that we always create upon receiving a resolver update.
+
+In all languages, `weighted_round_robin` policy de-duplicates redundant addresses in resolver updates. This behavior is different from `round_robin` where it uses the number of representations in the list as a resolver supplied weight.
```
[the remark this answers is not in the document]
yousukseung: I removed the second part for now, that's not part of the design and I think we can drop.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+
+In Go, `baseBalancer` handles the address update as in `round_robin`. When the address set changes, the `baseBuilder` calls the picker builder of `weighted_round_robin` if the aggregate connectivity state is `READY`. When the connectivity state of a subchannel changes, it triggers the same process as when the address set changes.
+
+In all languages, when a subchannel state is updated to `IDLE`, the LB policy immediately requests a connection and treats it as if it is in `CONNECTING` for the aggregation purpose.
+
+### Subchannel Weights
+
+To enforce `blackout_period` and `weight_expiration_period`, the LB policy tracks two timestamps along with the weight. `last_updated` tracks the last time a non-zero report was received. `non_empty_since` tracks when the first non-zero load report was received, and it is reset when a subchannel comes back to `READY` or the weight expires so that it can be updated with the next non-zero load report. The blackout and expiration are enforced when the weight value is looked up.
```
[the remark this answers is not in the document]
yousukseung: Updated.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+
+### OOB Load Reporting
+
+When OOB reporting is enabled, the LB policy gets data via an OOB watch on each subchannel with the requested reporting interval set to the value of the `oob_reporting_period` LB config field. The watcher is started whenever the `enable_oob_load_report` LB config field becomes true and stopped whenever it becomes false, and it is restarted whenever the value of the `oob_reporting_period` field changes. (In C++, these changes happen automatically, because we create a new subchannel for each resolver update, and we create a new OOB watcher if appropriate for each new subchannel object.)
+
+In C/C++,  processing OOB load reports is supported with `OobBackendMetricWatcher`. `weighted_round_robin` implements this interface. The watcher takes the reference of the subchannel weight object and update weights to it.
+
+In Java, `OrcaOobUtil` is available in `io.grpc.xds.orca` for OOB load reports but a new API is needed to set and unset the listener in `OrcaOobUtil` to toggle the OOB load report processing.
```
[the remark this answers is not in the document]
yousukseung: Done.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+
+In Go, `orca.OOBListener` is available for OOB load reporting but a new API is needed to support enabling and disabling OOB load reporting.
+
+
+### xDS Integration
+
+`weighted_round_robin` is enabled as the endpoint picking policy of `WrrLocality` of [gRFC A52][A52].
```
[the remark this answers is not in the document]
yousukseung: Updated.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -0,0 +1,249 @@
...
+
+`weighted_round_robin` is added as a new LB policy per [gRFC A52][A52] and is added to `XdsLbPolicyRegistry`. The name is prefixed with `ClientSide` to avoid confusion with the pending [gRFC A34][A34] where the endpoint weights are sent by the control plane.
```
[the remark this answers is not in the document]
yousukseung: @markdroth I don't know how we mention experimental stuff in gRFC, do we want to s/weighted_round_robin/weighted_round_robin_experimental/ everywhere here elsewhere?  Maybe having a section at the end that mentions this is yet experimental and is env var proetected?

A58-client-side-weighted-round-robin-lb-policy.md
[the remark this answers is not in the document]
ejona86: I'm not concerned with the set of backends changing. In that case we'll throw away any data we have for the backend so it'll show up as "new" when it gets re-added.

If the server has MAX_CONNECTION_AGE (or MAX_CONNECTION_IDLE, but that's less useful for round-robin like policies) enabled, then it can disconnect the client periodically. That is used to force clients to do DNS lookups to find more backends, for example. There's a few other rarer cases it could be used.

There's also the concern of cases like "disconnected because keepalive failure." Let's say a server is over 100% utilized and falling over. It may not respond promptly and the connection fails. When the client reconnects the client will use average load instead of avoiding it. It is unlikely a single instance will get in that situation when using WRR. But it is still foreign to me (with round-robin-style LB) to tie state to the particular connection. (With pick-first, there's proxies, so such things are easily appropriate.)

A58-client-side-weighted-round-robin-lb-policy.md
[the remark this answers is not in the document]
markdroth: For the LB policy itself, we used the `_experimental` suffix instead, so it didn't seem necessary to also use an env var.  But we do use an env var for the xDS functionality.  Once we pass interop tests, we will rename the policy to remove the `_experimental` suffix and remove the env var protection from the xDS code.

I agree that we need to call out this protection in the gRFC.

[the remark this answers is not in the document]
yousukseung: Updated, only added `GRPC_EXPERIMENTAL_XDS_WRR_LB`. We have `GRPC_EXPERIMENTAL_XDS_CUSTOM_LB_CONFIG` but I think this belongs to A52.

And I did not update/add anything regarding the name `weighted_round_robin_experimental`, let me know if we still need to include this in gRPC (i.e. this is currently experimental).

A58-client-side-weighted-round-robin-lb-policy.md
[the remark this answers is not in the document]
ejona86: I don't know what you mean by grace period. I don't see any existing references to such a period. And it isn't clear to me how providing configuration for such a thing would help. I'm fine revisiting this if there become issues mixing with polling resolvers.

A58-client-side-weighted-round-robin-lb-policy.md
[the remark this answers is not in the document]
markdroth: Let's also mention that the LB policy itself will be registered with an `_experimental` suffix until the feature is deemed stable.

A58-client-side-weighted-round-robin-lb-policy.md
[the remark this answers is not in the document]
markdroth: Eric, I think you have a good point that MAX_CONNECTION_AGE might not play well here.  If that becomes a problem, users can set the blackout period to 0.  This would admittedly be a trade-off between improving handling of MAX_CONNECTION_AGE and potentially getting bad data when a server first restarts (which is what the blackout period is meant to avoid), but at least there's a knob for it.  And I do agree that this is something we can improve later if it actually becomes a problem in practice.

A58-client-side-weighted-round-robin-lb-policy.md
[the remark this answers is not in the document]
yousukseung: Updated.

A58-client-side-weighted-round-robin-lb-policy.md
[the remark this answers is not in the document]
yousukseung: Thanks, I think we are good to close this discussion for now then. Agree with what Mark replied.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -237,7 +237,9 @@ In the xDS configuration for the client, `client_side_weighted_round_robin` is w
 }
 ```
 
-### Temporary environment variable protection
+### Experimental Status and Environment Variable Protection
```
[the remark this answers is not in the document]
yousukseung: Updated, I took A42- as a reference.

yousukseung:
Thanks all for reviewing this.

PR #345, opened 2023-02-08
A51 update: add ServerMetricRecorder API and QPS field

yousukseung:


A51-custom-backend-metrics.md
```diff
@@ -380,21 +415,122 @@ public interface OrcaOobReportListener {
...
-for both per-RPC and OOB mechanisms.  However, the C-core LB policy API
-is not yet ready to be made public, so this gRFC does not describe those
-APIs.  They will be described in a subsequent gRFC when the LB policy
-API is ready to be made public.
+#### Server
 
-This section describes only the C++ server-side APIs.
+C++ will provide `ServerMetricRecorder` to record metrics machine-wide.
+The user creates an instance and records metrics to it. All the methods are thread safe.
+This is optional for per-call load reporting and required for OOB load reporting
```
markdroth, line 422: Need a period at the end of this sentence.

A51-custom-backend-metrics.md
```diff
@@ -380,21 +415,122 @@ public interface OrcaOobReportListener {
...
+  CallMetricRecorder* GetCallMetricRecorder();
+};
+```
 
-C++ provides the following API for running the ORCA service:
+To enable per-call load reporting, an application must enable call metric recording when building the server.
+The application may optionally pass `ServerMetricRecorder` to also include machine-wide metrics in per-call load reports.
+
+```c++
+std::unique_ptr<ServerMetricRecorder> server_metric_recorder_ =
```
markdroth, line 525: No need for the trailing underscore in the variable name here.

Same thing on line 530.

markdroth (commented):
This looks really good!  Comments are all minor.

I'll let @ejona86 be the approver for this PR, since he was the approver for this gRFC.

A51-custom-backend-metrics.md
```diff
@@ -133,10 +122,47 @@ function setMemoryUtilizationMetric(double value);
...
+as part of intercepting the trailing metadata sent from the server.
+Note that the API must be structured such that if multiple LB policies are interested in the data (e.g., both xDS and WRR),
+the protobuf deserialization needs to happen only once. The exact API for this may differ across languages,
+but here are a couple of possible approaches:
+* The deserialized data can be attached to the metadata element, so that it can be deserialized once and accessed multiple times.
+* The deserialized data can be stored in the call data and accessed by the LB policy code when it intercepts the trailing metadata.
+
+#### The Server API
+The server will add a channel filter (C/C++) or an interceptor (Java/Go) that
+adds recorded metrics to the trailing metadata of the response. It reads from both the per-call and machine-wide metrics recorder if they exist. When the same metric is recorded with both the per-query and the machine-wide recorder, metrics recorded with per-query recorder takes a higher precedence.
```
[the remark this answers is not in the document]
yousukseung: Done.

A51-custom-backend-metrics.md
```diff
@@ -133,10 +122,47 @@ function setMemoryUtilizationMetric(double value);
...
+To periodically receive metrics data from a backend server, the client opens a stream on the
+established connection to the server to request the load report.
+The request should specify metrics reporting interval. The interval will be validated by the server,
+see validation details below.
+Meanwhile, the server registers an OOB streaming service to emit the metrics.
+
+#### The Server API
+We will provide a builtin implementation of [xds.service.orca.v3.OpenRCAService](https://github.com/cncf/xds/blob/eded343319d09f30032952beda9840bbd3dcf7ac/xds/service/orca/v3/orca.proto#L27)
+streaming service defined by ORCA in each supported language.
+The implementation will accept the above mentioned machine-wide metrics recorder implementation to read recorded metrics. This is required.
```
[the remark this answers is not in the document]
yousukseung: Done, and made it shorter.

A51-custom-backend-metrics.md
```diff
@@ -265,10 +291,44 @@ but this can be done in the future if needed without impacting the public interf
...
 
 ### Java
-#### Per-Query Reporting 
 
-At the server side, Java will provide a per-RPC metrics recorder for the user applications to record both utilization and
-request cost metrics. They can obtain a reference to the `CallMetricRecorder` using the static 
+#### Server
+Java will provide a machine-wide metrics recorder on the server. The user
+creates an instance and records metrics to it. All the methods are thread safe.
+This is optional for per-call load reporting and required for OOB load reporting
```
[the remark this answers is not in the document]
yousukseung: Done.

A51-custom-backend-metrics.md
yousukseung: Done.

A51-custom-backend-metrics.md
yousukseung: Done.

markdroth (approved):
Looks great!  @ejona86 should also approve before merging.

yousukseung:
Updated after today's sync:

- Use "per-request" and "per-server" everywhere. This doc already was using per-RPC|Query|Request interchangeably and I updated all to "per-request".
- Java ServerMetricRecorder -> MetricRecorder
- Moved "per-request recording" to before "per-server recording" to keep the current order.
- Made "per-XXX reporting on client" into "per-XX reporting" to cover both the client and server, and restored back what's already there in Java sections on server-side reporting.

A51-custom-backend-metrics.md
```diff
@@ -78,27 +77,36 @@ function recordRequestCostMetric(String name, double value);
 // Records a utilization metric measurement for the call.
 function recordUtilizationMetric(String name, double value);
 
+// Records an opaque named metric measurement for the call.
+function recordNamedMetricsMetric(String name, double value);
```
markdroth, line 81: The named_metrics and eps changes do not belong in this PR.  We'll make those changes in a separate PR.

A51-custom-backend-metrics.md
```diff
@@ -78,27 +77,36 @@ function recordRequestCostMetric(String name, double value);
...
+
 // Records the CPU utilization metric measurement for the call.
 function recordCPUUtilizationMetric(double value);
 
 // Records the memory utilization metric measurement for the call.
 function recordMemoryUtilizationMetric(double value);
 
+// Records the memory utilization metric measurement for the call.
+function recordMemoryUtilizationMetric(double value);
```
markdroth, line 90: This is a duplicate of the method above.

yousukseung: Fixed.

A51-custom-backend-metrics.md
yousukseung: Removed. I think I accidentally rebased on top of the follow-up change (eps/named metrics).

A51-custom-backend-metrics.md
```diff
@@ -380,21 +423,131 @@ public interface OrcaOobReportListener {
...
+  // Records the server CPU utilization in the range [0, 1].
+  // Values outside of the valid range are rejected.
+  // Overrides the stored value when called again with a valid value.
+  void SetCpuUtilization(double value);
+  // Records the server memory utilization in the range [0, 1].
+  // Values outside of the valid range are rejected.
+  // Overrides the stored value when called again with a valid value.
+  void SetMemoryUtilization(double value);
+  // Records number of queries per second to the server in the range [0, infy).
+  // Values outside of the valid range are rejected.
```
sergiitk, line 504: Looking at the [WIP grpc java implementation](https://github.com/grpc/grpc-java/pull/9902), they are ignored, not rejected. Should we clarify here if by "rejected" we mean "ignored" - because rejected may imply the method could throw.

On the other hand, I'm a bit afraid if we just ignore the value. I agree that throwing is a bad idea, but just swallowing an error leads to reporting incorrect data. I don't think we can judge if skewing a particular metric is critical to a customer.

One example that come to mind: a customer is storing qps internally in an int, the service gets a huge number of requests, and the int gets overflowed. In this case, they'll call `setQps()` with a negative value, when in fact there was a huge spike. If another service goes down because of a qps spike, but our service doesn't report the qps increase, an engineer responding to the incident won't be able to identify the source of the request spike correctly.

I wonder if it'd make sense require logging a warning when an incorrect value is presented to the MetricRecorder/CallMetricRecorder. However, depending on how often the data is reported, it might lead to the logs getting flooded. Obviously, we can restrict how often to produce these warnings, but this may be going to far out-of-scope.

yousukseung: This section is for C++ and it's from comments in code so I don't think "rejected" may confuse. Java snippet does not have this word.

Regarding the range validation, I believe the caller can validate or log in such cases. Note that recorder APIs already ignore with no feedback today (before this change) and this PR just clarifies that. Today the value will be rejected during proto's range validation and/or when consumed by the client where neither is explicit.

A51-custom-backend-metrics.md
```diff
@@ -279,67 +303,86 @@ public final class CallMetricRecorder {
...
-```
-
-#### Out-of-Band Reporting 
-At the server side, Java implemented the OOB service `OrcaServiceImpl`: users can optionally provide the minimum
-reporting interval configuration when creating the object. Then they can update utilization metrics
-data using the exposed APIs as follows. All the methods are thread safe.
+#### Per-Server Recording on Server
+Java will provide a per-server metrics recorder on the server as follows. The user
+creates an instance and records metrics to it. All the methods are thread safe.
+This is optional for per-request load reporting and required for OOB load reporting.
```
ejona86, line 323: s/This/An instance/

A51-custom-backend-metrics.md
```diff
@@ -279,67 +303,86 @@ public final class CallMetricRecorder {
...
 
 ```java
 public class MetricRecorder {
   // Update the metrics value corresponding to the specified key.
   public void putUtilizationMetric(String key, double value);
 
   // Replace the whole metrics data using the specified map.
-  public void setAllUtilizationMetrics(Map<String, Double> metrics);
+  public void putAllUtilizationMetrics(Map<String, Double> metrics);
```
ejona86, line 331: Don't change the Java API... These are sets, and the qps metric should also be a set.

A51-custom-backend-metrics.md
```diff
@@ -279,67 +303,86 @@ public final class CallMetricRecorder {
...
 
   // Update the memory utilization metrics data.
-  public void setMemoryUtilizationMetric(double value);
+  public void putMemoryUtilizationMetric(double value);
 
   // Clear the memory utilization metrics data.
   public void clearMemoryUtilizationMetric();
+
+  // Update the queries-per-second metrics data.
+  public void putQpsMetric(double value);
```
ejona86, line 349: @YifeiZhuang, this has "metric" in the method name. The API you added in https://github.com/grpc/grpc-java/pull/9866 was `setQps()`. Should it be `setQpsMetric()`?

YifeiZhuang: yes,  `setQpsMetric()` is better. I will make that change in java.

A51-custom-backend-metrics.md
```diff
@@ -133,10 +120,47 @@ function setMemoryUtilizationMetric(double value);
...
+as part of intercepting the trailing metadata sent from the server.
+Note that the API must be structured such that if multiple LB policies are interested in the data (e.g., both xDS and WRR),
+the protobuf deserialization needs to happen only once. The exact API for this may differ across languages,
+but here are a couple of possible approaches:
+* The deserialized data can be attached to the metadata element, so that it can be deserialized once and accessed multiple times.
+* The deserialized data can be stored in the call data and accessed by the LB policy code when it intercepts the trailing metadata.
+
+#### The Server API
+The server will add a channel filter (C/C++) or an interceptor (Java/Go) that
+adds recorded metrics to the trailing metadata of the response. It reads from both the per-request and per-server metrics recorder if they exist. When the same metric is recorded with both the per-request and the per-server recorder, metrics recorded with per-request recorder take a higher precedence.
```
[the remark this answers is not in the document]
yousukseung: Updated.

A51-custom-backend-metrics.md
yousukseung: Reverted. FYI left `putUtilizationMetric` as is.

ejona86 (approved):

PR #346, opened 2023-03-09
A59: gRPC Audit Logging

rockspore:
Still lacking the language-specific API sections. But I'd like to open this up so that folks commenting on https://github.com/envoyproxy/envoy/pull/26001 can have more context.

### DIFF
--- a/A59-audit-logging.md
+++ b/A59-audit-logging.md
@@ -65,7 +65,7 @@ xDS clients (namely Envoy) should benefit from as well.
 We will add an audit condition enum (see [PR](https://github.com/envoyproxy/envoy/pull/26001))
 in the [xDS RBAC policy][RBAC policy] as below:
 
-```
+```proto
 package envoy.config.rbac.v3;
 
 message RBAC {
--- a/A59-audit-logging.md
+++ b/A59-audit-logging.md
@@ -4,7 +4,7 @@ A59: gRPC Audit Logging
 * Approver: [Eric Anderson](https://github.com/ejona86)
 * Status: In Review {Draft, In Review, Ready for Implementation, Implemented}
 * Implemented in:
-* Last updated: 2023-03-09
+* Last updated: 2023-03-23
 * Discussion at: https://groups.google.com/g/grpc-io/c/LbnuqGqEwuM
 
 ## Abstract
@@ -120,6 +120,9 @@ allowing multiple audit loggers is useful when users plan to roll out a new
 logger. By also keeping the existing logger along for a period they will be able
 to prevent interruption.
 
+The RBAC filter will be NACKed if the configured logger types are not supported
+by gRPC, unless the filter it made optional. This is consistent with [A39: xDS HTTP Filter Support](https://github.com/grpc/proposal/blob/master/A39-xds-http-filters.md).
+
 Some readers may immediately notice that since an HTTP filter chain can contain
 an arbitrary number of RBAC filters, more than one of them could meet its audit
 condition for the same RPC and multiple audit log entries occur. When this happens,
@@ -194,7 +197,7 @@ to:
 
 ```proto
 message AuthorizationPolicy {
-  …
+  ...
 
   message AuditLoggingOptions {
     enum AuditCondition {
@@ -221,6 +224,9 @@ message AuthorizationPolicy {
 Note that the definition above is only for illustration purposes and does not
 actually exist anywhere as a `.proto` file. Nor does gRPC process the policy
 as protobuf by any means.
+
+Similar to the xDS case, unsupported logger types or any other misconfiguration
+will cause the server to fail to start.
 
 The authorization policy is backed by two RBAC filters, a DENY followed by an
 ALLOW. The audit logger configurations will be duplicated in both generated
--- a/A59-audit-logging.md
+++ b/A59-audit-logging.md
@@ -46,12 +46,13 @@ It is worth noting that audit happens right after the authorization
 decision is made, so the status only reflects that, but not the eventual
 one from the server. 
 
-|Fields             |Additional Information                                                             |Example                |
-|-------------------|-----------------------------------------------------------------------------------|-----------------------|
-|RPC Method         |Method the RPC hit.                                                                |"/pkg.service/foo"     |
-|Principal          |Identity the RPC. Currently only available in certificate-based TLS authentication.|"spiffe://foo/user1"   |
-|Policy Name        |Name of the policy that invokes audit logging.                                     |"policy_name"          |
-|Authorized         |Boolean indicating whether the RPC is authorized or not.                           |true                   |
+|Fields             |Additional Information                                                                 |Example                |
+|-------------------|---------------------------------------------------------------------------------------|-----------------------|
+|RPC Method         |Method the RPC hit.                                                                    |"/pkg.service/foo"     |
+|Principal          |Identity the RPC. Currently only available in certificate-based TLS authentication.    |"spiffe://foo/user1"   |
+|Policy Name        |The authorization policy name (or the xDS RBAC filter name).                           |"example-policy"       |
+|Matched Rule       |The matched rule (or the matched policy name in [RBAC][RBAC policy]. Empty if no match.|"admin-access"         |
+|Authorized         |Boolean indicating whether the RPC is authorized or not.                               |true                   |
 
 ### xDS API Changes
 
@@ -118,6 +119,14 @@ message HttpConnectionManager {
 The list of configured audit loggers will all be invoked by an RBAC filter when
 the audit condition is met. This is analogus to the semantics that all access
 loggers will be run after the response is sent.
+
+Some readers may immediately notice that since an HTTP filter chain can contain
+an arbitrary number of RBAC filters, more than one of them could meet its audit
+condition for the same RPC and multiple audit log entries occur. When this happens,
+it would be impossible to uniquely identify a specific RPC from a log entry so as
+to apply any sort of de-duplication. The design in this gRFC does not intend to
+address this issue as we argue it is a rare use case. This will be elaborated more
+in [one of the design alternatives](#audit-condition-in-the-http-connection-manager).
 
 ### gRPC Authorization Policy Changes
 
@@ -274,17 +283,31 @@ access log won't necessarily happen till hours later.
 
 ### Audit Condition in the HTTP Connection Manager
 
-As an HTTP filter chain can contain multiple RBAC filters in general. The typical
+As an HTTP filter chain can contain multiple RBAC filters in general, the typical
 behavior that users will expect if they want to audit on allow is audit should
-only happen once after the evaluation of the last RBAC filter. From this
-perspective, the audit condition could be considered as something on top of all
-RBAC filters and thus configured in the [HTTP Connection Manager][HttpConnectionManager proto].
+only happen once after the evaluation of the last RBAC filter. The gRPC
+authorization policy API assumes this expectation and handles the audit condition
+for the users.
 
-However, given that xDS users normally do not craft RBACs on their own but instead
-rely on the control plane APIs, such as Istio Authorization Policy, it is
-reasonable to let the control plane handle the translation properly.
-Meanwhile, configuring audit condition in individual RBACs offers the flexibility
-when users really want to audit events related to a particular RBAC.
+In the xDS cases, however, the audit condition could be considered as something
+on top of all RBAC filters and thus configured in the [HTTP Connection Manager][HttpConnectionManager proto].
+We decided not to take this approach because the component managing the filter
+chain would have to be aware of the last RBAC filter constantly and inform it
+if performing the audit logging. In other words, this would require some heavy
+lifting implementation-wise which does not make too much sense for such a
+specicialized case as audit logging.
+
+In practice, xDS users normally do not craft RBACs on their own but instead
+rely on the control plane APIs, such as Istio's Authorization Policy, to apply
+RBAC filters to the workloads. Whether the control plane explicitly sets the
+constraint or not, the common practice is to apply at most one authorization
+policy to a workload. So it generally suffices to leave the audit condition
+inside individual RBACs.
+
+To summarize, we acknowledge the possibility of multiple log entries for the
+same RPC and argue this to be an uncommon scenario. Our APIs will not disallow
+such configurations but also intentionally not provide any mechanism for log
+entry de-duplication.
 
 ### Audit Log Configured In gRPC Bootstrap
 
--- a/A59-audit-logging.md
+++ b/A59-audit-logging.md
@@ -284,8 +284,8 @@ use cases.
 ### Built-in logger types
 
 We plan to implement the stdout logger as a built-in logger type. The type is
-named `stdout_logger`. This logger will not support any configuration and it
-outputs log entries to stdout in JSON format.
+named `stdout_logger`. No configuration fields are currently defined for this
+logger and it outputs log entries to stdout in JSON format.
 
 Here is the example JSON configuration in the xDS case (see [PR](https://github.com/envoyproxy/envoy/pull/26453)).
 
@@ -328,9 +328,9 @@ will expose for users to implement their own types of audit loggers. Below are
 two important features that apply to all the languages.
 
 As is also documented in all the languages below, the logging function is
-executed by gRPC syncrhonously during the authorization process. Therefore,
+executed by gRPC synchronously during the authorization process. Therefore,
 its implementation should not block the RPC. When needed, it should invoke any
-long running operations asyncrhonously so that the function itself returns
+long running operations asynchronously so that the function itself returns
 promptly. For this reason, no error or status code will be returned from it.
 
 The logger factory (aka builder/provider) needs to implement a config parsing
@@ -577,14 +577,14 @@ for the users.
 In the xDS cases, however, the audit condition could be considered as something
 on top of all RBAC filters and thus configured in the [HTTP Connection Manager][HttpConnectionManager proto].
 We decided not to take this approach because the component managing the filter
-chain would have to be aware of the last RBAC filte and inform it if performing
+chain would have to be aware of the last RBAC filter and inform it if performing
 the audit logging. In other words, this would require more engineering effort
 which does not make too much sense for such a particular case as audit logging.
 
 In practice, xDS users normally do not craft RBACs on their own but instead
 rely on the control plane APIs, such as Istio's Authorization Policy, to apply
 RBAC filters to the workloads. We hope that when these control plane APIs
-start to support audit logging, they will only have a single RBAC policy o
+start to support audit logging, they will only have a single RBAC policy or
 will handle the logging behavior as what we do in the gRPC authorization policy
 case.
 
--- a/A59-audit-logging.md
+++ b/A59-audit-logging.md
@@ -329,9 +329,10 @@ two important features that apply to all the languages.
 
 As is also documented in all the languages below, the logging function is
 executed by gRPC synchronously during the authorization process. Therefore,
-its implementation should not block the RPC. When needed, it should invoke any
-long running operations asynchronously so that the function itself returns
-promptly. For this reason, no error or status code will be returned from it.
+its implementation **must** not block the RPC. If a logger implementation needs
+to trigger any long-running work, it **must** do so asynchronously so that the
+function itself returns promptly. For this reason, no error or status code will
+be returned from it.
 
 The logger factory (aka builder/provider) needs to implement a config parsing
 function for parsing and validation logger configs, and a build function to
--- a/A59-audit-logging.md
+++ b/A59-audit-logging.md
@@ -331,7 +331,8 @@ namespace grpc {
 namespace experimental {
 
 // This class contains useful information to be consumed in an audit logging
-// event.
+// event. It does not own any of the information so users should make copies
+// when they need to use them outside the scope of this object.
 class AuditContext {
  public:
   grpc::string_ref rpc_method() const;
@@ -348,6 +349,9 @@ class AuditLogger {
   // This function will be invoked synchronously when applicable during the
   // RBAC-based authorization process. It does not return anything and thus will
   // not impact whether the RPC will be rejected or not.
+  // Implementers should ensure this method does not block the RPC. Specifically,
+  // time-consuming processes should make a copy of the audit context and be fired
+  // asynchronously such that this function itself can return immediately. 
   virtual void Log(const AuditContext& audit_context) = 0;
 };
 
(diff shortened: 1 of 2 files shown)


A59-audit-logging.md
```diff
@@ -0,0 +1,243 @@
...
+decision is made, so the status only reflects that but not the eventual
+one from the server. 
+
+|Fields             |Additional Information                                                             |Example                |
+|-------------------|-----------------------------------------------------------------------------------|-----------------------|
+|RPC Method         |Method the RPC hit.                                                                |"/pkg.service/foo"     |
+|Principal          |Identity the RPC. Currently only available in certificate-based TLS authentication.|"spiffe://foo/user1"   |
+|Timestamp          |Time when the RPC happens.                                                         |1649376211             |
+|Denying Policy Name|Policy name that denied the RPC. Empty if RPC is allowed.                          |"policy_name"          |
+|Status             |gRPC status from the authorization enforcement, not the eventual RPC status.       |OK or PERMISSION_DENIED|
```
markdroth, line 55: Should this be a status or just a bool indicating whether the request was allowed?  Unless we expect that individual authz rules are going to reject a request with a particular status code, then a bool might be fine.

A59-audit-logging.md
```diff
@@ -0,0 +1,243 @@
...
+
+It is worth noting that audit happens right after the authorization
+decision is made, so the status only reflects that but not the eventual
+one from the server. 
+
+|Fields             |Additional Information                                                             |Example                |
+|-------------------|-----------------------------------------------------------------------------------|-----------------------|
+|RPC Method         |Method the RPC hit.                                                                |"/pkg.service/foo"     |
+|Principal          |Identity the RPC. Currently only available in certificate-based TLS authentication.|"spiffe://foo/user1"   |
+|Timestamp          |Time when the RPC happens.                                                         |1649376211             |
```
markdroth, line 53: What does "happens" mean here?  The time at which the server first saw the RPC?  Or the time at which the authz policy made the decision about it?  Those two things will not happen at the same moment.

If it's the time that the authz policy made the decision, then do we actually need to make this available in the audit context?  It seems like the audit logger impl could just check the current time itself.

A59-audit-logging.md
```diff
@@ -0,0 +1,243 @@
...
+|Denying Policy Name|Policy name that denied the RPC. Empty if RPC is allowed.                          |"policy_name"          |
```
markdroth, line 54: Isn't it possible that we also have a particular policy name response for allowing a request, if we're using an RBAC filter with an ALLOW action?

Also, aren't there possible configurations where we'll wind up generating multiple log entries for each RPC, one for each RBAC filter?  Don't we need some way to know that those duplicate messages all refer to the same RPC?  In particular, note that there is no restriction that prevents two RBAC filters from using the same policy name.

A59-audit-logging.md
```diff
@@ -0,0 +1,243 @@
...
+|Status             |gRPC status from the authorization enforcement, not the eventual RPC status.       |OK or PERMISSION_DENIED|
+
+### xDS API Changes
+
+xDS API changes are needed because gRPC authorization is entirely based on [xDS RBAC policy][RBAC policy]. Moreoever, audit logging is a common feature
+that other xDS clients (namely Envoy) should benefit from as well.
+
+We will add an audit condition enum in the [xDS RBAC policy][RBAC policy] as
```
markdroth, line 62: You can add a link to the PR in the Envoy repo here.

A59-audit-logging.md
```diff
@@ -0,0 +1,243 @@
...
+    NONE = 0;
+    ON_DENY = 1;
+    ON_ALLOW = 2;
+    ALWAYS = 3;
+  }
+
+  AuditCondition audit_condition = 1;
+
+  // See https://github.com/cncf/xds/blob/main/xds/core/v3/extension.proto.
+  xds.core.v3.TypedExtensionConfig audit_log = 2;
```
markdroth, line 186: I don't think we should use the xDS `TypedExtensionConfig` message here.  Instead, I think we should do something like this:

```proto3
message AuditLogger {
  string name = 1;
  google.protobuf.Struct config = 2;  
}
AuditLogger audit_logger = 2;
```

markdroth (commented):

A59-audit-logging.md
rockspore: Done.

A59-audit-logging.md
rockspore: Good point! I changed it to `Authorized` which will be a boolean. A new field indicating the detailed status code could always be added if needed I think. Currently, IIRC, in Envoy RBAC always rejects with a 403.

A59-audit-logging.md
```diff
@@ -0,0 +1,243 @@
+A59: gRPC Audit Logging
+----
+* Author(s): [Luwei Ge](https://github.com/rockspore)
+* Approver:
```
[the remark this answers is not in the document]
rockspore: Done. Thanks!

A59-audit-logging.md
rockspore: Done. Note I used `AuditLog` and `audit_log` instead of `...logger` to be consistent with the xDS RBAC change which is to be consistent with `access_log` field.

A59-audit-logging.md
rockspore: I revised the description to be `when the server saw the RPC`. For audit logging, my understanding is we generally want to know when the RPC was attempted not when the authorization decision is made or to be made.

A59-audit-logging.md
rockspore: First of all, given the scenario you used as as example, I realized it's still necessary to add some more explanation about the `ON_DENY` and `ON_ALLOW`. Basically, I think it'd be quite intuitive to consider `ON_DENY` condition met when a RBAC rejects the RPC, even if the action of it is `ALLOW`. Likewise, I think we should also consider `ON_ALLOW` condition met when a RBAC lets the RPC through, even if the action could be `DENY`. If the overloading of `DENY` and `ALLOW` for action and decision here could cause confusion, we can think about using other words like `ON_AUTHORIZED` or so. WDYT?

About logging the policy name, I changed it to log the allowing policy name as well for now. But I think some discussion related to this is still needed:
1. Like you said, one caveat is when there could be multiple log entries. I believe we discussed it before where we assumed this was not what users would generally want. In the gRPC authz policy section, I specifically made this assumption and added more paragraphs to explain the configuration. PTAL.
2. In xDS, particularly where users can have arbitrary numbers of RBAC filters, they could audit the same RPC multiple times if they really want. Let me know if I am wrong but I did't find gRPC generating any sort of request UUID on the server side to identify a RPC. So if users really want to do this, they would need to think about how to dedup the audits if needed. I guess if feasible, we could include some RPC metadata which users could configure out a way with.
3. Now let's assume users only want to audit once even with multiple RBACs, so they would only configure `ON_ALLOW` at the last RBAC. The question is how meaningful it would be to log just the policy name of the last RBAC? I don't think we should simply accumulate policy names here for audit on allow.

I don't know where users would reasonably want to use the same policy name in multiple active RBAC filters, _and_ also want to audit log multiple times. I'd prefer we consider this as an anti-pattern and I don't see much we could do to resolve the confusion.

rockspore:
Thanks, @markdroth, for the review! I revised the doc and replied to the comments. PTAL.

A59-audit-logging.md
```diff
@@ -0,0 +1,281 @@
...
+The advantage of that apporoach is the control plane would not need to know
+the platform-specific configuration options in the data plane, if any.
+But we concluded that in audit logging, the logger configuration should
+generally be uniform across the data plane and thus centralizing the configuration
+makes more sense. Moreover, the existing access log follows such a pattern
+in xDS, albeit not currently supported by gRPC.
+
+## Implementation
+
+The implementation order will be C++, Go, Java and then wrapped languages.
```
erm-g, line 281: Would it make sense to change it to 
"C++, Go, and then Java and wrapped languages"?

A59-audit-logging.md
```diff
@@ -0,0 +1,281 @@
...
+use cases.
+
+### Language Specific APIs
+
+_This section is to be done once the language-agnostic APIs are more finalized._
+
+_What is certain for now is we need to support third-party audit logger
+implementations registered by users._
+
+## Rationale
```
erm-g, line 239: I really like how [https://github.com/grpc/proposal/blob/4187b46efa08eea782685efb91509ad77cc5e096/A43-grpc-authorization-api.md#rationale](same section in authorization) describes a high level flow from a user perspective. Do you think a similar section will add value here?

A59-audit-logging.md
```diff
@@ -0,0 +1,281 @@
...
+  // [#extension-category: envoy.audit_loggers]
+  repeated xds.core.v3.TypedExtensionConfig audit_log = 53;
+}
+```
+
+The list of configured audit loggers will all be invoked by an RBAC filter when
+the audit condition is met. This is analogus to the semantics that all access
+loggers will be run after the response is sent.
+
+### gRPC Authorzation Policy Changes
```
erm-g, line 120: minor typo 
< Authorzation

A59-audit-logging.md
```diff
@@ -0,0 +1,281 @@
...
+will never happen twice. 
+
+If users want to audit on allow, then the `DENY` RBAC will have no audit enabled
+and the `ALLOW` RBAC with `ON_ALLOW`. So audit logging will at most happen once
+when the RPC passes through both RBACs.
+
+If users want to audit on both cases, then the `DENY` RBAC needs `ON_DENY` and
+the `ALLOW` RBAC needs `ON_DENY_AND_ALLOW`.
+
+Following is the table summarzing the combinations.
```
erm-g, line 219: minor typo
`summarzing `

A59-audit-logging.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Proposal
+
+### Audit Context
+
+The table below lists the metadata we will include in the audit context,
+which will be made available to audit loggers during an audit event.
+
+It is worth noting that audit happens right after the authorization
+decision is made, so the status only reflects that but not the eventual
```
larry-safran, line 46: s/that but/that, but/

A59-audit-logging.md
```diff
@@ -0,0 +1,281 @@
...
+
+  // Condition for the audit logging to be happen, specific to HTTP filters.
+  // If condition is met, all the audit loggers configured in the HCM will be invoked.
+  //
+  AuditCondition audit_condition = 3;
+}
+```
+
+Note that `DENY` and `ALLOW` in the enum refer to the RBAC decisions, not
+RBAC actions. For example, we consider it as `DENY` decision when a RBAC
```
larry-safran, line 94: This is a little confusing.  I would recommend instead:
For example, we consider it to be a `DENY` decision when a RBAC rejects the RPC, regardless of whether it is because there is no policy match from a RBAC with the `ALLOW` action, or because there is a match for a RBAC with a `DENY` action.  Likewise, it is an `ALLOW` decision when there are no matches to RBACs with a `DENY` action and at least one match to an RBAC with an `ALLOW` action.

A59-audit-logging.md
rockspore: Good idea. Working on it.

A59-audit-logging.md
rockspore: Thank you, Larry! I took your suggested piece.

A59-audit-logging.md
rockspore: Added some paragraphs. PTAL.

A59-audit-logging.md
markdroth: I don't think we need to be consistent with xDS here, since this is a different API.  I think it makes more sense to call it `AuditLogger` and to call the field `audit_logger`.

A59-audit-logging.md
markdroth: Is it actually important to distinguish between the moment that the server first saw the RPC and the moment that the authz decision is made?  Note that the authz decision is made as soon as the server sees the initial metadata from the client, which is always the first operation on the call.  So there are only a small set of cases in which these two timestamps will be significantly different (e.g., if for some reason the transport reads the start of the initial HEADERS frame but then there's a long delay before it reads the rest of the HEADERS frame), and I'm not sure anyone will care about that from the perspective of the audit logs.

My inclination here is to just remove this field and let the logger use the current time for the log entry.

If we do include this field here, then I think we need to be very specific about what timestamp it indicates, because "when the server saw the RPC" is not very precise.  I think it would need to say something like "the time at which the transport reads the start of the stream" or "the time at which the call is created".  And I think the need for being this precise is actually a good reason not to include this here, since it will ultimately over-specify the design, which may unnecessarily constrain implementation choices.

A59-audit-logging.md
markdroth: 1. I don't think it will be a common config, but it is a possible config, so we need to consider the implications when that config is used.
2. Correct, there is no UUID for each RPC.  And I agree that that's an issue for the multiple-logging case.  I suppose we could try to generate some sort of UUID, although I'm not crazy about it.
3. I don't fully follow the case you're describing here.  In general, I think that if there is a particular policy name, it seems right to log it.  But I think there may be cases where they would not be a specific policy name to log: for example, if an RBAC policy has action `DENY` but has audit condition `ON_ALLOW`, then we will log a message if none of the policies match.

I'd like to get @ejona86's thoughts on this.

rockspore: In #3, what I meant was if there are two RBACs present, an authorized request would need to be allowed by both of them. Now if users only want one log entry on allow, which is the assumption we have in SDK, the audit condition `ON_ALLOW` will only be added to the second RBAC policy. It will not be accurate to only log the second RBAC policy name in this case though, since the first RBAC technically contributed to allowing the request as well.

A59-audit-logging.md
rockspore: Makes sense! Renamed the field.

A59-audit-logging.md
rockspore: Thanks, Mark, for the detailed explanation. As we also discussed offline, we should be fine by removing this timestamp field and have the logger to use its current time. Change has been made.

A59-audit-logging.md
rockspore: During our offline discussion, we agreed that it was a reasonable expectation to audit at most once per RPC and the gRPC authorization API (SDK) makes this assumption for the users. We also concluded that while the possibility of configuring multiple audit logs per RPC is acknowledged, we would not consider providing any mechanism to de-duplicate those log entries in this gRFC.

As suggested, I also added more paragraphs elaborating on this in the `xDS API changes` section + `Audit Condition in the HTTP Connection Manager` section in `Rationale`. PTAL. @markdroth @ejona86

larry-safran: I think that the policy name is useful in the DENY case, but questionable in the ALLOW case.  If you were outputting policy names for ALLOW, you would want the policy names of all ALLOW rules that had matched, which doesn't seem worth the complication of supporting.  If there is only 1 ALLOW rule, then the name doesn't need to be logged since it must always be that rule's policy.

A59-audit-logging.md
larry-safran: You could emphasize C++ & Go with "Implementation will first be done for C++ and Go.  Following that will be Java, and finally wrapped languages will be implemented."

A59-audit-logging.md
rockspore: To avoid confusion, the DENY and ALLOW cases in your comment above refer to the RBAC actions, not decisions. 

According to the definition in [RBAC](https://github.com/envoyproxy/envoy/blob/main/api/envoy/config/rbac/v3/rbac.proto), "A match occurs when at least one policy matches the request. The policies are evaluated in lexicographic order of the policy name.", so the behavior is actually deterministic regarding which policy (aka rule in SDK) gets matched. So I think it's reasonable to just log that policy name.

A59-audit-logging.md
rockspore: Thank you, Larry! Done.

A59-audit-logging.md
```diff
@@ -0,0 +1,349 @@
...
+    },
+  },
+  "required": ["name", "allow_rules"]
+}
+```
+
+For readers that are more familiar with proto messages, the change translates
+to:
+
+```
```
ejona86, line 188: FYI: You can use the \`\`\`proto (or "protobuf") here to get syntax highlighting. Ditto above.

A59-audit-logging.md
```diff
@@ -0,0 +1,349 @@
...
+
+  AuditLoggingOptions audit_logging_options = 4;
+}
+```
+
+Note that the definition above is only for illustration purposes and does not
+actually exist anywhere as a `.proto` file. Nor does gRPC process the policy
+as protobuf by any means.
+
+The authorization policy is backed by two RBAC filters, so the audit logger
```
ejona86, line 218: You say "RBAC with DENY" later, but don't really introduce it. Somewhere here, where we introduce "two RBAC filters" it may be good to say that one is DENY and one is ALLOW.

Maybe:
> The authorization policy is backed by two RBAC filters, a DENY followed by an ALLOW. The audit logger...

A59-audit-logging.md
```diff
@@ -0,0 +1,349 @@
...
+|ON_ALLOW              |NONE               |ON_ALLOW             |
+|ON_DENY_AND_ALLOW     |ON_DENY            |ON_DENY_AND_ALLOW    |
+
+Again this is a subset of what users can technically do with xDS RBACs in the way
+described in the earlier section. But we think this represents the most common
+use cases.
+
+### Built-in logger types
+
+We plan to implement the stdout logger as a built-in logger type. More types of
```
ejona86, line 253: What is "the stdout logger"? It needs to be defined. What is the type used in the Struct? What output format does it use? Does it have any configuration (e.g., to use stderr)?

A59-audit-logging.md
```diff
@@ -0,0 +1,349 @@
...
+As an HTTP filter chain can contain multiple RBAC filters in general, the typical
+behavior that users will expect if they want to audit on allow is audit should
+only happen once after the evaluation of the last RBAC filter. The gRPC
+authorization policy API assumes this expectation and handles the audit condition
+for the users.
+
+In the xDS cases, however, the audit condition could be considered as something
+on top of all RBAC filters and thus configured in the [HTTP Connection Manager][HttpConnectionManager proto].
+We decided not to take this approach because the component managing the filter
+chain would have to be aware of the last RBAC filter constantly and inform it
```
ejona86, line 317: I don't know how to parse "constantly" in this context. Should it just be removed?

A59-audit-logging.md
```diff
@@ -0,0 +1,349 @@
...
+if performing the audit logging. In other words, this would require more
+engineering effort which does not make too much sense for such a particular
+case as audit logging.
+
+In practice, xDS users normally do not craft RBACs on their own but instead
+rely on the control plane APIs, such as Istio's Authorization Policy, to apply
+RBAC filters to the workloads. We hope that when these control plane APIs
+start to support audit logging, they will handle the logging behavior as what
```
ejona86, line 325: "they will only have a single RBAC policy or will handle..." ?

ejona86 (commented):
Mostly editorial comments, but we do need to define the stdout audit behavior.

Looks good.

A59-audit-logging.md
rockspore: Done.

A59-audit-logging.md
rockspore: Removed.

A59-audit-logging.md
rockspore: Done. Thank you, Eric!

A59-audit-logging.md
```diff
@@ -232,12 +281,249 @@ Again this is a subset of what users can technically do with xDS RBACs in the wa
...
+Following is an example log entry.
 
-_This section is to be done once the language-agnostic APIs are more finalized._
+```json
+{"grpc_audit_log":{"timestamp":1649376211,"rpc_method":"/pkg.Service/Foo","principal":"spiffe://foo/user1","policy_name":"example_policy","matched_rule":"admin_access","authorized":true}}
+```
 
-_What is certain for now is we need to support third-party audit logger
-implementations registered by users._
+The timestamp, as the number of seconds from the Unix Epoch, is generated from
```
markdroth, line 315: Are we sure that seconds since the Unix epoch is the right time format to use here?  The output is JSON, so it's presumably meant to be a bit more user-friendly.  Maybe we should format this as a human-readable time string?  Or is that just going to open up problems due to (e.g.) locale-specific formatting?

rockspore: When timestamp was still a field in the audit context, unix seconds was given as an example. So I assumed it was preferred but I don't have any opinion at all. 

We could change to a different format, e.g., RFC3339, which seems to be recommended generally.

rockspore: cc @erm-g who's authoring the stdout logger in Go. We need to ensure the output is consistent across languages.

erm-g: @markdroth I don't have a strong opinion, but I think in most use cases the output will be fed to some aggregation pipeline for further analysis. So Unix epoch format is an ok default.

rockspore: I have checked with @ejona86 about using RFC3339 seconds and that seems fine. Aggregation pipelines should be able to process either format. 

I have changed the example here.

ejona86 (approved):
Only editorial notes.

@markdroth, were you happy with this gRFC? To my knowledge it has been implemented in C and Go, so apparently only editorial issues remain?

A59-audit-logging.md
```diff
@@ -0,0 +1,612 @@
...
+|ON_DENY_AND_ALLOW     |ON_DENY            |ON_DENY_AND_ALLOW    |
+
+Again this is a subset of what users can technically do with xDS RBACs in the way
+described in the earlier section. But we think this represents the most common
+use cases.
+
+### Built-in logger types
+
+We plan to implement the stdout logger as a built-in logger type. The type is
+named `stdout_logger`. This logger will not support any configuration and it
```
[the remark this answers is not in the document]
rockspore: Suggestion taken. Thanks.

A59-audit-logging.md
```diff
@@ -0,0 +1,612 @@
...
+As an HTTP filter chain can contain multiple RBAC filters in general, the typical
+behavior that users will expect if they want to audit on allow is audit should
+only happen once after the evaluation of the last RBAC filter. The gRPC
+authorization policy API assumes this expectation and handles the audit condition
+for the users.
+
+In the xDS cases, however, the audit condition could be considered as something
+on top of all RBAC filters and thus configured in the [HTTP Connection Manager][HttpConnectionManager proto].
+We decided not to take this approach because the component managing the filter
+chain would have to be aware of the last RBAC filte and inform it if performing
```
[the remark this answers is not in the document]
rockspore: Done.

rockspore (commented):
I missed the latest reviews till now. Fixed. Let me know if it's good to merge. Thanks! @ejona86 @markdroth

A59-audit-logging.md
```diff
@@ -0,0 +1,612 @@
...
+### Language Specific APIs for third-party logger implementation
+
+Some users may want to have their own audit logging logic which built-in
+loggers do not fullfil. This section is specifically about public APIs we
+will expose for users to implement their own types of audit loggers. Below are
+two important features that apply to all the languages.
+
+As is also documented in all the languages below, the logging function is
+executed by gRPC synchronously during the authorization process. Therefore,
+its implementation should not block the RPC. When needed, it should invoke any
```
markdroth, line 332: Please change "should not block" to "must not block".  This is not optional; it's a hard requirement.

A59-audit-logging.md
```diff
@@ -0,0 +1,612 @@
...
+long running operations asynchronously so that the function itself returns
```
markdroth, line 333: I suggest changing the wording of this sentence to make its meaning a bit clearer:

If a logger implementation needs to trigger any long-running work, it must do so asynchronously so that the function itself returns promptly.

markdroth (approved):
This looks great!  I have just a couple of minor wording changes that I'd like to see before merging.  Thanks!

A59-audit-logging.md
rockspore: Done.

A59-audit-logging.md
rockspore: Suggestion taken.

rockspore (commented):
Thanks for the review @markdroth!

PR #348, opened 2023-03-23
A60: xDS WeightedClusters support for Stateful Session Affinity

sanjaypujare:
CC @ejona86 @dfawley PTAL


markdroth:
A58 and A59 are already in flight in other PRs.  This will be A60.  Please update the filename and title accordingly.

ejona86:
@markdroth, do you think us other languages need to look at this much? We've seen things at a high level, and they seemed fine. This is focused on C++, but we know we can do the same in the other languages, albeit maybe with some tweaks based on what's easiest.

A60-xds-stateful-session-affinity-weighted-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+### Temporary environment variable protection
+
+[A55: xDS-Based Stateful Session Affinity for Proxyless gRPC][A55] was only
+recently implemented and the implementation is still guarded by the
+`GRPC_EXPERIMENTAL_XDS_ENABLE_OVERRIDE_HOST` environment variable. We will
+continue to use the same environment variable to also guard this feature.
+This environment variable protection will be removed once the feature has
+proven stable.
+
+These design proposes some structural or declarative changes in the source code
```
markdroth, line 222: The only things that we need to guard with the env var are things that interact differently on the wire.  For this design, I think the only thing we need to guard is the behavior in the stateful session filter to add the cluster name to the cookie and to read the cluster name override from the cookie.

I think we can safely make the structural changes to the ConfigSelector and introduce the cluster selection filter without needing any env var guard, because those code-paths will be executed on every RPC on a xDS channel, so we'll have plenty of indication from our existing tests if those changes break anything.

A60-xds-stateful-session-affinity-weighted-clusters.md
```diff
@@ -0,0 +1,270 @@
+A60: xDS-Based Stateful Session Affinity for Weighted Clusters
+----
+* Author(s): Sanjay Pujare (@sanjaypujare)
+* Approver: markdroth
+* Status: In Review
+* Implemented in: <language, ...>
+* Last updated: 2023-03-19
+* Discussion at:
```
markdroth, line 8: Please create the discussion thread and add a link to it here.  This starts the 10-day clock for merging the gRFC.

markdroth (commented):
This looks really good!

The only really substantive comments are the ones about the env var protection.

markdroth:
@ejona86 No, I don't think we need further review from you or @dfawley.  I think we've already discussed all the relevant cross-language details here.

A60-xds-stateful-session-affinity-weighted-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+selection takes place inside the config selector - before any of the http
+filters are run, including the [StatefulSessionFilter][]. After having selected
+`Cluster V2` gRPC cannot send the REQ2 to `v1s1` and session affinity is
+broken.
+
+This proposal addresses the limitation.
+
+### Related Proposals: 
+
+* [A55: xDS-Based Stateful Session Affinity for Proxyless gRPC][A55]
```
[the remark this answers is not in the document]
sanjaypujare: done

A60-xds-stateful-session-affinity-weighted-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Notes
+
+This design mainly deals with how two existing features - stateful session
+affinity and weighted clusters - should work together. As a result this feature
+does not require a new API or impact any existing APIs.
+
+### Temporary environment variable protection
+
+[A55: xDS-Based Stateful Session Affinity for Proxyless gRPC][A55] was only
```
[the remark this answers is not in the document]
sanjaypujare: I think the users (like @va7ish and others) mainly want to use stateful session affinity (SSA) with weighted clusters support and without that support they probably consider the feature as incomplete. Since the basic SSA feature is already there, users may do limited testing but will wait for  weighted clusters support before using it or doing full testing.

In any case let's wait for input from @va7ish .

A60-xds-stateful-session-affinity-weighted-clusters.md
sanjaypujare: done

A60-xds-stateful-session-affinity-weighted-clusters.md
sanjaypujare: I am okay with that. Will make that change

A60-xds-stateful-session-affinity-weighted-clusters.md
```diff
@@ -219,11 +220,10 @@ continue to use the same environment variable to also guard this feature.
 This environment variable protection will be removed once the feature has
 proven stable.
 
-These design proposes some structural or declarative changes in the source code
-which cannot be guarded by a runtime environmental variable so these will be
-present in the source code. These are:
+This design proposes some structural or declarative changes in the source code
```
markdroth, line 223: I don't think we need to list the things that are *not* guarded by the env var, only the things that *are*.  There should only be two bullets needed here.

A58-xds-stateful-session-affinity-weighted-clusters.md
```diff
@@ -0,0 +1,270 @@
...
+
+### New `ClusterSelectionFilter`
+
+Because of the changes we made to the config selector we need a new filter
+to finalize cluster selection and to release refs to the cluster map and
+route config. This new filter is called `ClusterSelectionFilter` and it runs
+just before the Router filter. This is an internal implementation detail of
+gRPC so the xDS config is not aware of this new filter. Note:
+
+ * This is the second-last filter (retry/router filter being last).
```
va7ish: I am not too familiar with the internals of this logic so pardon for potentially dumb question here. Why does it need to run just before the router filter? Wouldn't we expect the HTTPRoute logic to need to pick the cluster according to route configuration first and then only if we have an overwrite value from the session cookie (as part of StatefulSessionFilter) it would need to be modified by ClusterSelectionFilter.
Perhaps the notion here is to let StatefulSessionFilter provide the cluster value from the cookie (if such is available) and store it in the call attributes and only if not set then Route would pick the cluster according to weights configuration?

A60-xds-stateful-session-affinity-weighted-clusters.md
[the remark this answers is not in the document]
sanjaypujare: @va7ish could you provide your input here?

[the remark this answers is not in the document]
va7ish: Yes, I think we should consider SSA incomplete if the weighted traffic routing configuration breaks SSA.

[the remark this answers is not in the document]
sanjaypujare: So just to confirm and to answer @markdroth 's question in the first comment: it is unlikely that you will use the existing stateful session affinity support without the new weighted cluster logic that is being added. So a single env var to guard both sets of code should be sufficient according to you?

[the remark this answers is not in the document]
markdroth: Okay, super.  Then using the same env var guard sounds good.

A58-xds-stateful-session-affinity-weighted-clusters.md
markdroth: In the xDS data model, the workflow is as follows:
1. We choose the route and cluster based on the RouteConfig.
2. We iterate over the HTTP filters, executing each one in order.  Each filter can access the currently chosen route and cluster, so it can do things like get per-route config overrides, get which cluster the request will be sent to, or (in the case of the StatefulSession filter) override the cluster that the request will be sent to.
3. The last HTTP filter in the chain is the router filter, which is responsible for supporting retries and for sending the request to the chosen cluster.

In gRPC, unlike Envoy, the router filter doesn't actually exist as a distinct piece of code, because gRPC already has retry functionality built into the client channel, and we handle clusters in the LB policy tree.  So when we see the router filter in the xDS config, we just treat it as shorthand to tell us to use that existing functionality.

The new ClusterSelectionFilter being added here is basically just another step that will occur right before the retry code, so it will occur logically as the very first part of step 3.  Note that this filter is an internal implementation detail in gRPC, not something that will be exposed in the xDS API, and it will always be executed right before the retry code, so you can think of it as part of the router filter implemention in gRPC.  And despite its name, this filter is not actually selecting which cluster to send the request to; that decision is still being made in step 1, and the only thing this filter is doing is handling some ref-counting details that we need internally.

I hope that answers your question.  Please let me know if I haven't explained that in a clear way.

va7ish: Thank you @sanjaypujare and @markdroth

A60-xds-stateful-session-affinity-weighted-clusters.md
markdroth: I don't think we need the env var to guard the following items:

- setting the call attributes for the ref-counted cluster map and route-config
- config selector will set the new call-attributes for the cluster-map, current route entry and the route config
- the new `ClusterSelectionFilter` will be added to the filter registry and added as the second-last filter in the filter chain

Those things will all happen unconditionally, because that approach supports the current functionality just fine, so it will be covered by our existing tests.

The only behavior that we need to guard with the env var are the two remaining items:

- the `StatefulSessionFilter` will add the cluster name to the cookie in the response side processing
- the `StatefulSessionFilter` will look for the cluster name in the cookie and set the cluster call-attribute in the request side processing

sanjaypujare: done

markdroth (approved):
Looks great!

PR #349, opened 2023-03-30
Add details on how Core performs proxy determination

yashykt:
Also formats the doc with google's `mdformat`.

### DIFF
--- a/A1-http-connect-proxy-support.md
+++ b/A1-http-connect-proxy-support.md
@@ -224,9 +224,9 @@ If none of the above are set, then no HTTP proxy will be used.
 The allowed format is an RFC3986 URI string where the scheme is expected to be
 "http" and the authority portion is used to determine the proxy to be used. For
 example, for an HTTP proxy setting of
-"http://username:password@proxy.google.com:443", "username:password" would be
+`http://username:password@proxy.google.com:443`, `username:password` would be
 used as user credentials for proxy authentication as per RFC 7617 and
-"proxy.google.com:443" would be the host:port HTTP proxy target. If the port
+`proxy.google.com:443` would be the host:port HTTP proxy target. If the port
 part of the authority is omitted, a default port of 443 is used. Note that user
 credential can also be omitted if the proxy does not need authentication.
 
@@ -242,10 +242,10 @@ If none of the above are set, then the previously found HTTP proxy is used.
 The format takes a comma-separated list of names, and if any of these names
 matches as a suffix of the server host (provided in the channel target), then
 the proxy will not be used for that target. For example, with a `grpc_proxy`
-setting of "proxy.google.com" and a `no_grpc_proxy` setting of "example.com,
-google.com", channel targets such as "dns:///foo.google.com:50051" and
-"bar.example.com:1234" will not use the proxy, but "baz.googleapis.com:443"
-would still use the configured proxy "proxy.google.com".
+setting of `proxy.google.com` and a `no_grpc_proxy` setting of `example.com,
+google.com`, channel targets such as `dns:///foo.google.com:50051` and
+`bar.example.com:1234` will not use the proxy, but `baz.googleapis.com:443`
+would still use the configured proxy `proxy.google.com`.
 
 The lookup and subsequent usage of an HTTP proxy for a specific channel can also
 be disabled by setting the channel arg `GRPC_ARG_ENABLE_HTTP_PROXY` to 0.
--- a/A1-http-connect-proxy-support.md
+++ b/A1-http-connect-proxy-support.md
@@ -199,6 +199,8 @@ would yield priviledged access on all internal servers).
 
 ## Implementation
 
+### C-Core
+
 In C-core, the HTTP CONNECT handshaker was already implemented. The following
 additional changes have also been made:
 
@@ -210,19 +212,31 @@ additional changes have also been made:
     hook for name rewriting)
 
 C-Core checks the following places to determine the HTTP proxy to use, stopping
-at the first one that is set: 1. `GRPC_ARG_HTTP_PROXY` channel arg 2.
-`grpc_proxy environment` variable 3. `https_proxy` environment variable 4.
-`http_proxy` environment variable If none of the above are set, then no HTTP
-proxy will be used.
+at the first one that is set:
 
-C-Core then checks the follows places to exclude traffic destined to listed
-hosts from going through the proxy determined above, again stopping at the first
-one that is set: 1. `no_grpc_proxy` environment variable 2. `no_proxy`
-environment variable
+1.  `GRPC_ARG_HTTP_PROXY` channel arg
+2.  `grpc_proxy environment` variable
+3.  `https_proxy` environment variable
+4.  `http_proxy` environment variable
+
+If none of the above are set, then no HTTP proxy will be used.
+
+If an HTTP proxy to be used is found, C-Core then checks the follows places to
+exclude traffic destined to listed hosts from going through the proxy determined
+above, again stopping at the first one that is set:
+
+1.  `no_grpc_proxy` environment variable
+2.  `no_proxy`environment variable
+
+If none of the above are set, then the previously found HTTP proxy is used.
+
+### Go
 
 In Go, this functionality is being provided via a custom dialer:
 
 -   [grpc/grpc-go#1098](https://github.com/grpc/grpc-go/pull/1098)
+
+### Java
 
 In Java for case 1, Java's `java.net.ProxySelector` is observed and usable
 starting in v1.11.0. `ProxySelector` is generally configured with the Java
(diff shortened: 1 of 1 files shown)


A1-http-connect-proxy-support.md
```diff
@@ -1,222 +1,242 @@
...
-* Discussion at: https://groups.google.com/d/topic/grpc-io/zG7iYpDw6lk/discussion
+## HTTP CONNECT Proxy Support
+
+*   Author(s): Mark D. Roth (roth@google.com), Eric Anderson (ejona@google.com)
+*   Approver: a11r
+*   Status: Implemented
+*   Implemented in: C-core (Java and Go are in progress)
+*   Last updated: 2017-03-20
+*   Discussion at:
+    https://groups.google.com/d/topic/grpc-io/zG7iYpDw6lk/discussion
```
sanjaypujare, line 9: These 8/9 lines are just source formatting change and does not fix anything AFAICS. Is that caused by the tooling?

A1-http-connect-proxy-support.md
```diff
@@ -1,222 +1,242 @@
...
+-   [grpc/grpc#9372](https://github.com/grpc/grpc/pull/9372) (adds proxy mapper
+    hook for address rewriting)
+-   [grpc/grpc#9557](https://github.com/grpc/grpc/pull/9557) (adds proxy mapper
+    hook for name rewriting)
+
+C-Core checks the following places to determine the HTTP proxy to use, stopping
+at the first one that is set:
+
+1.  `GRPC_ARG_HTTP_PROXY` channel arg
+2.  `grpc_proxy environment` variable
```
sanjaypujare, line 218: `grpc_proxy` environment variable 

(BTW: is it really lower case or upper case `GRPC_PROXY` ? Do we not read `GRPC_PROXY`? )

A1-http-connect-proxy-support.md
```diff
@@ -1,222 +1,242 @@
...
+3.  `https_proxy` environment variable
+4.  `http_proxy` environment variable
+
+If none of the above are set, then no HTTP proxy will be used.
+
+If an HTTP proxy to be used is found, C-Core then checks the follows places to
```
sanjaypujare, line 224: s/the follows places/the following places/

sanjaypujare: In "If an HTTP proxy to be used is found" what is "found"? Or are you just saying "if an HTTP proxy is to be used as dictated by the channel args or environment. variables"?

A1-http-connect-proxy-support.md
```diff
@@ -1,222 +1,242 @@
...
+exclude traffic destined to listed hosts from going through the proxy determined
+above, again stopping at the first one that is set:
+
+1.  `no_grpc_proxy` environment variable
+2.  `no_proxy`environment variable
```
sanjaypujare, line 229: Can we document how the hosts are listed in these environment variables? Comma separated list? IP address only? Names? Can names and IP addresses be mixed? How are names resolved to IP addresses (or they are not?) Where is the matching of listed hosts to actual gRPC connection attempt being done? If the `no_proxy` list includes a name `foobar` which resolves to `10.1.2.3` and for an xDS channel if there is an endpoint `10.1.2.3` will it be properly excluded? The other way? e.g. if the exclude list contains IP address `10.1.2.3` and the caller uses name `foobar` to connect and `foobar` resolves to multiple addresses (through DNS) one of which happens to be `10.1.2.3` will it only exclude the subchannel to that IP address?

A1-http-connect-proxy-support.md
yashykt: Yes

A1-http-connect-proxy-support.md
yashykt: Updated

A1-http-connect-proxy-support.md
yashykt: Thanks! Yeah, we only read the lower case environment variables

A1-http-connect-proxy-support.md
yashykt: Added details on how this works. https://github.com/grpc/grpc/pull/31119 improves the working of `no_proxy` but that is not part of a release yet.

A1-http-connect-proxy-support.md
```diff
@@ -1,222 +1,265 @@
...
+
+If none of the above are set, then the previously found HTTP proxy is used.
+
+The format takes a comma-separated list of names, and if any of these names
+matches as a suffix of the server host (provided in the channel target), then
+the proxy will not be used for that target. For example, with a `grpc_proxy`
+setting of `proxy.google.com` and a `no_grpc_proxy` setting of `example.com,
+google.com`, channel targets such as `dns:///foo.google.com:50051` and
+`bar.example.com:1234` will not use the proxy, but `baz.googleapis.com:443`
```
sanjaypujare, line 250: Thanks for the additional description and the example. Is it possible to say something about IP addresses? For example, if the `no_proxy` list contains IP addresses will that work in any way? And the example is using `dns:///` scheme but if one were to use `xds:///` scheme how will this feature behave or this feature just cannot be used? This is mainly for my information but if we have that info it might be useful to document here as well.

yashykt: So it all depends on what host:port was mentioned when creating the channel. We don't do anything special for `xds` and still use the path mentioned in the channel target. The only way IP addresses would be useful is if the channel target specified an IP address. Let me know if I should some more examples.

Note that I'm not saying that this is how it *should* work. I'm just documenting how it works right now. We could very well change the behavior for `xds` targets.

sanjaypujare: Correct. We just want to describe how it currently works - nothing more. Example or not, it might be worth mentioning that if the user includes an IP address in the `no_proxy` list then the only way it will be excluded is when used in the channel creation target string (and not if the target string resolves to that IP address).

A1-http-connect-proxy-support.md
```diff
@@ -1,222 +1,265 @@
...
-  create a single subchannel to the proxy, as opposed to one
-  subchannel for each backend server.)
-- Add the state described above triggering the use of the HTTP CONNECT
-  handshaker, with the argument set to the server name requested via
-  the client API.
+Case 1 will be addressed using the proxy mapper hook for overriding the server
+name to resolve. Sites that may want to send only external servers through the
+proxy can implement their own proxy mapper to do this. We will provide a default
+proxy mapper implementation that looks for the existence of the `http_proxy`
+environment variable (or equavalent signal). If that signal is present, it will
```
sanjaypujare, line 112: s/equavalent/equivalent/

A1-http-connect-proxy-support.md
```diff
@@ -1,222 +1,265 @@
...
+do the following: - Ask the resolver to resolve the proxy name instead of the
```
sanjaypujare, line 113: bulleted list formatting is lost in this version. Can you restore it so each line starting with `-` appears as a separate bullet item?

A1-http-connect-proxy-support.md
```diff
@@ -1,222 +1,265 @@
...
-addresses, it is impossible to use the normal gRPC client-side per-call
-load balancing.  It *is* possible to do load balancing on a per-connection
-basis, but that may not spread out the load evenly if different clients
-impose a different amount of load.  (Note that even if we established
-multiple connections to the proxy, we would have no guarantee that each
-connection would go to a different backend server.)
+addresses, it is impossible to use the normal gRPC client-side per-call load
+balancing. It *is* possible to do load balancing on a per-connection basis, but
+that may not spread out the load evenly if different clients impose a different
+amount of load. (Note that even if we established multiple connections to the
```
sanjaypujare, line 124: The wording "...even if we established multiple connections to the..." implies that we currently don't do this which is probably not the intent. Basically with the current implementation (as described in the previous lines 122-124) the gRPC client will create a separate connection to the same proxy when asked to create a channel to the target.

It is important to clarify this point for the use-case we have been discussing and for which this PR has been created to explain the functioning.

markdroth:
Rather than modifying a gRFC this old, I think we should just create a new doc in the grpc/grpc repo.

yashykt:
Alright, moving the new added details to https://github.com/grpc/grpc/pull/32789

A1-http-connect-proxy-support.md
yashykt: Addressed this in https://github.com/grpc/grpc/pull/32789

A1-http-connect-proxy-support.md
yashykt: Not proceeding with changes to this doc anymore

A1-http-connect-proxy-support.md
yashykt: Not proceeding with changes to this doc anymore

A1-http-connect-proxy-support.md
yashykt: Not proceeding with changes to this doc anymore

PR #353, opened 2023-04-10
eps and named metrics

yousukseung:


A51-custom-backend-metrics.md
```diff
@@ -349,6 +368,12 @@ public class MetricRecorder {
 
   // Clear the queries-per-second metrics data.
   public void clearQpsMetric();
+
+  // Update the errors-per-second metrics data.
+  public void putEpsMetric(double value);
```
ejona86, line 373: This should be `setEpsMetric(double)`. We use set single entries and put for mapped entries. That's why only `putUtilizationMetric(key, value)` uses put in its name.

ejona86 (commented):
Once small change, and then I think we can merge.

A51-custom-backend-metrics.md
yousukseung: Forgot to update this here as well, fixed.

PR #358, opened 2023-04-26
A58 update: Allow cpu_utilization larger than 1.0

yousukseung:


yousukseung:
@markdroth @dfawley FYI

ejona86 (approved):
We'll delay merging a bit to let https://github.com/cncf/xds/pull/60 go in.

For posterity: we noticed that if the CPU saturates at 100% the WRR weights become essentially "qps". Then you get a runaway effect where the backend with highest qps gets more and more qps sent to it. If the system allows the CPU to exceed 100%, it is better to just communicate that to keep the qps:cpu ratio consistent.

PR #359, opened 2023-05-03
A63: xDS StringMatcher in Header Matching

markdroth:


ejona86 (commented):
I see no issues. I'll approve after giving time for community feedback.

ejona86 (approved):

PR #372, opened 2023-05-24
A65: mTLS Credentials in xDS Bootstrap File

markdroth:
Addresses https://github.com/grpc/grpc/issues/32977.

### DIFF
--- a/A65-xds-mtls-creds-in-bootstrap.md
+++ b/A65-xds-mtls-creds-in-bootstrap.md
@@ -42,20 +42,20 @@ config will look like this:
 
 ```json
 {
-  // Path to identity certificate file.
+  // Path to CA certificate file.
+  // If unset, system-wide root certs are used.
+  "ca_certificate_file": <string>,
+
+  // Paths to identity certificate file and private key file.
+  // If either of these fields are set, both must be set.
   // If set, mTLS will be used; if unset, normal TLS will be used.
   "certificate_file": <string>,
-
-  // Paths to private key and CA certificate files.
-  // If either of these fields are set, both must be set.
-  // If neither is set, system-wide root certs are used.
   "private_key_file": <string>,
-  "ca_certificate_file": <string>,
 
   // How often to re-read the certificate files.
   // Value is the JSON format described for a google.protobuf.Duration
   // message in https://protobuf.dev/programming-guides/proto3/#json.
-  // If unset, defaults to 10 minutes.
+  // If unset, defaults to "600s".
   "refresh_interval": <string>
 }
 ```
@@ -72,7 +72,10 @@ functionality.
 
 ### Temporary environment variable protection
 
-This feature is not enabled via I/O, so no env var protection is needed.
+This feature is not enabled via remote I/O, and we don't have a good way
+to interop test it, so we will not use env var protection for this feature.
+Unit tests in individual languages should be sufficient to verify the
+functionality.
 
 ## Rationale
 
--- a/A65-xds-mtls-creds-in-bootstrap.md
+++ b/A65-xds-mtls-creds-in-bootstrap.md
@@ -1,0 +1,91 @@
+A65: mTLS Credentials in xDS Bootstrap File
+----
+* Author(s): @markdroth
+* Approver: @ejona86
+* Status: {Draft, In Review, Ready for Implementation, Implemented}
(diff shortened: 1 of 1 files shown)


A65-xds-mtls-creds-in-bootstrap.md
```diff
@@ -0,0 +1,97 @@
...
+config will look like this:
+
+```json
+{
+  // Path to identity certificate file.
+  // If set, mTLS will be used; if unset, normal TLS will be used.
+  "certificate_file": <string>,
+
+  // Paths to private key and CA certificate files.
+  // If either of these fields are set, both must be set.
```
ejona86, line 50: These shouldn't be paired together. `certificate_file` is the public certificate chain for the private key `private_key_file`. Those two should be paired together. `ca_certificate_file` is the trust root for the remote's certificate.

A65-xds-mtls-creds-in-bootstrap.md
```diff
@@ -0,0 +1,97 @@
...
+  // If neither is set, system-wide root certs are used.
+  "private_key_file": <string>,
+  "ca_certificate_file": <string>,
+
+  // How often to re-read the certificate files.
+  // Value is the JSON format described for a google.protobuf.Duration
+  // message in https://protobuf.dev/programming-guides/proto3/#json.
+  // If unset, defaults to 10 minutes.
```
ejona86, line 58: Specify `"600s"` (maybe also still saying 10 minutes), since that's the form it will be in?

A65-xds-mtls-creds-in-bootstrap.md
```diff
@@ -0,0 +1,97 @@
...
+specified, whereas in this configuration, it is acceptable to specify
+neither one.
+
+Implementations should be able to internally configure the use of the
+file-watcher certificate provider for the certificate-reloading
+functionality.
+
+### Temporary environment variable protection
+
+This feature is not enabled via I/O, so no env var protection is needed.
```
ejona86, line 75: It is enabled via I/O, because we read the bootstrap file, and newer bootstrap generators may want to support older clients.

I'd have to dig in to see if that distinction matters in this case though. I know we don't support a list of credentials here, so old clients would appear to fail anyway. There could be complications if we supported multiple servers in the list, but we don't do that either. So I could agree it isn't needed here (especially since this isn't a difficult feature and we're probably not going to have interop tests for it), but the reasoning would need to change.

A65-xds-mtls-creds-in-bootstrap.md
markdroth: Doh!  Yeah, I somehow got these totally conflated in my head when I wrote this.  Thanks for catching this.

Fixed.

A65-xds-mtls-creds-in-bootstrap.md
markdroth: Done.

A65-xds-mtls-creds-in-bootstrap.md
markdroth: Good point -- what I really meant here was that it's not enabled via *remote* I/O.  I think the local file isn't necessarily a problem for everyone, because even though we use a centrally maintained bootstrap generator for TD, other environments might just distribute the bootstrap config with the clients.  But I do take your point that reading a file is still I/O and can be an issue in some environments.

FWIW, note that we *do* actually support a list of credentials for each server, and the client will stop at the first one that it supports, so there is a mechanism here for handling older clients.  But that doesn't help if there's a bug in parsing the config or in actually configuring the credentials, which is the sort of thing that we'd normally use an env var guard for.

You're also right that we don't yet support multiple servers, but I don't think that would actually cause any significant complications here, because there's a separate list of creds for each server.

Anyway, I've updated the language here.

ejona86 (approved):

A65-xds-mtls-creds-in-bootstrap.md
ejona86: Oh, we do support a list; good. I couldn't understand how we had such an oversight. Good to see we didn't.

For a feature like this, it is useful for an implementation to fully-implement the feature and be tested before registering it as an possible option in the bootstrap. That way all the keys here are supported, or not. (Modulo Java which has tls but none of these.) If it is merged as a single PR, there's no problem, but if the work is split up...

PR #373, opened 2023-06-05
A58 update: Use application_utilization instead of cpu_utilization

yousukseung:


dfawley:
"Instead of"?  Don't we want to prefer `application_utilization` and then fall back to `cpu_utilization` so that there is some kind of migration strategy?  OK, it seems there's one sentence in there that explains this, but maybe it should be more prominent and part of the pseudo-code?

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -125,20 +125,20 @@ In all languages, when a subchannel state is updated to `IDLE`, the LB policy im
 
 ### Subchannel Weights
 
-To enforce `blackout_period` and `weight_expiration_period`, the LB policy tracks two timestamps along with the weight. `last_updated` tracks the last time a report with non-zero `cpu_utilization` and `qps` was received. `non_empty_since` tracks when the first non-zero load report was received, and it is reset when a subchannel comes back to `READY` or the weight expires so that it can be updated with the next non-zero load report. The blackout and expiration are enforced when the weight value is looked up.
+To enforce `blackout_period` and `weight_expiration_period`, the LB policy tracks two timestamps along with the weight. `last_updated` tracks the last time a report with non-zero `application_utilization` and `qps` was received. `non_empty_since` tracks when the first non-zero load report was received, and it is reset when a subchannel comes back to `READY` or the weight expires so that it can be updated with the next non-zero load report. The blackout and expiration are enforced when the weight value is looked up.
```
markdroth, line 128: Even if the `application_utilization` field is zero, if the `cpu_utilization` field is non-zero, then it still updates `last_updated`.  So we need to refer to both fields here.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -125,20 +125,20 @@ In all languages, when a subchannel state is updated to `IDLE`, the LB policy im
...
 
-The weight of a backend is calculated using `qps` (queries per second), `eps` (errors per second) and `cpu_utilization` reported by the backend. `eps` is only used when `error_utilization_penalty` is set to non-zero and applies a penalty based on the error rate calculated as `eps/qps`. The formula used is as follows:
+The weight of a backend is calculated using `qps` (queries per second), `eps` (errors per second) and `application_utilization` reported by the backend. `application_utilization` is a metric defined by the server application. This value is expected to range between [0.0, 1.0] normally but may exceed 1.0 when the usage exceeds the reporter dependent soft limits. The policy defaults to `cpu_utilization` when `application_utilization` is not defined in the load report. `eps` is only used when `error_utilization_penalty` is set to non-zero and applies a penalty based on the error rate calculated as `eps/qps`. The formula used is as follows:
```
markdroth, line 130: Suggest writing this as:

"""
The weight of a backend is calculated using utilization, QPS, and EPS reported by the backend.  The utilization value will come from the `application_utilization` field, if set; otherwise, the `cpu_utilization` field will be used instead.  The QPS and EPS values come from the `qps` and `eps` fields, respectively.  The `eps` is used only when `error_utilization_penalty` is set to non-zero and applies a penalty based on the error rate calculated as `eps/qps`. The formula used is as follows:
"""

I don't think we need to describe what `application_utilization` is; instead, I think that can be described in the ORCA proto, as you're doing in https://github.com/cncf/xds/pull/62.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -125,20 +125,20 @@ In all languages, when a subchannel state is updated to `IDLE`, the LB policy im
...
 
-$$weight = \dfrac{qps}{cpu\\_utilization + \dfrac{eps}{qps} * error\\_utilization\\_penalty}$$
+$$weight = \dfrac{qps}{application\\_utilization + \dfrac{eps}{qps} * error\\_utilization\\_penalty}$$
```
markdroth, line 132: I think this formula should just say "utilization" instead of referring specifically to `application_utilization`.

A58-client-side-weighted-round-robin-lb-policy.md
```diff
@@ -125,20 +125,20 @@ In all languages, when a subchannel state is updated to `IDLE`, the LB policy im
...
 
 In Java and Go, the weight will be stored in a wrapped subchannel. However, in C++, that approach won't work, because we create a new subchannel object for each address whenever we get a new address list, and we don't want to lose existing weight information when that happens. So instead, we store the weights in a map keyed by address, and each subchannel list will take its own ref to the entry in the map for each subchannel in the list.
 
 ```
-function UpdateWeight(qps, eps, cpu_utilization) {
-  if (cpu_utilization > 0 && qps > 0)
-    cpu_utilization += eps / qps * config.error_utilization_penalty;
+function UpdateWeight(qps, eps, application_utilization) {
```
markdroth, line 137: Let's make the behavior explicit by passing in both `application_utilization` and `cpu_utilization`, and then handling both cases:

```
function UpdateWeight(qps, eps, application_utilization, cpu_utilization) {
  var utilization =
      application_utilization > 0 ? application_utilization : cpu_utilization;
  if (utilization > 0 && qps > 0) {
    utilization += eps / qps * config.error_utilization_penalty;
  }
  var new_weight = utilization == 0 ? 0 : qps / utilization;
  if (new_weight == 0) return;
  if (non_empty_since_ == infy) non_empty_since_ = now();
  last_updated_ = now();
  weight_ = new_weight;
}
```

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: I moved what you suggested below to here to explain how the utilization (without quotes) is loaded once first, then updated to use the term utilization everywhere. I think this is simpler to follow later in text and code, but let me know what you think.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Thanks, updated (but moved up, please see my comment above).

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: Updated.

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: (please see my comment above) I instead explained how we read the utilization (without quotes) once first and use the term utilization without quotes

A58-client-side-weighted-round-robin-lb-policy.md
yousukseung: +@dfawley  as well, this is the only section we explain on load metrics and we explain how the utilization is acquired first now, PTAL.

PR #378, opened 2023-07-18
A28 update: fake content-type header and exclude -bin headers

markdroth:


A28-xds-traffic-splitting-and-routing.md
```diff
@@ -254,8 +254,14 @@ tricky.
 Note that only [custom
 metadata](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests)
 is considered during header matching, unlike in Envoy, where all headers
-(including HTTP/2 pseudo-headers) are included. This is because pick happens
-above gRPC’s transport (HTTP/2) layer, so transport headers are not accessible.
+(including HTTP/2 pseudo-headers) are included. This is because the pick
+happens above gRPC’s transport (HTTP/2) layer, so transport headers are
+not accessible. However, we will support matching against the
+`Content-type` header; if a gRPC implementation does not add that header
```
murgatroid99, line 260:
```suggestion
`content-type` header; if a gRPC implementation does not add that header
```

Header names are all lower-case

dfawley (approved):

ejona86 (commented):
LGTM

ejona86:
This was something that we discovered Java/C/Go were aligned on, but we didn't find it documented. It was clearly agreed but then we failed to document it.

markdroth:
For the record, the behavior in Node is consistent as well.

PR #380, opened 2023-07-21
A66: OpenTelemetry Metrics

yashykt:


A66-otel-stats.md
```diff
@@ -0,0 +1,305 @@
...
+*   Count : 0, 1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192,
+    16384, 32768, 65536
+
+### Attributes
+
+*   `grpc.method` : Full gRPC method name, including package, service and
+    method, e.g. "google.bigtable.v2.Bigtable/CheckAndMutateRow"
+*   `grpc.status` : gRPC server status code received, e.g. "OK", "CANCELLED",
+    "DEADLINE_EXCEEDED"
+*   `grpc.target` : Target URI used when creating gRPC Channel, e.g.
```
sanjaypujare, line 57: So we repeat this on all the RPCs on that channel so there is a lot of redundancy, is it not?

A66-otel-stats.md
```diff
@@ -0,0 +1,305 @@
...
+    "dns:///pubsub.googleapis.com:443", "xds:///helloworld-gke:8000"
+*   `grpc.authority` : Authority used by the call/attempt, e.g.
+    "pubsub.googleapis.com", "helloworld-gke"
```
sanjaypujare, line 60: In most cases the `grpc.authority` is going to be the same as the authority part of `grpc.target` (unless override host is used on a call) so there is another redundancy here. 

Also if your examples are supposed to match the previous example for `grpc.target` wouldn't the correct values be `pubsub.googleapis.com:443` and `helloworld-gke:8000` i.e. include the port?

A66-otel-stats.md
```diff
@@ -0,0 +1,305 @@
...
+    End timestamp - Ends at the first point where the transport considers the stream done. For HTTP2, this would be when scheduling a trailing header with END_STREAM to be written, or RST_STREAM, or a connection abort. Note that this wouldn’t necessarily mean that the bytes have also been immediately scheduled to be written by TCP. <br>
+    *Attributes*: grpc.method, grpc.authority, grpc.status <br>
+    *Type*: Histogram (Latency Buckets) <br>
+
+## Language Specifics
+
+Each language implementation will provide an API for registering an
+OpenTelemetry plugin. Overall, the APIs should have the following capabilities -
+
+*   Allow installing multiple OpenTelemetry plugins.
```
sanjaypujare, line 119: Is there a strong use-case for that? One can have a single plugin and multiple exporters to address most of these use-cases?

A66-otel-stats.md
```diff
@@ -0,0 +1,305 @@
...
+
+*   OpenTelemetry instrument names don’t allow ‘/’ so we use ‘.’ as the
+    separator. We also get rid of the “.io” suffix in “grpc.io” as it doesn’t
+    seem to add any value and is consistent with other names in the metrics spec
+    from OpenTelemetry.
+*   We also use this opportunity to resolve ambiguities from the gRPC OpenCensus
+    spec (detailed below).
+*   OpenTelemetry has attributes similar to tags in OpenCensus, and the
+    OpenCensus tag names already seem to match the OpenTelemetry spec - except
+    for ‘_’ vs ‘.’ for namespaces. So we just replace the ‘_’ with ‘.’. Note
```
murgatroid99, line 189:
```suggestion
    for ‘\_’ vs ‘.’ for namespaces. So we just replace the ‘\_’ with ‘.’. Note
```

Underscores are markdown formatting characters so they need to be escaped when used like this.

A66-otel-stats.md
```diff
@@ -0,0 +1,305 @@
...
+  // Enables base set of metrics by default
+  OpenTelemetryPluginBuilder() = default;
+  // If `SetMeterProvider` is called, \a meter_provider will be used instead of the global default.
+  OpenTelemetryPluginBuilder& SetMeterProvider(shared_ptr<opentelemetry::metrics::MeterProvider> meter_provider);
+  // Enable metric \a metric_name.
+  OpenTelemetryPluginBuilder&  EnableMetric(absl::string_view metric_name);
+  // Disable metric \a metric_name
+  OpenTelemetryPluginBuilder&  DisableMetric(absl::string_view metric_name);
+  // Builds and registers a OpenTelemetry Plugin
+  void BuildAndRegisterGlobal();
```
sanjaypujare, line 150: Since this is a builder pattern why not have a `build` return an instance and then a separate `register`? Also should e provider a "shutdown" hook for a plugin?

A66-otel-stats.md
```diff
@@ -0,0 +1,305 @@
...
+    *Attributes*: grpc.method, grpc.authority, grpc.status <br>
+    *Type*: Histogram (Size Buckets) <br>
+*   **grpc.server.call.duration** <br>
+    This metric aims to measure the end2end time an RPC takes from the server transport’s (HTTP2/ inproc / cronet) perspective. <br>
+    Start timestamp - After the transport knows that it's got a new stream. For HTTP2, this would be after the first header frame for the stream has been received and decoded. Whether the timestamp is recorded before or after HPACK is left to the implementation. <br>
+    End timestamp - Ends at the first point where the transport considers the stream done. For HTTP2, this would be when scheduling a trailing header with END_STREAM to be written, or RST_STREAM, or a connection abort. Note that this wouldn’t necessarily mean that the bytes have also been immediately scheduled to be written by TCP. <br>
+    *Attributes*: grpc.method, grpc.authority, grpc.status <br>
+    *Type*: Histogram (Latency Buckets) <br>
+
+## Language Specifics
```
markdroth, line 114: I think we need a subsection to describe the overall architecture of the OT plugin in a language-agnostic way, more than just stating the required capabilities.  This needs to describe things like the following:
- The CallTracer approach: An object is created for a call.  The object's methods will be invoked when various events happen over the lifetime of the call.  On the client side, the CallTracer knows about multiple attempts on the same call.  Needs to support more than one CallTracer per call.
- The OT plugin will basically be a way of configuring a CallTracer factory on gRPC clients and servers.  On the client side, the CallTracer will be created by an interceptor; on the server side, it will be a ServerCallTracerFactory.

We do want to describe this generically enough that it applies to all languages (and please feel free to tweak the langauge above as needed to accomplish that).  However, the goal here should be to be specific enough that we can hand this design to a junior engineer to implement and get the right result.

A66-otel-stats.md
```diff
@@ -0,0 +1,305 @@
...
+  OpenTelemetryPluginBuilder&  EnableMetric(absl::string_view metric_name);
+  // Disable metric \a metric_name
+  OpenTelemetryPluginBuilder&  DisableMetric(absl::string_view metric_name);
+  // Builds and registers a OpenTelemetry Plugin
+  void BuildAndRegisterGlobal();
+};
+
+```
+
+In the future, additional API might be provided to allow registering the plugin
```
markdroth, line 155: Why not support this right away?  Is it just because we don't yet have a story for the C++ interceptor API?

A66-otel-stats.md
markdroth, line 114: I think that the "Language Specifics" and "Migration From OpenCensus" sections should be nested underneath the "Proposal" section rather than being their own top-level headings.  I suggest the following overall structure for this section:

- Proposal
  - Metric Schema
    - Units
    - Attributes
    - Client Per-Attempt Instruments
    - Client Per-Call Instruments
    - Server Instruments
  - OT Plugin Architecture
  - Language-Specific Details
    - C++
    - Java
    - Go
    - Python
  - Migration From OpenCensus
    - Metric Schema Comparison
      - Metric Schema Differences Between OpenCensus and OpenTelemetry
      - Metric With Corresponding Equivalent
      - Metrics With Nuanced Differences
      - OpenCensus Metrics Not Initially Supported In OpenTelemetry
    - Migration Strategies
      - Migrate on a Per-Client Basis
      - Duplicate Metrics During Migration

markdroth (commented):

A66-otel-stats.md
```diff
@@ -0,0 +1,325 @@
...
+
+The following OpenCensus metrics have an equivalent in the OpenTelemetry spec
+(with the above noted differences) allowing for receivers of the telemetry data
+to join the views from the two metrics for continuity.
+
+gRPC OpenCensus                  | gRPC OpenTelemetry
+-------------------------------- | ---------------------------------------------
+grpc.io/client/started_rpcs      | grpc.client.attempt.started
+grpc.io/client/completed_rpcs    | (Derivable from grpc.client.attempt.duration)
+grpc.io/client/roundtrip_latency | grpc.client.attempt.duration
```
joybestourous, line 230: is there a particular reason for this naming convention you're using? it seems like if client metrics are always `grpc.client.attempt.*` and server metrics are always `grpc.server.call.*`, that dimension (`attempt`/`call`) is unnecessary.

A66-otel-stats.md
```diff
@@ -0,0 +1,325 @@
...
+grpc.io/server/server_latency    | grpc.server.call.duration
+
+#### Metrics with Nuanced Differences
+
+Unfortunately, the implementations of the gRPC OpenCensus spec in the various
+languages do not agree on the definition of the following size metrics. Go
+records uncompressed message bytes for the OpenCensus metric, while C++ and Java
+record the compressed message bytes. The OpenTelemetry spec proposed here calls
+for recording the compressed message bytes, resulting in an equivalence between
+the metrics definitions for C++ and Java, but not for Go.
```
joybestourous, line 242: it looks like the grpc-go implementation does have a compressed bytes metric (https://sourcegraph.com/github.com/grpc/grpc-go/-/blob/stats/opencensus/client_metrics.go?L39) which sounds like it would be equivalent here. is there a reason not to use that library/metric rather than [ocgrpc](https://pkg.go.dev/go.opencensus.io/plugin/ocgrpc)? we're setting up opencensus now with the intention of migrating when this work is complete, and we'd like to make sure the migration is as smooth as possible.

joybestourous (commented):
thanks for getting started on this!

akshayjshah:
I was a little surprised that this proposal doesn't reference [the existing OpenTelemetry specification for RPC metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md), especially since that doc includes a section just for gRPC. Fully acknowledging that the OTel doc is still experimental, could this proposal adopt/extend OTel's semantic conventions rather than starting from scratch?

A66-otel-stats.md
yashykt: Yes, it would be the same as the `method` and `status`

A66-otel-stats.md
```diff
@@ -0,0 +1,325 @@
...
+    *Unit*: {attempt} <br>
+*   **grpc.client.attempt.duration** <br>
+    End-to-end time taken to complete an RPC attempt including the time it takes to pick a subchannel. <br>
+    *Attributes*: grpc.method, grpc.target, grpc.status <br>
+    *Type*: Histogram (Latency Buckets) <br>
+*   **grpc.client.attempt.sent_total_compressed_message_size** <br>
+    Total bytes (compressed but not encrypted) sent across all request messages (metadata excluded) per RPC attempt; does not include grpc or transport framing bytes. <br>
+    Attributes: grpc.method, grpc.target, grpc.status <br>
+    Type: Histogram (Size Buckets) <br>
+*   **grpc.client.attempt.rcvd_total_compressed_message_size** <br>
```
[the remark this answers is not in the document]
yashykt: Yeah, that is true. This is one of the things we considered when deciding between `rcvd` and `recv`. The argument was that `rcvd` corresponds to `received` while `recv` corresponds to `receive`. 
We are already changing the naming when moving from OpenCensus to OpenTelemetry, so why not fix this as well. That being said, if there is a strong opinion to maintain consistency in naming with OpenCensus, I'll take that into consideration.

A66-otel-stats.md
yashykt: Consider the case where you have two plugins configured with different `MeterProviders`. These `MeterProviders` could be configured with different views and be configured to export data to different backends.

A66-otel-stats.md
yashykt: Yeah, this is still under flux. Suggestions welcome

A66-otel-stats.md
yashykt: That is one, and as discussed online, it is still not clear how we would want the registration look like for per-channel or per-server building if we take ordering into consideration.

A66-otel-stats.md
yashykt: Yeah, this gRFC doesn't list all of them, but there are potentially many more metrics we would want, for example - 
1) `grpc.client.call.duration`
2) `grpc.client.method.sent_messages` \ `grpc.client.method.rcvd_messages`
3) `grpc.client.connection.opened`
4) `grpc.client.http.pings_sent`

A66-otel-stats.md
yashykt: Yeah, that is a newer metric recently added. If that's the metrics being used then yes it is equivalent. The one listed here is the one from the original list at https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md

yashykt:
> I was a little surprised that this proposal doesn't reference [the existing OpenTelemetry specification for RPC metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md), especially since that doc includes a section just for gRPC. Fully acknowledging that the OTel doc is still experimental, could this proposal adopt/extend OTel's semantic conventions rather than starting from scratch?

@akshayjshah There were a bunch of conversations about this with the OTel folks, and the understanding was that it would be hard for different RPC systems to export a single set of metrics given the different nuances that each protocol would have. For example, gRPC semantics make a difference between attempts and calls. There are also differences between the various metrics exporting compressed/uncompressed sizes.

akshayjshah:
> @akshayjshah There were a bunch of conversations about this with the OTel folks, and the understanding was that it would be hard for different RPC systems to export a single set of metrics given the different nuances that each protocol would have.

I'm poking through issues on `grpc/grpc` and `grpc/proposals`, and I'm having a hard time finding these conversations. Were they in a place that's public, and do you have any links? There are _lot_ of places to search, so I'm probably just looking in the wrong spot.

> For example, gRPC semantics make a difference between attempts and calls. There are also differences between the various metrics exporting compressed/uncompressed sizes.

Every RPC (and REST) framework I'm aware of supports compression, so this doesn't strike me as gRPC-specific. Amending the OTel metrics schema to distinguish compressed from uncompressed sizes (and clarifying that protocol-specific framing shouldn't be counted) would be a nice upstream improvement. If gRPC's notion of "attempts" is unique, it seems reasonable to handle that as an additive set of metrics beyond what's in the OTel semantic conventions.

Many of the metrics proposed here - e.g., call latency - do have direct analogs in other systems. If the gRPC team feels that the metrics schema designed by the OpenTelemetry committers isn't appropriate, it would be nice to explain why in  the Background and Related Proposals sections of this document. As written, this proposal suggests (by omission) that there's no prior art for collecting OpenTelemetry metrics from gRPC.

A66-otel-stats.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Metrics Schema
+
+#### Units
+
+Following the
+[OpenTelemetry Metrics Semantic Conventions](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/),
+the following units are used -
+
+*   Latencies are measured in float64 seconds, `s`
```
joshcarp, line 37: Might want to be more specific about which part of the OpenTelemetry Metrics Semantic Conventions this follows; 
- [http.server.duration is specified as seconds (s)](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/rpc-metrics/#rpc-server)
- [rpc.server.duration is specified as milliseconds (ms)](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/rpc-metrics/#rpc-server)

yashykt:
> I'm poking through issues on grpc/grpc and grpc/proposals, and I'm having a hard time finding these conversations. Were they in a place that's public, and do you have any links? There are lot of places to search, so I'm probably just looking in the wrong spot.

No, these were on video conferences unfortunately. I'll tag @jsuereth as the OTel Metrics Conventions spokesperson.

> If the gRPC team feels that the metrics schema designed by the OpenTelemetry committers isn't appropriate, it would be nice to explain why in the Background and Related Proposals sections of this document. 

Thanks! Yeah, I'll update the rationale. I don't know why I skipped that part.

A66-otel-stats.md
```diff
@@ -0,0 +1,325 @@
...
+    Start timestamp - After the transport knows that it's got a new stream. For HTTP2, this would be after the first header frame for the stream has been received and decoded. Whether the timestamp is recorded before or after HPACK is left to the implementation. <br>
+    End timestamp - Ends at the first point where the transport considers the stream done. For HTTP2, this would be when scheduling a trailing header with END_STREAM to be written, or RST_STREAM, or a connection abort. Note that this wouldn’t necessarily mean that the bytes have also been immediately scheduled to be written by TCP. <br>
+    *Attributes*: grpc.method, grpc.authority, grpc.status <br>
+    *Type*: Histogram (Latency Buckets) <br>
+
+## OpenTelemetry Plugin Architecture
+
+To-be-filled
+
+This section describes a CallTracer approach to collect the client and server
```
markdroth, line 120: We should probably describe the CallTracer API in more detail, so that its nature is clear to the reader.  Obviously, the details will be language-specific, but the semantics should be similar enough across languages to allow a common description.

A66-otel-stats.md
yashykt: Removed `authority`. Explanation below.

A66-otel-stats.md
yashykt: Done

yashykt:
We initially added authority as an attribute to servers, but the value add is not clear at the moment, and there is a potential of opening the servers to attack since the clients can send arbitrary authority values. (We could potentially build out an API where servers provide the ability to register which authorities should be accepted by the stats plugin, but we punt this for the future, for when we get a more clear idea as to the benefit here.)

A66-otel-stats.md
```diff
@@ -0,0 +1,395 @@
...
+upto the user to choose the right bucket boundaries.
+
+Also note that, as per an
+[OpenTelemetry proposal on stability](https://docs.google.com/document/d/1Nvcf1wio7nDUVcrXxVUN_f8MNmcs0OzVAZLvlth1lYY/edit#heading=h.dy1cg9doaq26)
+though, changes to bucket boundaries might not be considered a breaking change.
+Depending on the proposal, this recommendation would change to use
+`ExponentialHistogram`s instead, which would allow for automatic adjustments of
+the scale to better fit the data.
+
+#### Attributes
```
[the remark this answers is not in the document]
yashykt: Do you mean attributes (key-value pairs) that remain constant for the life of the plugin?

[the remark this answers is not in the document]
yashykt: Does `OTEL_RESOURCE_ATTRIBUTES` satisfy your usecase?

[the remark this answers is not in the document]
joybestourous: I don't think it does. From what I understand using that environment variable is only useful for static global flags. What we're hoping for is something more dynamic. For example, some of our users like to specify a `source` tag that identifies a group of clients s.t. problems can be localized when they arise. Tags like these may differ within a given environment, so an environment variable is insufficient for setting those tags. Without the flexibility to add tags more dynamically in code, the metrics become rather inflexible and force us to separately define metrics when we otherwise would like to use the ones provided by grpc.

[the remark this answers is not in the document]
yashykt: Clarifying questions : Are you thinking of a case where these labels be the same within the whole process, or would different clients/servers in the same process also want different labels?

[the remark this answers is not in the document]
joybestourous: The latter. The labels may be different for clients/servers within the same process. They may also be different across requests for the same client/server within a process (e.g. if we want to classify requests in some way). The more flexible it can be the better.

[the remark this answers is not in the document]
yashykt: This is the first I'm hearing about such a usecase, but that doesn't mean that we can't add it. Can you talk more about what kind of labels you are imagining? The source labels would not change for the life of the channel/server, so you should be able to add those at the point of configuring the plugin for the respective channel/server. As for the method, we already have the method name (and also the target name for clients).
((That would also help determine what information you would need for adding those labels.)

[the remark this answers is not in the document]
joybestourous: Can that be added at the point of configuration? To my understanding source labels might change within the life of the server because they may differ per request. There are other per-request attributes that may be useful such as tags related to request content or authorization.

A66-otel-stats.md
```diff
@@ -144,12 +144,53 @@ the scale to better fit the data.
...
+the incoming call. Depending on the implementation details, the method may be
+propagated as part of the initial metadata.
+
+The following call-outs are needed on the `CallTracer` -
+
+*   When the call has been created. This call-out should be before payload
+    serialization.
+*   When new attempts are created on the call along with information on whether
+    the attempt was a transparent retry or not. (Attempts are created after name
+    resolution but before the pick.) This is also when it's expected for the
```
markdroth, line 168: s/pick/LB pick/

A66-otel-stats.md
```diff
@@ -144,12 +144,53 @@ the scale to better fit the data.
...
+    `CallAttemptTracer` to be created.
+*   When an attempt ends. This will be needed future stats around retries and
```
markdroth, line 170: s/needed/needed for/

A66-otel-stats.md
markdroth, line 168: Maybe also mention that it's after xDS HTTP filters are executed, and reference gRFC A39.

A66-otel-stats.md
```diff
@@ -144,12 +144,53 @@ the scale to better fit the data.
...
+    hedging. This information can also be propagated through the
+    `CallAttemptTracer` if the `CallAttemptTracer` keeps a reference to the
+    parent `CallTracer` object.
+*   When the call ends. This along with the call creation call-out allows the
+    `CallTracer` to calculate the call duration.
+
+The following call-outs are needed on the `CallAttemptTracer` -
```
markdroth, line 177: Do we want to say that all of these should be as close to the transport as possible?

A66-otel-stats.md
```diff
@@ -144,12 +144,53 @@ the scale to better fit the data.
...
+
+*   When a new message is sent/received. The message should be in its compressed
+    form.
+*   When the trailing metadata/status is received for the attempt. Receipt of
+    this indicates that the attempt has ended. Implementations may choose to
+    delegate the responsibility of notifying the `CallTracer` about the attempt
+    end to the `CallAttemptTracer`.
+
+The following call-outs are needed on the `ServerCallTracer` -
```
markdroth, line 186: Same question here: Do we want to say that all of these should be as close to the transport as possible?

A66-otel-stats.md
```diff
@@ -0,0 +1,468 @@
...
+    *Attributes*: grpc.method, grpc.status <br>
+    *Type*: Histogram (Size Buckets) <br>
+*   **grpc.server.call.duration** <br>
+    This metric aims to measure the end2end time an RPC takes from the server transport’s (HTTP2/ inproc / cronet) perspective. <br>
+    Start timestamp - After the transport knows that it's got a new stream. For HTTP2, this would be after the first header frame for the stream has been received and decoded. Whether the timestamp is recorded before or after HPACK is left to the implementation. <br>
+    End timestamp - Ends at the first point where the transport considers the stream done. For HTTP2, this would be when scheduling a trailing header with END_STREAM to be written, or RST_STREAM, or a connection abort. Note that this wouldn’t necessarily mean that the bytes have also been immediately scheduled to be written by TCP. <br>
+    *Attributes*: grpc.method, grpc.status <br>
+    *Type*: Histogram (Latency Buckets) <br>
+
+## OpenTelemetry Plugin Architecture
```
markdroth, line 142: Every section header from here until the "Rationale" section needs to be one level lower (i.e., add one more `#` character`).  All of these sections belong under the "Proposal" section.

A66-otel-stats.md
```diff
@@ -0,0 +1,468 @@
...
+    *   grpc.io/server/sent_bytes_per_method
+    *   grpc.io/server/received_bytes_per_method
+*   Retry Views
+    *   grpc.io/client/retries_per_call
+    *   grpc.io/client/retries
+    *   grpc.io/client/transparent_retries_per_call
+    *   grpc.io/client/transparent_retries
+    *   grpc.io/client/retry_delay_per_call
+
+## Migration Strategies
```
markdroth, line 407: This and its subsections should be indented one level more so that they are underneath the "Migration from OpenCensus" header.  This should be at the same level as the "Metric Schema Comparison" header.

A66-otel-stats.md
```diff
@@ -0,0 +1,468 @@
...
+
+The [Meter](https://opentelemetry.io/docs/specs/otel/metrics/api/#get-a-meter)
+creation should use a `name` that identifies the library, for example,
+"grpc-c++", "grpc-java", "grpc-go". The `version` should be the same as the
+release version of the gRPC library, for example, "1.57.1".
+
+### C++
+
+```c++
+Class OpenTelemetryPluginBuilder {
```
markdroth, line 239: Please update this to match the latest changes in the C++ impl.

A66-otel-stats.md
```diff
@@ -0,0 +1,468 @@
...
+     * Otherwise it will be recorded as "other". */
+    public OpenTelemetryBuilder targetFilter(Predicate<String> targetFilter);
+
+    public OpenTelemetryModule build();
+}
+```
+
+### Go
+
+To be filled
```
markdroth, line 308: @zasweq Do you have something we can show here for the Go API?

A66-otel-stats.md
```diff
@@ -0,0 +1,468 @@
...
+
+### Python
+
+To be filled
```
markdroth, line 312: @XuanWang-Amos Do you have some idea of what the python API for this will look like?  If not, we can add this later.

A66-otel-stats.md
```diff
@@ -0,0 +1,468 @@
...
+sense for all systems. For gRPC specifically, the following differences are
+immediately obvious -
+
+*   gRPC differentiates between the concept of a `call` and an `attempt`. Each
+    `call` can have multiple `attempts` with retries/hedging.
+*   The various gRPC implementations can record the compressed message lengths,
+    but not all implementations can get the uncompressed message length (as
+    recommended by OTel RPC conventions.)
+
+This gRFC, hence, intends to override the [General RPC conventions] for gRPC's
```
markdroth, line 453: This link isn't formatted correctly.

markdroth (commented):
This looks very good overall!  Most comments are minor.

I'd like to get feedback from @ejona86 and @dfawley as well.

A66-otel-stats.md
yashykt: The position of the callout matters the most for the start and end timestamp, which for a client-side attempt is at the attempt top and not close to the transport.

The position for recording when a new message is sent/received does not matter for stats.

A66-otel-stats.md
yashykt: Done

A66-otel-stats.md
```diff
@@ -144,12 +144,53 @@ the scale to better fit the data.
 This section describes a `CallTracer` approach to collect the client and server
 per-attempt/call metrics. A CallTracer is a class that is instantiated for every
 call. This class has various methods that are invoked during the lifetime of the
-call. On the client side, the CallTracer knows about multiple attempts on the
+call. On the client-side, the CallTracer knows about multiple attempts on the
```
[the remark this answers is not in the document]
yashykt: Done

A66-otel-stats.md
yashykt: Done

A66-otel-stats.md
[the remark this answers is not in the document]
yashykt: This is tricky. For example, request content is not going to be available at the point of `started` because at that point you would only have the initial metadata. It won't be available at the point where the RPC ends either, since the RPC could potentially have been a streaming RPC.

I think we might be able to provide an API to inject labels based on the initial metadata. The issue with that currently is that atleast in gRPC Core, the types used in stats plugins like OTel plugins are not stable, and they are still going through enough change. This might be possible in other language implementations though.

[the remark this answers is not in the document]
yashykt: @markdroth @ejona86 @dfawley what do you think?

A66-otel-stats.md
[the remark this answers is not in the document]
markdroth: You added a link in the "related proposal" section, but I think we need to include the word "retry" in the description here somehow.  Otherwise, I'm not sure it's clear to the reader that "attempt" refers to the retry code.

A66-otel-stats.md
markdroth: > The position for recording when a new message is sent/received does not matter for stats.

That's probably fine for stats, but what about tracing and logging?  We'll be using this same API for tracing and logging, so if it matters for those, then we need to specify the requirements for this API correctly.

A66-otel-stats.md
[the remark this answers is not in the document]
markdroth: It's not clear to me whether the request here is to allow setting labels based on headers or based on message payloads.

I am pretty strongly opposed to the idea of allowing labels to be set based on message payloads, for the following reasons:
- gRPC at its core treats payloads as opaque blobs and does not make assumptions about what format the payloads are in.  While most users use gRPC with protobuf payloads, gRPC can be used with other serialization formats, and we don't want to break that by baking in assumptions that the payloads are in protobuf form.
- There are a bunch of layering problems here.  At the layer at which gRPC core is interacting with the CallTracer, the payloads may not be represented in a form suitable to export in any sort of public API.  And the payloads are typically just serialized bytes, so even if they were in protobuf form, it would add quite a bit of overhead to deserialize them just to figure out what label to add.
- As Yash mentioned, the entire idea of setting labels based on message contents inherently seems to assume that the requests are unary (i.e., there is only one message payload per RPC).  However, in gRPC, streaming is a first-class use-case (in fact, unary is actually implemented as a special case of streaming), and we don't want to build features that work only for unary.

Using headers instead of payloads is less objectionable to me than using message payloads, but still not a slam dunk.  Here are some factors we'd need to consider:
- My main concern is that if we give users the ability to add arbitrary labels based on the headers or message payloads, it will be very hard to prevent them from hitting cardinality limits for the resulting timeseries.  This can be a problem both in OT storing the data in memory and in whatever monitoring system you eventually export the OT data to.  It's not clear to me how we can provide this flexibility without making it too easy for users to shoot themselves in the foot.
- Even for headers, there can be significant overhead here, especially if we would need to support things like regex matches.  I'm concerned that we would again be giving users the ability to shoot themselves in the foot performance-wise.
- There are some API/layering concerns here similar to those that I mentioned for message payloads, although they are somewhat more tractable for headers than they are for message payloads.  We might be able to find a way to make this workable, but it will take some careful thought.

A66-otel-stats.md
yashykt: Yeah, for tracing, I was hoping the tracing gRPC makes it more clear, but we would actually want two callouts. One for the compressed size, and one for the non-compressed size, and this would happen at the point of message compression.

A66-otel-stats.md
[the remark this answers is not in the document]
yashykt: added

A66-otel-stats.md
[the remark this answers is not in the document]
s-matyukevich: How about we do this:
* allow setting labels only based on headers.
* require the header to have full key=value pair definition, no regexes and no additional parsing allowed.

This still should cover the main use-case  we have in mind for this. If we need to group clients on a server by client type, we'll simply add a header "source=clientA" via an interceptor on every client.  Then on the server side we will be able to filter requests by client type, which is super useful IMO.  In our case client type is the name of the client service, it is always static and it has very low cardinality, but I agree that people may misuse this feature and put some very dynamic data in it. 

I think this addresses some of the Mark's concerns:
* the overhead should be minimal, as we are doing no additional processing and parsing headers is cheap and should be done early in the call stack anyway.
* It is harder to accidentally get explosive cardinality if you are forced to put the exact values in the header. We can also provide a hardcoded limit and completely remove a tag on the server after its cardinality exceeds the threshold. In this case users will either notice a warn message in the server logs or notice the the tag is gone, so they could either fix the cardinality of their tag or bump the threshold via some configurable parameter.

A66-otel-stats.md
XuanWang-Amos: Not yet, I need discuss with Richard first, this can be added later.

A66-otel-stats.md
```diff
@@ -0,0 +1,540 @@
...
+    gRPC implementations allow server to handle generic method names, i.e., not
+    registering method names in advance with the server. This allows clients to
+    send arbitrary method names that could potentially open up the server to
+    malicious attacks that result in metrics being stored with a high
+    cardinality. To prevent this, unregistered/generic method names should by
+    default be reported with "other" value instead. Implementations should
+    provide the option to override this behavior to allow recording generic
+    method names as well.
+*   `grpc.status` : gRPC server status code received, e.g. "OK", "CANCELLED",
+    "DEADLINE_EXCEEDED"
```
XuanWang-Amos, line 82: [Minor] Can we add a link to all `grpc.status`?

A66-otel-stats.md
```diff
@@ -0,0 +1,540 @@
...
+    implementations should *NOT* fallback on this global.
+*   Optionally allow enabling/disabling metrics. This would allow optimizations
+    to avoid computation and collection of expensive stats within the gRPC
+    library. Note that even without this capability, users of OpenTelemetry
+    would be able to customize
+    [views](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view) through
+    the MeterProvider.
+*   Optionally allow setting of a OpenTelemetry plugin for a specific channel or
+    server, instead of setting it globally.
+*   Optionally allowing setting of a map of constant attributes that are
```
XuanWang-Amos, line 230: [Minor] `Optionally allow`?

A66-otel-stats.md
```diff
@@ -0,0 +1,540 @@
...
+          target_attribute_filter);
+  // If set, \a generic_method_attribute_filter is called per call with a
+  // generic method type to decide whether to record the method name or to
+  // replace it with "other". Non-generic or pre-registered methods remain
+  // unaffected. If not set, by default, generic method names are replaced with
+  // "other" when recording metrics.
+  OpenTelemetryPluginBuilder& SetGenericMethodAttributeFilter(
+      absl::AnyInvocable<bool(absl::string_view /*generic_method*/) const>
+          generic_method_attribute_filter);
+  void BuildAndRegisterGlobal();
```
XuanWang-Amos, line 280: Can we add comments to explain what this API does?

A66-otel-stats.md
```diff
@@ -0,0 +1,540 @@
...
+To be filled
+
+### Migration from OpenCensus
+
+The following sections show the differences between the gRPC OpenCensus spec and
+the proposed gRPC OpenTelemetry spec and the mapping of metrics between the two.
+It also presents metrics present in OpenCensus spec that do not map to a metric
+in the OpenTelemetry spec at present. Two migration strategies are also proposed
+for customers who are satisfied with the stats coverage provided by the current
+OpenTelemetry spec.
```
XuanWang-Amos, line 392: Do we mean `current OpenCensus spec` here?

A66-otel-stats.md
yashykt: Done

A66-otel-stats.md
yashykt: No, the `current OpenTelemetry spec` refers to this spec. Updated the wording.

A66-otel-stats.md
yashykt: Done

A66-otel-stats.md
yashykt: Done

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+    might again blow up the cardinality. Implementations should provide the
+    option to override recorded target names with "other".
+
+#### Client Per-Attempt Instruments
+
+*   **grpc.client.attempt.started** <br>
+    The total number of RPC attempts started, including those that have not completed. <br>
+    *Attributes*: grpc.method, grpc.target <br>
+    *Type*: Counter <br>
+    *Unit*: {attempt} <br>
```
DNVindhya, line 99: Is `{attempt}` supposed to be recorded as OpenTelemetry metric [unit](https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/#units) as is?

XuanWang-Amos (approved):

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+*   Author: Yash Tibrewal (@yashykt)
+*   Approver: Mark Roth (@markdroth)
+*   Status: In Review
+*   Implemented in: <language, ...>
+*   Last updated: Jul 20, 2023
+*   Discussion at: https://groups.google.com/g/grpc-io/c/po-deqYEQzE
+
+## Abstract
+
+Propose a metrics data model for gRPC OpenTelemetry metrics.
```
zasweq, line 12: Nit: this also specifies gRPC Instrumentation Component requirements and APIs

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+    16777216, 67108864, 268435456, 1073741824, 4294967296
+*   Count : 0, 1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192,
+    16384, 32768, 65536
+
+These buckets were chosen to maintain compatibility with the gRPC OpenCensus
+spec. The OpenTelemetry API has added an experimental feature for
+[advice](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advice)
+that would allow the gRPC library to provide these buckets as a hint. Since this
+is still an experimental feature and not yet implemented in all languages, it is
+upto the user to choose the right bucket boundaries.
```
zasweq, line 60: Can you mention our instrumentation component creates views which pick up the default bounds if not specified by user? Also somewhere in this document specify a user set view takes precedence over the OTel view.

murgatroid99 (approved):

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+*   `grpc.method` : Full gRPC method name, including package, service and
+    method, e.g. "google.bigtable.v2.Bigtable/CheckAndMutateRow". Note that some
+    gRPC implementations allow server to handle generic method names, i.e., not
+    registering method names in advance with the server. This allows clients to
+    send arbitrary method names that could potentially open up the server to
+    malicious attacks that result in metrics being stored with a high
+    cardinality. To prevent this, unregistered/generic method names should by
+    default be reported with "other" value instead. Implementations should
+    provide the option to override this behavior to allow recording generic
+    method names as well.
```
zasweq, line 80: This talks about generic methods, but does mention arbitrary methods from a headers frame (unrelated to if the server supports generic methods). I guess technically this is semantically covered by the statement "allows clients to send arbitrary method names", but I don't like the language, since it's just arbitrary headers frames from the wire, not whether the client is allowed to or not.

A66-otel-stats.md
dfawley, line 60: What does this mean, "up to the user"?  How would the user do this?  Or do you mean _we_ are "the user" here?

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+  // grpc.client.attempt.duration
+  // grpc.client.attempt.sent_total_compressed_message_size
+  // grpc.client.attempt.rcvd_total_compressed_message_size
+  // grpc.server.call.started
+  // grpc.server.call.duration
+  // grpc.server.call.sent_total_compressed_message_size
+  // grpc.server.call.rcvd_total_compressed_message_size
+  OpenTelemetryPluginBuilder& EnableMetric(absl::string_view metric_name);
+  OpenTelemetryPluginBuilder& DisableMetric(absl::string_view metric_name);
+  OpenTelemetryPluginBuilder& DisableAllMetrics();
```
zasweq, line 264: Nit: I find this API a lot more clunky than just passing a list (and exporting a list of defaults). Yash mentioned this went through design review though, so I will keep the Go API as is. It feels clunky to by default get them all, and if you want to disable a subset of metrics you have to call DisableMetric() multiple times. Rather than multiple operations in the case where you want all metrics but two (Disable(A), Disable(B)) or in the case where you want 2 metrics (Yash mentioned DisableAllMetrics() Enable(A) Enable(B)), to me it's much cleaner/simpler and more readable to have oneOperation(list of metrics you want).

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+  // If set, \a target_attribute_filter is called per channel to decide whether
```
zasweq, line 265: Nit: s/\a/a

zasweq (changes requested):
Some comments.

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+    malicious attacks that result in metrics being stored with a high
+    cardinality. To prevent this, unregistered/generic method names should by
+    default be reported with "other" value instead. Implementations should
+    provide the option to override this behavior to allow recording generic
+    method names as well.
+*   `grpc.status` : gRPC server status code received, e.g. "OK", "CANCELLED",
+    "DEADLINE_EXCEEDED".
+    [(Full list)](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)
+*   `grpc.target` : Canonicalized target URI used when creating gRPC Channel,
+    e.g. "dns:///pubsub.googleapis.com:443", "xds:///helloworld-gke:8000".
```
dfawley, line 85: Did we decide this needs to have 3(+) slashes, or that it was fine to use the user's text if they leave out the authority, e.g. `"dns:host:port"`, or that it doesn't matter?

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+A CallTracer is a class that is instantiated for every call. This class has
+various methods that are invoked during the lifetime of the call. On the
+client-side, the CallTracer knows about multiple attempts on the same call (due
+to retries or hedging), and creates a `CallAttemptTracer` object for each
+attempt, and the `CallAttemptTracer` gets invoked during the lifetime of the
+attempt. On the server-side, we have an equivalent `ServerCallTracer`. (There is
+no concept of an attempt on the server-side.)
+
+The OTel plugin will basically be a way of configuring CallTracer factories on
+gRPC channels and servers.
```
dfawley, line 161: Try to avoid words like "basically".  Cleaner is: `The OTel plugin will configure CallTracer factories on gRPC channels and servers.`

Also, should we avoid `OTel` as a lingo-ish term and use `Open Telemetry` everywhere instead?

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+
+*   Allow installing multiple OpenTelemetry plugins.
+*   Allow setting a
+    [MeterProvider](https://opentelemetry.io/docs/specs/otel/metrics/api/#meterprovider)
+    on individual plugins. Implementations should require a MeterProvider being
+    set. A MeterProvider not being set should result in a no-op. Some
+    OpenTelemetry language APIs have a global MeterProvider. gRPC
+    implementations should *NOT* fallback on this global.
+*   Optionally allow enabling/disabling metrics. This would allow optimizations
+    to avoid computation and collection of expensive stats within the gRPC
```
dfawley, line 224: In a vacuum this doesn't make sense.  Simply don't register the plugin if you don't want metrics enabled...  But I think the idea is that the same plugin will do tracing in the future, right?

A66-otel-stats.md
```diff
@@ -0,0 +1,542 @@
...
+// Client applications interested in instrumenting their grpc.ClientConn should
+// pass the dial option returned from this function as a dial option to
+// grpc.Dial().
+//
+// For the metrics supported by this instrumentation code, a user needs to
+// specify the client metrics to record in metrics options. A user also needs to
+// provide an implementation of a MeterProvider. If the passed in Meter Provider
+// does not have the view configured for an individual metric turned on, the API
+// call in this component will create a default view for that metric.
+func DialOption(mo MetricsOptions) grpc.DialOption {}
```
dfawley, line 367: `DialOptionOptions`?  😆 

Seriously though I think we'll want this to take a struct that contains `MetricsOptions` instead so that we can include tracing options in the future.  Or a name besides `MetricsOptions` that includes tracing options as well.  Probably just `Options`?

A66-otel-stats.md
yashykt: Yes

A66-otel-stats.md
yashykt: https://opentelemetry.io/docs/specs/otel/#project-naming It's official :D

A66-otel-stats.md
zasweq: This means the caller of the API by setting a registered view on their meter provider. I agree, need to expand this language out. Also commented in another comment to please talk about the precedence of set meter providers views > api call views particularly with respect to bounds but please mention all the precedence ordering of the view properties (I think it's all).

A66-otel-stats.md
zasweq: Yeah, I will scale this up to Options { metrics options, tracing options } (OpenCensus just had tracing options). Or should I change this and scale upward to not break users eventually (i.e. have an options struct with metrics options as a knob but no tracing options until tracing)?

A66-otel-stats.md
zasweq: I think the idea of this language is if a user wants a subset of the metrics, avoiding computation on the metrics the user doesn't want.

A66-otel-stats.md
yashykt: That's just OTel stuff, not specific to gRPC.

A66-otel-stats.md
yashykt: Yes, as per https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/

A66-otel-stats.md
yashykt: I don't see an authoritative document on this, but I think it'd be nice to use the same format across languages. Core uses the three slash format even without authority.

A66-otel-stats.md
yashykt: The \a was to just say that `target_attribute_filter` is a function argument

A66-otel-stats.md
yashykt: We went through a design review on the API here among the Core/C++ folks, and the current API was what was preferred over passing in a list of metrics. The two APIs are really equivalent in functionality, so I don't care much.

A66-otel-stats.md
yashykt: This is all explained in `OTel SDK`. I'll add a link to that here and fix the wording here

A66-otel-stats.md
murgatroid99: The [naming doc](https://github.com/grpc/grpc/blob/master/doc/naming.md) lists the DNS name format as `dns:[//authority/]host[:port]`, which implies to me that the slashes should only be included if the authority is non-empty.

A66-otel-stats.md
DNVindhya: Can we also add unit and description for this as well other metrics?

yashykt: Following the recommendations at https://opentelemetry.io/docs/specs/otel/metrics/semantic_conventions/#instrument-units

A66-otel-stats.md
yashykt: > Try to avoid words like "basically". Cleaner is: The OTel plugin will configure CallTracer factories on gRPC channels and servers.

done

A66-otel-stats.md
yashykt: Discussed offline and updated wording

A66-otel-stats.md
```diff
@@ -57,7 +57,8 @@ spec. The OpenTelemetry API has added an experimental feature for
 [advice](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advice)
 that would allow the gRPC library to provide these buckets as a hint. Since this
 is still an experimental feature and not yet implemented in all languages, it is
-upto the user to choose the right bucket boundaries.
+upto the user to choose the right bucket boundaries and set it through the
```
markdroth, line 60: s/upto/up to/

A66-otel-stats.md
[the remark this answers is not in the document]
markdroth: I'm not really keen on having an API that explicitly provides the functionality that allows header values to dictate metric label values directly.  I think that makes it a little too easy for users to shoot themselves in the foot in terms of cardinality.  I think I'd rather provide an API hook that calls some user-supplied code to read the headers and return the set of labels to add, which would need to return the result synchronously.  That does still give users the ability to shoot themselves in the foot, but at least they have to write a bit of code to do so, so at least we won't be directly encouraging dangerous functionality.

I think this probably needs a bit more consideration before we decide.  Let's keep talking about this separately, but we need to get the basic OT metric functionality nailed down fairly soon, so I think we'll exclude this from this gRFC for now.  However, there will definitely be more gRFCs adding more OT metrics functionality on top of the foundation laid by this gRFC, so we'll definitely have opportunities to add this once we have consensus on it.

A66-otel-stats.md
markdroth: Why would we want two separate call-outs for tracing?

I think we would still want the call-out to be as close to the transport as possible, because even for tracing, it's important to reflect timing.  Someone looking at a trace may want to see how long it takes messages to be sent or received, and how much of that time is in the network vs. in gRPC itself.

A66-otel-stats.md
yashykt: Added the unit. I don't think description matters so much but for consistency, we can use the description at https://github.com/grpc/grpc/pull/34441

A66-otel-stats.md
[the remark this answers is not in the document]
yashykt: As discussed, I've stripped the proposal to the bare minimal stuff that we all agree on.

A66-otel-stats.md
yashykt: Added

A66-otel-stats.md
yashykt: Outdated since these APIs have been removed

A66-otel-stats.md
```diff
@@ -0,0 +1,525 @@
...
+};
+
+```
+
+In the future, additional API might be provided to allow registering the plugin
+for a particular channel or server builder.
+
+#### Java
+
+```
```
DNVindhya: s/\```/\```java

A66-otel-stats.md
```diff
@@ -0,0 +1,525 @@
...
+     * Otherwise it will be recorded as "other". */
+    public OpenTelemetryBuilder targetFilter(Predicate<String> targetFilter);
+
+    public OpenTelemetryModule build();
+}
+```
+
+#### Go
+
+```
```
DNVindhya: s/\```/\```go

A66-otel-stats.md
```diff
@@ -0,0 +1,525 @@
...
+    public OpenTelemetryModuleBuilder openTelemetry(OpenTelemetry openTelemetry);
+
+    /* If targetFilter returns true for a target, target is recorded as is.
+     * Otherwise it will be recorded as "other". */
+    public OpenTelemetryBuilder targetFilter(Predicate<String> targetFilter);
+
+    public OpenTelemetryModule build();
+}
+```
+
```
DNVindhya, line 328: Can you add the following, 
"Note: For non-generated methods, method names are recorded as "other" for `grpc.method` attribute. If you are interested in recording the method names for these methods, set [`isSampledToLocalTracing`](https://grpc.github.io/grpc-java/javadoc/io/grpc/MethodDescriptor.html#isSampledToLocalTracing()) to `true` while defining your methods in [`HandlerRegistry`](https://grpc.github.io/grpc-java/javadoc/io/grpc/HandlerRegistry.html).

A66-otel-stats.md
yashykt: outdated

A66-otel-stats.md
yashykt: outdated

A66-otel-stats.md
yashykt: Added Python API

A66-otel-stats.md
yashykt: Done, thanks!

A66-otel-stats.md
```diff
@@ -0,0 +1,525 @@
...
+spec. The OpenTelemetry API has added an experimental feature for
+[advice](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advice)
+that would allow the gRPC library to provide these buckets as a hint. Since this
+is still an experimental feature and not yet implemented in all languages, it is
+up to the user to choose the right bucket boundaries and set it through the
+[OTel SDK](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view).
+
+Also note that, as per an
+[OpenTelemetry proposal on stability](https://docs.google.com/document/d/1Nvcf1wio7nDUVcrXxVUN_f8MNmcs0OzVAZLvlth1lYY/edit#heading=h.dy1cg9doaq26)
+though, changes to bucket boundaries might not be considered a breaking change.
```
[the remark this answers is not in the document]
yashykt: removed "though"

A66-otel-stats.md
```diff
@@ -0,0 +1,525 @@
...
+    "DEADLINE_EXCEEDED".
+    [(Full list)](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)
+*   `grpc.target` : Canonicalized target URI used when creating gRPC Channel,
+    e.g. "dns:///pubsub.googleapis.com:443", "xds:///helloworld-gke:8000".
+    Canonicalized target URI is its form with the scheme if the user didn't
+    mention the scheme. For channels such as inprocess channels where a target
+    URI is not available, implementations can synthesize a target URI. It is
+    possible for some channels to use IP addresses as target strings and this
+    might again blow up the cardinality. Implementations should provide the
+    option to override recorded target names with "other".
```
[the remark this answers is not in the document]
yashykt: Added the default behavior

A66-otel-stats.md
yashykt: done

A66-otel-stats.md
yashykt: As discussed, I think this is something that needs to be covered by the tracing grfc

A66-otel-stats.md
```diff
@@ -0,0 +1,569 @@
...
+
+### Language-Specific Details
+
+Each language implementation will provide an API for registering an
+OpenTelemetry plugin. Overall, the APIs should have the following capabilities -
+
+*   Allow installing multiple OpenTelemetry plugins.
+*   Allow setting a
+    [MeterProvider](https://opentelemetry.io/docs/specs/otel/metrics/api/#meterprovider)
+    on individual plugins. Implementations should require a MeterProvider being
```
DNVindhya, line 248: Setting a meter provider is optional based on default no-op behavior. Can you rephrase it to say "Implementations must provide an option to set MeterProvider" or something similar?

A66-otel-stats.md
```diff
@@ -0,0 +1,569 @@
...
+grpc.io/client/completed_rpcs    | (Derivable from grpc.client.attempt.duration)
+grpc.io/client/roundtrip_latency | grpc.client.attempt.duration
+grpc.io/server/started_rpcs      | grpc.server.call.started
+grpc.io/server/completed_rpcs    | (Derivable from grpc.server.call.duration)
+grpc.io/server/server_latency    | grpc.server.call.duration
+
+##### Metrics with Nuanced Differences
+
+Unfortunately, the implementations of the gRPC OpenCensus spec in the various
+languages do not agree on the definition of the following size metrics. Go
```
DNVindhya, line 465: s/size/message size

DNVindhya (commented):
Looks great. Few minor comments / clarifications.

A66-otel-stats.md
DNVindhya: Nit: I agree with @dfawley. Although, `OTel` is an official acronym imo using project name `OpenTelemetry` instead of the acronym in gRFC sounds better.

A66-otel-stats.md
```diff
@@ -0,0 +1,569 @@
...
+    "DEADLINE_EXCEEDED".
+    [(Full list)](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)
+*   `grpc.target` : Canonicalized target URI used when creating gRPC Channel,
+    e.g. "dns:///pubsub.googleapis.com:443", "xds:///helloworld-gke:8000".
+    Canonicalized target URI is its form with the scheme if the user didn't
+    mention the scheme. For channels such as inprocess channels where a target
+    URI is not available, implementations can synthesize a target URI. It is
+    possible for some channels to use IP addresses as target strings and this
+    might again blow up the cardinality. Implementations should provide the
+    option to override recorded target names with "other". If no such override
```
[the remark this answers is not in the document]
yashykt: No, by default, we record the target name as is. If the filter decides to not record it, it should show up as "other".

A66-otel-stats.md
yashykt: done

A66-otel-stats.md
yashykt: Done

A66-otel-stats.md
```diff
@@ -0,0 +1,569 @@
...
+to retries or hedging), and creates a `CallAttemptTracer` object for each
+attempt, and the `CallAttemptTracer` gets invoked during the lifetime of the
+attempt. On the server-side, we have an equivalent `ServerCallTracer`. (There is
+no concept of an attempt on the server-side.)
+
+The OTel plugin will configure CallTracer factories on gRPC channels and
+servers.
+
+A CallTracer needs to know the channel's target in the canonical form, and the
+full qualified method name for filling in the attributes needed on the metrics.
```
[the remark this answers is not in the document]
yashykt: Done

A66-otel-stats.md
```diff
@@ -0,0 +1,569 @@
...
+    cardinality. To prevent this, unregistered/generic method names should by
+    default be reported with "other" value instead. Implementations should
+    provide the option to override this behavior to allow recording generic
+    method names as well.
+*   `grpc.status` : gRPC server status code received, e.g. "OK", "CANCELLED",
+    "DEADLINE_EXCEEDED".
+    [(Full list)](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)
+*   `grpc.target` : Canonicalized target URI used when creating gRPC Channel,
+    e.g. "dns:///pubsub.googleapis.com:443", "xds:///helloworld-gke:8000".
+    Canonicalized target URI is its form with the scheme if the user didn't
```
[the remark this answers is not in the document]
yashykt: Updated the language for clarification

A66-otel-stats.md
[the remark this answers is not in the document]
yashykt: Updated the language

A66-otel-stats.md
```diff
@@ -0,0 +1,569 @@
...
+These buckets were chosen to maintain compatibility with the gRPC OpenCensus
+spec. The OpenTelemetry API has added an experimental feature for
+[advice](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advice)
+that would allow the gRPC library to provide these buckets as a hint. Since this
+is still an experimental feature and not yet implemented in all languages, it is
+up to the user of the gRPC OpenTelemetry plugin to choose the right bucket
+boundaries and set it through the
+[OTel SDK](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view).
+
+Also note that, as per an
```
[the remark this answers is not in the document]
yashykt: Done

A66-otel-stats.md
```diff
@@ -0,0 +1,569 @@
...
+package opentelemetry
+
+// MetricsOptions are the metrics options for OpenTelemetry instrumentation.
+type MetricsOptions struct {
+  // MeterProvider is the MeterProvider instance that will be used for access
+  // to Named Meter instances to instrument an application. To enable metrics
+  // collection, set a meter provider. If unset, no metrics will be recorded.
+  // Any implementation knobs (i.e. views, bounds) set in the passed in object
+  // take precedence over the API calls from the interface in this component
+  // (i.e. it will create default views for unset views).
```
zasweq, line 372: Nit: this precedence logic is only mentioned in this Go docstring. Can you please mention this somewhere in top level API documentation or a section earlier in the document?

zasweq (approved):
LGTM modulo precedence nit

A66-otel-stats.md
yashykt: This statement should actually be removed since we no longer API to override that

yashykt: Note that there is a statement up above that says - `Users of the gRPC OpenTelemetry plugin will use the OTel SDK's MeterProvider to
[control the views](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view)
and customize the metrics that will be exported.`

zasweq: What? I still call out into API. I thought we needed to since these metrics are unconditionally recorded? That docstring doesn't mention API.

A66-otel-stats.md
```diff
@@ -0,0 +1,566 @@
...
+     *         .setMeterProvider(sdkMeterProvider)
+     *         .build();
+     *
+     * If MeterProvider is not configured, no-op meterProvider will be used by default.
+     * It provides meters which do not record or emit.
+     */
+    public OpenTelemetryModuleBuilder openTelemetry(OpenTelemetry openTelemetry);
+
+    /* If targetFilter returns true for a target, target is recorded as is.
+     * Otherwise it will be recorded as "other". */
```
DNVindhya, line 340: Can you add this in a new line i.e L341:
If targetFilter is not set, by default target is recorded as is.

A66-otel-stats.md
yashykt: Yes, you use the OTel API, but there is no longer any knob to control which instrument gets created and recorded on.  I think the statement "Any implementation knobs (i.e. views, bounds) set in the passed in object take precedence over the API calls from the interface in this component (i.e. it will create default views for unset views)." is just going to confuse people. The term "API" is pretty loaded in this context, and if we want to explain this we should probably mention the whole thing as stated above in the `Language-Specific Details` section.

yashykt: If we want to mention it, instead of mentioning precedence, we should use the word "customize", same as the OTel documentation at https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view. Let me know if you want to update the text here.

A66-otel-stats.md
```diff
@@ -0,0 +1,566 @@
...
+     *     OpenTelemetry openTelemetry = OpenTelemetrySdk.builder()
+     *         .setMeterProvider(sdkMeterProvider)
+     *         .build();
+     *
+     * If MeterProvider is not configured, no-op meterProvider will be used by default.
+     * It provides meters which do not record or emit.
+     */
+    public OpenTelemetryModuleBuilder openTelemetry(OpenTelemetry openTelemetry);
+
+    /* If targetFilter returns true for a target, target is recorded as is.
```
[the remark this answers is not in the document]
yashykt: done

A66-otel-stats.md
yashykt: done

A66-otel-stats.md
```diff
@@ -0,0 +1,566 @@
...
+    *Attributes*: grpc.method, grpc.status <br>
+    *Type*: Histogram (Size Buckets) <br>
+    *Unit*: `By` <br>
+*   **grpc.server.call.rcvd_total_compressed_message_size** <br>
+    Total bytes (compressed but not encrypted) received across all request messages (metadata excluded) per RPC; does not include grpc or transport framing bytes. <br>
+    *Attributes*: grpc.method, grpc.status <br>
+    *Type*: Histogram (Size Buckets) <br>
+    *Unit*: `By` <br>
+*   **grpc.server.call.duration** <br>
+    This metric aims to measure the end2end time an RPC takes from the server transport’s (HTTP2/ inproc / cronet) perspective. <br>
```
ejona86, line 151: nit: cronet isn't a server transport.

A66-otel-stats.md
```diff
@@ -0,0 +1,566 @@
...
+similarly along with a server call scoped context object, to get
+`ServerCallTracer` equivalent functionality.
+
+### Language-Specific Details
+
+Each language implementation will provide an API for registering an
+OpenTelemetry plugin. Overall, the APIs should have the following capabilities -
+
+*   Allow installing multiple OpenTelemetry plugins.
+*   Implementations must provide an option to set
```
ejona86, line 247: Why this mutate API? Would it be fine (e.g., Java) to just pass it to the constructor and then it's guaranteed to be set? I see that is the case in the Java section below, but that seems counter to this requirement.

A66-otel-stats.md
```diff
@@ -0,0 +1,566 @@
...
+     * If MeterProvider is not configured, no-op meterProvider will be used by default.
+     * It provides meters which do not record or emit.
+     */
+    public OpenTelemetryModuleBuilder openTelemetry(OpenTelemetry openTelemetry);
+
+    /* If targetFilter is set, and returns true for a target, target is recorded as is. Records "other" on false.
+    If targetFilter is not set, target is recorded as is. */
+    public OpenTelemetryBuilder targetFilter(Predicate<String> targetFilter);
+
+    public OpenTelemetryModule build();
```
ejona86, line 343: We won't want simple "build" to install globally (although it's not clear if that is the intention here). I'm fine with fixing up this API afterward, as long as others are. The sketch is fair, but the spellings will probably morph a bit.

(Similarly, I don't think we'll use the method name `openTelemetry()` above.)

ejona86 (approved):

A66-otel-stats.md
yashykt: Yes, that sounds fine. Whether it is a builder API or just passed through the constructor seems like detail. Do you prefer some different wording here?

A66-otel-stats.md
yashykt: sounds fine to me

A66-otel-stats.md
```diff
@@ -0,0 +1,544 @@
...
+
+Propose a metrics data model for gRPC OpenTelemetry metrics.
+
+## Background
+
+There are a collection of
+[metrics](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md)
+proposed by OpenCensus for gRPC. OpenCensus is no longer being actively
+maintained and is being
+[deprecated](https://opentelemetry.io/blog/2023/sunsetting-opencensus/#:~:text=Compatibility%20specification%204.-,What%20to%20Expect%20After%20July%2031st%2C%202023,found%20will%20not%20be%20patched.),
```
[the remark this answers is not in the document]
yashykt: done. (link was fixed as well)

A66-otel-stats.md
yashykt: removed

A66-otel-stats.md
yashykt: done

A66-otel-stats.md
dfawley: What is accepted by the library and what we consider "canonical" are two different things.  Someone should make a gRFC for "canonical target URI strings" I guess

A66-otel-stats.md
dfawley: So then "Optionally allow enabling/disabling _individual_ metrics" please

A66-otel-stats.md
markdroth: One thing to keep in mind here is that when we later add tracing and logging support, it is conceivably possible that a user may want to enable tracing or logging without enabling metrics.  In that case, the API should not force them to specify a `MeterProvider`.

A66-otel-stats.md
markdroth: I'm not sure what the offline discussion was here, but I think Zach is right: This gRFC does describe more than just the metric data model, and I think this line needs to be updated to reflect that.

A66-otel-stats.md
```diff
@@ -0,0 +1,566 @@
...
+
+### Related Proposals:
+
+*   [A6: gRPC Retry Design](A6-client-retries.md)
+*   [A39: xDS HTTP Filter Support](A39-xds-http-filters.md)
+*   [A45: Exposing OpenCensus Metrics and Tracing for gRPC retry](A45-retry-stats.md)
+
+## Proposal
+
+### Metrics Schema
```
markdroth, line 30: I think it might make sense to move this section down to line 412, right before the "Migration from OpenCensus" section.  Readers will probably care more about the APIs and architecture than about what metrics are being provided.

markdroth (approved):
This looks great!

A66-otel-stats.md
```diff
@@ -0,0 +1,566 @@
...
+    default be reported with "other" value instead. Implementations should
+    provide the option to override this behavior to allow recording generic
+    method names as well.
+*   `grpc.status` : gRPC server status code received, e.g. "OK", "CANCELLED",
+    "DEADLINE_EXCEEDED".
+    [(Full list)](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)
+*   `grpc.target` : Canonicalized target URI used when creating gRPC Channel,
+    e.g. "dns:///pubsub.googleapis.com:443", "xds:///helloworld-gke:8000".
+    Canonicalized target URI is the form with the scheme included if the user
+    didn't mention the scheme ([scheme]:///[target]). For channels such as
```
dfawley, line 90: I think this should be `scheme://[authority]/path`?  No brackets on scheme or path as those are required.

A66-otel-stats.md
yashykt: Java takes in an `OpenTelemetry` object on which you, I believe, you can choose whether to set MeterProvider or not, so yeah, both constructor and builder patterns work.

A66-otel-stats.md
yashykt: Done

A66-otel-stats.md
yashykt: moved

A66-otel-stats.md
yashykt: Offline, we discussed that it was fine as it is, and there weren't any strong opinions, so added it now.

A66-otel-stats.md
yashykt: Yeah, I think that makes sense, or maybe this gRFC will be setting "precedence" for what canonical means

A66-otel-stats.md
yashykt: We removed this since we want it to be barebones for now, we'll consider it in a subsequent gRFC

A66-otel-stats.md
DNVindhya: I avoided using global in the build method, because [`buildAndRegisterGlobal`](https://javadoc.io/doc/io.opentelemetry/opentelemetry-sdk/latest/io/opentelemetry/sdk/OpenTelemetrySdkBuilder.html#buildAndRegisterGlobal()) has different meaning wrt OpenTelemetry java. In OpenTelemetry it also sets the sdk instance invoking the instance as Global instance which I didn't want to inferred here as well. 

I named method as `openTelemetry()` since we are taking an openTelemetry instance. How does `openTelemetrySdk()` or `sdk()` sound?

A66-otel-stats.md
dfawley: Even better!

dfawley (approved):

yashykt:
Thanks for all the reviews!

A66-otel-stats.md
ejona86: Oh, this is a single unified API for the various OTel components. Gotcha. I didn't piece that together.

I think the way earlier discussion of "don't stabilize the API until Metric and Tracing are both implemented" applies. But this does seem fair for now.

kingcb11:
> LGTM modulo precedence nit

PR #381, opened 2023-07-21
xDS dynamic parameters

allanrbo:


A67-xds-dynamic-parameters.md
```diff
@@ -0,0 +1,246 @@
...
+  foreach resource_names, dynamic_parameters in authority_resource_names
+    if dynamic_parameters is empty
+      request.resource_names = resource_names
+    else
+      foreach name in resource_names
+        request.resource_locators.add(name, dynamic_parameters)
+  ...
+```
+
+### Temporary environment variable protection
```
allanrbo, line 206: @markdroth You mentioned out of band that you think we do need a flag, so I changed it to say so. (This actually almost takes away the reason for not being able to reuse the metadata.)

You also mentioned that we will need an interop test. Should I mention this, or is that just a given, perhaps? Grep'ing for "interop" in the proposals repo, it doesn't seem like this is something that is usually explicitly mentioned.

A67-xds-dynamic-parameters.md
```diff
@@ -0,0 +1,246 @@
...
+      [type.googleapis.com/envoy.config.listener.v3.Listener] {
+        name: "xdstp://xds.example.com/envoy.config.listener.v3.Listener/foo"
+        ...
+      }
+    }
+  }
+}
+```
+
+### Related Proposals:
```
markdroth, line 103: It's probably also relevant to add a link to [gRFC A47: xDS Federation](https://github.com/grpc/proposal/blob/master/A47-xds-federation.md), which describes the common structure of the XdsClient.

A67-xds-dynamic-parameters.md
markdroth: Yeah, we definitely want env var protection here.

I don't think this actually eliminates the reason for not using the metadata.  In practice, I think we're going to want to explictly control when users upgrade to dynamic parameters, not just based on whether they're using a version of gRPC that has passed interop tests.

We definitely need an interop test definition for interop with TD, but that's Google-specific, not part of this OSS gRFC, so I don't think it needs to be included in this doc.  Please take that as a separate action item.

A67-xds-dynamic-parameters.md
```diff
@@ -0,0 +1,246 @@
...
+    else
+      foreach name in resource_names
+        request.resource_locators.add(name, dynamic_parameters)
+  ...
+```
+
+### Temporary environment variable protection
+
+During initial development, the GRPC_EXPERIMENTAL_XDS_DYNAMIC_PARAMETERS
+environment variable will control whether ADS requests will use the
```
markdroth, line 209: The overall goal of the env var is to ensure that until the feature is tested and considered stable, we don't want to enable any new functionality that can be triggered via something out of gRPC's control.  And we need to be very precise here about what parts of the code are guarded by the env var, because we've had bugs in the past where we failed to guard things, and we really don't want to repeat that mistake.

Note that the goal here is *not* to catch problems where the bootstrap config is updated to use dynamic parameters but the xDS server doesn't support them.  I would not actually expect that case to happen in practice, because whoever maintains the bootstrap file should not configure it to use a feature with a server that doesn't support that feature.  And if they did do that, there's really nothing we can do to avoid it.

Also, note that this document is for OSS gRPC, but the bootstrap generator you're linking to here is TD-specific.  It's fine to cite that as an example, but I think we should describe the problem more generally: it happens in cases where the bootstrap file is generated by a separate tool with an independent release process.

So, putting all of this together, I suggest rewording this paragraph as follows:

"""
During development, the dynamic parameter functionality will be guarded by the GRPC_EXPERIMENTAL_XDS_DYNAMIC_PARAMETERS env var.  Specifically, this env var will control two things:
- Whether we read the new `dynamic_parameters` fields in the bootstrap config (and therefore whether we populate the `resource_locators` field in ADS requests).  This guards against cases where the bootstrap file is generated by a separate tool with a release process independent of that of the application (e.g., the [Traffic Director bootstrap generator](https://github.com/GoogleCloudPlatform/traffic-director-grpc-bootstrap)), such that a newer bootstrap config may be used with an older version of gRPC.
- Whether we look at the new `resource_names` field in ADS responses.  This guards against xDS servers that may incorrectly start sending new fields to clients that are not yet ready to understand them.

The env var protection will be removed once this feature passes interop tests.
"""

A67-xds-dynamic-parameters.md
```diff
@@ -0,0 +1,246 @@
...
+
+### No changas to dynamic parameters after startup
+
+For our initial support of dynamic parameters, we can let the parameter
+key-value pairs be only read at startup. A client restart will be required when
+changing the parameters.
+
+### No metadata reuse
+
+An idea in TP2 is to reuse the node metadata field as dynamic parameters. I
```
markdroth, line 228: I think there are a few key points to call out here:
- We need an explicit knob in the bootstrap config that tells us whether or not to use dynamic parameters, and using the presence of the `dynamic_parameters` field gives us that.  If we tried to automatically convert node metadata into dynamic parameters, we'd need a separate field to control whether dynamic parameters should be used.
- The suggestion in TP2 was really intended as a possible migration path for Envoy, where basically everyone needs to use node metadata because of the way that Envoy does wildcard LDS and CDS subscriptions, where the server uses node metadata to decide what resources to send.  In contrast, node metadata is not as important in the gRPC xDS ecosystem, so this migration path is not considered important.
- Converting node metadata to dynamic parameters would be somewhat complex to implement, since node metadata can have structured values, whereas dynamic parameter values are always just strings.

A67-xds-dynamic-parameters.md
```diff
@@ -0,0 +1,246 @@
+## xDS Dynamic Parameters
+
+*   Author(s): @allanrbo
+*   Approver: @markdroth
+*   Status: Draft
+*   Implemented in:
+*   Last updated: 2023-07-21
+*   Discussion at:
```
markdroth, line 8: Please go ahead and create the mailing list thread, and then add a link to it here, as per the gRFC process.

markdroth (commented):
Sorry for the delay on this!  The doc looks good overall, comments are relatively minor.

A67-xds-dynamic-parameters.md
allanrbo: Added.

A67-xds-dynamic-parameters.md
allanrbo: Thanks for the rewording suggestion. Applied.

A67-xds-dynamic-parameters.md
allanrbo: Rewrote this section to include these key points.

A67-xds-dynamic-parameters.md
markdroth: Still need to take care of this.

A67-xds-dynamic-parameters.md
```diff
@@ -104,6 +104,9 @@ resources {
 
 *   CNCF xDS proposal
     [TP2: Dynamically Generated Cacheable xDS Resources](https://github.com/cncf/xds/blob/main/proposals/TP2-dynamically-generated-cacheable-xds-resources.md).
+*   gRPC proposal[gRFC A47: xDS Federation](https://github.com/grpc/proposal/blob/master/A47-xds-federation.md).
```
markdroth, line 107: I think we can remove the words "gRPC proposal".  Anyone reading this already knows what a gRFC is. :)

markdroth (commented):
This looks really good!

@ejona86 @dfawley, I'd like reviews from you as well.

A67-xds-dynamic-parameters.md
```diff
@@ -0,0 +1,266 @@
...
+
+  // Resources for authority "zzz.example.com"
+  { resource_names: ["yyy"], dynamic_parameters: {"env": "prod", "version": "v2"} },
+]
+
+function ConstructADSRequest(authority_resource_names)
+  ...
+  foreach resource_names, dynamic_parameters in authority_resource_names
+    if dynamic_parameters is empty
+      request.resource_names = resource_names
```
ejona86, line 202: Doesn't this overwrite all but the last authority's resource names? Is this supposed to be add() as well like below? You either have to do two passes or mix both `resource_names` and `resource_locators` in a single request. It really isn't clear which way is intended here (based on the text above I thought only one of the two fields would have any entries, but the add() below makes me think maybe that wasn't the intention).

A67-xds-dynamic-parameters.md
```diff
@@ -0,0 +1,266 @@
...
+complex to implement, since node metadata can have structured values, whereas
+dynamic parameter values are always just strings.
+
+Additionally, an explicit knob in the bootstrap config that determines whether
+to use dynamic parameters is desirable for control. Using the presence of the
+`dynamic_parameters` field gives us that.
+
+### No merging of top-level and per-authority dynamic parmeters
+
+The `dynamic_parameters` at root level only apply to the non-xdstp case. There
```
ejona86, line 256: That is not clear in the proposal text. It just says "To support non-xdstp use cases, we can add a map at the root level too". That doesn't require that it is used exclusively in non-xdstp.

I do wonder if there's something we can do here to avoid confusion where someone thinks they are configuring the dynamic parameters for xdstp, but actually aren't. However, I see the value in using the same dynamic_parameters name, just like we are doing with xds_servers. Although for xds_servers, if you don't specify it in the authorities section the top-level xds_servers is used. Maybe we should do that; it becomes the default, but there's still no merging. That's probably the least surprising. Although it does cause problems for disabling the dynamic parameters within authorities, because empty list is semantically the same as not present.

ejona86 (commented):
This seems fine. Some comments to work out, but I'm not all that concerned how they are resolved.

PR #382, opened 2023-07-25
A69: Certificate Revocation List Enhancements

gtcooke94:


AXX-crl-enhancements.md
```diff
@@ -0,0 +1,220 @@
...
+* Author(s): Gregory Cooke (@gtcooke94, @gregorycooke)
+* Approver: TODO 
+* Status: TODO
+* Implemented in: C, Go
+* Last updated: 2023-07-25
+* Discussion at: TODO
+
+## Abstract
+
+gRPC has an MVP implementation for CRLs (Certificate Revocation Lists). We wish to provide a more generic interface. In addition, we will provide two concrete implementations of this interface similar to Credential Reloading - a static provider and a directory reloader.
```
erm-g, line 12: I'd say 'more generic and flexible to enable developers to implement desired business functionality fitting their landscape'

erm-g: Let's also add a link to Credentials Reloading docs/gRFCs

AXX-crl-enhancements.md
```diff
@@ -0,0 +1,220 @@
...
+
+
+## Background
+
+[RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280) defines the desired behavior for Certificate Revocation Lists (CRLs). gRPC has an MVP implementation of this RFC for CRL handling in several languages, each with their own flavors - for example, C++ uses OpenSSL, whereas Go uses an implementation written in Go (and the stack in Java is significatly different). There are mismatches and deficiencies between the current Go and C++ implementations. Providing CRL reloading as a feature allows for more internal cohesion with the ideas behind credential reloading.
```
erm-g, line 17: I'd drop 'There are mismatches and deficiencies between the current Go and C++ implementations.' statement as the next one looks sufficient for me

AXX-crl-enhancements.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Related Proposals:
+
+* [L46: New TLS Credentials API](https://github.com/grpc/proposal/pull/205)
+
+## Proposal
+We seek to provide an alternate option for users to customize CRL handling. In a similar way that `CredentialReloading` is implemented, we will add a `CrlProvider`` interface through which gRPC fetches CRLs. In gRPC we will implement two providers (these match what we provide for certificates as well):
+
+1. A `StaticCrlProvider` that takes a list of strings that represent CRL file content
+1. A `CrlReloaderProvider` that periodically and asynchronously reloads a directory of CRLs.
```
erm-g, line 27: I'd add 'local directory' or 'FS directory' for clarification

AXX-crl-enhancements.md
```diff
@@ -0,0 +1,220 @@
...
+
+Users can also implement the interface with whatever behaviors they specifically desire.
+
+Per RFC5280, during revocation checking one of three status should be returned - `RevocationUnrevoked`, `RevocationRevoked`, and `RevocationUndetermined`. The user must specify whether `RevocationUndetermined` should be treated as `RevocationRevoked` or `RevocationUnrevoked` (failing open vs. failing closed).
```
erm-g, line 31: I'd add links here to RFC and statuses

AXX-crl-enhancements.md
```diff
@@ -0,0 +1,220 @@
...
+
+Assumptions:
+1. There should be no direct linkage between this interface and the creation/distribution of CRLs. gRPC is a user of CRLs and credentials, not a PKI itself.
```
erm-g, line 34: I'd add some 'high level overview' section (maybe as a comment to a diagram?) saying that existing impls use map as an in-memory storage and a provider to update it using some IOs. PKIs are completely out of scope of this gRFC

erm-g: Also let's add an assumption that we can't guess CRL provider by a name of a file / metadata of strings. However if someone wants to rely on it they can do a specific impl

AXX-crl-enhancements.md
```diff
@@ -0,0 +1,220 @@
...
+We will also require that users perform safe and atomic file/directory operations with their CRL files using filesystem primitives (for example, the copy-rename pattern). If users do not exercise good file management hygiene, we cannot guarantee any exact behavior. For example, if a user creates a directory and writes files one-by-one, there is a chance the directory reloader would update when only 1 file was present and this would be seen as a success (so this would overall be unspecified behavior because of bad file hygiene).
+If a CRL is not found for an issuer this is fine - the verification code will return RevocationUndetermined per RFC5280, then the user will have a setting to fail-open or fail-closed indicating whether RevocationUndetermined should be treated as Unrevoked or Revoked respectively.
```
erm-g, line 91: at the end I'd add 'if your landscape/PKI requires different business logic, feel free to provide your own implementation of directory reloader'

AXX-crl-enhancements.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### API Outcomes and Error Handling
+Because CRLs involve reading/writing from the filesystem, we will have to deal with potential edge cases of bad files, bad updates, etc. From a high-level perspective, we are going to follow the Authz Policy design patterns - use existing information in case an update contains bad data. Startup behavior is not different from updates. Particularly in the case of updating CRLs, we don't want to error and crash the server if there is a bad update. However, we still want this error to be known, so we will have an optional and overridable error callback when CRLs fail to be read. This will let users tie in whatever alerting/monitoring they may want in these failure cases. This callback will be for notifications/alerting, it will not be a decision-making callback (it is expected the IO for CRLs will already be done in the background, so these errors would not be on the main path by design anyways). Users should use `RevocationUndertermined` combined with the FailOpen/FailClosed knob for decision making on uncertain CRLs.
```
erm-g, line 94: let's add a link to Authz gRFC

AXX-crl-enhancements.md
```diff
@@ -0,0 +1,220 @@
...
+
+This approach adds complexity to the user facing API, and a user overriding the provider interface poorly could result in undefined or inefficient behaviors. To counteract this, we will provide generally useful implementations and good documentation on requirements if a user is writing their own implementation.
+
+We explicitly will not be supporting OSCP-style revocation . We want to stay inside the realm of RFC5280 and X509-style CRLs, and OSCP represents a departure from that standard. Notably, the API surface for OSCP is different - for OSCP one makes a call to an external service that returns the revocation status. Rather, we expect to be given CRLs and will check the revocation status in gRPC. In addition, OSCP requires external calls during the handshake, whereas with X509 CRL files everything can done locally.
```
erm-g, line 113: I'd add that gRPC is optimized for high performance so making external calls is always an approach to avoid

erm-g (changes requested):
looks good, suggested a bunch of changes related to wording / moving statements around

AXX-crl-enhancements.md
gtcooke94: Changed - link is already in the Related Proposals section

AXX-crl-enhancements.md
gtcooke94: Done

AXX-crl-enhancements.md
gtcooke94: Added

AXX-crl-enhancements.md
gtcooke94: Done

AXX-crl-enhancements.md
gtcooke94: Added the second

For the first, this document has a structure defined by a template. I could have a High Level Overview sub-section in the Proposal section.

> existing impls use map as an in-memory storage and a provider to update it using some IOs

I'm a little confused at this comment - the existing CRL impls don't do this?

AXX-crl-enhancements.md
gtcooke94: Done

AXX-crl-enhancements.md
gtcooke94: Done, and added to the related proposals at the top as well

AXX-crl-enhancements.md
gtcooke94: Done

A69-crl-enhancements.md
```diff
@@ -0,0 +1,222 @@
...
+![Basic Flow Diagram](A69_graphics/basic_diagram.png)
+
+CrlProvider Interface Pseudocode:
+```
+public:
+// Get the CRL that is associated with a cert. Provides good semantic
+// consistency as most X509 APIs that try to get a CRL expect to get a cert
+// passed in. In each language we may use an abstraction here that makes it easy
+// to get CRL-lookup information from the certificate.
+// Expected to use as read-only
```
ejona86, line 49: What is expected to be read-only?

A69-crl-enhancements.md
```diff
@@ -0,0 +1,222 @@
...
+  return revocationStatus 
+}
+```
+
+### Provider Implementation - Static Provider
+
+This is a simple provider that takes in all CRLs to be used as raw strings during initialization, then returns them in the appropriate format when called. The CRLs will be stored in a mapping of a hash of the issuer to the CRL content.
+
+### Provider Implementation - Directory Provider
+This provider will periodically read CRL files in a given directory and update gRPCs internal representation of those CRLs. We expect this will be heavily used, as a directory of CRLs is very common for X509 CRL files. The CRLs will be stored in a mapping of a hash of the issuer to the CRL content - `map<issuer_hash, crl_object>`.
```
ejona86, line 89: Are these supposed to be multiple paragraphs or a single paragraph? Right now it is a single paragraph.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,222 @@
...
+
+These cases are all on a per-CRL level - a given CRL having an issue will not keep other good CRLs in an update batch to fail. The opposite behavior - failing the whole batch or dropping all the handshakes in the case of a bad CRL - might enable a DDOS-type attack vector.
```
ejona86, line 98: nit: It'd just be a DoS, not a DDoS.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,222 @@
...
+type CRLProvider interface {
+	// Callers are expected to use the returned value as read-only.
+	CRL(cert x509.Certificate) (*CertificateListExt, error)
+
+	// Close cleans up resources allocated by the Provider.
+	Close()
+}
+```
+
+Regarding sharing between connections - since the providers are in a high-level configuration, the same provider should be used across many individual connections and handshakes. [Here is an example of how the credential reloading provider is configured](https://github.com/grpc/grpc-go/blob/master/security/advancedtls/examples/credential_reloading_from_files/server/main.go#L63). In this example, the server would use this provider for all connections.
```
ejona86, line 151: This needs to be a permalink as with time it won't be clear what you are referring to. On the left of the line in github are three dots, click those dots and "copy permalink" and use that URL here.

ejona86 (approved):
A few more comments on top of @erm-g's. But looks good.

@markdroth, @dfawley, can y'all take a look?

A69-crl-enhancements.md
erm-g, line 89: @gtcooke94 WDYT about adding as a second statement in this paragraph `Initial load of the directory is performed when the Directory Provider is created.`?

A69-crl-enhancements.md
```diff
@@ -0,0 +1,222 @@
...
+
+#### Directory Reloader Specifics
+![Directory Reloader Flow](A69_graphics/golang_reloader.png)
+```go
+type CrlDirectoryReloadingProvider struct {
+	ReloadInterval time.Duration
+	CrlDirectory string
```
erm-g, line 158: The resulting struct we implemented looks a little different (for example, most of the params are wrapped in Options struct), but I think this sample represents the overall idea quite good so I don't think we need to change it. @gtcooke94 WDYT?

A69-crl-enhancements.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+The core CrlProvider interface will look as follows. Since x509 is a built in go library, we can safely use it in our API.
+```go
+type CRLProvider interface {
+	// Callers are expected to use the returned value as read-only.
+	CRL(cert x509.Certificate) (*CertificateListExt, error)
+
+	// Close cleans up resources allocated by the Provider.
+	Close()
```
erm-g, line 147: @gtcooke94  We don't have `Close` at interface level (I don't think it's relevant for StaticProvider for example) - are you ok to drop it from here?

rockspore: I'm actually curious about this. If we don't have this at the interface level, how does the consumer of the CRLProvider call this func? I don't think casting the interface back to a specific implementation is a good idea. Did I miss anything?

erm-g: @rockspore IMHO this function is relevant only for some of the CRLProviders but not to all of them (for example, StaticProvider, or a provider which reads a directory only once).
However, there is another approach - keep `Close()` on interface level but use a no-op impl if it's not relevant. Such pattern is used in gRPC Go [repo](https://github.com/grpc/grpc-go/blob/70f1a4045da95b93f73b6dbdd7049f3f053c0680/balancer/balancer.go#L388)

rockspore: This pattern is not a nice-to-have but a must. The consumer will store an instance of the interface and has no knowledge about the specific implementation. If `Close` is not defined at the interface level, it won't be able to call it.

erm-g: My point is that not all the needed methods will be present at the interface level. Let's say a consumer develops a `CRLDirectoryFileMaskProvider` with an extra `UpdateFileMaskAndScan` method, which accepts file extensions to be included in the next directory scans. This method won't be part of the interface, so the consumer will need to store the instance (or to work with casting).

rockspore: I think it all depends on who calls the `Close()` and I was thinking this would be handled during the server shutdown. So gRPC needs to be able to stop whatever CRL provider it's given. But if the intent is for gRPC users that created the provider to recycle it, then it probably doesn't matter.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,222 @@
...
+
+[RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280) defines the desired behavior for Certificate Revocation Lists (CRLs). gRPC has an MVP implementation of this RFC for CRL handling in several languages, each with their own flavors - for example, C++ uses OpenSSL, whereas Go uses an implementation written in Go (and the stack in Java is significatly different). Providing CRL reloading as a feature allows for more internal cohesion with the ideas behind credential reloading.
+
+### Related Proposals:
+
+* [L46: New TLS Credentials API](https://github.com/grpc/proposal/pull/205)
+* [A43: gRPC authorization API](https://github.com/grpc/proposal/blob/master/A43-grpc-authorization-api.md)
+
+## Proposal
+We seek to provide an alternate option for users to customize CRL handling. In a similar way that `CredentialReloading` is implemented, we will add a `CrlProvider`` interface through which gRPC fetches CRLs. In gRPC we will implement two providers (these match what we provide for certificates as well):
```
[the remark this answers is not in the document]
gtcooke94: Fixed

A69-crl-enhancements.md
```diff
@@ -0,0 +1,222 @@
...
+
+#### API Outcomes and Error Handling
+Because CRLs involve reading/writing from the filesystem, we will have to deal with potential edge cases of bad files, bad updates, etc. From a high-level perspective, we are going to follow the [Authz Policy design patterns](https://github.com/grpc/proposal/blob/master/A43-grpc-authorization-api.md) - use existing information in case an update contains bad data. Startup behavior is not different from updates. Particularly in the case of updating CRLs, we don't want to error and crash the server if there is a bad update. However, we still want this error to be known, so we will have an optional and overridable error callback when CRLs fail to be read. This will let users tie in whatever alerting/monitoring they may want in these failure cases. This callback will be for notifications/alerting, it will not be a decision-making callback (it is expected the IO for CRLs will already be done in the background, so these errors would not be on the main path by design anyways). Users should use `RevocationUndertermined` combined with the FailOpen/FailClosed knob for decision making on uncertain CRLs.
```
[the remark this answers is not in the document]
gtcooke94: Added

A69-crl-enhancements.md
gtcooke94: Added to clarify that this is about the return value

A69-crl-enhancements.md
gtcooke94: Added newlines to make them separate paragraphs, thanks for the catch

A69-crl-enhancements.md
gtcooke94: Changed

A69-crl-enhancements.md
gtcooke94: Fixed

A69-crl-enhancements.md
gtcooke94: Added

A69-crl-enhancements.md
gtcooke94: I think it's okay, this is meant to be a design idea, not precise and exact code

A69-crl-enhancements.md
gtcooke94: We decided to remove this after discussing it in the Golang PR

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+
+Regarding sharing between connections - since the providers are in a high-level configuration, the same provider should be used across many individual connections and handshakes. [Here is an example of how the credential reloading provider is configured](https://github.com/grpc/grpc-go/blob/7935c4f759419cb0ab4cc50070f5487b33a19bb7/security/advancedtls/examples/credential_reloading_from_files/server/main.go#L58-L63). In this example, the server would use this provider for all connections.
+
+#### Directory Reloader Specifics
+![Directory Reloader Flow](A69_graphics/golang_reloader.png)
+```go
+type CrlDirectoryReloadingProvider struct {
+	ReloadInterval time.Duration
+	CrlDirectory string
+     // maps issuer hash to crl
```
ejona86, line 163: nit: Tabs vs spaces. Make this a tab?

ejona86 (approved):

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+
+Assumptions:
+1. There should be no direct linkage between this interface and the creation/distribution of CRLs. gRPC is a user of CRLs and credentials, not a PKI itself.
+1. The CRL Providers simply provide CRLs. The actual verification will be done by other pieces of the code (these are already written).
+1. We cannot know users' PKI details, so our implementations should not enforce non-X509 metadata mattering (for example, file names). A user could certainly implement their own provider that cares about filenames.
+1. We will not support OSCP-style checking (OSCP is an alternative to CRLs, not another form of CRLs).
+
+![Basic Flow Diagram](A69_graphics/basic_diagram.png)
```
markdroth, line 40: Please also include the source form (e.g., SVG files) for these diagrams, so that we can modify them in future if needed.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+  return revocationStatus 
+}
+```
+
+### Provider Implementation - Static Provider
+
+This is a simple provider that takes in all CRLs to be used as raw strings during initialization, then returns them in the appropriate format when called. The CRLs will be stored in a mapping of a hash of the issuer to the CRL content.
+
+### Provider Implementation - Directory Provider
+This provider will periodically read CRL files in a given directory and update gRPCs internal representation of those CRLs. We expect this will be heavily used, as a directory of CRLs is very common for X509 CRL files. The CRLs will be stored in a mapping of a hash of the issuer to the CRL content - `map<issuer_hash, crl_object>`. The initial read of the directory and filling of this map will be synchronous during server startup when the provider is created.
```
markdroth, line 89: It's probably worth explicitly stating that the reason that the initial read is synchronous is to ensure that the CRL data is already loaded before the server starts up, so that we don't delay handshakes while waiting for the CRLs to load.  We should explicitly state that the creation does *not* fail if there is an error in this initial load.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+// to get CRL-lookup information from the certificate.
+// Return value expected to be as read-only 
+// Returns null if there isn't an associated CRL (as null is different than an
+// empty CRL)
+  CRL Crl(certficate/certificateInfo)
+private:
+  // Implementation for getting and storing CRLs
+```
+
+Example feature use:
```
markdroth, line 57: Please extend this example to show setting the fail open/closed knob, so that it's clear that this knob is intended to be part of the TlsCreds API, not part of a specific CRL provider implementation.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+### Map Content and API Outcomes
+
+The diagram below describes the behavior of the API and why this or that outcome is produced. Here, "Map" is a generic way to represent some in-memory representation of the CRLs in a key-value store. “CRL Provider API Return” refers to the CRL return of the `CrlProvider` in the Pseudocode above.
+![Basic API Outcomes](A69_graphics/basic_table.png)
+
+The pseudo code snippet below illustrates the same idea:
+```
+<pseudo code in the handshaker when verifying cert>
+
+RevocationStatus GetRevocationStatus(cert, crlProvider, settings) {
```
markdroth, line 74: Suggest changing this pseudo-code as follows:
- Rename `settings` to `opts`, to clarify that this is the same options structure created in the previous pseudo-code block.
- Instead of passing in `crlProvider` as a separate parameter, use `opts.provider`, since that's where we set the provider in the previous pseudo-code block.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+
+In the case that the directory and all files within are read completely during an update, we will treat the directory as the exact truth. This means if a CRL was in-memory and is no longer in the directory after the update, it will be removed in-memory.
+
+In the case that there is an issue reading a file, we cannot be certain what exactly is happening. In particular, we enforce no naming convention on the CRL files, so if it cannot be read, we cannot know what it contains or what issuer it belongs to. We will do a best-effort safe update - for files which are read correctly, we can update those individual entries. Since we can't know the issuer if a file can't be read, it is unsafe to do deletion in this case. An overridable error callback will be called so the user can receive a signal that something has gone wrong with CRLs.
```
markdroth, line 93: Please explicitly state that this error callback will be passed as a parameter when instantiating the directory-based provider, so that it's clear that this is not part of the TlsCreds API itself.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+
+These cases are all on a per-CRL level - a given CRL having an issue will not keep other good CRLs in an update batch to fail. The opposite behavior - failing the whole batch or dropping all the handshakes in the case of a bad CRL - might enable a DOS-type attack vector.
+
+Example behavior for a directory reloader:
```
markdroth, line 107: This line should probably be a section header (i.e., prefix with `####`).

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+
+At startup we read 3 CRL files (A, B, C) and construct a map as described above. During the handshakes, we return the CRLs from the map.
+Then, a bad update happens and CRL C ends up corrupted, but A and B are good. When this directory is reloaded, the resulting map will have the newer A and B CRL, but the older C CRL (as this will not see any update). In addition, the CrlReadErrorCallback would be invoked when CRL C fails to be read. There is no linkage between the map's CRL C and the fact that there was a read failure. It is the user's responsibility to act on file issues after being notified via the error callback.
```
markdroth, line 110: The text "but the older C CRL (as this will not see any update)" is confusing.  Suggest changing that to something like "but it will also include the original C CRL, since we do not delete entries from the map when there is a failure reading any of the files in the directory".

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+
+We will also require that users perform safe and atomic file/directory operations with their CRL files using filesystem primitives (for example, the copy-rename pattern). If users do not exercise good file management hygiene, we cannot guarantee any exact behavior. For example, if a user creates a directory and writes files one-by-one, there is a chance the directory reloader would update when only 1 file was present and this would be seen as a success (so this would overall be unspecified behavior because of bad file hygiene). If a user has needs that don't match this, they can implement the interface per their needs.
+
+If a CRL is not found for an issuer this is fine - the verification code will return RevocationUndetermined per RFC5280, then the user will have a setting to fail-open or fail-closed indicating whether RevocationUndetermined should be treated as Unrevoked or Revoked respectively.
+
+
+#### API Outcomes and Error Handling
```
markdroth, line 100: We need to specify the behavior of this callback a little more precisely:
- The callback will be invoked at most once per update pass.  If there are errors reading more than one file in a given pass, all of those errors will be returned in a single invocation of the callback.
- The callback will be passed only one parameter, which is an indication of the error in a language-specific way (e.g., `absl::Status` in C++).  The error text should include the list of files for which there were errors and what those errors were, but that information will not be passed in a structured way for the callback to understand those individual underlying errors.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+
+This approach adds complexity to the user facing API, and a user overriding the provider interface poorly could result in undefined or inefficient behaviors. To counteract this, we will provide generally useful implementations and good documentation on requirements if a user is writing their own implementation.
+
+We explicitly will not be supporting OSCP-style revocation . We want to stay inside the realm of RFC5280 and X509-style CRLs, and OSCP represents a departure from that standard. Notably, the API surface for OSCP is different - for OSCP one makes a call to an external service that returns the revocation status. Rather, we expect to be given CRLs and will check the revocation status in gRPC. In addition, OSCP requires external calls during the handshake, whereas with X509 CRL files everything can done locally. Handshakes are on a path for which performance in critical, so making external calls during them is an approach to avoid.
```
markdroth, line 122: In "performance in critical", change "in" to "is".

Also, I would say "important" instead of "critical" -- we do care about it, but handshakes are not on the per-call path, so their performance is not truly what I would consider critical.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+
+> "sets the function to get the crl for a given certificate x. When found, the crl must be assigned to *crl. This function must return 0 on failure and 1 on success. If no function to get the issuer is provided, the internal default function will be used instead."
+
+ We will set this function to get the CRL from the CRLProvider if it is configured. We will add a `grpc::experimental::CrlProvider` that can be configured in `grpc::experimental::TlsChannelCredentialOptions`. This value and `set_crl_directory` cannot both be configured. The Provider will be configured in these options by the user and supplied as a pointer to grpc. Its lifetime needs to be managed by the user, and any handshake happening with a provider configured should share this provider, so the user can configure this to be shared across many connections if desired.
```
markdroth, line 182: The statement "Its lifetime needs to be managed by the user" is inconsistent with the fact that we're passing it in as a `shared_ptr<>`.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,226 @@
...
+func (provider *CrlDirectoryReloadingProvider) run(ctx context.Context) {
+	// golang ticker that refreshes the Crls
+}
+```
+
+Reference Code:
+* [IdentityCertificateOptions](https://github.com/grpc/grpc-go/blob/2aa261560586eab6795301a3670d9dfdd7308625/security/advancedtls/advancedtls.go#L115)
+* [certificateListExt](https://github.com/grpc/grpc-go/blob/2aa261560586eab6795301a3670d9dfdd7308625/security/advancedtls/crl.go#L88)
+
+### C/C++
```
markdroth, line 177: I think this section is probably a little *too* detailed.  I think the main things we need to cover here are:
- APIs that we're adding, both at the C++ and C-core layers.
- A brief description of where the functionality will be implemented (i.e., the provider and the fail open/closed option will be passed down from TlsCreds to the security connector and from there to the TSI layer, where they will be used when interacting with the SSL library).

I don't think we need to cover the deeper implementation details in this doc.

markdroth (commented):
The overall design looks good!  My comments here are mostly just details that need to be spelled out a little more clearly.

A69-crl-enhancements.md
gtcooke94: Removed that statement

A69-crl-enhancements.md
gtcooke94: Simplified implementation specifics significantly, and added text about us doing the CRL checks relating to RFC 5280 6.3.3

A69-crl-enhancements.md
```diff
@@ -99,15 +100,18 @@ If a CRL is not found for an issuer this is fine - the verification code will re
...
 
 These cases are all on a per-CRL level - a given CRL having an issue will not keep other good CRLs in an update batch to fail. The opposite behavior - failing the whole batch or dropping all the handshakes in the case of a bad CRL - might enable a DOS-type attack vector.
 
-Example behavior for a directory reloader:
+### Example Behavior for a Directory Reloader:
```
markdroth, line 111: This should have one more `#` character, so that it's underneath the "Provider Implementation - Directory Provider" heading.

A69-crl-enhancements.md
```diff
@@ -195,32 +194,28 @@ class CrlProvider {
...
+
+These will be passed down from the `TlsCreds` layer to the security connector layer, and from there to the TSI and the ssl_transport_security layer which actually constructs the handshakers and interacts with BoringSSL.
+
+The footprint of the TlsCredentialProvider is more complex than needed for the scope of our CRL work, largely because we are supporting X509 CRL files and not something that needs to make calls such as OSCP. Thus, we can look at the footprint of the custom tls CertificateVerifier to inform how these user-facing options cascade through the code - the provider will have an API that takes certificate information and returns the CRL content. There is already plumbing in these spots for the `crl_directory`, so it shouldn't be too difficult to plumb this around.
```
markdroth, line 201: I suggest removing this paragraph, since it doesn't really seem to add any useful information from the perspective of this document.

A69-crl-enhancements.md
```diff
@@ -195,32 +194,28 @@ class CrlProvider {
...
+
+Due to constraints from supporting both OpenSSL and BoringSSL, our best option for doing the revocation checks is to treat it as post-chain building validation. We will tie into BoringSSL via the `SSL_CTX_set_cert_verify_callback`. In that callback, we will iterate over the certificate chain that has been built, fetch the CRLs for those certificates from the provider, and do the associated checks per [RFC5280 6.3.3](https://datatracker.ietf.org/doc/html/rfc5280#section-6.3.3). Specifically, gRPC will do the following:
```
markdroth, line 203: Suggest changing "Specifically, gRPC will do the following:" to "Specifically, here's how gRPC will handle each of the points in that section of the RFC:".

markdroth (approved):
This looks good!  Remaining comments are just minor wording suggestions.

A69-crl-enhancements.md
gtcooke94: Done

A69-crl-enhancements.md
gtcooke94: Done

A69-crl-enhancements.md
gtcooke94: Done

A69-crl-enhancements.md
```diff
@@ -0,0 +1,219 @@
...
+
+1. A `StaticCrlProvider` that takes a list of strings that represent CRL file content
+1. A `CrlReloaderProvider` that periodically and asynchronously reloads a local directory of CRLs.
+
+Users can also implement the interface with whatever behaviors they specifically desire.
+
+Following [RFC5280](https://datatracker.ietf.org/doc/html/rfc5280), during revocation checking one of three status should be returned - `RevocationUnrevoked`, `RevocationRevoked`, and `RevocationUndetermined` ([RFC5280 defines `Undetermined` in the CRL Processing section](https://datatracker.ietf.org/doc/html/rfc5280#section-6.3.3). There are many reasons a certificate can be revoked, but in the end gRPC cares about a certificate being revoked or not, thus we differentiate between `RevocationUnrevoked` and `RevocationRevoked`. The user must specify whether `RevocationUndetermined` should be treated as `RevocationRevoked` or `RevocationUnrevoked` (failing open vs. failing closed).
```
dfawley, line 32:
```suggestion
Following [RFC5280](https://datatracker.ietf.org/doc/html/rfc5280), during revocation checking one of three status should be returned - `RevocationUnrevoked`, `RevocationRevoked`, and `RevocationUndetermined` ([RFC5280 defines `Undetermined` in the CRL Processing section](https://datatracker.ietf.org/doc/html/rfc5280#section-6.3.3)). There are many reasons a certificate can be revoked, but in the end gRPC cares about a certificate being revoked or not, thus we differentiate between `RevocationUnrevoked` and `RevocationRevoked`. The user must specify whether `RevocationUndetermined` should be treated as `RevocationRevoked` or `RevocationUnrevoked` (failing open vs. failing closed).
```

A69-crl-enhancements.md
```diff
@@ -0,0 +1,219 @@
...
+	// Directory format must match OpenSSL X509_LOOKUP_hash_dir(3).
+	RootDir string
+	// AllowUndetermined controls if certificate chains with RevocationUndetermined
+	// revocation status are allowed to complete.
+	AllowUndetermined bool
+	// Cache will store CRL files if not nil, otherwise files are reloaded for every lookup.
+	Cache Cache
+
+  // crlProvider and RootDir/Cache cannot both be set
++ CrlProvider crlProvider
```
dfawley, line 147:
```suggestion
+	// crlProvider and RootDir/Cache cannot both be set
+	CRLProvider crlProvider
```

A69-crl-enhancements.md
```diff
@@ -0,0 +1,219 @@
...
+type CRLProvider interface {
+	// Callers are expected to use the returned value as read-only.
+	CRL(cert x509.Certificate) (*CertificateListExt, error)
+}
+```
+
+Regarding sharing between connections - since the providers are in a high-level configuration, the same provider should be used across many individual connections and handshakes. [Here is an example of how the credential reloading provider is configured](https://github.com/grpc/grpc-go/blob/7935c4f759419cb0ab4cc50070f5487b33a19bb7/security/advancedtls/examples/credential_reloading_from_files/server/main.go#L58-L63). In this example, the server would use this provider for all connections.
+
+#### Directory Reloader Specifics
+![Directory Reloader Flow](A69_graphics/golang_reloader.png)
```
dfawley, line 162: There's similar capitalization issues in the image, but that's fine.

A69-crl-enhancements.md
```diff
@@ -0,0 +1,219 @@
...
+  CrlDirectory string
+  // maps issuer hash to crl
+	crls map[string]*CertificateListExt
+}
+func (provider *CrlDirectoryReloadingProvider) Crl(x509.Certificate) (*CertificateListExt, error) {...}
+
+func (provider *CrlDirectoryReloadingProvider) run(ctx context.Context) {
+	// golang ticker that refreshes the Crls
+}
+```
```
dfawley, line 175:
```suggestion
```go
type CRLDirectoryReloadingProvider struct {
	ReloadInterval time.Duration
	CRLDirectory string
	// maps issuer hash to crl
	crls map[string]*CertificateListExt
}

func (provider *CRLDirectoryReloadingProvider) CRL(x509.Certificate) (*CertificateListExt, error) {...}

func (provider *CRLDirectoryReloadingProvider) run(ctx context.Context) {
	// golang ticker that refreshes the CRLs
}
```
```

dfawley (approved):
Some formatting things otherwise LGTM!

PR #387, opened 2023-08-17
Update A48: remove reference to Go's base balancer

dfawley:
Implementing this using the Go "base" balancer package turned out to be more trouble than it was worth, most likely.


markdroth (approved):

ejona86 (approved):
I don't think it was unfair to say "you can share common code," so that didn't need to be removed. But it also doesn't have to be called out; that sort of thing always applies and most other balancers are like RR so it isn't surprising.

PR #388, opened 2023-08-21
L105: Python Add New Error Types

XuanWang-Amos:
Add two new errors in grpc public API.
Discussion thread: https://groups.google.com/g/grpc-io/c/pG34X9nAa3c

### DIFF
--- a/L105-python-expose-new-error-types.md
+++ b/L105-python-expose-new-error-types.md
@@ -4,8 +4,8 @@ L105: Python Add New Error Types
 * Approver: gnossen
 * Status: In Review
 * Implemented in: Python
-* Last updated: 08/21/2023
-* Discussion at: TODO
+* Last updated: 08/31/2023
+* Discussion at: https://groups.google.com/g/grpc-io/c/pG34X9nAa3c
 
 ## Abstract
 
--- a/L105-python-expose-new-error-types.md
+++ b/L105-python-expose-new-error-types.md
@@ -10,8 +10,8 @@ L105: Python Add New Error Types
 ## Abstract
 
 Add two new errors in grpc public API:
-  * SyncBaseError
-  * SyncAbortError
+  * BaseError
+  * AbortError
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


L105-python-expose-new-error-types.md
```diff
@@ -0,0 +1,30 @@
...
+In case of abort, currently we don't log anything, exposing those error types allows user to catch and handle aborts if they want.
+
+## Proposal
+
+Add AbortError and BaseError in public API.
+
+## Rationale
+
+The Async API [has similar errors](https://github.com/grpc/grpc/blob/v1.57.x/src/python/grpcio/grpc/aio/__init__.py#L23,L24). We're refactoring code so those errors will also be used in Sync API. Adding them to the Sync API will help us keep the two stacks in sync and allow users of the Sync implementation to catch and handle aborts.
+
```
gnossen, line 27: We probably want to mention that adding this will allow users to type:

```python
try:
  do_grpc_stuff()
except grpc.BaseError as e:
  # handle error
```

and be confident that they're catching _all_ gRPC exceptions.

L105-python-expose-new-error-types.md
```diff
@@ -0,0 +1,30 @@
+L105: Python Add New Error Types
+----
+* Author(s): XuanWang-Amos
+* Approver: gnossen
+* Status: In Review
+* Implemented in: Python
+* Last updated: 08/21/2023
+* Discussion at: TODO
```
gnossen, line 8: Reminder to populate this.

gnossen (approved):
LGTM. But please do not merge until 2 weeks after the thread has been posted to the mailing list.

L105-python-expose-new-error-types.md
XuanWang-Amos: We also have `RpcError` exposed to user, I think we should mention their differences, when should user expect RpcError?

gnossen: If RpcError is not currently a subclass of BaseError, I think we should make it a subclass of BaseError. Ditto for all other exception types. The ultimate goal is to enable the snippet above.

XuanWang-Amos: Make sense, I'll change `RpcError` to be a subclass of `BaseError` and add it to this gRFC.

PR #389, opened 2023-08-22
A72: OpenTelemetry Tracing

YifeiZhuang:
Design doc: go/grpc-open-telemetry-tracing-design

cc. @yashykt @zasweq @XuanWang-Amos


yashykt:
I don't see the span name mentioned here. I was going to ask whether we can be consistent on the naming. @dfawley had suggested we "rcvd" instead of "recv" as the shortform for received in the OTel metrics https://github.com/grpc/proposal/pull/380

YifeiZhuang:
> I don't see the span name mentioned here. I was going to ask whether we can be consistent on the naming. @dfawley had suggested we "rcvd" instead of "recv" as the shortform for received in the OTel metrics #380

The span name will inherit what census uses, like in server `Recv.<service name>.<method name>`, client side `Sent.<service name>.<method name>` ([A45](https://github.com/grpc/proposal/blob/master/A45-retry-stats.md)), as mentioned in **New Function in OpenTelemetry Plugin**  
> In the new function we will produce the same tracing information as we produce for Census

But it looks A45 does not mention `Recv` at all.
`rcvd` seems fine. When there is decision there, we can certainly make the name consistent here.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+
+TODO: add pseudocode here in C++ for `grpcTraceBinPropagator`,`GrpcCommonSetter` 
+and `GrpcCommonSetter`.
+
+The `GrpcCommonSetter.set()` and `GrpcCommonGetter.get()` method in C++
+should handle both binary (`-bin`) header 
+(e.g. `grpc-trace-bin`) and ASCII header from other text map 
+propagators that users configure into gRPC OpenTelemetry, e.g. w3c.
+
+### Grpc OpenTelemetry Tracing API
```
yashykt, line 301: s/Grpc/gRPC

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+type(Received) and wire message size.
+* When the call is closed, end the parent span with RPC status.
+
+On attempt span:
+* When span is created, add attribute `previous-rpc-attempts` that captures the 
+number of preceding attempts for the RPC, and attribute `transparent-retry` that 
+shows whether stream is a transparent retry.
+* When the stream is created on transport, annotate delayed load balancer pick 
+complete, if any.
+* When an outbound message has been sent, add a message event to capture seq no.,
```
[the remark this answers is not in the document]
YifeiZhuang: Interesting, it looks in Java, `MessageEvent` is used in `outboundMessageSent()` stream tracer methods for both client and server. Other places are using `annotation`.  
But, in OT the API is changed to only have `Event`, so it does not matter whether we used message event or annotation, everything will be an `Event`.
[A45 retry stats](https://github.com/grpc/proposal/blob/master/A45-retry-stats.md) does not mention that many details about cross language tracing information. I think it would be fine as long as languages keep generating the same information in OT tracing. Let me change the description of what is the current tracing information in more general way to avoid confusion.

Same for the wire size. Because java [StreamTracer](https://grpc.github.io/grpc-java/javadoc/io/grpc/StreamTracer.html) has those information, tracing just record it. So in C++ it what size information is captured in tracing?

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+  @Override
+  public <C> void inject(Context context, @Nullable C carrier, TextMapSetter<C> setter) {
+    SpanContext spanContext = Span.fromContext(context).getSpanContext();
+    byte[] value = BinaryFormat.toBytes(spanContext);
+    if (setter instanceof GrpcCommonSetter) {
+      // Fast path in Java and Go that passes bytes directly through API boundaries using
+      // the overloaded set(Metadata, String, byte[]) method added by gRPC.
+      ((GrpcCommonSetter) setter).set((Metadata) carrier, "grpc-trace-bin", value);
+    } else {
+      // Slow path for C++. gRPC C++ does not have type checking, so we encode bytes to
```
ejona86, line 176: s/type checking/runtime type inspection/

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+// This interface will be implemented by gRPCCommonSetter/Getter as an optimization path 
+// to avoid base64 encoding between TextMap APIs due to lack of
+// OpenTelemetry binary propagator API.
+private interface GrpcBinarySetter<Metadata> {
+    void set(Metadata header, String key, byte[] value);
+}
+
+```
+
+The `GrpcCommonSetter` adds an overloaded `set()` method to directly take `bytes[]`
```
ejona86, line 267:
```suggestion
The `GrpcCommonSetter` adds an overloaded `set()` method to directly take `byte[]` 
```

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+private interface GrpcBinarySetter<Metadata> {
+    void set(Metadata header, String key, byte[] value);
+}
+
+```
+
+The `GrpcCommonSetter` adds an overloaded `set()` method to directly take `bytes[]` 
+(Java and Go) to avoid extra base64 encoding. For the normal `set()` method it 
+should handle both binary (`-bin`) header and ASCII header from any TextMap 
+propagators that users may config.
```
ejona86, line 270:
```suggestion
propagators that users may configure.
```

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+The `GrpcCommonGetter` adds new method `getBinary()` for the optimized path for 
+the same reason in Java and Go. Similarly, the normal `get()` method handles both 
+binary header and TextMap propagators.
+
+#### Context Propagation APIs in C++
+C++ will also support the propagators API, because this imposes API
```
ejona86, line 276:
```suggestion
C++ will also support the propagators API, because this provides API 
```

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+The `GrpcCommonGetter` adds new method `getBinary()` for the optimized path for 
+the same reason in Java and Go. Similarly, the normal `get()` method handles both 
+binary header and TextMap propagators.
+
+#### Context Propagation APIs in C++
+C++ will also support the propagators API, because this imposes API 
+uniformity among languages. Due to the language restriction, 
+C++ can not take the optimization path to workaround lacking the binary 
+propagator API. That means using propagators API with C++ needs base64 encoding 
+and therefore is slower compared with just using metadata API. However, C++ can be
```
ejona86, line 280: "can be" isn't clear what point is being made, so "will" makes that more clear.
> However, C++ will have an API that enables adding grpc-trace-bin to the metadata directly, without using the propagators API.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+
+The `GrpcCommonSetter.set()` and `GrpcCommonGetter.get()` method in C++
+should handle both binary (`-bin`) header 
+(e.g. `grpc-trace-bin`) and ASCII header from other text map 
+propagators that users configure into gRPC OpenTelemetry, e.g. w3c.
+
+### Grpc OpenTelemetry Tracing API
+This section talks about enabling and configuring OpenTelemetry tracing.
+The OpenTelemetry API will coexist with the 
+OpenCensus API until the latter is dropped. Only one "grpc-trace-bin" header
```
ejona86, line 304: Let's remove "until the latter is dropped." That isn't part of this gRFC, and we don't have a schedule for it yet.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+The `GrpcCommonSetter.set()` and `GrpcCommonGetter.get()` method in C++
+should handle both binary (`-bin`) header 
+(e.g. `grpc-trace-bin`) and ASCII header from other text map 
+propagators that users configure into gRPC OpenTelemetry, e.g. w3c.
+
+### Grpc OpenTelemetry Tracing API
+This section talks about enabling and configuring OpenTelemetry tracing.
+The OpenTelemetry API will coexist with the 
+OpenCensus API until the latter is dropped. Only one "grpc-trace-bin" header 
+will be sent for a single RPC.
```
ejona86, line 305: as long as only one of OTel or OC is enabled for the channel?

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+TODO: add Go API.
+
+### Migrate to OpenTelemetry: cross-process networking concerns
+When clients first introduce gRPC OpenTelemetry, for the time window when the 
+gRPC client and server have mixed plugins of OpenTelemetry and OpenCensus, 
+with `grpcTraceBinPropagator` users can do migration easily with the compatibility
+guaranteed. It is encouraged to use `grpcTraceBinPropagator` that propagates
+`grpc-trace-bin` header for migration because of the following advantages:
+* Simplified migration path, no migration phase deployments.
+* Binary header is more efficient.
```
ejona86, line 371: This sort of argues against some other statements made. And it seems it could be left unsaid. Using the same header greatly simplifies rollout.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+
+A binary formatter implementation for OpenTelemetry is needed in each language,
```
ejona86, line 373: It isn't clear this is talking about grpc-trace-bin. You can call it a grpc-trace-bin formatter.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+gRPC will expose a custom `grpcTraceBinPropagator` that implements `TextMapPropagator`.
+This grpc-provided propagator still uses the `grpc-trace-bin` header for context 
+propagation. The OpenCensus spanContext and OpenTelemetry spanContext transmitted
+in binary header over the wire are identical, therefore a gRPC OpenCensus client can 
+speak with a gRPC OpenTelemetry server and vice versa. Users can provide a 
+single composite propagator that combines one or multiple `TextMapPropagator` 
+for their client and server separately. This way, users can define their own 
+migration path for context propagators in distributed components, see detailed
+discussion in the later session. Configuring gRPC OpenTelemetry with this 
+propagator when dealing with cross-cutting concerns during migration is
```
ejona86, line 157:
```suggestion
propagator when dealing with cross-process concerns during migration is 
```

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+2. Configure the client with the desired new propagators and to drop the old propagator.
+3. Make the server only accept the new propagators and complete the migration.
+
+### Migrate to OpenTelemetry: in binary
+The OpenCensus [shim](https://github.com/open-telemetry/opentelemetry-java/tree/main/opencensus-shim)
+(currently available in Java, Go, Python) allows binaries that have a mix of 
+OpenTelemetry and OpenCensus dependencies to export trace spans from both frameworks,
+and keep the correct parent-child relationship. This is the recommended approach to migrate
+to OpenTelemetry in one binary gradually. Note that the in-binary migration and 
+cross-cutting concerns migration can be done in parallel.
```
ejona86, line 395:
```suggestion
cross-process migration can be done in parallel.
```

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+
+The shim packages that bridge two libraries works as follows, considering the
+following migration scenarios example: 
+
+`|-- Application - Configured OpenCensus ------------------------------- |`\
```
ejona86, line 400: Put each of these in a single code block. I think that'll make it draw nicer.

\`\`\`
Bunch
of
text
\`\`\`

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+`|--  Application -> Using OpenCensus to generate a sub Trace B--------- |`
+
+The application may use a bridge package in the outermost layer first:\
+`|-- Application - Configured Otel w/ OpenCensus Shim ------------------- |`\
+`|--  gRPC -> Using OpenCensus to generate Trace A  --------------------- |`\
+`|--  Application -> Using OpenCensus to generate a sub Trace B---------- |`
+
+Then the application changes the instrumentation to OpenTelemetry:\
+`|-- Application - Configured Otel w/ OpenCensus Shim ---------------------- |`\
+`|--  gRPC -> Using Otel to generate Trace A  ------------------------------ |`\
```
ejona86, line 411: Looks like this should still have been OpenCensus.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,428 @@
...
+implement the OpenTelemetry propagators API interface. The API composites the 
+propagators and gRPC puts all the propagator data into the wire through metadata. 
+This allows users to easily migrate a group of applications with an old propagator to 
+a new propagator. An example migration path can be:
+1. Configure server to accept both old and new propagators.
+2. Configure the client with the desired new propagators and to drop the old propagator.
+3. Make the server only accept the new propagators and complete the migration.
+
+### Migrate to OpenTelemetry: in binary
+The OpenCensus [shim](https://github.com/open-telemetry/opentelemetry-java/tree/main/opencensus-shim)
```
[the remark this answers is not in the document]
YifeiZhuang: Yea. The shim tool replaces OC span with an OT span and maintains the parent-child span relationship. It also makes OC exporter no-op so that everything is exported through OT exporter. In java they implement the shim through reflection that injects a temporary implementation of related APIs like span and trace that does the conversion. C++ applications would need to implement similar. Note that this is only needed when there are mixed OC/OT dependencies in one application binary. And this is orthogonal to the propagator migration we talked. This topic mainly raises awareness of the problem and mentions the solution/tool hand waving. The grpc OT migration likely won't be blocked on this.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,475 @@
...
+| Uncompressed message size             | `message.event.size.uncompressed`       |
+| Compressed message size               | `message.event.size.compressed`         |
+
+OpenCensus span annotation description maps to OpenTelemetry event name, and 
+annotation attributes keys are mapped to event attributes keys:
+
+| OpenCensus Trace Annotation Attribute Key | OpenTelemetry Trace Event Attribute Key |
+|-------------------------------------------|-----------------------------------------|
+| `id`                                      | `message.event.type`                    |
+| `type`                                    | `message.message.id`                    |
```
zasweq, line 126: I think the attribute key is mixed up here :).

YifeiZhuang: oh i see, thanks for catching it!

A72-open-telemetry-tracing.md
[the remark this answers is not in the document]
yashykt: Isn't the `Event` just a string? We need to specify the contents of the string in that case, and I think the size should be the same as that of stats `Total bytes (compressed but not encrypted) sent across all request messages (metadata excluded) per RPC attempt; does not include grpc or transport framing bytes.`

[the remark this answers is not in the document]
yashykt: Also, I don't think we need "seq no.". I'm not even sure what it means.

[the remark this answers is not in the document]
zasweq: https://github.com/grpc/grpc-go/blob/master/stats/opencensus/trace.go#L107

We have something in our OpenCensus support that essentially annotates the message sent/recv events with an increasing number so we can see what order they were sent and received in. IIRC there was some async ordering issue where you could record these not in order so it is important to provide order. This was called out in the OpenCensus design, and I think this is what Ivy is referring to here.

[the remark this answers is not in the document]
yashykt: I don't think we have such an ordering issue in C++. Given that you just see it in trace UI in the right order, I don't see the benefit to mentioning the seq number, but it's simple enough that I can add it.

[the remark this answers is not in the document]
zasweq: I don't remember what the root reason was why this was included in the OpenCensus spec, but I'm sure it was called out for go tracing.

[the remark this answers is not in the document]
YifeiZhuang: 1. Regarding seq.no, I believe Zach is right. From OpenCensus spec, a span records `MessageEvent` [span.addMessageEvent()](https://www.javadoc.io/doc/io.opencensus/opencensus-api/0.16.1/io/opencensus/trace/Span.html#addMessageEvent-io.opencensus.trace.MessageEvent-), and [messageId](https://opencensus.io/tracing/span/time_events/message_event/#id) identifies the message of the event. The purpose of the messageID is to 
 > The identifier for the message to help correlate between SENT and RECEIVED message events

Not that is not the http stream id, but "the sequential number of the message within the stream, starting from 0",  Java added it specifically for OpenCensus [#3461](https://github.com/grpc/grpc-java/pull/3461). It would be good that C++ can add that, but i'm fine to have it as is and add them later.

2. > Isn't the Event just a string? 

In OpenCensus, a span can record MessageEvent. It does not have a string name.
In OpenTelemetry, span can record Event. An Event has a string name and attributes (key/value pairs). I'll suggest a name update grfc, e.g. "Outbound message sent" and "Inbound message read". Java/Go also record both compressed and uncompressed message sizes when possible. The size is per message, and yes it does not include grpc framing. Mostly we just copy from the census implementation.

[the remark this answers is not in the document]
zasweq: Yeah, thinking about this, I think it was partly to handle the inordering between messages sent and received. A sequence number allows you to order between sent and received.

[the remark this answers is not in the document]
yashykt: I don't understand why we would want to put an order on sent and received messages. For example, if it is a bidi-stream, the application might not have enforced a specific ordering between the sent and received messages. gRPC *would* see them in *a* specific order but I still don't see the point of exposing that order to the users.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,548 @@
...
+####  Propagator API in Java/Go
+The pseudocode below demonstrates `GrpcTraceBinPropagator` and the corresponding 
+gRPC Getter/Setter with an optimization path.
+
+```Java
+public class GrpcTraceBinPropagator implements TextMapPropagator {
+  @Override
+  public <C> void inject(Context context, @Nullable C carrier, TextMapSetter<C> setter) {
+    SpanContext spanContext = Span.fromContext(context).getSpanContext();
+    byte[] value = BinaryFormat.toBytes(spanContext);
```
yashykt, line 170: as discussed offline, the value that we use for "grpc-trace-bin" follows a particular format. We should follow the same format. I've updated the C++ section with the right format. Whether you want to write the entire code is upto you, but you should atleast mention that it will be transformed to the "grpc-trace-bin" format

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,548 @@
...
+      bytes = ((GrpcCommonGetter) textMapGetter).getBinary((Metadata) c, "grpc-trace-bin");
+    } else {
+      // Slow path for C++. gRPC C++ does not have runtime type inspection, so 
+      // we decode String from TextMapGetter API to bytes. This code path applies 
+      // to the situation where GrpcTraceBinPropagator is used with a TextMapGetter
+      // externally.
+      String contextString = textMapGetter.get(c, "grpc-trace-bin");
+      bytes = Base64.getDecoder().decode(contextString);
+    }
+    SpanContext spanContext = BinaryFormat.parseBytes(bytes);
```
yashykt, line 197: Same comment as above on transforming this from the "grpc-trace-bin" format

A72-open-telemetry-tracing.md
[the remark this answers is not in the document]
YifeiZhuang: Yash, I agree with you that the stream id doesn't seem to have critical values, maybe only for reference,  e.g. refer to message `1` is more convenient than referring to message `Thu, Jan 18, 2024, 11:39 AM GMT-8`. And also server side can match up with the client side message id.

The official description from OpenCensus for the use case of the sequence number:
> For example it can be useful when matching sequence/state numbers between protocol handshakes or for a streaming RPC. It is recommended to be unique within a span.

I don't see much point arguing the OpenCensus API for now. For Java/Go since users already have the sequence number in Census we just keep having it in OTel. For C++, no need to add if you don't already have it. And we move on. I'll add that.

A72-open-telemetry-tracing.md
YifeiZhuang: Yea, I verified that with java's binary format implementation (in opencensus repo). Actually in `###Propagator Wire Format` section 3rd paragraph I mentioned similar thing: "When using `grpc-trace-bin` the OpenCensus spanContext and 
OpenTelemetry spanContext are identical, therefore a gRPC OpenCensus client can 
speak with a gRPC OpenTelemetry server and vice versa."

YifeiZhuang:
Hi @markdroth @fengli79 @dfawley,

FYI I've addressed all the AI from the recent 01/08 meeting
1. [done] Add that OpenTelemetry module should document that things/configurations will change as OT changes.
2.  [done] Move C++ fast path to rationale. Capture two alternatives about the fast path that may have in the future.
3. [done] Update C++ API about adding a new method in OpenTelemetryPluginBuilder, and reference to the current code and the gRFC.
4. [done] complete the C++ GrpcTraceBinPropagator and GrpcTextMapCarrier

The PR is ready for review. Please comment and approve.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,554 @@
...
+## Proposal
+gRPC users depending on the grpc-census plugin have non-trivial migration paths 
+to OpenTelemetry. Consider the following use cases:
+1. Migrate an application binary where both OpenCensus and OpenTelemetry maybe exist
+in the dependency tree. This can be the application’s own tracing code, or gRPC 
+OpenCensus, or other dependencies that involve OpenCensus and/or OpenTelemetry.
+2. Compatibility between a gRPC client and server as two distributed components, 
+where during migration one will use OpenCensus and the other will use OpenTelemetry.
+
+All related APIs here are experimental until OpenTelemetry metrics [gRPC A66][A66]
```
markdroth, line 98: Given how far along A66 is, we probably don't need this paragraph anymore.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,554 @@
...
+    }
+  }
+
+ private:
+  grpc_metadata_batch* metadata_;
+};
+
+}  // namespace internal
+
+std::unique_ptr<opentelemetry::context::TextMapCarrier>
```
markdroth, line 391: Shouldn't this be `TextMapPropagator` instead of `TextMapCarrier`?  @yashykt

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,554 @@
...
+MakeGrpcTraceBinTextMapPropagator() {
+  return std::make_unique<internal::GrpcTraceBinTextMapPropagator>();
+}
+
+}  // namespace grpc
+
+```
+
+### gRPC OpenTelemetry Tracing API
```
markdroth, line 400: I think this subsection should be the first one in the "proposal" section, because it describes the main new API that we're adding here.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,554 @@
...
+     ContextPropagators.create(GrpcTraceBinPropagator.getInstance()))
+  .build();
+
+// Alternatively, use auto configuration:
+// OpenTelemetry openTelemetry = 
+// AutoConfiguredOpenTelemetrySdk.getOpenTelemetrySdk().
+
+// Create a module that hosts tracing infrastructures. Should document that 
+// the module implementation may change as OpenTelemetry evolves.
+OpenTelemetryModule otModule = OpenTelemetryModule.getInstance(openTelemetry);
```
markdroth, line 439: Why is this not using `OpenTelemetryPluginBuilder` as in other languages?  I think we want that to be the main API for all things we configure via OT, in all languages.  Tracing should not have a fundamentally different API from metrics.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,554 @@
...
+* Last updated: 2024-01
+* Discussion at: https://groups.google.com/g/grpc-io/c/e_ByaRmtJak
+
+## Abstract
+This proposal adds support for OpenTelemetry tracing and suggests migration
+paths away from OpenCensus tracing. Discussions include:
+* Context propagation between a gRPC client and server during migration.
+* Migrate a gRPC involved software binary that has both OpenTelemetry and 
+OpenCensus dependency.
+* The API surface to enable and configure OpenTelemetry tracing.
```
markdroth, line 18: General comment about the organization of this doc:

The doc seems to be currently organized to present OT tracing primarily in the context of OC tracing.  However, this is likely to be confusing to new users that are not already using OC and just want to understand how OT tracing works.  And in the long reun, I think this doc will be read by more users like than than by users who are already using OC.

Given that, I think it would make more sense to first describe the OT tracing functionality from scratch, without referring to OC.  Then after that, we can have a section describing migration from OC, which will include things like the differences in trace-span annotations, `grpc-trace-bin` propagation, migration use-cases, etc.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,554 @@
...
+  @Override
+  public <C> void inject(Context context, @Nullable C carrier, TextMapSetter<C> setter) {
+    SpanContext spanContext = Span.fromContext(context).getSpanContext();
+    byte[] value = BinaryFormat.toBytes(spanContext);
+    if (setter instanceof GrpcCommonSetter) {
+      // Fast path in Java and Go that passes bytes directly through API boundaries using
+      // the overloaded set(Metadata, String, byte[]) method added by gRPC.
+      ((GrpcCommonSetter) setter).set((Metadata) carrier, "grpc-trace-bin", value);
+    } else {
+      // Slow path for C++. gRPC C++ does not have runtime type inspection, so we
```
markdroth, line 178: It seems strange to have a comment about C++ in the java/go section.  Might be better to defer discussion of C++ until the next subsection.

Same thing throughout this subsection.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,554 @@
...
+C++ will also support propagator APIs to provides API uniformity among
+languages. Since gRPC C++ avoids RTTI, it can not use the same optimization path
+as Java/Go. This will result in an extra base64 encoding/decoding step to
+satisfy `TextMapPropagator` requirement that the key/value pair be a valid HTTP
+field. There are possible optimizations C++ might pursue in the future, for
+example, providing an explicit knob on `GrpcTraceBinTextMapPropagator` that
+assumes that this propagator is being used with gRPC and can hence skirt `TextMapPropagator` compatibility requirements.
+
+```C++
+std::unique_ptr<opentelemetry::context::TextMapCarrier>
```
markdroth, line 292: Shouldn't this be `TextMapPropagator` instead of `TextMapCarrier`?  @yashykt

markdroth (commented):
The design content looks great!  However, I do have some comments about the organization of the doc.

A72-open-telemetry-tracing.md
YifeiZhuang: Just a naming issue. The APIs are the same: tracing and metrics share the same plugin. Updated.

A72-open-telemetry-tracing.md
YifeiZhuang: Hm, describing OT first followed by the migration stuff in later section seems indeed very clear. Re-arranged. Thanks for the suggestion.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,560 @@
...
+### Related Proposals and Documents:
+* [gRFC L29: C++ API Changes for OpenCensus Integration][L29]
+* [gRFC A45: Exposing OpenCensus Metrics and Tracing for gRPC retry][A45]
+* [Microservices observability overview][grpc-observability-public-doc]
+* [gRFC A66: OpenTelemetry Metrics][A66]
+
+## Proposal
+### gRPC OpenTelemetry Tracing API
+We will add tracing functions in grpc-open-telemetry plugin, along with OpenTelemetry
+metrics [gRPC A66][A66]. Languages will keep using gRPC infrastructures, e.g. interceptor and
```
DNVindhya, line 50: s/gRPC/gRFC

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,560 @@
...
+`grpc-trace-bin` header will be sent for a single RPC as long as only one of
+OpenTelemetry or OpenCensus is enabled for the channel. The APIs to enable and
+configure OpenTelemetry tracing are different among languages due to different 
+underlying infrastructures.
+
+#### Java
+In Java, it will be part of global interceptors, so that the interceptors are
+managed in a more sustainable way and user-friendly. As a prerequisite, the stream
+tracer factory API will be stabilized. OpenTelemetryModule will be created with
+an OpenTelemetryAPI instance passing in for necessary configurations.
```
DNVindhya, line 63: s/OpenTelemetryAPI/OpenTelemetry API

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,560 @@
...
+Users can also rely on SDK autoconfig extension that configures the sdk object
+through environmental variables or Java system properties, then obtain the sdk
+object passed in to gRPC.
```
DNVindhya, line 66: Do you mean _passing in the obtained sdk object to gRPC_? As the snippet does not show an API to get the SDK object provided to OpenTelemetryModule.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,560 @@
...
+(Java and Go) to avoid extra base64 encoding. For the normal `set()` method it 
+should handle both binary (`-bin`) header and ASCII header from any TextMap 
+propagators that users may configure.
+The `GrpcCommonGetter` in Java and Go adds a new method `getBinary()` for the 
+optimized path for the same reason. Similarly, the normal `get()` method handles
+both binary headers and TextMap propagators.
+
+#### Context Propagation APIs in C++
+C++ will also support propagator APIs to provides API uniformity among
+languages. Since gRPC C++ avoids RTTI, it can not use the same optimization path
```
DNVindhya, line 311: This is the first time we are referring to Run-time type information. Can we add it along with acronym?

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,560 @@
...
+1. Configure server to accept both old and new propagators.
+2. Configure the client with the desired new propagators and to drop the old propagator.
+3. Make the server only accept the new propagators and complete the migration.
+
+### Migrate to OpenTelemetry: In Binary
+The OpenCensus [shim](https://github.com/open-telemetry/opentelemetry-java/tree/main/opencensus-shim)
+(currently available in Java, Go, Python) allows binaries that have a mix of 
+OpenTelemetry and OpenCensus dependencies to export trace spans from both 
+frameworks, and keep the correct parent-child relationship. This is the 
+recommended approach to migrate to OpenTelemetry in one binary gradually.
```
DNVindhya, line 477: Instead of saying "in one binary gradually", how about we say "incrementally within a single binary"?

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,560 @@
...
+
+OpenCensus span annotation description maps to OpenTelemetry event name, and
+annotation attributes keys are mapped to event attributes keys:
+
+| OpenCensus Trace Annotation Attribute Key | OpenTelemetry Trace Event Attribute Key |
+|-------------------------------------------|-----------------------------------------|
+| `type`                                    | `message.event.type`                    |
+| `id`                                      | `message.message.id`                    |
+
+And OpenTelemetry no longer has span end options as OpenCensus does.
```
DNVindhya, line 532: Does that mean we cannot get RPC status from traces?

A72-open-telemetry-tracing.md
YifeiZhuang: Java has `OpenTelemetryModule` equivalent to `OpenTelemetryPluginBuilder`. Align with A66. Let me know if you have concerns on the name.

YifeiZhuang:
@markdroth thanks for the suggestions. They are great. Please take a look again.

@dfawley kind reminder please review and approve.

A72-open-telemetry-tracing.md
YifeiZhuang: The status will be recorded in a separate API `.setStatus()` for OTel Java. In Census Java the status was in span end option.
However, I noticed Otel C++ and Census C++ always has `.setStatus()` API.
So I removed this line and rephrased it in the `Tracing Information` section in the proposal. I think it is more clear. Thanks!

A72-open-telemetry-tracing.md
YifeiZhuang: Yeah. Thanks!

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
+* [Microservices observability overview][grpc-observability-public-doc]
+* [gRFC A66: OpenTelemetry Metrics][A66]
+
+## Proposal
+### gRPC OpenTelemetry Tracing API
+We will add tracing functions in grpc-open-telemetry plugin, along with OpenTelemetry
+metrics [gRFC A66][A66]. Languages will keep using gRPC infrastructures, e.g. interceptor and
+stream tracer to implement the feature, the same as Census.
+
+The OpenTelemetry API will coexist with the OpenCensus API. Only one
```
markdroth, line 52: Let's move this paragraph down to the "Migrate from OpenCensus to OpenTelemetry" section.

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
+  .builder()
+  .addSpanProcessor(
+     BatchSpanProcessor.builder(exporter).build())
+  .build();
+
+OpenTelemetry openTelemetry = OpenTelemetrySdk.builder()
+  .setTracerProvider(sdkTracerProvider)
+  .setMeterProvider(...)
+  .setPropagators(
+     ContextPropagators.create(GrpcTraceBinPropagator.getInstance()))
```
markdroth, line 80: Let's use the provided W3C propagator in these examples, so that we don't talk about grpc-trace-bin until the migration section.

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
+#### Java
+In Java, it will be part of global interceptors, so that the interceptors are
+managed in a more sustainable way and user-friendly. As a prerequisite, the stream
+tracer factory API will be stabilized. OpenTelemetryModule will be created with
+an OpenTelemetry API instance passing in for necessary configurations.
+Users can also rely on SDK autoconfig extension that configures the sdk object
+through environmental variables or Java system properties, then pass the 
+obtained sdk object to gRPC.
+
+```Java
```
markdroth, line 67: This java section shows an example of how to use the API rather than defining the API itself, whereas the C++ section below does the opposite.  Let's pick one of the two approaches and be consistent.  @yashykt

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
+  OpenTelemetryPluginBuilder& SetTextMapPropagator(
+      std::unique_ptr<opentelemetry::context::propagation::TextMapPropagator>
+          text_map_propagator);
+};
+```
+
+#### Go
+In Go, the OpenTelemetry stream tracers and interceptors will be provided for users to install.
+
+TODO: add Go API.
```
markdroth: Can we fill in the Go section before we merge this?  @zasweq

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
+
+### Tracing Information
+With the new OpenTelemetry plugin we will produce the same tracing information as we 
+produce for Census. The following tracing information during an RPC lifecycle
```
markdroth, line 159: Please describe this from scratch without referring to OC.  For example, we should have a brief description of the difference between the parent call span vs. the attempt span on the client side.

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
+### Tracing Information
+With the new OpenTelemetry plugin we will produce the same tracing information as we 
+produce for Census. The following tracing information during an RPC lifecycle 
+should be captured:
 
 At the client, on parent span:
-* When the call is started, annotate name resolution completed if the RPC had 
-name resolution delay.
-* When the call is closed, end the parent span with RPC status.
+* When the call is started, annotate name resolution completed if the RPC had
```
markdroth, line 163: For each of these items, please provide a detailed spec as to how the info will be recorded in the trace span (e.g., attribute key, value type and format, etc).  (The attribute names are already listed in the table below that shows the mapping from OC, but the same info should also be here.)

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
-to capture seq no., type(Received), wire message size, and uncompressed message size,
-if any decompression.
-* When the stream is closed, end the span with RPC status.
+  to capture seq no., type(Received), wire message size, and uncompressed message size,
+  if any decompression.
+* When the stream is closed, set the RPC status and end the span.
 
-Note that C++ is missing the seq no. information. And Java has issue of reporting 
-decompressed message size upon receiving messages, 
+Note that C++ is missing the seq no. information. And Java has issue of reporting
```
markdroth, line 192: Is this paragraph talking about limitations in the existing OC impl or limitations that we're encoding into the new OT functionality?  If the former, it should probably just be removed, or maybe moved down to the "Migrate from OpenCensus to OpenTelemetry" section.  If the latter, we need to explain why we're going to have these limitations.

markdroth (commented):
This new organization looks much better!  I have a few more specific comments along the same lines.

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
+* [gRFC L29: C++ API Changes for OpenCensus Integration][L29]
+* [gRFC A45: Exposing OpenCensus Metrics and Tracing for gRPC retry][A45]
+* [Microservices observability overview][grpc-observability-public-doc]
+* [gRFC A66: OpenTelemetry Metrics][A66]
+
+## Proposal
+### gRPC OpenTelemetry Tracing API
+We will add tracing functions in grpc-open-telemetry plugin, along with OpenTelemetry
+metrics [gRFC A66][A66]. Languages will keep using gRPC infrastructures, e.g. interceptor and
+stream tracer to implement the feature, the same as Census.
```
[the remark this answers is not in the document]
YifeiZhuang: fixed

A72-open-telemetry-tracing.md
```diff
@@ -36,98 +38,125 @@ and OpenCensus's binary format for (de)serialization. The header name is unique
...
-
-| OpenCensus Trace Annotation Attribute Key | OpenTelemetry Trace Event Attribute Key |
-|-------------------------------------------|-----------------------------------------|
-| `type`                                    | `message.event.type`                    |
-| `id`                                      | `message.message.id`                    |
-
-And OpenTelemetry no longer has span end options as OpenCensus does.
+  type(Received) and uncompressed message size.
 
 ### Propagator Wire Format
```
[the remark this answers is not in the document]
YifeiZhuang: fixed

A72-open-telemetry-tracing.md
YifeiZhuang: Good call, fixed. I added the API itself.

A72-open-telemetry-tracing.md
YifeiZhuang: @zasweq please take a look. I added the Go API, similar to Otel metrics, and comments about how users will expect from it.

YifeiZhuang:
> This new organization looks much better! I have a few more specific comments along the same lines.
> 
> Please let me know if you have any questions. Thanks!

@markdroth  Thanks for the quick feedbacks, I should have resolved all the comments. Please take a look again.

A72-open-telemetry-tracing.md
```diff
@@ -105,94 +133,181 @@ class OpenTelemetryPluginBuilder {
...
+    message size. The size is the total attempt message bytes without encryption,
+    not including grpc or transport framing bytes, the same below.
+  * If any compression, key `message.event.size.compressed` with integer value
+    of compressed message size.
+* When an inbound message has been received from the transport, add an Event
+  with name "Inbound message read" and the following attributes:
+  * key `message.event.type` with String value "RECEIVED".
+  * key `message.message.id` with integer value of the seq no.
+  * key `message.event.size.compressed` with integer value of wire message size. 
+  * If any compression, key `message.event.size.uncompressed` with integer value
```
markdroth, line 223: Isn't this backwards?  If there is no compression, I'd expect us to report the uncompressed size, not the compressed size.

Same for server side.

A72-open-telemetry-tracing.md
```diff
@@ -105,94 +133,181 @@ class OpenTelemetryPluginBuilder {
...
+* When span is created, set the attribute with key `previous-rpc-attempts` and an 
+  integer value representing the count of previous attempts made for the RPC.
+* When span is created, set the attribute with key `transparent-retry` and a 
+  boolean value indicating whether the stream is undergoing a transparent retry.
+* If the RPC experienced load balancer pick delay, add an Event with the name 
+  "Delayed LB pick complete" upon creation of the stream on the transport. 
+* When the application sends an outbound message to the transport, add an Event 
+  with name "Outbound message sent" and the following attributes:
+  * key `message.event.type` with string value "SENT".
+  * key `message.message.id` with integer value of the seq no. The seq no. is a sequence
```
markdroth, line 211: We should explicitly state that the seq no is incremented once for each sent message.  Suggest something like:

The seq no indicates the order of the sent messages on the attempt (i.e., it starts at 0 and is incremented by 1 for each message sent).

Same for server side.

A72-open-telemetry-tracing.md
```diff
@@ -105,94 +133,181 @@ class OpenTelemetryPluginBuilder {
...
+    of integer numbers starting from 0 to identify sent messages within the stream, the same below.
+  * key `message.event.size.uncompressed` with integer value of uncompressed 
+    message size. The size is the total attempt message bytes without encryption,
+    not including grpc or transport framing bytes, the same below.
+  * If any compression, key `message.event.size.compressed` with integer value
+    of compressed message size.
+* When an inbound message has been received from the transport, add an Event
+  with name "Inbound message read" and the following attributes:
+  * key `message.event.type` with String value "RECEIVED".
+  * key `message.message.id` with integer value of the seq no.
```
markdroth, line 221: We should say this in a way that clarifies that the recv seq no is independent of the send seq no.  Suggest similar wording to above:

The seq no indicates the order of the received messages on the attempt (i.e., it starts at 0 and is incremented by 1 for each message received).

Same for server side.

A72-open-telemetry-tracing.md
```diff
@@ -105,94 +133,181 @@ class OpenTelemetryPluginBuilder {
...
+  * key `message.event.size.compressed` with integer value of wire message size.
+  * If any compression, key `message.event.size.uncompressed` with integer value
+    of uncompressed message size.
 * When the stream is closed, set the RPC status and end the span.
 
-Note that C++ is missing the seq no. information. And Java has issue of reporting
-decompressed message size upon receiving messages,
+### Limitations
+Note that C++ is missing the seq no. information due to lack of transport support.
```
markdroth, line 246: Where does this limitation come from?  I don't see why we would have trouble reporting the sequence numbers.  @yashykt

A72-open-telemetry-tracing.md
```diff
@@ -105,94 +133,181 @@ class OpenTelemetryPluginBuilder {
...
+    of uncompressed message size.
 
-### Propagator Wire Format
-While gRPC OpenCensus directly interacts with the metadata API, gRPC OpenTelemetry 
-will use the standardized propagators API for context propagation, for the 
-following benefits:
-1. Full integration with OpenTelemetry APIs that is easier for users to reason about.
-2. Make it possible to plugin other propagators that the community supports.
-3. Flexible API that allows clean and simple migration paths to a different propagator.
+## Migrate from OpenCensus to OpenTelemetry
```
markdroth, line 258: s/Migrate/Migration/

A72-open-telemetry-tracing.md
```diff
@@ -41,22 +41,34 @@ tracing instrumentation.
...
-stream tracer to implement the feature, the same as Census.
+metrics [gRFC A66][A66]. Internally, the tracing functionality will be implemented
+using existing gRPC infrastructure such as interceptors and stream tracers.
 
-The OpenTelemetry API will coexist with the OpenCensus API. Only one
-`grpc-trace-bin` header will be sent for a single RPC as long as only one of
-OpenTelemetry or OpenCensus is enabled for the channel. The APIs to enable and
-configure OpenTelemetry tracing are different among languages due to different 
-underlying infrastructures.
+#### Propagator Wire Format
```
markdroth, line 52: Suggest de-indenting this header 2 levels and moving this section down to right before the migration section.

A72-open-telemetry-tracing.md
```diff
@@ -105,94 +133,181 @@ class OpenTelemetryPluginBuilder {
...
-propagation and also support other propagators. gRPC will expose a custom 
-`GrpcTraceBinPropagator` that implements `TextMapPropagator`. However, there is a caveat: 
+Here are the suggested solutions for both use cases. 
 
-Currently, OpenTelemetry propagator API only supports `TextMapPropagator`, 
-that is to send string key/value pairs between the client and server, which is 
-different from the binary header that gRPC currently uses. The future roadmap 
-to support binary propagators at OpenTelemetry is unclear. So, gRPC will use 
-propagator API in TextMap format with an optimization path (Go and Java) to work 
+### Migrate to OpenTelemetry: Cross-process Networking Concerns
```
markdroth, line 279: s/Migrate/Migration/

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,643 @@
...
+Configuring gRPC OpenTelemetry with this propagator when dealing with 
+cross-process concerns during migration is straightforward and recommended. 
+In the long term, community standardized propagators, e.g. W3C is more encouraged than `GrpcTraceBinPropagator`.
+This also allows users to easily migrate a group of applications with an old propagator to 
+a new propagator. An example migration path can be:
+1. Configure server to accept both old and new propagators.
+2. Configure the client with the desired new propagators and to drop the old propagator.
+3. Make the server only accept the new propagators and complete the migration.
+
+### Migrate to OpenTelemetry: In Binary
```
markdroth, line 558: s/Migrate/Migration/

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,643 @@
...
+3. Flexible API that allows clean and simple migration paths to a different propagator.
+
+In order for the propagator to perform injecting and extracting spanContext value 
+from the carrier, which is the Metadata in gRPC, languages will 
+implement Getter and Setter corresponding to the propagator type.
+Currently, OpenTelemetry propagator API only supports `TextMapPropagator`,
+that is to send string key/value pairs between the client and server.
+Therefore, adding Getter and Setter is to implement the TextMap carrier interface: 
+`TextMapCarrier` (For C++/Go), or `TextMapGetter`/`TextMapSetter` (For Java). (see
+pseudocode in section [Migrate to OpenTelemetry](#migrate-to-opentelemetry--cross-process-networking-concerns)).
```
markdroth, line 66: s/Migrate/Migration/ (both in text and in the link)

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,643 @@
...
+Using the same header greatly simplifies rollout. However, there is a caveat about 
+`GrpcTraceBinPropagator`:
+
+Currently, OpenTelemetry propagator API only supports `TextMapPropagator`,
+that is to send string key/value pairs between the client and server, which is
+different from the binary header that gRPC currently uses. The future roadmap
+to support binary propagators at OpenTelemetry is unclear. So, gRPC will use
+propagator API in TextMap format with an optimization path (Go and Java) to work
+around the lack of binary propagator API to support `grpc-trace-bin`. In fact,
+TextMap propagator does not show visible performance impact for C++, which is
```
markdroth, line 298: I think we should be a little clearer here.  Specifically, I belive that what we actually found is that using the W3C text format was performant enough that it's a viable alternative to our existing binary format.  I don't think we actually checked the performance of the slow path for the binary header, and I wouldn't expect the performance to be very good for that.  So what we really decided here was that we'd live with this bad performance for now just for the migration cases, and if that posed a performance problem for someone, we can consider implementing an alternative.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,643 @@
...
+the most sensitive language to performance, based on internal micro benchmarking.
+Therefore, gRPC will only support propagating `grpc-trace-bin` in TextMap propagator.
+Only one `grpc-trace-bin` header will be sent for a single RPC as long as only one of
```
markdroth, line 301: @yashykt How are we going to implement this?  Are we just going to have the two CallTracers set the same metadata, and the last one wins?  Or are we going to do something more clever to make sure that we don't need to encode it multiple times?

A72-open-telemetry-tracing.md
```diff
@@ -105,94 +133,181 @@ class OpenTelemetryPluginBuilder {
...
+    of uncompressed message size.
 * When the stream is closed, set the RPC status and end the span.
 
-Note that C++ is missing the seq no. information. And Java has issue of reporting
-decompressed message size upon receiving messages,
+### Limitations
+Note that C++ is missing the seq no. information due to lack of transport support.
+While it's not critical, we can include these information if users request it in the future.
+
+Java has an issue of reporting decompressed message size upon receiving messages,
```
markdroth, line 249: Maybe we should be slightly less specific here?  Rather than saying that Java will do this differently, maybe we should say that implementations that cannot report both the uncompressed size at the transport layer will instead add this second event.

The reason I suggest this is that I think we may have a similar concern in C-core, where we don't uncompress until above the transport.  @yashykt

markdroth (commented):
This is looking better and better!  Comments are mostly fine details at this point.

A72-open-telemetry-tracing.md
yashykt: I think we should remove the mention of "transport" and we should use a more loose language similar to what we added in https://github.com/grpc/proposal/blob/master/A66-otel-stats.md. 

For example, in gRPC Core, the message events (both compressed and decompressed) are added in the message compression filter which is above the transport.

One thing I want to stress and want mentioned in the gRPC is that I want the flexibility to report the compressed and decompressed events separately. This will allow users to see the amount of time it took to do compression/decompression. For implementations that cannot do this, it's fine to use a single event or not report at all.

A72-open-telemetry-tracing.md
yashykt: Yeah, I can add it, but I just don't see the point of it. Why do we want to keep track of the sequence number of a message? We don't reorder messages.

A72-open-telemetry-tracing.md
YifeiZhuang: **Send/receive messages to/from where**
IIRC In Java it is different than C++: reporting compressed/uncompressed happens _at_ the transport framer/deframer, except for reporting inbound uncompressed message sizes that happens later above transport. 
Earlier Mark had question about the context of send/receive, like it is pass flow control or on the wire. So I mentioned application. Since C++ does that above transport, I will not mention "to transport".

**Combined or separate events for reporting compressed and uncompressed message size**
C++ is already doing that separately. Java needs some changes to support reporting separate messages on before-compression at sending. Similar for Go I believe.

A72-open-telemetry-tracing.md
YifeiZhuang: when received from wire, we don't know there is compression or not, so we call it in general "message.event.size.compressed", indicating the wire size. This is similar to OpenCensus tracing info in java.

YifeiZhuang:
@ejona86 @yashykt @markdroth @zasweq @dfawley updated tracing info based on conversation yesterday. Please review. Thanks!

A72-open-telemetry-tracing.md
```diff
@@ -205,57 +186,75 @@ On attempt span:
   boolean value indicating whether the stream is undergoing a transparent retry.
 * If the RPC experienced load balancer pick delay, add an Event with the name 
   "Delayed LB pick complete" upon creation of the stream on the transport. 
-* When the application sends an outbound message to the transport, add an Event 
+* When the application sends an outbound message to the transport, add Event(s)
+  (it depends on implementation whether there is a single event or two separate 
+  events for compressed/uncompressed message sizes, the same below) 
   with name "Outbound message sent" and the following attributes:
```
markdroth, line 192: If the implementation chooses to record separate events for compressed and uncompressed message sizes, presumably the events will have different names.  We should probably specify the name of the other event (e.g., maybe something like "Outbound message compressed").

Also, can we specify anything about the order of the events if there are two of them?  Alternatively, if we think this will be different in different impls, we can say that the events can be in either order.

A72-open-telemetry-tracing.md
```diff
@@ -205,57 +186,75 @@ On attempt span:
...
-* When an inbound message has been received from the transport, add an Event
-  with name "Inbound message read" and the following attributes:
-  * key `message.event.type` with String value "RECEIVED".
-  * key `message.message.id` with integer value of the seq no.
-  * key `message.event.size.compressed` with integer value of wire message size. 
-  * If any compression, key `message.event.size.uncompressed` with integer value
-    of uncompressed message size.
+  * If compression needed, add key `message.event.size.compressed` with integer 
+    value of compressed message size.
+* When an inbound message has been received from wire, add Event(s) with name
```
markdroth, line 202: I think impls might want to split up the compressed and uncompressed sizes into separate events for inbound messages as well, so we should use the same language here as we do for outbound messages.

A72-open-telemetry-tracing.md
```diff
@@ -205,57 +186,75 @@ On attempt span:
...
+    the order of the received messages on the attempt (i.e., it starts at 0 and is
+    incremented by 1 for each message received), the same below.
+  * key `message.event.size.compressed` with integer value of wire message size.
+  * If the message needs decompression, add key `message.event.size.uncompressed`
+    with integer value of uncompressed message size.
 * When the stream is closed, set RPC status and end the attempt span.
 
 At the server:
-* When the application sends an outbound message to the transport, add an Event
+* When the application sends an outbound message to the transport, add Event(s)
```
markdroth, line 214: Same language here about allowing splitting up the events.

A72-open-telemetry-tracing.md
```diff
@@ -205,57 +186,75 @@ On attempt span:
...
   * key `message.message.id` with integer value of the seq no.
   * key `message.event.size.uncompressed` with integer value of uncompressed
     message size.
-  * If any compression, key `message.event.size.compressed` with integer value
-    of compressed message size.
-* When an inbound message has been read from the transport, add an Event
-  with name "Inbound message read" and the following attributes:
+  * If compression needed, add key `message.event.size.compressed` with integer
+    value of compressed message size.
+* When an inbound message has been received from wire, add Event(s) with name
```
markdroth, line 222: Same language here about allowing splitting up the events.

A72-open-telemetry-tracing.md
```diff
@@ -205,57 +186,75 @@ On attempt span:
   boolean value indicating whether the stream is undergoing a transparent retry.
 * If the RPC experienced load balancer pick delay, add an Event with the name 
   "Delayed LB pick complete" upon creation of the stream on the transport. 
-* When the application sends an outbound message to the transport, add an Event 
+* When the application sends an outbound message to the transport, add Event(s)
```
markdroth, line 189: If we're intentionally being imprecise about where in the stack we record this event, then let's not say "to the transport".

A72-open-telemetry-tracing.md
markdroth, line 202: If we're intentionally being imprecise about where in the stack we record this event, then let's not say "from wire".

A72-open-telemetry-tracing.md
markdroth, line 214: As above, suggest removing "to the transport".

A72-open-telemetry-tracing.md
markdroth, line 222: As above, suggest removing "from wire".

A72-open-telemetry-tracing.md
markdroth: I think this approach results in somewhat inconsistant naming when compression is not used: we will wind up reporting only `message.event.size.compressed` for inbound messages but only `message.event.size.uncompressed` for outbound messages.

I can think of a couple possible alternatives here (other suggestions welcome):
1. If compression is not used, report message size as just "message.event.size", without labelling it as either compressed or uncompressed.
2. Instead of calling these attributes "compressed" and "uncompressed", report them as "application" and "wire" sizes.  Always report the wire size, and report the application size only if compression is used.

markdroth (commented):
Just a few details left here!

PLMK if you have any questions.  Thanks!

A72-open-telemetry-tracing.md
yashykt: I believe we said that using two plugins that both use the same metadata "grpc-trace-bin" is equivalent to shooting yourself in the foot, and should not be done.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,641 @@
...
+  integer value representing the count of previous attempts made for the RPC.
+* When span is created, set the attribute with key `transparent-retry` and a 
+  boolean value indicating whether the stream is undergoing a transparent retry.
+* If the RPC experienced load balancer pick delay, add an Event with the name 
+  "Delayed LB pick complete" upon creation of the stream on the transport. 
+* When the application sends an outbound message to the transport, add Event(s)
+  (it depends on implementation whether there is a single event or two separate 
+  events for compressed/uncompressed message sizes, the same below) 
+  with name "Outbound message sent" and the following attributes:
+  * key `message.event.type` with string value "SENT".
```
yashykt, line 193: is this attribute "message.event.type" useful? The message already says sent?

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,641 @@
...
+* When span is created, set the attribute with key `previous-rpc-attempts` and an 
+  integer value representing the count of previous attempts made for the RPC.
+* When span is created, set the attribute with key `transparent-retry` and a 
+  boolean value indicating whether the stream is undergoing a transparent retry.
+* If the RPC experienced load balancer pick delay, add an Event with the name 
+  "Delayed LB pick complete" upon creation of the stream on the transport. 
+* When the application sends an outbound message to the transport, add Event(s)
+  (it depends on implementation whether there is a single event or two separate 
+  events for compressed/uncompressed message sizes, the same below) 
+  with name "Outbound message sent" and the following attributes:
```
yashykt, line 192: nit: feel free to ignore
Is "outbound" adding any value here?

A72-open-telemetry-tracing.md
yashykt: +1

yashykt: +1

yashykt (approved):
I think my comments are mostly minor assuming we can easily change the formatting of the event messages and attributes

A72-open-telemetry-tracing.md
YifeiZhuang: Sure. (I purposely added _the same below_ at the outbound message so I can avoid duplicated words for similar situation for inbound case, and  server side cases. )

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,641 @@
...
+  * key `message.event.type` with string value "SENT".
+  * key `message.message.id` with integer value of the seq no. The seq no. indicates
```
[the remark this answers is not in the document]
YifeiZhuang: It was from the Otel java opencensus shim. Languages can be different. 
I think whenever possible, we can make the name generated from the shim and the one generated directly from otel be consistent. This is not happening because the naming in our census needs improvements. And I am fine with that. @markdroth @ejona86 

(https://github.com/open-telemetry/opentelemetry-java/blob/main/opencensus-shim/src/main/java/io/opentelemetry/opencensusshim/SpanConverter.java#L24-L27)

A72-open-telemetry-tracing.md
YifeiZhuang: I think we should consider parity from opencensus first, and make possible changes when we see there is improvements to make. So report `message.event.size` without knowing it needs compression or decompression, and "compressed/uncompressed" for a second event sounds fair to me. This would be less of a problem for Java/Go that reports one event only.

A72-open-telemetry-tracing.md
yashykt: @veblush I think Esun did do a microbenchmark that showed that doing millions of encoding and decoding was actually not that bad.

A72-open-telemetry-tracing.md
markdroth: I think we should define a schema that is most clear for users rather than being bound by what OpenCensus did.

As currently written, if compression is enabled, then `message.event.size` will be the uncompressed size for outbound messages but the compressed size for inbound messages.  I think that's going to be very confusing.  I think we should define a schema that makes sense both when compression is enabled and disabled.

I would prefer one of the following options:
1. We use different names depending on whether or not compression is enabled:
   -  If compression is enabled, then we use `message.event.size.compressed` and `message.event.size.uncompressed`.
   - If compression is disabled, then we use just `message.event.size`.
2. Instead of calling these attributes "compressed" and "uncompressed", report them as "application" and "wire" sizes. Always report the wire size, and report the application size only if compression is used.

markdroth (commented):
Just one issue remaining here!

PLMK if you have any questions.  Thanks!

A72-open-telemetry-tracing.md
YifeiZhuang: I did option1. The "application"/"wire" sounds more literal but is less precise. I'm nervous they will introduce further discussions.

markdroth (approved):
Thanks for all the edits!  This looks great!

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,659 @@
...
+// the server option returned from this function as an argument to
+// grpc.NewServer().
+//
+// Using this option will always lead to instrumentation, however in order to
+// use the data a SpanExporter must be registered with the TraceProvider option.
+// Server side does not have retries, so a registered Stats Handler is the only
+// option that is returned.
+func ServerOption(to TraceOptions) grpc.ServerOption {}
+```
+
```
YifeiZhuang, line 166: @XuanWang-Amos I think it would be appropriate to add a `###Python` section here about API. The propagator part it would be similar to java/go, please review.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,680 @@
...
+The APIs to enable and configure OpenTelemetry tracing are different among 
+languages due to different underlying infrastructures.
+
+### Java
+In Java, it will be part of global interceptors, so that the interceptors are
+managed in a more sustainable way and user-friendly. As a prerequisite, the stream
+tracer factory API will be stabilized. OpenTelemetryModule will be created with
+an OpenTelemetry API instance passing in for necessary configurations.
+Users can also rely on SDK autoconfig extension that configures the sdk object
+through environmental variables or Java system properties, then pass the
```
dfawley, line 60: Nit: "environment" variables, not "environmental".

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,680 @@
...
+//
+// Server applications interested in instrumenting their grpc.Server should pass
+// the server option returned from this function as an argument to
+// grpc.NewServer().
+//
+// Using this option will always lead to instrumentation, however in order to
+// use the data a SpanExporter must be registered with the TraceProvider option.
+// Server side does not have retries, so a registered Stats Handler is the only
+// option that is returned.
+func ServerOption(to TraceOptions) grpc.ServerOption {}
```
dfawley, line 184: These bits are pre-existing from after Metrics support is added.  I think instead we'll just be adding a field called `TraceOptions` to `opentelemetry.Options`

dfawley (approved):

A72-open-telemetry-tracing.md
zasweq: Yeah we will. Options will have Metrics Options as the in flight PR, I'll add tracing options, and eventually Logging Options.

YifeiZhuang: oh I see this in the open https://github.com/grpc/grpc-go/pull/7166/, This will be outdated but i think not necessary to keep all updated - we are racing with ourselves.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,680 @@
...
+We will add tracing functions in grpc-open-telemetry plugin, along with OpenTelemetry
+metrics [gRFC A66][A66]. Internally, the tracing functionality will be implemented
+using existing gRPC infrastructure such as interceptors and stream tracers.
+The APIs to enable and configure OpenTelemetry tracing are different among 
+languages due to different underlying infrastructures.
+
+### Java
+In Java, it will be part of global interceptors, so that the interceptors are
+managed in a more sustainable way and user-friendly. As a prerequisite, the stream
+tracer factory API will be stabilized. OpenTelemetryModule will be created with
```
ejona86, line 57: We aren't going to stabilize the tracer API. And I don't see a reason we need to, here.

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,680 @@
...
+  "Delayed LB pick complete" upon creation of the stream on the transport. 
+* When the application sends an outbound message, add Event(s)
+  (it depends on implementation whether there is a single event or an additional 
+  separate event with name "Outbound message compressed" for compressed message size) 
+  with name "Outbound message sent" and the following attributes:
+  * key `message.event.type` with string value "SENT".
+  * key `message.message.id` with integer value of the seq no. The seq no. indicates
+    the order of the sent messages on the attempt (i.e., it starts at 0 and is 
+    incremented by 1 for each message sent), the same below.
+  * key named `message.event.size.uncompressed` if the message needs compression,
```
ejona86, line 217: "needs compression" is a bit difficult. I assume that is "if this message will be compressed." But I'm not certain all languages can do this (know before the message is compressed whether the message will be compressed).

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,680 @@
...
+On the server side, there is only per-call traces.  With the new OpenTelemetry 
+plugin we will produce the following tracing information during an RPC lifecycle:
+
+At the client, on parent span:
+* If the RPC experienced name resolution delay, add an Event at the start of the
+  call with the name "Delayed name resolution complete" upon completion of the
+  name resolution process.
+* When the call is closed, set RPC status and end the parent span. gRPC status "OK" 
+  is recorded with status "OK", while other gRPC statuses are marked as "ERROR". 
+  Non-"OK" statuses include their code as a description, the same below.
```
ejona86, line 200: Do they use the integer code or the code's name?

ejona86, line 200: Are we not including the status description? Seems that is useful, and was included with OpenCensus.

A72-open-telemetry-tracing.md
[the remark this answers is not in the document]
ejona86: I don't think we have to copy the shim's behavior.

ejona86 (commented):
Sending what I have. I think yashykt had some good comments.

A72-open-telemetry-tracing.md
YifeiZhuang: @yashykt could you help confirm whether C++ is able to tell the message needs compression or not when it is reporting the first Event?

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,680 @@
...
+    order of the event must be after the first event that reports the message size.
+* When an inbound message has been received, add Event(s) (it depends on 
+  implementation whether there is a single event or an additional separate event
+  with name "Inbound message uncompressed" for uncompressed message size) with name 
+  "Inbound message read" and the following attributes:
+  * key `message.event.type` with string value "RECEIVED".
+  * key `message.message.id` with integer value of the seq no. The seq no. indicates
+    the order of the received messages on the attempt (i.e., it starts at 0 and is
+    incremented by 1 for each message received), the same below.
+  * key named `message.event.size.compressed` if the message needs decompression,
```
YifeiZhuang, line 233: @yashykt same here for inbound.

A72-open-telemetry-tracing.md
YifeiZhuang: oh this may have to change if we don't want to overload description with both description and status code.
What about inherit `opentelemetry.StatuCode` and add grpc statuscode there?

opentelemetry does not have API to specify detailed status code(their enum only has UNSET OK ERROR)
https://javadoc.io/doc/io.opentelemetry/opentelemetry-api/latest/io/opentelemetry/api/trace/Span.html#setStatus(io.opentelemetry.api.trace.StatusCode,java.lang.String)

A72-open-telemetry-tracing.md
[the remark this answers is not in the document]
YifeiZhuang: I proposed new names. Let's move forward and fix/vote in API stabilization.

A72-open-telemetry-tracing.md
YifeiZhuang: Fair enough, removed.

A72-open-telemetry-tracing.md
YifeiZhuang: The language seems common and clear to me.

> The phrase "outbound message sent" is clear in the sense that it conveys a simple meaning: a message has been transmitted from one system or user to an external destination.
 
-From Gemini

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,630 @@
...
+* When span is created, set the attribute with key `previous-rpc-attempts` and an 
+  integer value representing the count of previous attempts made for the RPC.
+* When span is created, set the attribute with key `transparent-retry` and a 
+  boolean value indicating whether the stream is undergoing a transparent retry.
+* If the RPC experienced load balancer pick delay, add an Event with the name 
+  "Delayed LB pick complete" upon creation of the stream on the transport. 
+* When the application sends an outbound message, add Event(s)
+  (it depends on implementation whether there is a single event or an additional 
+  separate event with name "Outbound message compressed" for compressed message size) 
+  with name "Outbound message sent" and the following attributes:
```
yashykt, line 166: I think we talked about this earlier, but "sent" in "Outbound message sent" is confusing. Can we drop it?
For C++, this would otherwise mean that we would first see - 
"Outbound message sent" and then "Outbound message compressed"

A72-open-telemetry-tracing.md
yashykt: Eric is right. Till we actually go and compress, we won't know whether we will compress it.

A72-open-telemetry-tracing.md
yashykt: On the inbound message, we know from the wire format whether the message was compressed, so this is fine.

But, it is illegal to have an attribute named `message.event.size` if there is another attribute `message.event.size.uncompressed` where  `message.event.size` is the namespace.

A72-open-telemetry-tracing.md
yashykt: Also, it is illegal to have an attribute named `message.event.size` if there is another attribute `message.event.size.uncompressed` where  `message.event.size` is the namespace.

Refer https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/

yashykt (changes requested):

A72-open-telemetry-tracing.md
YifeiZhuang: > "Outbound message sent" and then "Outbound message compressed"

It looks fine. "sent" means this trace happens somewhere from application to the wire, and we made that vague and general(decision made from a discussion). Are you confused because compression should not happen after "sent"?

A72-open-telemetry-tracing.md
YifeiZhuang: > it is illegal to have an attribute named message.event.size if there is another attribute message.event.size.uncompressed where message.event.size is the namespace.

Good point. I have two naming proposals:
1. Extending from opencensus and use "-" to form the message: e.g. "message-sequence-no", "message-size","message-size-compressed".  
3. Follow ot spec and use ".", then we may end up with "grpc.previous.rpc.attempts", "grpc.transparent.retry", "grpc.message.sequence.no", "grpc.message.size", "grpc.compressed.message.size", or similar.

I'm fine either way. Does anyone have any other naming suggestions?

A72-open-telemetry-tracing.md
YifeiZhuang: I gave an example of overloading the description with grpc status code.

YifeiZhuang:
@yashykt  Thanks for the review. Eric and I made improvements on the otel message size reporting details (mainly https://github.com/grpc/proposal/pull/389/commits/8afa3173b673ea3b930336f2ef40a4d650b36f53), please take a look and comment whether it is feasible in C++.

cc. @ejona86 @dfawley

A72-open-telemetry-tracing.md
```diff
@@ -0,0 +1,675 @@
...
+std::unique_ptr<opentelemetry::context::TextMapPropagator>
+MakeGrpcTraceBinTextMapPropagator() {
+  return std::make_unique<internal::GrpcTraceBinTextMapPropagator>();
+}
+
+}  // namespace grpc
+
+```
+
+With gRPC OpenTelemetry API, users can provide a single composite propagator that
```
ejona86, line 575: This looks to be in the C++ section. Should it be there?

ejona86 (approved):

PR #390, opened 2023-08-23
A42 and A52 update: clarify aggregate cluster handling, and LB policy registry algorithm

murgatroid99:
This change clarifies two points that came up while I was implementing A52 in Node:

 1. If the `xds_cluster_resolver` config is generated from an aggregate cluster tree, the `xds_lb_policy` field should be determined by the root cluster config.
 2. When converting a `LoadBalancingPolicy` message to an internal LB policy config, if a `TypedStruct` policy is encountered, we should check whether the type name is in the gRPC LB policy registry, and if it is not, the policy should be skipped.

There are some other changes with no substance because my editor automatically removes trailing whitespace.

### DIFF
--- a/A42-xds-ring-hash-lb-policy.md
+++ b/A42-xds-ring-hash-lb-policy.md
@@ -257,6 +257,10 @@ hard-coded LB policy config fields in the CDS response:
 - For `RING_HASH`, the policy name will be `"RING_HASH"`, and the config will
   be the one for the `ring_hash_experimental` LB Policy described below.
 
+If the `xds_cluster_resolver` config is generated from an aggregate cluster
+tree, the value of this field will be determined from the CDS update for
+the root cluster.
+
 Despite this new config field being structured as a `LoadBalancingConfig`
 message, it will not actually indicate a child policy; instead, the
 functionality to understand the differences between `ROUND_ROBIN` and
--- a/A52-xds-custom-lb-policies.md
+++ b/A52-xds-custom-lb-policies.md
@@ -2,7 +2,7 @@ A52: gRPC xDS Custom Load Balancer Configuration
 ----
 * Author(s): temawi
 * Approver: markdroth
-* Implemented in: 
+* Implemented in:
 * Last updated: 2022-04-22
 * Discussion at: https://groups.google.com/g/grpc-io/c/BoRazEVUEz0
 
@@ -50,7 +50,7 @@ The XdsClient will transfer the responsibility of converting the xDS configurati
 
 The CDS resource is rejected and a NACK returned to the control plane in either of the following cases:
 
-- xDS LB policy registry cannot convert the configuration and returns an error 
+- xDS LB policy registry cannot convert the configuration and returns an error
 - gRPC LB policy registry cannot parse the converted configuration.
 
 ### xDS LB Policy Registry
@@ -67,7 +67,7 @@ The registry will provide the following function:
 //
 // An error/exception will be returned if the message type
 // is not recognized, or if policy configuration is invalid in a way that
-// can't be determined by looking at the PolicyConfig output. 
+// can't be determined by looking at the PolicyConfig output.
 string convertToServiceConfig(envoy.config.cluster.v3.LoadBalancingPolicy loadBalancingPolicy)
 ```
 
@@ -89,6 +89,7 @@ This function will iterate over the list of `Policy` messages in `LoadBalancingP
 - `xds.type.v3.TypedStruct` or `udpa.type.v1.TypedStruct`
   - This type represents a custom policy, deployed with the gRPC client by a user
   - The gRPC policy name will be the "type name" part of the value of the `type_url` field in the `TypedStruct`. We get this by using the part after the last `/` character.
+  - If the type name derived in the above bullet point does not correspond to a registered policy in the gRPC LB policy registry, this policy is unsupported and should be skipped.
   - The `Struct` contained in the `TypedStruct` will be returned as-is as the configuration JSON object.
   - Note that when registering custom policy implementations in the gRPC load balancer registry, the name should follow valid protobuf message naming conventions and use a custom package, e.g. "`myorg.MyCustomLb`".
 
@@ -144,7 +145,7 @@ The cluster resolver configuration documentation in service_config.proto is upda
 ```protobuf
 // Configuration for xds_cluster_resolver LB policy.
 message XdsClusterResolverLoadBalancingPolicyConfig {
-  // xDS LB policy. Will be used as the child config of the xds_cluster_impl LB policy. 
+  // xDS LB policy. Will be used as the child config of the xds_cluster_impl LB policy.
   repeated LoadBalancingConfig xds_lb_policy = 2;
 
 }
@@ -153,6 +154,10 @@ message XdsClusterResolverLoadBalancingPolicyConfig {
 To provide the `xds_wrr_locality` load balancer information about locality weights received from EDS, the cluster resolver will populate a new locality weight attribute for each address. The attribute will have the weight (as an integer) of the locality the address is part of. This alongside the already existing locality attribute will be used to generate the `weighted_target` configuration by `xds_wrr_locality`. Note that a change in this attribute would still allow associated subchannels to be reused - it will not affect their uniqueness.
 
 Since an xDS configuration can place a given locality under multiple priorities, it is possible to see locality weight attributes with different values for the same locality. We do not support this kind of an edge case and just use the weight in the first attribute we encounter. A warning should be logged about any locality weights being discarded.
+
+### CDS Load Balancer
+
+In the `xds_cluster_resolver` config, the `xds_lb_policy` field comes directly from the field of the same name in the CDS update struct. If the `xds_cluster_resolver` config is generated from an aggregate cluster tree, the value of the `xds_lb_policy` field will be determined from the CDS update for the root cluster.
 
 ### Configuration Example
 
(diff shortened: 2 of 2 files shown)


A42-xds-ring-hash-lb-policy.md
```diff
@@ -257,6 +257,10 @@ hard-coded LB policy config fields in the CDS response:
 - For `RING_HASH`, the policy name will be `"RING_HASH"`, and the config will
   be the one for the `ring_hash_experimental` LB Policy described below.
 
+If the `xds_cluster_resolver` config is generated from an aggregate cluster
```
markdroth, line 260: Hmm... Now that I'm looking at the text here, why is this necessary to say in A42?  The aggregate cluster functionality was actually introduced in A37, and the code described in this text was added to the CDS policy as part of implementing A37.  Nothing in A42 actually changes this.

I think this text should instead go in this section in A37:

https://github.com/grpc/proposal/blob/master/A37-xds-aggregate-and-logical-dns-clusters.md#cds-lb-policy-support-for-aggregate-and-logical-dns-clusters

A52-xds-custom-lb-policies.md
```diff
@@ -154,6 +155,10 @@ To provide the `xds_wrr_locality` load balancer information about locality weigh
 
 Since an xDS configuration can place a given locality under multiple priorities, it is possible to see locality weight attributes with different values for the same locality. We do not support this kind of an edge case and just use the weight in the first attribute we encounter. A warning should be logged about any locality weights being discarded.
 
+### CDS Load Balancer
+
+In the `xds_cluster_resolver` config, the `xds_lb_policy` field comes directly from the field of the same name in the CDS update struct. If the `xds_cluster_resolver` config is generated from an aggregate cluster tree, the value of the `xds_lb_policy` field will be determined from the CDS update for the root cluster.
```
markdroth, line 160: I don't think we need the second sentence here.  As mentioned above, I think this should be stated in A37, and nothing in A52 actually changes it, so I don't think it needs to be stated here.

A42-xds-ring-hash-lb-policy.md
murgatroid99: A37 does not describe any mechanism at all for populating the `locality_picking_policy` and `endpoint_picking_policy` fields of the `xds_cluster_resolver` config, so it doesn't make sense to me to add a clarification there of how those fields should be populated in one specific scenario. A42 is the first gRFC that describes such a mechanism, so it makes the most sense to me to clarify its behavior there.

A52-xds-custom-lb-policies.md
murgatroid99: As I said in the other comment, I don't think it makes sense to say this in A37, and since A42 and A52 can each be implemented without the other, I think it makes sense to make this clarification in both.

markdroth: I don't think it's true that A52 can be implemented without A42.  (It doesn't require the ring_hash policy itself, but it does require the structural changes in A42.)  Without A42, the `xds_cluster_resolver` config still encodes the LB policy config in the split `locality_picking_policy` and `endpoint_picking_policy` fields, and that does not provide a single way to send the combined LB policy config down.

I think A37 is a prereq for A42, and A42 is a prereq for A52.  And neither A42 nor A52 actually change the aggregate cluster behavior described in this text, which was actually introduced in A37.

A42-xds-ring-hash-lb-policy.md
markdroth: Also, I think the "one specific scenario" you mention is a reference to aggregate clusters, and that's exactly what A37 introduces, so I think it's totally on-topic to cover it there.

murgatroid99: "Mechanism" may not have been exactly the right word here. A37 does not describe what fields in the CDS update should be used to determine the values of the `locality_picking_policy` and `endpoint_picking_policy` fields.

A52-xds-custom-lb-policies.md
murgatroid99: A52 explicitly says that the `xds_cluster_resolver` config has a single `xds_lb_policy` field that is "used as the child config of the xds_cluster_impl LB policy". The only thing there that actually depends on A42 is the part about "Move Load Reporting into xds_cluster_impl Policy", because A52 needs that change for the same reason that A42 needs it.

A42-xds-ring-hash-lb-policy.md
murgatroid99: If we add a statement in A37 that says "In the aggregate cluster case, the values of the `locality_picking_policy` and `endpoint_picking_policy` fields should be determined from the root cluster update," then an obvious next question most readers will have is "*How* should the values of those fields be determined from the root cluster update?" Since A37 does not answer this question, this change would increase confusion rather than resolve it.

markdroth: Okay, I had to go digging back through some old PRs to remind myself of the history here, but I now see what happened.  The aggregate cluster functionality was implemented in C-core in https://github.com/grpc/grpc/pull/25078.  At the time, we didn't actually support any LB policies except for ROUND_ROBIN, so the CDS validation code rejected any resource with any other policy, and there actually wasn't any field in the CDS update indicating the LB policy, because there was no possible value other than ROUND_ROBIN and therefore no reason to specify it.  The CDS policy was simply hard-coding the values in the `xds_cluster_resolver` config: it set `locality_picking_policy` to `weighted_target` and `endpoint_picking_policy` to `round_robin`.  (We originally added those two fields in the `eds_experimental` policy that was created in the internal design go/grpc-xds-policy-refactoring, which was then split into `xds_cluster_resolver` and `xds_cluster_impl` in A37.  One of the goals of go/grpc-xds-policy-refactoring was to lay the groundwork for hierarchical LB policies, which is why these fields were added there.)

So you're right in the sense that the question of which cluster's LB policy config to use in an aggregate cluster was not relevant until A42, since that's the first time we had a choice of LB policies.  But at the same time, I still think that the fact that the LB policy comes from the aggregate cluster instead of the underlying clusters is a key part of the definition of an aggregate cluster and therefore needs to be specified in A37.  So, I propose the following:

- In A37, add text saying something like this: "In an aggregate cluster, the LB policy config comes from the aggregate cluster, not the underlying clusters.  At present, we do not support any LB policy except ROUND_ROBIN, so this decision is moot, and the CDS policy will hard-code the `locality_picking_policy` field to `weighted_target` and the `endpoint_picking_policy` field to `round_robin`.  However, in the future, when we add support for other LB policies, the CDS policy will be changed to get the LB policy config from the aggregate cluster."
- Here in A42, say something like this: "As mentioned in gRFC A37, the LB policy config for an aggregate cluster comes from the aggregate cluster, not the underlying clusters.  The CDS policy will select that config when constructing the `xds_lb_policy` field for the `xds_cluster_resolver` policy."

A52-xds-custom-lb-policies.md
markdroth: A42 explicitly covers replacing the `locality_picking_policy` and `endpoint_picking_policy` fields with the `xds_lb_policy` field:

https://github.com/grpc/proposal/blob/e045287a90083bd448280c94499e256bd2b12bc4/A42-xds-ring-hash-lb-policy.md#change-child-policy-config-generation-in-xds_cluster_resolver-policy

A52 does not make sense without that.

A42-xds-ring-hash-lb-policy.md
murgatroid99: My interpretation of A37 when I implemented it was that the `locality_picking_policy` and `endpoint_picking_policy` could be empty (since the documentation describes them as "optional") and that `xds_cluster_resolver` policy should know that the defaults are `weighted_target` and `round_robin` respectively, and fall back to those defaults if the fields are empty. So my CDS policy just left those fields empty and let the child policy sort it out.

markdroth: Sure, that's fine.  I don't think that affects the substance of what we're talking about here.  Regardless of whether the CDS policy leaves the fields unset and lets them default or explicitly sets them to the same defaults, the overall structure is still the same in terms of where decisions are being made.

murgatroid99: I made the change to A37, and updated the wording of the others. I kept the mention of this point in A52, because A52 is making some changes to the CDS update and the logic for converting it to an `xds_cluster_resolver` config, and I think it's worth mentioning explicitly that this part of that logic is not changing.

A37-xds-aggregate-and-logical-dns-clusters.md
```diff
@@ -239,6 +239,15 @@ a single priority with no endpoints. An EDS discovery mechanism should do the
 same if its watcher reports that the resource does not exist (whether or not
 it has previously reported a result).
 
+In an aggregate cluster, the `locality_picking_policy` and
```
markdroth, line 242: I think this paragraph actually belongs in the "CDS LB Policy Support for Aggregate and Logical DNS Clusters" section above, since it's talking about a responsibility of the CDS policy, not the xds_cluster_resolver policy.

Since that section is before the description of the xds_cluster_resolver LB policy config, you can either say something like "the `locality_picking_policy` and `endpoint_picking_policy` fields in the `xds_cluster_resolver` policy config (see below)", or you can just avoid mentioning the individual field names and instead say something like "the LB policy config passed down to the `xds_cluster_resolver` policy".

markdroth (approved):
Looks great!  Thanks for doing this!

PR #391, opened 2023-08-25
L106: Node Heath Check Library 2.0

murgatroid99:


L106-node-heath-check-library.md
```diff
@@ -0,0 +1,70 @@
...
+
+We will publish version 2.0 of the grpc-health-check library with the following API, defined in TypeScript:
+
+```ts
+type ServingStatus = 'UNKNOWN' | 'SERVING' | 'NOT_SERVING';
+interface ServingStatusMap {
+  [serviceName: string]: ServingStatus;
+}
+
+class HealthImplementation {
```
wenbozhu, line 35: Do we need to call out that this is server-only, e.g. ServerHealthImplementation ?

wenbozhu (approved):

L106-node-heath-check-library.md
murgatroid99: The current API uses a similar name to refer to an object with similar functionality, and no user that I have seen has expressed any confusion, so I think the name is fine as it is.

PR #393, opened 2023-09-08
A42 update: Clarify a few points in the design

murgatroid99:
This clarifies a few points that I found confusing:

 - How hashes from multiple hash policies should be combined
 - How the `terminal` bit should be handled
 - What number "8M" refers to
 - How the ring should be constructed


markdroth (approved):
Thanks for the clarifications!

PR #395, opened 2023-10-03
L107: Node: Make `Server#start` a no-op

murgatroid99:

PR #397, opened 2023-10-10
L108: Node.js Reflection Server Library

jtimmons:
adding proposal as discussed in grpc/grpc-node#79

### DIFF
--- a/L108-node-grpc-reflection-library.md
+++ b/L108-node-grpc-reflection-library.md
@@ -27,7 +27,7 @@ This feature was initially requested in grpc/grpc-node#79
 
 ## Proposal
 
-We are proposing the creation of a new `@grpc/reflection-server` package with the following external interface:
+We are proposing the creation of a new `@grpc/reflection` package with the following external interface:
 
 ```ts
 import type { Server as GrpcServer } from '@grpc/grpc-js';
--- a/L108-node-grpc-reflection-library.md
+++ b/L108-node-grpc-reflection-library.md
@@ -35,14 +35,17 @@ import type { PackageDefinition } from '@grpc/proto-loader';
 
 type MinimalGrpcServer = Pick<GrpcServer, 'addService'>;
 
+interface ReflectionServerOptions {
+  services?: string[]; // whitelist of fully-qualified service names to expose. Default: expose all
+}
+
 export interface ReflectionServer {
-  constructor(pkg: PackageDefinition);
-  reload(pkg: PackageDefinition);
+  constructor(pkg: PackageDefinition, options?: ReflectionServerOptions);
   addToServer(server: MinimalGrpcServer);
 }
 ```
 
-this class will be responsible for managing requests for the various published versions of the gRPC Reflection Specification. At the time of writing, this includes `v1` and `v1alpha` but may include more in the future. These version-specific handlers will be isolated into their own services, such as the following which can be used for both `v1` and `v1alpha` due to their identical interface:
+this `ReflectionServer` class will be used to expose information about the user's gRPC package according to the gRPC Reflection Specification via their existing gRPC Server. Internally, the class will be responsible for managing incoming requests for each of the various published versions of the gRPC Reflection Specification: at the time of writing, this includes `v1` and `v1alpha` but may include more in the future. These version-specific handlers will be isolated into their own services in order to preserve backwards-compatibility, and will look like the following:
 
 **reflection.v1.ts**
 ```ts
@@ -82,8 +85,6 @@ server.addService(proto.sample.SampleService.service, { ... });
 reflection.addToServer(server)
 
 server.bindAsync('0.0.0.0:5001', grpc.ServerCredentials.createInsecure(), () => { server.start(); });
-
-reflection.reload({ ... }) // can reload later on if additional services are added
 ```
 
 ## Rationale
@@ -93,6 +94,14 @@ several reflection implementations linked above leverage protoc in order to gene
 
 ### 2. Design Decision: support multiple reflection implementations
 currently not all reflection clients request the `v1` version of the spec so we need to include handlers for both `v1` and `v1alpha` to support both during the transition. For this reason we separate the reflection handling logic itself to allow for reuse across multiple service versions
+
+### 3. Design Decision: restrict user to loading a single PackageDefinition
+
+ideally we would be able to support the user adding multiple `PackageDefinition` objects at a time in a similar way to the gRPC server itself, however due to some internal protobuf behavior discussed in [this thread](https://github.com/grpc/proposal/pull/397#discussion_r1357181337) this is currently difficult to accomplish. For that reason we will be restricting the user to loading a single PackageDefinition for the time being to avoid any confusing behavior or bugs. Detail on the issue is described below for completeness:
+
+**background**: when loading a `PackageDefinition` object via the `protoLoader.load(...)` function, proto-loader/protobufjs will rename the input `.proto` files based on their [protobuf package](https://protobuf.dev/programming-guides/proto3/#packages) name. For example a file named _file.proto_ in the `sample` package will actually be referred to as _sample.proto_ in all `FileDescriptorProto` objects in the resulting `PackageDefinition`.
+
+This behavior can cause issues when multiple files exist within the _same_ package as there can be confusion about what is the "real" contents of a file (which is critical information for the reflection API). Proto-loader/protobufjs handles this for a _single_ `load()` call by unifying all files into a single package-file; for example: if we have files _vendor/a.proto_ and _vendor/b.proto_ which are both in the `vendor` protobuf namespace then contents from both files will be combined into a single _vendor.proto_ file descriptor in the `PackageDefinition`. The issue arises when we attempt multiple invocations of `protoLoader.load()` as each may only fetch a subset of the package (in this example from _a.proto_ in the first invocation and _b.proto_ in the second). In these cases we have multiple different _vendor.proto_ references which breaks the assumptions of the reflection specification in which files are often looked up by name.
 
 ## Implementation
 
--- a/L108-node-grpc-reflection-library.md
+++ b/L108-node-grpc-reflection-library.md
@@ -1,0 +1,115 @@
+Node.js Reflection Server Library
+----
+* Author(s): jtimmons
+* Approver:
+* Status: Draft
+* Implemented in: Node.js
+* Last updated: 2023-10-09
+* Discussion at:
+
+## Abstract
+
+Create a canonical implementation of the gRPC reflection API for Node.js based on the logic from the [nestjs-grpc-reflection library](https://gitlab.com/jtimmons/nestjs-grpc-reflection-module/-/blob/30b67a78ff99e31ae54a0ab34c3784316579c665/src/grpc-reflection/grpc-reflection.service.ts)
+
+## Background
+
+Since its introduction in 2017 there have been a variety of external node.js implementations for the [gRPC Reflection Specification](https://github.com/grpc/grpc/blob/ce75ec23a1a9c5239834b92da4ce0992d367a39c/doc/server-reflection.md), each of which is in various states of maintenance. A few examples are linked at the bottom of this section.
+
+This feature was initially requested in grpc/grpc-node#79
+
+* https://gitlab.com/jtimmons/nestjs-grpc-reflection-module
+* https://github.com/deeplay-io/nice-grpc/tree/master/packages/nice-grpc-server-reflection
+* https://github.com/AckeeCZ/grpc-server-reflection
+
+### Related Proposals:
+* Initial reflection proposal: https://github.com/grpc/proposal/blob/master/A15-promote-reflection.md
+* Proposal of API for similar library: https://github.com/grpc/proposal/blob/master/L106-node-heath-check-library.md
+
+## Proposal
+
+We are proposing the creation of a new `@grpc/reflection-server` package with the following external interface:
+
+```ts
+import type { Server as GrpcServer } from '@grpc/grpc-js';
+import type { Options as ProtoLoaderOptions } from '@grpc/proto-loader';
+
+type MinimalGrpcServer = Pick<GrpcServer, 'addService'>;
+
+interface ReflectionOptions {
+  filename: string | string[];
+  loader: ProtoLoaderOptions;
+}
+
+export interface ReflectionServer {
+  constructor(options?: ReflectionOptions);
+  load(options: ReflectionOptions);
+  loadSync(options: ReflectionOptions);
+
+  addToServer(server: MinimalGrpcServer);
+}
+```
+
+this class will be responsible for managing requests for the various published versions of the gRPC Reflection Specification. At the time of writing, this includes `v1` and `v1alpha` but may include more in the future. These version-specific handlers will be isolated into their own services, such as the following which can be used for both `v1` and `v1alpha` due to their identical interface:
+
+**reflection.v1.ts**
+```ts
+import {
+  ExtensionNumberResponse,
+  FileDescriptorResponse,
+  ListServiceResponse,
+} from './proto/grpc/reflection/v1/reflection';
+
+export interface ReflectionV1Implementation {
+  async load(options: ReflectionOptions);
+  loadSync(options: ReflectionOptions);
+
+  listServices(listServices: string): ListServiceResponse;
+  fileContainingSymbol(symbol: string): FileDescriptorResponse;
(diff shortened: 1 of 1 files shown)


murgatroid99:
I took the liberty of updating the PR title. The proposal itself is for the "Node.js Reflection Server Library", and the status can be managed within the text of the proposal document. Please list me as the approver, create a discussion thread as described in the [README's process section](https://github.com/grpc/proposal#process), and update the status in the document to "In Review".

In addition, L107 is already reserved for #395, so please update the PR title and text to L108.

L108-node-grpc-reflection-library.md
```diff
@@ -0,0 +1,115 @@
...
+
+```ts
+import type { Server as GrpcServer } from '@grpc/grpc-js';
+import type { Options as ProtoLoaderOptions } from '@grpc/proto-loader';
+
+type MinimalGrpcServer = Pick<GrpcServer, 'addService'>;
+
+interface ReflectionOptions {
+  filename: string | string[];
+  loader: ProtoLoaderOptions;
```
murgatroid99, line 40: Which of these options actually impact the reflection data?

L108-node-grpc-reflection-library.md
```diff
@@ -0,0 +1,115 @@
...
+* https://github.com/deeplay-io/nice-grpc/tree/master/packages/nice-grpc-server-reflection
+* https://github.com/AckeeCZ/grpc-server-reflection
+
+### Related Proposals:
+* Initial reflection proposal: https://github.com/grpc/proposal/blob/master/A15-promote-reflection.md
+* Proposal of API for similar library: https://github.com/grpc/proposal/blob/master/L106-node-heath-check-library.md
+
+## Proposal
+
+We are proposing the creation of a new `@grpc/reflection-server` package with the following external interface:
```
murgatroid99, line 30: If we ever want to provide reflection client functionality in the future, I think we would want it in the same package, so I would prefer to call the package `@grpc/reflection`.

murgatroid99 (commented):
As we discussed in grpc/grpc-node#79, it should be possible for the Server class to expose all of the necessary information to support reflection. The reflection library should have an API to take advantage of that by making it as simple as possible to add reflection, e.g. it could export a top-level function `enableReflection(server)`. This is also important because a server can have services based on proto files that the user does not have direct access to: in grpc-js, the `addAdminServicesToServer` adds the channelz service, and also the CSDS service if using xDS.

L108-node-grpc-reflection-library.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Implementation
+
+I (jtimmons) will implement this once the maintainers have approved
+
+## Open issues (if applicable)
+
+### 1. Package Publishing
+
+Not sure how the package versioning and publishing works but may need some help from a maintainer to set that up and publish things into the `@grpc` scope
```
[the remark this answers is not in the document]
jtimmons: okay great so I won't worry about that for now then

L108-node-grpc-reflection-library.md
jtimmons: makes sense, will rename 👍

L108-node-grpc-reflection-library.md
murgatroid99: I think the library should provide the minimum set of relevant options, and if something changes, that change should be made explicit with a change to the API.

L108-node-grpc-reflection-library.md
```diff
@@ -0,0 +1,115 @@
...
+
+### 2. keepCase Option
+One of the more awkward pieces in here is dealing with grpc-js/proto-loader's `keepCase` option, though this isn't unique to this library. Consumers can have their gRPC server configured with `keepCase` either on or off which changes the message handling substantially: for example if enabled then we would need to refer to the `listServices` field as `list_services` instead.
```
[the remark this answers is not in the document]
jtimmons: ah that's a good point, this may be unique to the NestJS ecosystem then since there's a global proto-loader config. I'll open an issue there

L108-node-grpc-reflection-library.md
jtimmons: sure, I can restrict that to just take in the include directory path. But to be clear: at the moment the reflection logic is not actually using *any* of these options itself and is instead just [passing them directly through to proto-loader](https://gitlab.com/jtimmons/nestjs-grpc-reflection-module/-/blob/30b67a78ff99e31ae54a0ab34c3784316579c665/src/grpc-reflection/grpc-reflection.service.ts#L66) so that we're guaranteed to get the exact same package definition that the consumer would use in their server

thinking about it more though it may actually make more sense to just take in the `PackageDefinition` and inspect that since we don't need the options anywhere else. This would also eliminate the need for separate async vs sync loading functions

murgatroid99: Do you just assume that every service in the package definition is being served? I don't see any API for specifying which services are being served. Instead of a `PackageDefinition`, it might make more sense to pass `ServiceDefinition` objects.

L108-node-grpc-reflection-library.md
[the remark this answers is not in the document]
jtimmons: for completeness this is now tracked in the NestJS repo under nestjs/nest#12560

L108-node-grpc-reflection-library.md
jtimmons: yes the assumption now is that every service in the `PackageDefinition` is being served which is a _bit_ awkward but we do need the full package in order to index all of the non-service symbols that may be looked up by name

I could add an option for an array of service names to expose if it seems likely that people would have package definitions with services that they don't support?

jtimmons: also for some additional context I'm intentionally only allowing a single `PackageDefinition` here because of [this bug](https://gitlab.com/jtimmons/nestjs-grpc-reflection-module/-/issues/15#note_1592086353) in the reflection module that was fixed last week

**problem**: proto-loader will rename the proto files based on the package so if we have files `vendor/a.proto` and `vendor/b.proto` both in the `vendor` namespace then those _should_ both be combined into a single `vendor.proto` file descriptor in the `PackageDefinition`. The issue with this is that different invocations of proto-loader may only fetch a subset of the package so the `vendor.proto` file descriptor can be different from one invocation to another

**solution**: internally we just used `protoLoader.load` on _all_ files simultaneously so that it would unify all of them into a single file definition. In this spec we can force the user to do the same by only allowing a single `PackageDefinition`

murgatroid99: > I could add an option for an array of service names to expose if it seems likely that people would have package definitions with services that they don't support?

The first example I thought of is our own [test.proto](https://github.com/grpc/grpc/blob/master/src/proto/grpc/testing/test.proto). This may be an artificial or unusual example, but that file defines 7 services. One of those services is by definition not provided by any server, and to my knowledge no single server implements all 6 of the others.

> also for some additional context I'm intentionally only allowing a single PackageDefinition here because of [this bug](https://gitlab.com/jtimmons/nestjs-grpc-reflection-module/-/issues/15#note_1592086353) in the reflection module that was fixed last week

I know this isn't a distinction that matters from that position, but where you said "`@grpc/proto-loader` does replace the actual file structure with one that's based on the package structure", there is no intentional code in proto-loader that does that, so it seems to be the behavior of Protubuf.js. If that behavior is causing problems for reflection, we should probably look into fixing it.

jtimmons: > The first example I thought of is our own [test.proto](https://github.com/grpc/grpc/blob/master/src/proto/grpc/testing/test.proto). This may be an artificial or unusual example, but that file defines 7 services. One of those services is by definition not provided by any server, and to my knowledge no single server implements all 6 of the others.

it's a bit of a tricky distinction because it's not really "what the server implements" so much as "what the server wants to expose". In the case of that test.proto's `UnimplementedService` I think we'd actually _want_ to expose that since the test is expressly for calling the method and getting an unimplemented message back from the server

maybe the right concept here is for blacklisting services to hide them from the reflected output? I've seen a case in the wild where people were accidentally exporting their service definition along with some common message definitions so the server was reflecting that as well. It wouldn't be too difficult to add a `{ hideServices: ['accidentally.added.Service'] }` to let people control that a bit better

> it seems to be the behavior of Protubuf.js. If that behavior is causing problems for reflection, we should probably look into fixing it

it would definitely solve a big class of issues if the file descriptors were mapped to the files-on-disk but I'm not sure how much work that would be and if anything relies on that current file-remapping behavior. For now, limiting the reflection server to only accept a single `PackageDefinition` at a time should prevent the user from hitting that case - if that seems good enough for a first API I can create an issue for the file remapping so that can be addressed later?

murgatroid99: Even ignoring `UnimplementedService`, the standard interop server exposes only the `Test` service, while the xDS interop client server exposes only the `LoadBalancerStatsService` and `XdsUpdateClientConfigureService` services. And an explicit list of services to include or exclude doesn't solve the problem that servers may expose services from multiple `PackageDefinition` objects, such as the health service, channelz, or even the reflection service itself.

Fundamentally, I think the goal here should be automatically detecting the services that are actually exposed by a server, and providing that information through the reflection service. If that requires changes to the server API, and proto-loader, and even Protobuf.js, then so be it.

jtimmons: > Even ignoring UnimplementedService, the standard interop server exposes only the Test service, while the xDS interop client server exposes only the LoadBalancerStatsService and XdsUpdateClientConfigureService services.

ah gotcha, missed that - that makes sense!

> And an explicit list of services to include or exclude doesn't solve the problem that servers may expose services from multiple PackageDefinition objects, such as the health service, channelz, or even the reflection service itself.
> 
> Fundamentally, I think the goal here should be automatically detecting the services that are actually exposed by a server, and providing that information through the reflection service. If that requires changes to the server API, and proto-loader, and even Protobuf.js, then so be it.

definitely agreed, but from what I'm seeing in practice it seems more like some slightly awkward UX for the consumer in the current state rather than a hard limitation? The `PackageDefinition` object that proto-loader returns doesn't seem to need to map 1-1 to a single grpc package and instead holds a collection of top-level symbols from any number of packages which were included in the `load` call. A single `PackageDefinition` from consumers of the nest module frequently includes multiple packages and this is reflected in our [sample application](https://gitlab.com/jtimmons/nestjs-grpc-reflection-module/-/tree/v0.1.1/src/sample/01-register?ref_type=tags) as well. Here's an example of what that looks like for the sample application:

image

it would definitely be ideal for the API to be able to do something similar to `addService` but it wouldn't be terribly difficult to add a note to the docs for the time being to combine those package definitions if the user wants to expose multiple disparate services

murgatroid99: I think the biggest problem here is [admin services](https://github.com/grpc/proposal/blob/master/A38-admin-interface-api.md). The user does not directly specify the services to add, and instead calls a generic API to add services that will be determined by some internal code.

jtimmons: ah I see, so yeah it seems like the only way to really account for that would be to be able to accept multiple `PackageDefinition` objects one-by-one then. Currently the biggest hurdle to that is that remapping of file names to ones based on the package name then since that breaks our indexing

I'm not particularly familiar with the various layers of the grpc server but from my perspective there's a few things that could help solve the issue:
  1. Creating `FileDescriptorProto` objects based on the file on disk rather than the package name would just outright fix the issue
  2. Exposing some ability to combine `FileDescriptorProto` objects would allow us to maintain a consistent index as we receive more files from a package. I assume that there's logic for this somewhere in either proto-loader or protobufjs already considering we currently unify all files in a single package into a single "file"

any thoughts on how this should be addressed?

murgatroid99: I suspect the problem is that Protobuf.js loses/discards source file information when loading `.proto` files, so when it creates the descriptor proto objects from the loaded objects, it has to make something up to generate the `FileDescriptorProto` objects. Fixing that would probably require plumbing source file information through multiple places.

Merging separate `FileDescriptorSet` objects might be as simple as concatenating their `file` fields and then loading the result back in to `protoLoader.loadFileDescriptorSetFromObject`. I don't know what would happen if they contain duplicate definitions, though.

In principle, I would prefer to fix the file information, because it would mean that you could in theory take multiple separate service objects and merge their information in a way that is useful for reflection without examining the content of the individual `FileDescriptorProto` objects.

jtimmons: > Merging separate FileDescriptorSet objects might be as simple as concatenating their file fields and then loading the result back in to protoLoader.loadFileDescriptorSetFromObject. I don't know what would happen if they contain duplicate definitions, though.

I prototyped this idea a bit [here](https://gitlab.com/jtimmons/nestjs-grpc-reflection-module/-/commit/121f85200888e868b9720dfe8f06b2637f33de2e?view=inline&w=1#4fe7c10adfe8e1f9005d64b781c8d8e6fb82660b_58_69) but hit a couple of issues with it:
  1. This does fail for duplicate definitions which means that at a minimum we need to dedupe files that are coming in. Obviously going by file name doesn't work since multiple files in the same namespace will map there, so if there's a good contents hash or something to go by that _might_ work - otherwise I haven't found anything that would work well there across multiple calls
  3. For some reason the `protoLoader.loadFileDescriptorSetFrom{Buffer,Object}` methods are failing on proto2 extension types. I've seen this exact error elsewhere where files were improperly marked as proto3 so there may be some incorrect assumptions on proto version in those methods somewhere?
image


> In principle, I would prefer to fix the file information, because it would mean that you could in theory take multiple separate service objects and merge their information in a way that is useful for reflection without examining the content of the individual FileDescriptorProto objects.

that definitely sounds ideal but I'm not sure that I have the time or the expertise to take something of that scope on, I'll keep poking at some of the simpler options to see if I can get anywhere. Otherwise might it be worth launching with a documented limitation that the reflection information is static at the server startup time for the time being?

jtimmons: @murgatroid99 I spent a couple days looking into options for multiple service loading+indexing and I just don't think I've got the expertise for that big of a change across the various proto-loader/protobufjs layers. For the sake of keeping this proposal moving along I'd like to either:
  1. Launch this with just the ability to load a single bundle of static packages, make a note in the docs, and circle back on adding support for adding services to it dynamically later. Or,
  2. I can contribute the base capability in a PR if someone more familiar with the codebases can contribute the dynamic package logic on top of it

thoughts on this?

murgatroid99: I think I would be OK with an API that accepts a single package definition object plus a list of the names of all services that are actually exposed on the server. That way we both avoid the need to curate the package definition to only contain the right services, and we can also implement some dynamic behavior by including an API to add or remove service names after initial construction. One optional behavior could be omitting the list of names, and automatically including every service in the package definition.

In addition, if we find a way to merge package definitions, that can both improve the power of that basic API, and give us the ability to add a new API to add more reflection data after starting.

I would still like to find a way to track everything automatically, but it doesn't seem like we have the foundations in place to accomplish that right now.

jtimmons: okay, updated the proposal with the final API and some reasoning about this decision

L108-node-grpc-reflection-library.md
```diff
@@ -0,0 +1,123 @@
...
+
+This behavior can cause issues when multiple files exist within the _same_ package as there can be confusion about what is the "real" contents of a file (which is critical information for the reflection API). Proto-loader/protobufjs handles this for a _single_ `load()` call by unifying all files into a single package-file; for example: if we have files _vendor/a.proto_ and _vendor/b.proto_ which are both in the `vendor` protobuf namespace then contents from both files will be combined into a single _vendor.proto_ file descriptor in the `PackageDefinition`. The issue arises when we attempt multiple invocations of `protoLoader.load()` as each may only fetch a subset of the package (in this example from _a.proto_ in the first invocation and _b.proto_ in the second). In these cases we have multiple different _vendor.proto_ references which breaks the assumptions of the reflection specification in which files are often looked up by name.
+
+## Implementation
+
+I (jtimmons) will implement this once the maintainers have approved
+
+## Open issues (if applicable)
```
murgatroid99, line 110: Don't include "open issues" that aren't actually open. Just exclude anything that is resolved.

However, there are actual open issues here regarding the behavior of proto-loader and the remaining obstacles to fully automatic reflection service generation. Please include explanations of those in this section.

jtimmons: done! Moved the `PackageDescription` issue from the design decisions area to open issues instead

murgatroid99 (approved):

PR #399, opened 2023-11-07
A73: Transport Selection

dfawley:


A73-transport-selection.md
```diff
@@ -0,0 +1,119 @@
...
+clients and servers. The exact details will be left up to gRPC implementations
+to decide; only the high-level method is included in this design.
+
+Some additional design goals of this proposal:
+
+1. Transport and stream APIs should expose gRPC semantics as defined in the
+   (pending) [Call Semantics
+   Specification](https://github.com/grpc/grpc/pull/15460) and hide HTTP/2
+   semantics and grpc's wire protocol semantics.
+1. Stream APIs should support object-based messages to facilitate an in-process
```
ejona86, line 44: Based on the end, it seems you're expecting to make a copy here. I'd suggest mentioning copying objects as the alternative to serialize+deserialize. Otherwise it isn't really clear what you're talking about. On the face of it it looked like you'd just pass the object from client to server, but that is fundamentally broken unless the objects are immutable.

A73-transport-selection.md
```diff
@@ -0,0 +1,119 @@
...
+* [A6: Client
+  Retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md)
+
+## Proposal
+
+This proposal is for a standard way of selecting which transport to use in gRPC
+clients and servers. The exact details will be left up to gRPC implementations
+to decide; only the high-level method is included in this design.
+
+Some additional design goals of this proposal:
```
ejona86, line 38: This is awkward. It is in the proposal section, and defines a mix of goals and design choices. I think "goals" isn't the right word here, as these are part of the proposal itself. Goals shouldn't be in "Proposal" anyway.

dfawley: Moved into a sub-section about transports.

A73-transport-selection.md
dfawley: Mentioned "copying".

dfawley (commented):
Updated; PTAL

PR #404, opened 2023-12-07
A74: xDS Config Tears

markdroth:


A74-xds-config-tears.md
```diff
@@ -0,0 +1,309 @@
...
+list for that aggregate cluster, but it will not have its own top-level
+entry.
+
+#### Changes in cds LB policy
+
+Instead of starting a watch on the XdsClient to get the cluster data,
+the cds LB policy will get its configuration by lookup up the cluster
+name in the XdsConfig passed to it via an attribute.
+
+To address the case where the cluster is determined dynamically (e.g., a
```
murgatroid99, line 225: Is this part about existing functionality? I don't see any existing mention of `ClusterSpecifierPlugin` in this repo.

murgatroid99: I think it would be good to have a note in this part about dynamic clusters that this is related to a non-public feature, don't think they missed something in some other gRFC.

murgatroid99 (approved):
Other than my one comment, this looks reasonable to me.

A74-xds-config-tears.md
markdroth: Done.

A74-xds-config-tears.md
```diff
@@ -0,0 +1,311 @@
...
+Therefore, when the XdsDependencyManager sees an error on a cluster for
+which it did not previously have data, or if it sees a does-not-exist
+notification on a cluster, it will report that by returning a non-OK
+status for the cluster in the XdsConfig.  Similarly, if there is a problem
+obtaining endpoint data from EDS or DNS, the XdsDependencyManager will
+report that problem via the cluster's resolution_note field.
+
+The xds resolver will pass the returned XdsConfig to the LB policies via
+an attribute, so that the LB policies can use the cluster data.  Note
+that we cannot simply pass the data via the cds LB policy's JSON config,
```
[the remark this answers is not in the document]
markdroth: In the dynamic cluster case, we don't have a way to pass the config down as JSON, because the config of the child is coming from the ClusterSpecifierPlugin policy, not from the xds resolver.

I've added some wording to try to clarify this.

A74-xds-config-tears.md
```diff
@@ -0,0 +1,311 @@
...
+every cluster that is either referenced in the route configuration or
+for which a dynamic subscription was registered.  If a cluster is
+present only underneath an aggregate cluster, it will be present in the
+list for that aggregate cluster, but it will not have its own top-level
+entry.
+
+#### Changes in cds LB policy
+
+Instead of starting a watch on the XdsClient to get the cluster data,
+the cds LB policy will get its configuration by lookup up the cluster
```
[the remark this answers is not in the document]
markdroth: Done.

A74-xds-config-tears.md
```diff
@@ -0,0 +1,311 @@
...
+the cds LB policy's configuration:
+
+```
+// Set this to true for a cluster that is determined dynamically
+// rather than being present in the route configuration.
+bool is_dynamic = 2;
+```
+
+If that field is set to true, then if the cluster name is not present in
+the XdsConfig, the cds policy will obtain a subscription handle from the
```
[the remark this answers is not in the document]
markdroth: The cds policy's config already has a field for the [cluster name](https://github.com/grpc/grpc-proto/blob/37330a01c4e1ac4d379bb81cc349df34aaf306cf/grpc/service_config/service_config.proto#L395).  When the ClusterSpecifierPlugin creates a cds policy for a given cluster, it sets that field so that the child policy knows which cluster it's for.

Today, that child cds policy gets the rest of the config by starting a watch on the XdsClient for that cluster name.  With this design, if the new `is_dynamic` field is set, then it will instead request a subscription handle for that policy from the XdsDependencyManager.  It will then wait for a subsequent address update (which the ClusterSpecifierPlugin will pass down to its children) that includes the updated xDS config with the new cluster.

A74-xds-config-tears.md
```diff
@@ -0,0 +1,311 @@
...
+XdsDependencyManager if it has not already done so.  This handle will
+not be released until the cds policy is destroyed.
+
+Note that when XdsDependencyManager returns a new configuration that
```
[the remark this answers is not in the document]
markdroth: I had the same idea when I was implementing this, but when I looked at the details, there were enough differences that it didn't really make sense to use a single mechanism for both.

The purpose of the cluster refs taken by the ConfigSelector is to make sure that we don't remove the cluster from the LB policy tree generated by the xds resolver if the cluster gets removed from the RouteConfig before the calls have finished their LB picks.  It's true that prior to this design, retaining the cds policy also had the side effect of retaining the CDS watch, but I'm not sure we actually need that -- in general, these refs should just keep the cds policy around for a fairly short amount of time, so it probably doesn't matter if we stop the CDS watch and just keep the policy around with its last-seen config, because there probably won't be any updates in that short window that we couldn't just ignore.  (Having just written this out, though, I would be open to an argument about edge cases that may make this window longer than we would normally expect, where we might not want to ignore subsequent CDS updates.)

In contrast, the clusters that are dynamically created by the ClusterSpecifierPlugin are never going to be present in the LB policy config generated by the xds resolver in the first place.  The purpose of the dynamic subscription is specifically to trigger the existence of the CDS watch, because unlike the ConfigSelector case, here we do need the CDS watch to exist for as long as the ClusterSpecifierPlugin decides to keep the cds child policy alive.

markdroth (commented):
Thanks for the review!  Please let me know if you have any questions.

A74-xds-config-tears.md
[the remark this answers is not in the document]
markdroth: As per our conversation today, I've changed this such that the cluster refs held by the ConfigSelector will also hold a subscription from the XdsDependencyManager, so we'll ensure that the CDS watch sticks around as long as the cluster is in the xds_cluster_manager LB policy config.

A74-xds-config-tears.md
```diff
@@ -0,0 +1,311 @@
...
+  struct ClusterConfig {
+    // Cluster name and resource.
+    std::string cluster_name;
+    XdsClusterResource cluster;
+    // Endpoint info.  If there was an error, endpoints will be null
+    // and resolution_note will be set.
+    XdsEndpointResource endpoints;
+    std::string resolution_note;
+  };
+  std::map<std::string, absl::StatusOr<std::vector<ClusterConfig>>> clusters;
```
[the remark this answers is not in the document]
markdroth: I thought about this a bit more and realized that I can actually back-port the changes I was going to make for A75 to this one without too much trouble.  I've changed this to store every cluster in its own entry in the map, and for aggregate clusters I am storing the flattened list of leaf cluster names, which the cds LB policy can look up in the map individually.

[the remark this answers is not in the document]
markdroth: ...And this also means that we can pull forward the JSON optimization for xds_cluster_impl and xds_override_host that I was going to do in A75, so I've pulled that change in here as well.

A74-xds-config-tears.md
```diff
@@ -0,0 +1,349 @@
...
+obtaining endpoint data from EDS or DNS, the XdsDependencyManager will
+report that problem via the cluster's resolution_note field.
+
+The xds resolver will pass the returned XdsConfig to the LB policies via
+an attribute, so that the LB policies can use the cluster data.  Note that
+we cannot simply pass the data via the cds LB policy's JSON config,
+because we need this to work in cases where the cluster is determined
+dynamically, such as where a route uses a ClusterSpecifierPlugin, in
+which case the LB policy's config comes from the ClusterSpecifierPlugin,
+not from the xds resolver.  (We have some internal functionality that
```
dfawley, line 216: I'm not following this fully.

ClusterSpecifierPlugin config is already determined in the xds resolver, is it not?  That creates a child cluster with the name `cluster_specifier_plugin:<plugin_name>` and using the child policy config generated by the CSP.  We do this here in Go:

https://github.com/grpc/grpc-go/blob/e20d5ce8c48ee70cfabf6d7dc73c4367ee529e04/xds/internal/resolver/xds_resolver.go#L312-L319

It sounds like you're saying some part of this is per-RPC and determines the structure of the LB policy tree somehow?

markdroth: When a route uses a ClusterSpecifierPlugin, the xds resolver does indeed create a `cluster_specifier_plugin:<plugin_name>` entry in the LB policy tree, which configures the policy for that plugin as a child of the xds_cluster_manager policy.  However, that plugin policy will then dynamically create cds policies as its children on the basis on individual RPCs.  The LB policy config for those cds child policies is generated by the ClusterSpecifierPlugin policy, not by the xds resolver, which is why we need some other way to pass the configuration down to them.

A74-xds-config-tears.md
```diff
@@ -0,0 +1,346 @@
+A74: xDS Config Tears
+----
+* Author(s): @markdroth
+* Approver: @ejona86
```
dfawley, line 4: :cry:

A74-xds-config-tears.md
```diff
@@ -0,0 +1,346 @@
...
+
+### Related Proposals: 
+* [A27: xDS Global Load Balancing][A27]
+* [A28: xDS Traffic Splitting and Routing][A28]
+* [A29: xDS mTLS Security][A29]
+* [A31: xDS Timeout Support and Config Selector Design][A31]
+* [A37: xDS Aggregate and Logical DNS Clusters][A37]
+* [A55: xDS-Based Stateful Session Affinity for Proxyless gRPC][A55]
+* [A56: Priority LB Policy][A56]
+* [A71: xDS Fallback (pending)][A71]
```
dfawley, line 60: Can be removed from pending status and not link to pull.

A74-xds-config-tears.md
```diff
@@ -0,0 +1,346 @@
...
+methods will be invoked only for problems with the listener and route
+configuration resources.  For cluster-specific errors, we need to
+retain the existing behavior, which is that it causes a problem for the
+individual cluster but does not affect traffic sent to any other cluster.
+Therefore, when the XdsDependencyManager sees an error on a cluster for
+which it did not previously have data, or if it sees a does-not-exist
+notification on a cluster, it will report that by returning a non-OK
+status for the cluster in the XdsConfig.  Similarly, if there is a problem
+obtaining endpoint data from EDS or DNS, the XdsDependencyManager will
+report that problem via the cluster's resolution_note field.
```
dfawley, line 207: Why not also make the `EndpointConfig` contain a `StatusOr`?

dfawley (approved):

A74-xds-config-tears.md
markdroth: Because there are cases where we set both the `endpoints` and the `resolution_note` field, such as when the EDS update contains an empty locality.

Note that this structure isn't really new; it just models the existing mechanism for passing endpoint addresses from the resolver to the LB policy tree, and we use the same representation for both.

A74-xds-config-tears.md
markdroth: Done.

A74-xds-config-tears.md
markdroth: Whoops -- sorry, not sure why you weren't listed here.  Fixed.

ejona86 (approved):

PR #405, opened 2023-12-14
A75: xDS Aggregate Cluster Behavior Fixes

markdroth:


A75-xds-aggregate-cluster-behavior-fixes.md
```diff
@@ -0,0 +1,312 @@
...
+  xds_override_host policy assumes ownership of the subchannel instead
+  of deleting it.
+- In the picker, when we iterate through the addresses in the cookie, if
+  we do not find any entries with a subchannel is states READY, IDLE, or
+  CONNECTING and we do find entries that do not have any subchannel, we
+  will trigger creation of a subchannel for one of those entries.
+
+The picker logic will now look like this:
+
+```
```
sergiitk, line 232: nit
```suggestion
```py
```

A75-xds-aggregate-cluster-behavior-fixes.md
```diff
@@ -0,0 +1,312 @@
+A75: xDS Aggregate Cluster Behavior Fixes
+----
+* Author(s): @markdroth
+* Approver: @ejona86, @dfawley
+* Status: {Draft, In Review, Ready for Implementation, Implemented}
+* Implemented in: <language, ...>
```
sergiitk, line 6:
```suggestion
* Implemented in: cpp since v1.61.0
```

A75-xds-aggregate-cluster-behavior-fixes.md
```diff
@@ -0,0 +1,312 @@
...
+def Pick(pick_args):
+  override_host_attribute = pick_args.call_attributes.get(attribute_key)
+  if override_host_attribute is not None:
+    entry_with_no_subchannel = None
+    idle_subchannel = None
+    found_connecting = False
+    for address in override_host_attribute.cookie_address_list:
+      entry = lb_policy.address_map[address]
+      if (entry found AND
+          entry.health_status is in policy_config.override_host_status):
```
sergiitk, line 242:
```suggestion
     if entry and entry.health_status in policy_config.override_host_status:
```

1. `if entry` implies it's found
2. lowercase and
3. `is in` -> `in`

A75-xds-aggregate-cluster-behavior-fixes.md
```diff
@@ -0,0 +1,312 @@
...
+        if entry.subchannel is set:
+          if entry.subchannel.connectivity_state == READY:
+            override_host_attribute.set_actual_address_list(entry.address_list)
+            return entry.subchannel as pick result
+          elif entry.subchannel.connectivity_state == IDLE:
+            if idle_subchannel is None:
+              idle_subchannel = entry.subchannel
+          elif entry.subchannel.connectivity_state == CONNECTING:
+            found_connecting = True
+        else if entry_with_no_subchannel is None:
```
sergiitk, line 252:
```suggestion
        elif entry_with_no_subchannel is None:
```

A75-xds-aggregate-cluster-behavior-fixes.md
```diff
@@ -0,0 +1,312 @@
...
+  override_host_attribute = pick_args.call_attributes.get(attribute_key)
+  if override_host_attribute is not None:
+    entry_with_no_subchannel = None
+    idle_subchannel = None
+    found_connecting = False
+    for address in override_host_attribute.cookie_address_list:
+      entry = lb_policy.address_map[address]
+      if (entry found AND
+          entry.health_status is in policy_config.override_host_status):
+        if entry.subchannel is set:
```
sergiitk, line 243:
```suggestion
        if entry.subchannel is not None:
```

A75-xds-aggregate-cluster-behavior-fixes.md
```diff
@@ -0,0 +1,312 @@
...
+          if entry.subchannel.connectivity_state == READY:
+            override_host_attribute.set_actual_address_list(entry.address_list)
+            return entry.subchannel as pick result
+          elif entry.subchannel.connectivity_state == IDLE:
+            if idle_subchannel is None:
+              idle_subchannel = entry.subchannel
+          elif entry.subchannel.connectivity_state == CONNECTING:
+            found_connecting = True
```
sergiitk, line 251:
```suggestion
                    match entry.subchannel.connectivity_state:
            case READY:
                override_host_attribute.set_actual_address_list(entry.address_list)
                return entry.subchannel   # as pick result
            case IDLE:
              if idle_subchannel is None:
                idle_subchannel = entry.subchannel
            case CONNECTING:
                found_connecting = True
```

we finally have switch/case now :)

A75-xds-aggregate-cluster-behavior-fixes.md
markdroth: This is not actually intended to be valid python code; it's just python-style pseudo-code.  I've add a parenthetical indicating this.

A75-xds-aggregate-cluster-behavior-fixes.md
markdroth: Done.

A75-xds-aggregate-cluster-behavior-fixes.md
sergiitk: Up to you - even if it's not python, it'll still be easier to ready with syntax highlighting. It's clear now from your clarification it's pseudo-code, IMO adding `py` just for readability might worth it.

A75-xds-aggregate-cluster-behavior-fixes.md
sergiitk: Python or not, switch-case just seems more readable to me.

A75-xds-aggregate-cluster-behavior-fixes.md
sergiitk: minor note: `elif` is used everywhere else in this code

A75-xds-aggregate-cluster-behavior-fixes.md
```diff
@@ -0,0 +1,312 @@
...
+There are two main parts to this proposal: aggregate cluster changes and
+stateful session affinity changes.
+
+### Aggregate Cluster Changes
+
+Instead of representing an aggregate cluster as a single cluster whose
+priorities come from different underlying clusters, we will instead
+represent an aggregate cluster as an instance of a priority LB policy
+([gRFC A56][A56]) where each child is a cds LB policy for the underlying
+cluster.
```
murgatroid99, line 86: If an aggregate cluster has a child that is also an aggregate cluster, should that be represented as a deep tree, with each child of the root priority LB policy corresponding to a direct child of the root aggregate cluster, or a flat tree, with each child of the root priority LB policy corresponding to an underlying cluster?

markdroth: The intent is that the code in the `XdsDependencyManager` will recursively resolve the aggregate cluster dependency tree, using the same algorithm that we'd already implemented in the cds LB policy prior to this design.  In the complete `XdsConfig` that is passed down to the LB policies, the data for an aggregate cluster will contain the ordered, flattened list of leaf policies, ignoring any intermediate aggregate clusters.

In other words, let's say you have the following clusters:

A: aggregate cluster pointing to [B, C]
B: EDS cluster
C: aggregate cluster pointing to [D, E]
D: EDS cluster
E: EDS cluster

The `XdsConfig` data for A will indicate the leaf clusters as [B, D, E].  It will not include cluster C.

There's a note about this in the "Rationale" section.

murgatroid99: Got it. I didn't expect that information to only be in the Rationale section.

So, does this mean that all the root cds LB policy does is read the list of underlying cds LB configs and generate a priority LB policy with one of those configs in each priority?

markdroth: Yes, the cds policy for the aggregate cluster will generate a config for a priority policy where each child is a cds policy for the underlying leaf cluster.  However, note that the LB policy configs for the cds policies for the underlying clusters do not need to include anything other than the name of the underlying cluster, because the rest of the info will come from the relevant entry in the `XdsConfig` that is passed down via an attribute.

The fact that the cds policy receives just the list of leaf clusters is actually covered in A74, not in this gRFC.  All we're doing here is changing the way we're handling that list in the cds policy.

ejona86 (commented):
I need to go over the session affinity changes, but sending what I have. This is another case of "really small and easy gRFC, until you talk about stateful session affinity" :smile: .

A75-xds-aggregate-cluster-behavior-fixes.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+### Temporary environment variable protection
+
+This design does not provide any new functionality that will be enabled
+by external I/O, so no environment variable guard is necessary.
+
+## Rationale
+
+We considered implementing nested aggregate clusters as nested cds LB
```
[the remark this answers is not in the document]
markdroth: That's correct.  This flattening is not part of the proposal; it's an alternative that was considered but ruled out for the reasons explained here.

(Most gRFCs don't mention alternatives outside of the "rationale" section.  The dualstack gRFC was an exception to this, since everything was so tightly coupled that it was hard to describe the "what" without the "why", but it was an outlier.)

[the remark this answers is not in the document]
ejona86: This rationale section mentions two options: tree or flattening. We have to choose one or the other. Neither is mentioned in the proposal. It sounded like you wanted flattening:

> However, since the XdsDependencyManager already needs to resolve the
aggregate cluster dependency graph to make sure the entire config is
loaded, it seemed fairly straightforward to pass down the flattened list
and deal with it all in a single layer.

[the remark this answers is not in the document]
markdroth: The flattening isn't mentioned in the proposal, because it's not actually new in this proposal.  We were already doing it that way previously, so we're just sticking with that approach.

I've changed the wording here to try to clarify this.

dfawley (approved):

ejona86 (approved):
I didn't look much at the new Pick pseudocode, but no need to hold this up.

PR #407, opened 2024-01-08
A71 update: Finish review

eugeneo:


A71-xds-fallback.md
```diff
@@ -83,7 +83,9 @@ target.
 The CSDS RPC service will be changed to get data from all XdsClient instances.
 A new string field named `client_scope` will be added to the CSDS
 [ClientConfig] message to indicate the channel target the data is associated
-with.
+with. For gRPC clients, this field will contain the key channels use to lookup
+their XdsClient instance, such as dataplane target for client channels or
+special value for xDS-enabled gRPC servers.
```
markdroth, line 88: I think the suggestion from @ejona86 was that we should specify the exact string of the key we are going to use for the servers.

eugeneo: We haven't chosen that yet. I suggest we delay this decision in case we see some special requirements, like format or something...

markdroth: I'm not sure what special requirements would arise for this.  Let's just pick a string like "server".

eugeneo: I feel like "server" would clash for way too many teams... I propose `grpc::server`.

eugeneo:
I updated the PR and also addressed [comments by @ejona86](https://github.com/grpc/proposal/pull/386#discussion_r1456490002)

A71-xds-fallback.md
markdroth: I don't think there's a concern about whatever we choose here conflicting with anything else, since this name is meaningful only in the context of the individual client.  In the case of a gRPC xDS client, we can say that "server" means the gRPC servers, and it wouldn't cause any problems even if other xDS clients used the same string to refer to something else.

"grpc::server" seems too C++-specific.  I think "server" is fine.

markdroth: As per subsequent offline discussion, we need to use a character that would not be valid in an authority, since otherwise there could still be a conflict with a data plane authority (e.g., if the target URI is "xds://xds.example.com/@server").  Let's go with "#server", since "#" is not a valid character in authorities and would therefore be percent-encoded.

eugeneo: Done

markdroth: Thanks!

In further offline discussions, we realized that there isn't necessarily a conflict here with the authority characters specifically, since the key for the clients will actually be the full target URI, not just the data plane authority.  But either way, "#server" is a good choice, since it's also not a valid URI.

dfawley (approved):
Nit:

Line 221:
> This approach would drastically include the likelihood of config tears.

Should say "drastically _increase_ the likelihood".

Everything else LGTM; thanks!

eugeneo:
> Should say "drastically _increase_ the likelihood".

Done

A71-xds-fallback.md
```diff
@@ -71,19 +71,19 @@ control plane, which is not desirable.
 gRPC servers using the xDS configuration will share the same XdsClient instance
 keyed with a dedicated well-known key value.
 
-Changes XdsClient to per-data plane target would enable gRPC to switch to
+Changing the XdsClient to per-data plane target enables gRPC to switch to
```
ejona86, line 74: This language still assumes somewhere else we are saying we are making this change. "Implementations will change the XdsClient to be per-data plane target, which enables gRPC to switch..." or "XdsClient will be changed to be per-data plane target, which enables..." (which could be misinterpreted, but you use similar language elsewhere so it should be fine) or similar.

eugeneo: Updated the text. Please take another look.

ejona86: I don't see any new changes.

eugeneo: See here: http://localhost:3000 😉 

(I pushed the commit I forgot to push)

ejona86 (approved):

PR #408, opened 2024-01-09
binder: Wireformat changes in support of stream flow control

jdcormie:
https://github.com/grpc/grpc-java/issues/9010


L73-java-binderchannel/wireformat.md
```diff
@@ -229,7 +253,53 @@ stream transactions until we receive an acknowledgement message from the transpo
...
+While transport flow control limits consumption of scarce buffer resources at
+the Android/binder layer, stream flow control limits the amount of buffering at
+the gRPC layer. The resulting back pressure lets an application adjust its rate
+of message production to match the remote reader's rate of stream consumption.
+
+### Sender Responsibilities
+Each stream sender (both client and server) must keep track of the amount of
+space remaining in its peer's receive window, decrementing this counter by the
+`bytes data` or `parcelable` size of each `rpc transaction` it sends and
+incrementing it upon receipt of a `stream receive window update`. A stream
```
markb74, line 267: It's worth nothing that parcelables throw a bit of a spanner in the works here, since we can't easily predict how large they'll be.

I'm not too concerned with that though since we never use parcelable messages internally, and if someone uses them they just need to set a super lenient window.

L73-java-binderchannel/wireformat.md
```diff
@@ -229,7 +253,53 @@ stream transactions until we receive an acknowledgement message from the transpo
...
+window updates are not allowed.
+
+Receivers must detect incoming transactions that exceed their advertised receive
+window and respond by "out of band" closing the stream with canonical code
+`INTERNAL`. Receivers that reduce their incoming transaction buffer size must
+take care to account for window updates messages they sent before the reduction.
+
+### Backwards Compatibility
+Although stream flow control is a core part of the gRPC abstraction, the
+earliest drafts of this document unfortunately did not support it. For
```
markb74, line 296: drafts -> versions? (Otherwise you might imply there's no production code which doesn't support this).

markb74 (approved):
Thanks!

L73-java-binderchannel/wireformat.md
jdcormie: I actually intentionally didn't say versions here because we have an int version field in the setup transaction which actually isn't changing with this update. Changed to "revisions" to make it clearer that the old wireformat has actually been widely deployed.

L73-java-binderchannel/wireformat.md
```diff
@@ -56,8 +56,17 @@ protocol extension flags = int;
 shutdown flags = int;
 num bytes = long;
 ping id = int;
-
-setup transport transaction = version, binder, [protocol extension flags];
+initial stream receive window size = int; (* The initial receive window size for
+                                             new streams created on this
+                                             transport, before any window
+                                             updates are sent. *)
```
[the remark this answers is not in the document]
jdcormie: good point done

L73-java-binderchannel/wireformat.md
```diff
@@ -229,7 +253,53 @@ stream transactions until we receive an acknowledgement message from the transpo
...
+Each stream sender (both client and server) must keep track of the amount of
+space remaining in its peer's receive window, decrementing this counter by the
+`bytes data` or `parcelable` size of each `rpc transaction` it sends and
+incrementing it upon receipt of a `stream receive window update`. A stream
+transaction must not be sent unless there's room for its data in the peer's
+window. Availability of peer receive window space should be exposed to the
+sending application using the language-specific stream readiness API. 
+
+The initial receive window size of all new streams is specified by a peer in its
+`setup transport transaction` and cannot be changed for the life of a transport.
```
larry-safran, line 273: In the next section it is described how the receiver can change the window size, so this should probably be rephrased to indicate that the sender can't change it.

L73-java-binderchannel/wireformat.md
```diff
@@ -229,7 +253,53 @@ stream transactions until we receive an acknowledgement message from the transpo
...
+A receiver may increase the size of the incoming transaction buffer of an existing
+stream simply by sending a window update message advertising the additional
+capacity. A receiver can also reduce the size of this buffer by suppressing
+window updates as incoming messages are consumed by the app. However, negative
+window updates are not allowed.
+
+Receivers must detect incoming transactions that exceed their advertised receive
+window and respond by "out of band" closing the stream with canonical code
+`INTERNAL`. Receivers that reduce their incoming transaction buffer size must
+take care to account for window updates messages they sent before the reduction.
```
larry-safran, line 292: s/window updates messages/window update messages/

jdcormie: done thanks

L73-java-binderchannel/wireformat.md
jdcormie: I rephrased to make a stronger distinction between the window size of an established stream and the initial window size for new streams. LMK if this doesn't address your comment.

L73-java-binderchannel/wireformat.md
jdcormie: This is a great point. I added a new "How to handle Parcelable Message Streams?" section to the design doc for discussion. Quick question though: I know we don't use parcelable messages internally but I noticed that ParcelableInputStream isn't public and we don't expose a Marshaller<T> that can use it. So ... could anybody be using Parcelable message Streams?

PR #412, opened 2024-01-22
A76: Improvements to the Ring Hash LB Policy

atollena:
Based on discussion in grpc/grpc#33356.


A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,176 @@
...
+After the addition of this field, the `ring_hash` LB policy config will be:
+
+    message RingHashLoadBalancingConfig {
+      // A client-side cap will limit these values.  If either of these values
+      // are greater than the client-side cap, they will be treated as the
+      // client-side cap.  The default cap is 4096.
+      uint64 min_ring_size = 1;        // Optional, defaults to 1024.
+      uint64 max_ring_size = 2;        // Optional, defaults to 4096, max is 8M.
+
+      string request_metadata_key = 3; // Optional, defaults to the empty string.
```
murgatroid99, line 129: In the context of a Protobuf 3 message, an unset string field is the same as an empty string value, so stating that the field "defaults to the empty string" is redundant with the default Protobuf behavior. In addition, "defaults to the empty string" seems to imply that the empty string value will be used in some way, but earlier it says that "not set" and "empty" are treated the same. I think it would be clearer to say something like "Optional, unused if unset".

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,176 @@
...
+resolver endpoints, it will set the `hash_key` name resolver attribute to the
+value of [LbEndpoint.Metadata][LbEndpoint.Metadata] `envoy.lb` `hash_key` field,
+as described in [Envoy's documentation for the ring hash load
+balancer](envoy-ring-hash).
+
+### LB Policy Config Changes
+
+After the addition of this field, the `ring_hash` LB policy config will be:
+
+    message RingHashLoadBalancingConfig {
```
murgatroid99, line 122: To get protobuf syntax highlighting, instead of indenting the code, use a [fenced code block](https://www.markdownguide.org/extended-syntax/#syntax-highlighting), with the language name `proto`.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,176 @@
...
+Explicitly setting the request hash key cannot possibly lead to problem with
+existing deployment because the new behavior requires setting a load balancing
+policy configuration field that did not exist before. Therefore, it is not gated
+behind an environment variable.
+
+The second behavior change will be enabled by the
+`GRPC_EXPERIMENTAL_XDS_RING_HASH_ENDPOINT_HASH_KEY` environment variable. This
+will protect from the case where an xDS control plane is already setting the
+`LbEndpoint.Metadata` `envoy.lb` `hash_key` field, in which case deploying this
+new behavior would churn all endpoint hash keys. This environment variable will
```
murgatroid99, line 144: Deploying a new version of gRPC will clear all ring_hash state anyway, so I don't understand how this environment variable would do anything to prevent churn.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,176 @@
...
+The `ring_hash` policy will be changed such that the hash key used for
+determining the locations of each endpoint on the ring will be extracted from a
+pre-defined name resolver attribute called `hash_key`. If this attribute is set,
+then the endpoint is placed on the ring by hashing its value. If this attribute
+is not set or empty, then the endpoint IP address is used (current
+behavior). The location of an existing endpoint on the ring changes if its
+`hash_key` name resolver attribute changes.
+
+The xDS resolver will be changed so that when converting EDS responses to
+resolver endpoints, it will set the `hash_key` name resolver attribute to the
```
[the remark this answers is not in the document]
atollena: OK, I interpreted your comment as two things: 
1. It's not the xDS resolver, but the cluster resolver LB policy that does this, so I replaced the term.
2. "name resolver attributes" are not really a thing. I replaced this with "Endpoint attributes", which is metadata attached to each endpoint, used for example to store locality IDs and weight, and which IIUC are language specific -- I don't know if this is clear enough for cross-language interpretations -- please let me know.

1 seems to be in motion with A74, which removes the cluster resolver LB policy in favor of the "xDS resolver".

A76-ring-hash-improvements.md
atollena: Yes, replaced with "Optional, unused if unset.". I also removed mentions of "not set" in the implementation section, since there is no way to distinguish this from empty.

A76-ring-hash-improvements.md
atollena: done.

A76-ring-hash-improvements.md
atollena: Deploying with the environment variable unset will keep the current behavior of hashing the IP address instead of the new behavior of using the value from LbEndpoint.Metadata. As a result, the locations of existing endpoints on the ring will be guaranteed to be the same as before, and requests will continue being routed to the same endpoint. It will churn connections, if that's what you mean, but not endpoint locations on the ring. Suddenly changing all locations on the ring may have unintended consequences, removing the locality that ring hash may have been used for in the first place.

I'm thinking of a better phrasing, and the best I could think of was to replace "churn" with "change". Let me know if this is still unclear.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+
+```proto
+    message RingHashLoadBalancingConfig {
+      // A client-side cap will limit these values.  If either of these values
+      // are greater than the client-side cap, they will be treated as the
+      // client-side cap.  The default cap is 4096.
+      uint64 min_ring_size = 1;        // Optional, defaults to 1024.
+      uint64 max_ring_size = 2;        // Optional, defaults to 4096, max is 8M.
+
+      string request_metadata_key = 3; // Optional, unused if unset.
```
markdroth, line 130: I don't think we need to spell out the whole config here, just the one new field that we're adding.

As mentioned above, I think this should be moved up to the "Explicitly setting the request hash key" section rather than being here in a separate section.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+
+The `ring_hash` policy will be changed such that the hash key used for
+determining the locations of each endpoint on the ring will be extracted from a
+pre-defined endpoint attribute called `hash_key`. If this attribute is set, then
+the endpoint is placed on the ring by hashing its value. If this attribute is
+not set or empty, then the endpoint IP address is hashed, matching the current
+behavior. The locations of an existing endpoint on the ring is updated if its
+`hash_key` endpoint attribute changes.
+
+The cluster resolver LB policy responsible for translating EDS responses into
```
markdroth, line 112: This paragraph should start with "For xDS, ".

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Temporary environment variable protection
+
+Explicitly setting the request hash key cannot possibly lead to problem with
+existing deployment because the new behavior requires setting a load balancing
+policy configuration field that did not exist before. Therefore, it is not gated
+behind an environment variable.
+
+The second behavior change will be enabled by the
+`GRPC_EXPERIMENTAL_XDS_RING_HASH_ENDPOINT_HASH_KEY` environment variable. This
```
markdroth, line 142: I think this should be structured as more of a migration tool than an experiment-enabling feature.  I suggest the following text for this paragraph:

"""
Adding support for the `hash_key` in xDS endpoint metadata could potentially break existing clients whose control plane is setting this key, because upgrading the client to a new version of gRPC would automatically cause the key to start being used.  We expect that this change will not cause problems for most users, but just in case there is a problem, we will provide a migration story by supporting a temporary mechanism to tell gRPC to ignore the `hash_key` endpoint metadata.  This will be enabled by setting the `GRPC_XDS_HASH_KEY_BACKWARD_COMPAT` environment variable to true. This mechanism will be supported for a couple of gRPC releases but will be removed in the long run.
"""

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,169 @@
+A76: Improvements to the Ring Hash LB Policy
+----
+* Author(s): @atollena
+* Approver: ?
+* Status: Draft
+* Implemented in: <language, ...>
+* Last updated: 2024-01-15
+* Discussion at: TODO
```
markdroth, line 8: Please start a mailing list thread and add a link to it here, as per the gRFC process.

markdroth (commented):
Thanks for writing this!

Overall, the design looks good, but I think there is one significant issue related to where we compute the request hash, since I don't think it will work to do it in the picker.  That will probably need a bit more discussion to resolve.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,169 @@
+A76: Improvements to the Ring Hash LB Policy
+----
+* Author(s): @atollena
+* Approver: ?
```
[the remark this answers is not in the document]
atollena: Done.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+
+This proposal extends the following existing gRFC:
+
+* [gRFC A42: xDS Ring Hash LB Policy][A42]
+
+## Proposal
+
+### Explicitly setting the request hash key
+
+A new string field `request_metadata_key` will be added to the `ring_hash`
```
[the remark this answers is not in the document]
atollena: I would like to understand this suggestion. What's the rational for preferring header rather than metadata in this context? The difference between the two is not 100% clear to me. For clients, headers and metadata are basically the same, no?

(I did make the change though)

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+`request_metadata_key` contains multiple values, then values are joined with a
+comma `,` character between each value before hashing.
+- If `request_metadata_key` is not empty, and the request has no value or an
+empty value associated with the metadata key defined, then the picker will
+generate a random hash for it. The use of a random hash key will effectively
+sends the request to a random endpoint.
+- If `request_metadata_key` is empty, then the request hash key will be based on
+the xDS hash policy in RDS, which keeps the existing LB configuration for ring
+hash working as before with xDS.
+- If `request_metadata_key` is empty but the xDS configuration does not provide
```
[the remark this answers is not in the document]
atollena: I simply removed this bullet point, I think the above one is sufficient. I'm likely to have to update that part anyway after we decide how to solve https://github.com/grpc/proposal/pull/412#discussion_r1462445670.

A76-ring-hash-improvements.md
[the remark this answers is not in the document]
atollena: Yes, I had noticed that. I'd propose to update this when A74 is finalized.

A76-ring-hash-improvements.md
atollena: I removed the whole config and moved this to "Explicitly setting the request hash key" section.

A76-ring-hash-improvements.md
atollena: Added.

A76-ring-hash-improvements.md
atollena: I took your wording here.

A76-ring-hash-improvements.md
atollena: Done.

A76-ring-hash-improvements.md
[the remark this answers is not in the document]
markdroth: In gRPC, the term "metadata" refers generically to either headers or trailers (sometimes referred to as "initial metadata" and "trailing metadata" respectively).  Saying "headers" makes it clearer which one we mean.  (It couldn't possibly be anything else in this case, since we don't send trailers on the client side, but in general I think it's better for the language to be clearer to avoid any possible confusion.)

In addition, since this gRFC also refers to xDS endpoint metadata, which is something completely different than gRPC metadata, it seems better to use different terms to avoid confusion.

markdroth (commented):
Thanks for making those updates!  We'll get back to you on the open question once Eric gets back.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+contains a value that is not a valid metadata key, then the configuration is
+rejected. If the `request_metadata_key` refers to a binary metadata (suffixed
+with `-bin`), the configuration is also rejected.
+
+At pick time:
+- If `request_metadata_key` is not empty, and the request metadata has a
+non-empty value, then the request hash key will be set to this value. If
+`request_metadata_key` contains multiple values, then values are joined with a
+comma `,` character between each value before hashing.
+- If `request_metadata_key` is not empty, and the request has no value or an
```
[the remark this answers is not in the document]
markdroth: We finally had a chance to chat about this today.  In addition to the two options I mentioned above, we came up with a few more options that would all be implemented purely within the ring_hash picker:

3. For a request that does not have the specified header, we compute a random hash in the picker.  If we trigger a connection attempt using the normal ring_hash picker logic, then before we queue the pick, we scan forward through the entire ring searching for a subchannel in state READY.  If we find one, we return it; otherwise, we queue the pick.  If we wind up queuing the pick, then we will still compute a new random hash on the next pick attempt and therefore trigger another connection attempt, but if there is at least one subchannel in state READY, then we won't need a second pick for the RPC, which means that we'll have eliminated the problem of increasing RPC latency.  And while this doesn't completely eliminate the problem of creating unnecessary connections, it should significantly reduce the number in the common case where connection attempts finish quickly.  (Also, note that there are already some other cases in ring_hash where we generate unnecessary connections, so this isn't a purely new problem.)  This option would result in slightly worse load distribution for these RPCs while there are few existing connections, but it converges to fairly reasonable behavior with a reasonable rate of traffic.
4. We could just fail the pick for an RPC without the specified header.  This seems sub-optimal, since we'd generally prefer the lack of a header to affect load balancing but not correctness.  (Also, note that for wait_for_ready RPCs, this would result in the RPC remaining queued until its deadline, unless the picker generates a "drop" result instead of a "fail" result.)
5. We could treat a missing header as being present but having the empty string for the value.  This would result in all RPCs without the header being hashed to the same endpoint, which is a significant difference to the current behavior of choosing randomly.  (Another way to say this is that ring_hash would basically devolve to pick_first for this case.)

Given this menu of options, we think that option 3 is the best trade-off in terms of how much work it is to design and implement relative to how close it gets us to ideal behavior.

It's also worth noting that all three of these new options are purely implemented within the ring_hash picker, which means that we can pick any one of them now, and then later we could add a config option to configure one of the other two options instead, if that turns out to be necessary.

What do you think?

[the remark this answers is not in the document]
atollena: Thank you for providing alternatives. Reading through option 3, I realize that you are concerned with usages of ring hash where a given channel connects to a subset of the endpoints on the ring. This is not something I had really considered: the primary use case we have is one where a given channel ends up connecting to all endpoints anyway (just like round robin), and ring hash provides affinity between the content of a header and the endpoint that gets picked.

Option 3 sounds ideal in terms of "theoretical" properties (converging to random load balancing when many requests lack the header, without a single request causing multiple new connections), but it also sounds a bit complicated to me. I find it a bit hard to wrap my head around edge cases, and I don't look forward to having to explain it to our users. I don't have a good sense of whether its behavior will need to be understood and troubleshooted.

I like the simplicity of option 4 actually. At least for our use case, which is to afford some locality based on request content, if the application fails to provide the header, the fall back to a random endpoint is likely to overload the system: the system may not be designed to resist lacking any locality. If users want the behavior of using a random endpoint when the request is lacking the header, then they can set it to a random value in the application.

I would rule out option 5 because if the traffic suddenly starts lacking the header, then all requests from all instances of the channel will end up picking the endpoint closest to `xxhash("")` on the ring, and there would be a high risk of overloading it, then the next closest, etc. This seems really dangerous, but maybe I'm missing something.

If you're all not too concerned about the complexity of option 3, then I'm happy to go with it, otherwise I would go for option 4.

A76-ring-hash-improvements.md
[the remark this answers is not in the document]
atollena: I replace "cluster resolver LB policy" with "xDS resolver" in this section.

A76-ring-hash-improvements.md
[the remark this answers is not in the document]
atollena: OK, that all makes sense. I updated the proposal to use option 3.

A76-ring-hash-improvements.md
```diff
@@ -93,13 +93,22 @@ At pick time:
 then the request hash key will be set to this value. If `request_hash_header`
```
markdroth, line 93: For clarity, suggest changing "If request_hash_header" to "If the header".  That way, it's clear that we're talking about the header specified by the `request_hash_header` config field, not the `request_hash_header` config field itself.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,187 @@
...
+endpoint hash key explicitly via EDS by reusing the configuration mechanism
+implemented in Envoy. To retain the advantage of being able to use `ring_hash`
+without xDS, custom gRPC name resolvers will be able to set this endpoint
+attribute through the language specific resolver attribute interface.
+
+### Related Proposals:
+
+This proposal extends the following existing gRFC:
+
+* [gRFC A42: xDS Ring Hash LB Policy][A42]
```
markdroth, line 70: It's probably also worth referencing A61, which [significantly simplified the ring_hash picker logic](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#ring-hash).

markdroth, line 70: Please also include A74 in this list, since we reference it below.

A76-ring-hash-improvements.md
```diff
@@ -93,13 +93,22 @@ At pick time:
...
+hash triggers a connection attempt (according to the rules described in [A42:
+Picker Behavior][A42-picker-behavior]), then before queuing the pick, the picker
+will scan forward searching for a subchannel in `READY` state. If it finds a
+subchannel in `READY` state, the picker returns it.
+
+This behavior ensures that a single request creates at most one new connection,
+and that missing header does not add extra latency in the common case where
+there is already at least one subchannel in `READY` state. It converges to
+picking a random host, since each request may create a new connection to a
+random endpoint.
```
markdroth, line 112: I think it would be helpful to provide updated pseudo-code for the picker here.  Based on the pseudo-code in A61, I think the updated code will look like this:

```
// Determine request hash.
using_random_hash = false;
if (config.request_hash_header.empty()) {
  request_hash = call_attributes.hash;
} else {
  header = headers.find(config.request_hash_header);
  if (header == null) {
    using_random_hash = true;
    request_hash = ComputeRandomHash();
  } else {
    request_hash = ComputeHash(header);
  }
}

// Do pick based on hash.
first_index = ring.FindIndexForHash(request_hash);
requested_connection = false;
for (i = 0; i < ring.size(); ++i) {
  index = (first_index + i) % ring.size();
  if (ring[index].state == READY) {
    return ring[index].picker->Pick(...);
  }
  if (requested_connection) continue;
  if (ring[index].state == IDLE) {
    ring[index].endpoint.TriggerConnectionAttemptInControlPlane();
    if (using_random_hash) {
      requested_connection = true;
      continue;
    }
    return PICK_QUEUE;
  }
  if (ring[index].state == CONNECTING) {
    return PICK_QUEUE;
  }
}
if (requested_connection) return PICK_QUEUE;
return PICK_FAIL;
```

markdroth (commented):
Sorry for the delay!  This looks really good overall.  Comments are mostly things to improve clarity of the doc.

A76-ring-hash-improvements.md
atollena: Done.

A76-ring-hash-improvements.md
atollena: OK, thanks for pointing this out, I wasn't aware of A61's impact on ring hash. I spent some time to understand how that affects the implementation I started in Go. Go implements A62 (pick first sticky transient failure) but not A61, so it cannot take advantage of this without further refactoring at least the ring hash policy. I'll sync up with the Go team to see what the best path forward is, but I imagine it'd be best to start with implementing at least part of A61 (delegating to `pick_first`).

A76-ring-hash-improvements.md
atollena: I added the algorithm.

markdroth (commented):
This looks great to me!

@ejona86 @dfawley Would you two please do a review pass as well?  Thanks!

A76-ring-hash-improvements.md
atollena: After a quick discussion with @dfawley I decided to try to implement this before A61. I think the proposal could be more clear in this case, at least for Go, because there is a difference between immediately attempting to connect (which happens only for subchannels in "idle" state), and queuing a connect for subchannels in transient failure, where a connection attempt will be triggered when the subchannel returns to `idle` state after the connection backoff.

I implemented the new behavior of scanning forward for a ready subchannel, either when:
1. the subchannel picked randomly is idle (immediate connection attempt triggered) or,
2. the subchannel picked randomly is in transient failure (connection attempt queued)

I think it has the desired behaviour of making sure we don't trigger more than one connection attempt on each pick if there is a ready subchannel, not adding latency if there is a ready subchannel, and converge to random. But I also considered only implementing the scanning forward for ready subchannel when we triggered an immediate connection attempt, which happens when either:
1. the picked subchannel is idle
2. the picked subchannel is in transient failure, and the second is idle
3. the picked suchannel is in transient failure, the second is also in transient failure, and we found an idle connection before a ready one when scanning the ring forward.

This ambiguity will disappear when all implementations have A61 implemented. This is planned for this quarter for Go, IIUC, but I imagine there may be reasons for some implementations to want to implement A76 before A61. My question is whether you think it's worth adding a note about it in this gRFC to lift this ambiguity.

markdroth: My inclination is that this logic is already complicated enough to understand, so I'd prefer to avoid muddying the waters by trying to do this without having first implemented the relevant parts of A61.  From first glance, what you describe here seems reasonable, but I think I'd have to stare at it for a lot longer to convince myself it didn't have any potential pitfalls.

@dfawley, is it not feasible to implement just the ring_hash part of A61 before implementing this change?

atollena: We discussed with the go team and we're going to implement delegating to pick first before this change.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,236 @@
...
+The xDS resolver, described in [A74][A74], will be changed to set the `hash_key`
+endpoint attribute to the value of [LbEndpoint.Metadata][LbEndpoint.Metadata]
+`envoy.lb` `hash_key` field, as described in [Envoy's documentation for the ring
+hash load balancer][envoy-ring-hash].
+
+### Temporary environment variable protection
+
+Explicitly setting the request hash key cannot possibly lead to problem with
+existing deployment because the new behavior requires setting a load balancing
+policy configuration field that did not exist before. Therefore, it is not gated
```
ejona86, line 177: That's not the condition. It's "whether remote I/O can trigger it" and it can. The reason to have an environment variable here is to handle bugs in implementations while they are being implemented. If there's a crasher bug, for example, you don't want to _prevent_ using the new feature because you might trigger old clients to crash.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,236 @@
...
+  if (ring[index].state == CONNECTING) {
+    return PICK_QUEUE;
+  }
+}
+if (requested_connection) return PICK_QUEUE;
+return PICK_FAIL;
+```
+
+This behavior ensures that a single request creates at most one new connection,
+and that a request missing the header does not add extra latency in the common
```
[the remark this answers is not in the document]
atollena: Thanks Eric for pointing out this flaw, and Mark for providing the updated pseudo code.

> But I also wonder if we want to take a completely alternate approach.

Alternative approaches were discussed in https://github.com/grpc/proposal/pull/412#discussion_r1462445670.
1. Use something like config selector to set the hash outside the picker, but we don't have a cross language API for this yet.
2. Make it possible for the picker to retrieve the hash across picks for a given request. Again IIUC we don't have an API for this.
3. What we chose here -- make sure we don't cause more than one connection attempt per request by taking the ring state into account.
4. Fail the request
5. Use a hardcoded hash.

There are details on the comment thread about the pros and cons of each approach. I can't really speak for 1 or 2, they are more work which may be worth it. I could live with 4 and 5, although 3 is arguably nicer in terms of behavior, as no hash values results in a random load distribution.

> So I think what we want is something like this: ...

In the code you propose, we end up failing picks when all children are in either `CONNECTING` or `TRANSIENT_FAILURE` state, even if this request actually caused one of the child to transition to `CONNECTING`. That doesn't seem right to me. Shouldn't we queue the pick in that case? I would have expected something like:

```
  // If we didn't find CONNECTING but found IDLE, start a connection attempt.
  if (!found_connecting && found_idle_index != -1) {
    ring[found_idle_index].endpoint.TriggerConnectionAttemptInControlPlane();
    if (found_ready_index == -1) return PICK_QUEUE;
  }
  // If we found READY, use it.
  if (found_ready_index != -1) {
    return ring[found_ready_index].picker->Pick(...);
  }  
  // If we found CONNECTING, queue the pick.
  if (found_connecting) {
    return PICK_QUEUE;
  }
  // Didn't find READY or CONNECTING, and didn't start a connection attempt.
  return PICK_FAIL;
```

But then, this would mean that multiple requests using a random hash being issued concurrently can only trigger one connection at a time. This behavior may be acceptable, but that's not my understanding of what Eric was suggesting. IIUC Eric suggested using the copy of the previous children states that we already have in ring hash ([here in c-core](https://github.com/grpc/grpc/blob/b1896a2136e9f60e026daeaaf393cb8cb143be72/src/core/load_balancing/ring_hash/ring_hash.cc#L212)), and compare it to the new state to avoid updating the picker in the case where the state update is due to a subchannel going from `IDLE` to `CONNECTING`. This would effectively avoid triggering picks as a result of triggering a connection attempt.

A76-ring-hash-improvements.md
atollena: OK, so I guess for this one it also makes sense to have it be opt-in to start with. I added an environment variable condition called `GRPC_EXPERIMENTAL_RING_HASH_SET_REQUEST_HASH_KEY`. That would let users disable the feature should it cause problems if the source of the service config is causing an undesirable behavior. I am not familiar with the process of when we decided to remove this kind of feature gating environment variable. Please let me know if that seems right.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,236 @@
...
+pre-defined endpoint attribute called `hash_key`. If this attribute is set, then
+the endpoint is placed on the ring by hashing its value. If this attribute is
+not set or empty, then the endpoint's first address is hashed, matching the
+current behavior. The locations of an existing endpoint on the ring is updated
+if its `hash_key` endpoint attribute changes.
+
+The xDS resolver, described in [A74][A74], will be changed to set the `hash_key`
+endpoint attribute to the value of [LbEndpoint.Metadata][LbEndpoint.Metadata]
+`envoy.lb` `hash_key` field, as described in [Envoy's documentation for the ring
+hash load balancer][envoy-ring-hash].
```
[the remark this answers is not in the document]
atollena: Fixed.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,236 @@
...
+At pick time:
+- If `request_hash_header` is empty, then the request hash key will be based on
+the xDS hash policy in RDS, which keeps the existing LB configuration for ring
+hash working as before with xDS.
+- If `request_hash_header` is not empty, and the header has a non-empty value,
+then the request hash key will be set to this value. If the header contains
+multiple values, then values are joined with a comma `,` character before
+hashing.
+- If `request_hash_header` is not empty, and the request has no value associated
+with the header or its value is empty, then the picker will generate a random
```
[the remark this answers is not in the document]
atollena: OK, you're right, we should simply accept that if the value is present but empty, we use it as-is. That also matches what was in the pseudo code with `if (header == null)`.

A76-ring-hash-improvements.md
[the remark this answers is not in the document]
ejona86: I need to catch up on the conversation, but we discussed it on Friday (after Mark's message here) and seems it didn't make it here yet. I'll be in meetings all day to day, so want to get it out now.

We came up with initializing `requested_connection` specially in the random hash case (note that it also has the change to CONNECTING that was mentioned in my other comment). When the picker is created, it'd go through the subchannels and compute `picker_has_a_subchannel_connecting`, and then use it at picking time:

```c++
first_index = ring.FindIndexForHash(request_hash);
requested_connection = using_random_hash && picker_has_a_subchannel_connecting;
for (i = 0; i < ring.size(); ++i) {
  index = (first_index + i) % ring.size();
  if (ring[index].state == READY) {
    return ring[index].picker->Pick(...);
  }
  if (requested_connection) continue;
  if (ring[index].state == IDLE) {
    ring[index].endpoint.TriggerConnectionAttemptInControlPlane();
    if (using_random_hash) {
      requested_connection = true;
      continue;
    }
    return PICK_QUEUE;
  }
  if (ring[index].state == CONNECTING) {
    if (using_random_hash) {
      continue;
    }
    return PICK_QUEUE;
  }
}
if (requested_connection) return PICK_QUEUE;
return PICK_FAIL;
```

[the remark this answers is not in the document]
ejona86: > We already share child state outside of the picker, so...

This comment of mine was wrong. So avoiding updating the Picker is probably a mess.

> 2. Make it possible for the picker to retrieve the hash across picks for a given request. Again IIUC we don't have an API for this.

We talked about this option a lot. We _love_ how simple [ring_hash's picker is since A61](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#ring-hash). And our initial pseudocode to fix this just within ring_hash was really, really ugly.

But I in particular was worried about adding this API because there's multiple architectural choices to be made, and each choice could impact whether it is useful for other cases. Needing to store state has come up multiple times, but without looking at multiple cases and confirming the design works for them, it seemed _unlikely_ that it would happen to be helpful for other cases.

Notably, in general we'd probably need to use the LB storing state as part of the key for the state. Yes, ring_hash is okay if the hash is shared across leafs, but that isn't generally true. (The concern is if a later pick takes a different path through the LB tree; we don't want the LB to be confused by the state.)

> We came up with initializing requested_connection specially in the random hash case

This caused the pseudocode to greatly simplify, such that it did seem an acceptable complexity in ring_hash to avoid moving the complexity into our architecture. That pseudocode I shared was a first cut; we were going to look at other simplifications, but it was clean enough to be acceptable.

One thing that might be preferable, is to just use a separate loop/function for the `using_random_hash` case to avoid all the ifs. This is the equivalent to what I posted Monday, but only for the `use_random_hash=true` case:
```c++
first_index = ring.FindIndexForHash(request_hash);
requested_connection = picker_has_a_subchannel_connecting;
for (i = 0; i < ring.size(); ++i) {
  index = (first_index + i) % ring.size();
  if (ring[index].state == READY) {
    return ring[index].picker->Pick(...);
  }
  if (requested_connection) continue;
  if (ring[index].state == IDLE) {
    ring[index].endpoint.TriggerConnectionAttemptInControlPlane();
    requested_connection = true;
    continue;
  }
}
if (requested_connection) return PICK_QUEUE;
return PICK_FAIL;
```

I think we'd consider them essentially equivalent, so feel free to put the combined or the separate psuedocode in the gRFC; whatever your preference. That's assuming you like this pseudocode at all.

A61-IPv4-IPv6-dualstack-backends.md
```diff
@@ -545,6 +545,7 @@ for (i = 0; i < ring.size(); ++i) {
     return PICK_QUEUE;
   }
 }
+return PICK_FAIL;
```
atollena, line 548: I took the liberty to add this clarification to A62 -- handle the case of transient failure. Although I presume that the aggregated state reported to the parent may prevent the case from ever happening?

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,227 @@
...
+  }
+}
+
+first_index = ring.FindIndexForHash(request_hash);
+if !using_random_hash {
+    // Use the logic from A62 unchanged.
+    // ...
+}
+
+requested_connection = picker_has_a_child_connecting;
```
ejona86, line 130: We should describe that `picker_has_a_child_connecting` is computed when the picker is created.

ejona86 (approved):

A76-ring-hash-improvements.md
ejona86: The normal process is, "when it is adequately tested, enable it by default." C++ core tends to remove the env var at that point. In Java/Go we tend to wait a release with the env var there so it can be disabled if something goes awry. When the env var is removed doesn't matter; we care about when the default changes here.

But the main point is "when tested." In most other gRFC's that's our interop test framework; here, I guess it'll just be unit tests? For Go in practice it might be "when someone tests it with a service" where "someone" is "Antoine and friends".

Again, the main thing we want to avoid is a bug during the initial implementation that causes really bad client behavior such that you can't enable the feature in the future, lest you trigger broken clients.

A76-ring-hash-improvements.md
atollena: done.

A61-IPv4-IPv6-dualstack-backends.md
atollena: C-core does essentially the same, attaching additional info: https://github.com/grpc/grpc/blob/master/src/core/load_balancing/ring_hash/ring_hash.cc#L356-L358

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,232 @@
...
+(`picker_has_a_child_connecting` in the pseudo code below). This avoids having
+to compute this information on every pick when using a random hash.
+
+The following pseudo code describes the updated picker logic:
+
+```
+// Determine request hash.
+using_random_hash = false;
+if (config.request_hash_header.empty()) {
+  request_hash = call_attributes.hash;
```
dfawley, line 118: Can't this also be empty?  In which case you'd use a random hash, too?

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,232 @@
...
+endpoint in `READY` state. It converges to picking a random endpoint, since each
+request may eventually cause a random endpoint to go from `IDLE` to `READY`.
+
+### Explicitly setting the endpoint hash key
+
+The `ring_hash` policy will be changed such that the hash key used for
+determining the locations of each endpoint on the ring will be extracted from a
+pre-defined endpoint attribute called `hash_key`. If this attribute is set, then
+the endpoint is placed on the ring by hashing its value. If this attribute is
+not set or empty, then the endpoint's first address is hashed, matching the
```
dfawley, line 162: "first address" -- So the A61 gRFC says the DNS resolver should do RFC-6724 sorting.  But the endpoints aren't necessarily coming from the DNS resolver, right?  So do we want to apply the same sorting here anyway, to ensure consistency?

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,232 @@
...
+if its `hash_key` endpoint attribute changes.
+
+The xDS resolver, described in [A74][A74], will be changed to set the `hash_key`
+endpoint attribute to the value of [LbEndpoint.Metadata][LbEndpoint.Metadata]
+`envoy.lb` `hash_key` field, as described in [Envoy's documentation for the ring
+hash load balancer][envoy-ringhash].
+
+### Temporary environment variable protection
+
+Explicitly setting the request hash key will be gated by the
```
dfawley, line 173: "During testing, ..." or "Until sufficient testing has passed, ..."?

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,232 @@
...
+`GRPC_EXPERIMENTAL_RING_HASH_SET_REQUEST_HASH_KEY` environment variable. This
+mechanism will be supported for a couple of gRPC releases but will be removed in
+the long run.
```
dfawley, line 176: If we use the above language, this last sentence is not necessary.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,232 @@
...
+Adding support for the `hash_key` in xDS endpoint metadata could potentially break
+existing clients whose control plane is setting this key, because upgrading the
+client to a new version of gRPC would automatically cause the key to start being
+used. We expect that this change will not cause problems for most users, but
+just in case there is a problem, we will provide a migration story by supporting
+a temporary mechanism to tell gRPC to ignore the `hash_key` endpoint
+metadata. This will be enabled by setting the
+`GRPC_XDS_ENDPOINT_HASH_KEY_BACKWARD_COMPAT` environment variable to true. This
+mechanism will be supported for a couple of gRPC releases but will be removed in
+the long run.
```
dfawley, line 187: Should we have a rollout plan here?  I.e. something like "The first release will have it off by default.  A subsequent release will enable it by default, but leave the opt-out.  A final release will remove the opt-out."

A76-ring-hash-improvements.md
atollena: > But the endpoints aren't necessarily coming from the DNS resolver, right? So do we want to apply the same sorting here anyway, to ensure consistency?

The DNS resolver will create an endpoint per address. The fact that it sorts them seems irrelevant, since we do not take the order of endpoints into consideration when placing them on the ring. I would expect other resolvers that cannot group endpoints to operate like the DNS resolver, and resolvers that *can* group endpoints to sort addresses within an endpoint in a consistent way (such as according to RFC-6724). To me the wording seems sufficient here: users that care about this consistency should either sort addresses in the resolver, or use the endpoint metadata key.

You do raise an interesting point which is that if the DNS resolver returns multiple addresses for the same endpoint, there is no way for ring hash to associate them to the same endpoint. So each address, regardless of family, will get its own place on the ring. Some addresses may be multi-home of the same endpoint, etc. That seems like a fundamental limitation of using DNS A/AAAA records for service discovery: we can't do things that rely on knowing the list of *endpoints* rather than the list of *addresses*. And I think this concern also applies to other LB policies, such as RR and WRR, which when used with DNS, would cause multiple connections to the same endpoint, potentially resulting in load imbalance.

A76-ring-hash-improvements.md
atollena: I'm fine either way, so I took your suggestion. The wording here was based on Mark's suggestion in https://github.com/grpc/proposal/pull/412#discussion_r1462478950.

A76-ring-hash-improvements.md
atollena: I took your first suggestion.

A76-ring-hash-improvements.md
atollena: Done.

A76-ring-hash-improvements.md
atollena: OK, that makes sense, and it also applies to the other environment variable then.

I'm happy to remove the environment variables when we have usage (we have internal usage of a fork of the existing ring hash that we'll be able to replace when this is implemented in Java and Go).

markdroth (commented):
Sorry for the delay on this review pass -- way too many things to review, and not enough time. :(

I like where we ended up on the resulting pseudo-code.  My remaining comments here are mostly minor things.

Hopefully, we're very close to getting this merged!

markdroth:
@atollena, can you address the few remaining open comments?  It would be good to get this merged.  Thanks!

atollena:
> @atollena, can you address the few remaining open comments? It would be good to get this merged. Thanks!

Hey Mark, sorry for the delay on this, this time it was me being swamped by other things.
I'll get to it in the next few days.

markdroth:
No worries, thank you!

FYI, I've thrown together an implementation of this in C-core in https://github.com/grpc/grpc/pull/38312.

A76-ring-hash-improvements.md
```diff
@@ -221,9 +230,7 @@ considered the following alternative solutions:
 
 ## Implementation
 
-Implemented in Go:
-- Allow setting the request hash key: https://github.com/grpc/grpc-go/pull/7170
```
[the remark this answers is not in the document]
atollena: I will add it back when I have an implementation that conforms to this spec, which was not the case of this PR, which predates the implementation of A61. This isn't yet implemented in Go, so I am waiting for it to update the implementation (cc @arjan-bal who was working on dual stack support, starting with pick first as leaf).

In the meantime I added your implementation for C-core.

A76-ring-hash-improvements.md
```diff
@@ -124,36 +135,33 @@ if (config.request_hash_header.empty()) {
   }
 }
 
-// Do pick based on hash.
 first_index = ring.FindIndexForHash(request_hash);
-requested_connection = false;
+if !using_random_hash {
+    // Return based on A62 unchanged.
```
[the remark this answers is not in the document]
atollena: Yes, fixed.

atollena:
> FYI, I've thrown together an implementation of this in C-core in [grpc/grpc#38312](https://github.com/grpc/grpc/pull/38312).

I have changed the pseudo-code to match the code in c-core, it maps very nicely.

A76-ring-hash-improvements.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Explicitly setting the request hash key
+
+A new field `request_hash_header` will be added to the `ring_hash` policy
+config:
+
+```proto
+    message RingHashLoadBalancingConfig {
+      // (existing fields omitted)
+      string request_hash_header = 3;
```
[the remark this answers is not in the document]
markdroth: No, this is fine as-is -- I was just noting this for when you implement it in Go, since we've occasionally gotten that wrong in the past. :)

markdroth (approved):
> I have changed the pseudo-code to match the code in c-core, it maps very nicely.

That's not surprising, since I based the C-core impl off of the pseudo-code here. :)


This looks great!  I'll double-check with Eric and Doug tomorrow that there are no other concerns, and then we can get this merged.

yousukseung:
@atollena, do you plan to revive and complete grpc/grpc-go#7170, and if so have the timeline?
I'm waiting for this to be implemented in Go and found your old PR.

atollena:
> @atollena, do you plan to revive and complete [grpc/grpc-go#7170](https://github.com/grpc/grpc-go/pull/7170), and if so have the timeline? I'm waiting for this to be implemented in Go and found your old PR.

I don't have a timeline, but the work involves converting the ring hash balancer to use pick first as leaf, first, and that wasn't ready until recently (https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#ring-hash). I'll check with the go team to see whether we're ready to start that work.

dfawley (approved):

PR #414, opened 2024-01-24
A77: xDS Server-Side Rate Limiting

sergiitk:
> [!WARNING]
> This is work-in-progress. Ready for review when promoted from a PR draft


A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+1. Inserting/reading a bucket from the bucket cache using `bucket_id`. Note
+   that `bucket_id` is represented as a `Map<String, String>`, which may
+   introduce complexities in efficient cache sharding for certain programming
+   languages.
+2. Incrementing `num_request_allowed`/`num_request_denied` bucket counters.
+
+Each gRPC implementation needs to consider what synchronization primitives are
+available in their language to minimize the thread lock time.
+
+### Code Samples
```
markdroth, line 407: These code snippets should be part of the descriptions of the RLQS filter (interceptor) and RLQS Filter State objects, respectively.

markdroth (commented):
Thanks for writing this up!  Overall, this looks good, but I have few comments around making it clearer to readers.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+[rlqs_filter_proto]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/rate_limit_quota/v3/rate_limit_quota.proto
+
+## Proposal
+
+### RLQS Components Overview
+
+The diagram below shows the conceptual components of the RLQS Filter. Note that
+the actual implementation may vary depending on the language.
+
+```mermaid
```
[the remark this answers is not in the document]
sergiitk: Done.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+    linkStyle 3,4 stroke: RoyalBlue, stroke-width: 2px;
+```
+
+##### RLQS HTTP Filter
+
+The filter parses the config, combines LDS filter config with RDS overrides, and
+generates the `onClientCall` handlers (aka interceptors in Java and Go, and
+filters in C++).
+
+##### RLQS Cache
```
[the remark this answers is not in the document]
sergiitk: Yes, I meant A83-based cache here, I've pointed it out in several places. This was supposed to describe how this persistent cache is _used_ by RLQS. I see how this can be confusing for the reader. Will update accordingly.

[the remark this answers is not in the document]
sergiitk: done.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+
+#### Security Considerations
+
+In [`GrpcService.GoogleGrpc`], xDS Control Plane provides the target
+URI, `channel_credentials`, and `call_credentials`. If the xDS Control Plane is
+compromised, the attacker could configure the xDS clients to talk to other
+malicious Control Plane, leading to such potential exploits as:
+
+1. Leaking customer's Application Default Credentials OAuth token.
+2. Causing MalOut/DDoS by sending bad data from the compromised RLQS (f.e. set
```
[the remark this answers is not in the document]
sergiitk: done.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+##### RLQS Cache
+
+RLQS Cache persists across LDS/RDS updates. It maps unique filter configs to
+RLQS Filter State instances, and provides the thread safety for creating and
+accessing them. Each unique filter config generates a unique RLQS Filter state,
+a 1:1 mapping.
+
+##### RLQS Filter State
+
+RLQS Filter State contains the business logic for rate limiting, and the current
```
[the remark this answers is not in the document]
sergiitk: done.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+LDS 2 update removes `r1` and `r2`, and adds new route r4 with the config
+identical to `c2`. While onCallHandlers for routes `r1` and `r2` are
+destroyed, `RlqsFilterState(c2)` is still used by two onCallHandlers, so it's
+preserved in RLQS Cache.
+
+##### Future considerations
+
+With this proposal, the filter state is lost if change is made to the filter
+config, including updates to inconsequential fields such as deny response
+status. Additional logic can be introduced to handle updates to such fields
```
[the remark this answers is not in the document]
sergiitk: done.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+##### Implementation
+
+For performance reasons, CEL variables should be resolved on demand. CEL Runtime
+provides the different variable resolving approaches based on the language:
+
+* CPP: [`BaseActivation::FindValue()`](https://github.com/google/cel-cpp/blob/9310c4910e598362695930f0e11b7f278f714755/eval/public/base_activation.h#L35)
+* Go: [`Activation.ResolveName(string)`](https://github.com/google/cel-go/blob/3f12ecad39e2eb662bcd82b6391cfd0cb4cb1c5e/interpreter/activation.go#L30)
+* Java: [`CelVariableResolver`](https://javadoc.io/doc/dev.cel/runtime/0.6.0/dev/cel/runtime/CelVariableResolver.html)
+
+### Persistent Filter Cache
```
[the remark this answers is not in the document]
sergiitk: Done.

A77-xds-rate-limiting-rlqs.md
sergiitk: Done.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+    rlqs_filter_state -- sendUsageReports --> rlqs_client
+    rlqs_filter_state -- CRUD --> rlqs_bucket_cache
+    rlqs_client -- onBucketsUpdate --> rlqs_filter_state
+    rlqs_client <-. gRPC Stream .-> rlqs_server
+    rlqs_bucket_cache -- " getOrCreate(bucketId)<br />Atomic Updates " --> rlqs_bucket
+    style request stroke: RoyalBlue, stroke-width: 2px;
+    linkStyle 3,4 stroke: RoyalBlue, stroke-width: 2px;
+```
+
+##### RLQS HTTP Filter
```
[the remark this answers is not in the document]
sergiitk: Done. Please review the updated version.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,490 @@
...
+preserved in RLQS Cache.
+
+##### Future considerations
+
+With this proposal, the filter state is lost if change is made to the filter
+config, including updates to inconsequential fields such as deny response
+status. Additional logic can be introduced to handle updates to such fields
+while preserving the filter state.
+
+### Multithreading
```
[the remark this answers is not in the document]
sergiitk: Done.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -40,10 +41,10 @@ which are covered in the proposal:
 3. RPCs will be matched into buckets using [Unified Matcher API].
 4. One of the matching mechanisms will be [CEL](https://cel.dev/) (Common
    Expression Language).
-5. RLQS filter state will persist across LDS/RDS updates using cache retention
-   mechanism similar to the one implemented for [gRFC A83].
+5. RLQS Filter State will persist across LDS/RDS updates using cache retention
+   mechanism described in [gRFC A83].
```
markdroth, line 45: See my comments in #491.  I think we want to move description of this server-side mechanism to this gRFC, and I think we need a bit more discussion about exactly how it will work.  In particular, I want to consider the following:

- According to @bsurber, Envoy doesn't support the per-route override config at all.  We should discuss this and either decide to add it to Envoy (which has some challenges) or remove it from the design entirely and say that we won't support it in gRPC either.
- It seems that Envoy will not actually retain RLQS filter state across LDS updates, so we should consider whether that's the behavior we want as well.  There is at least an argument that that behavior makes sense, since an LDS update on the server side will also cause the server to drain all connections that came in on that listener.
- If we do decide to retain state across LDS updates, we need to define the exact mechanism for matching up existing state with new configs.  As per my comments in #491, the proposed behavior of matching based on listener order does not seem right to me.
- The C-core mechanism is likely to need a bit more thought.  The approach that we're using on the client side would wind up with a separate RLQS filter state object for each incoming connection, not one for the HCM config instance that is shared across all connections, which probably isn't what we want.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -80,113 +83,383 @@ config:
...
+    the RLQS Filter State of updates to quota assignments.
+
+### Handling Events
+
+#### On LDS/RDS Updates
+
+When receiving an LDS/RDS update, the RLQS filter will perform the following 
+steps for each route:
+
+1.  Parse the new filter config and per-route overrides into an internal RLQS
```
markdroth, line 381: One thing that seems to be missing from this gRFC is config validation information.  We need to list all of the fields in the RLQS filter config that we support, and we need to say what validation we are going to do on each field.  In particular, we need to spell out which values will cause us to NACK.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -80,113 +83,383 @@ config:
...
+matcher tree from the filter config to determine which bucket to use for the
+RPC. It then looks for that RLQS Bucket in the RLQS Bucket Map, creating it if
+it doesn't already exist.
 
-RLQS Cache persists across LDS/RDS updates. It maps unique filter configs to
-RLQS Filter State instances, and provides the thread safety for creating and
-accessing them. Each unique filter config generates a unique RLQS Filter state,
-a 1:1 mapping.
+If a new bucket has been created, the RPC is rate-limited according to bucket's
+default configuration. Then the filter schedules a report on the RLQS stream
```
markdroth, line 400: We need to document what fields we will populate in the RLQS requests and what their values will be.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -80,113 +83,383 @@ config:
...
-##### RLQS Filter State
+If the bucket exists, the RPC is rate-limited according to its active quota
+assignment and usage counters. The bucket increments corresponding bucket usage
+counter.
 
-RLQS Filter State contains the business logic for rate limiting, and the current
-state of rate limit assignments per bucket. RLQS Filter State is what's passed
-to the `onCallHandler`. It exposes the public "`rateLimit()`" method, which
-takes request metadata as an argument.
+#### On RLQS Server Response
```
markdroth, line 408: We need to document which fields we will read in the RLQS responses, what values are considered valid, and how we will handle any case where the response is invalid (e.g., required field missing, field has an invalid value, etc).

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,618 @@
...
+
+When xDS Control Plane configures connection to another control plane
+via [`GrpcService.GoogleGrpc`] message, we'll inspect the allow-list for the
+matching target URI.
+
+1. If target URI is not present, we don't create the connection to the requested
+   Control Plane, and NACK the xDS resource.
+2. If target URI is present, we create the connection to the requested Control
+   Plane using the channel credentials provided in the bootstrap file. Transport
+   security configuration provided by the TD is ignored.
```
markdroth, line 479: s/TD/control plane/

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,618 @@
...
+
+#### Unified Matcher API
+
+RPCs will be matched into buckets using [Unified Matcher API] — an adaptable
+framework that can be used in any xDS component that needs matching features.
+
+Envoy provides two syntactically equivalent Unified Matcher
+definitions: [`envoy.config.common.matcher.v3.Matcher`](https://github.com/envoyproxy/envoy/blob/e3da7ebb16ad01c2ac7662758a75dba5cdc024ce/api/envoy/config/common/matcher/v3/matcher.proto)
+and [`xds.type.matcher.v3.Matcher`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto).
+We will only support the latter, which is the preferred version for all new APIs
```
markdroth, line 493: Any reason not to support both?  I think there are other places where we'll want to use the same matcher implementation in the future that use the old type, and it seems like we should be able to handle either one.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,618 @@
...
+contains the information needed to associate the RPC with `bucket_id` and the
+default rate limiting configuration.
+
+In this iteration the following Unified Mather extensions will be supported:
+
+1. Inputs:
+    1. [`HttpRequestHeaderMatchInput`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/matcher/v3/http_inputs.proto#type-matcher-v3-httprequestheadermatchinput)
+    2. [`HttpAttributesCelMatchInput`](https://www.envoyproxy.io/docs/envoy/latest/xds/type/matcher/v3/http_inputs.proto#envoy-v3-api-msg-xds-type-matcher-v3-httpattributescelmatchinput)
+2. Custom Matchers:
+    1. [`CelMatcher`](https://www.envoyproxy.io/docs/envoy/latest/xds/type/matcher/v3/cel.proto.html)
```
markdroth, line 507: Doesn't this matcher require using `HttpAttributesCelMatchInput` as the input?  If so, and if this is the only matcher we're going to support, then it doesn't really make sense to say that we're going to support `HttpRequestHeaderMatchInput`, since there's no possible way to use it.

We should probably also support any common matchers that are expected to be used with `HttpRequestHeaderMatchInput`.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,618 @@
...
+| `request.referer`   | `string`              | `metadata["referer"]`        | Referer request header.                                     |
+| `request.useragent` | `string`              | `metadata["user-agent"]`     | User agent request header.                                  |
+| `request.time`      | `timestamp`           | Not set                      | Time of the first byte received.                            |
+| `request.id`        | `string`              | `metadata["x-request-id"]`   | Request ID corresponding to `x-request-id` header value     |
+| `request.protocol`  | `string`              | Not set                      | Request protocol.                                           |
+| `request.query`     | `string`              | `""`                         | The query portion of the URL.                               |
+
+###### Footnotes
+
+**<sup>1</sup> `request.path`**
```
markdroth, line 562: The language-specific details in these footnotes don't seem necessary to spell out.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,618 @@
...
+
+**<sup>4</sup> `request.headers`**\
+As defined in [gRFC A41], "header" field.
+
+##### Implementation
+
+For performance reasons, CEL variables should be resolved on demand. CEL Runtime
+provides the different variable resolving approaches based on the language:
+
+* CPP: [`BaseActivation::FindValue()`](https://github.com/google/cel-cpp/blob/9310c4910e598362695930f0e11b7f278f714755/eval/public/base_activation.h#L35)
```
markdroth, line 585: I don't think we need to list the specific APIs in each language.

A77-xds-rate-limiting-rlqs.md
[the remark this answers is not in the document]
markdroth: This structure looks good, thanks!

markdroth (commented):

A77-xds-rate-limiting-rlqs.md
```diff
@@ -80,113 +83,383 @@ config:
...
+
+RLQS Bucket Map object will include the following data members:
+
+-   Bucket Map: Initialized empty at instantiation, thread-safe.
+    -   Entries retrieved on each data plane RPC and when report timers fire.
+    -   Entries inserted when we get the first data plane RPC for a given
+        bucket, or when instructed by the RLQS Server.
+    -   Entries deleted either by RLQS server via `abandon_action`, or on report
+        timers if a bucket's active assignment expires.
+-   Buckets Per Interval Map: Initialized empty at instantiation, thread-safe.
```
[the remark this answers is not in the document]
sergiitk: > if we already have a separate map by report interval to store the timers in, it's not clear to me why we are duplicating that map here

This is a basic lookup map so that we don't have to scan full Bucket Map every time. As you pointed out, doing this every time is pretty much an unnecessary waste of CPU.

The idea was something like this

```py
def timer_fire(self, interval):
  buckets_to_report: list[Bucket] = self.buckets_by_interval(interval)
  report = Report()
  for bucket in buckets_to_report:
    report.add(bucket.snapshot_stats())

  self.rlqs_client.send_report(report)
```

However, as I pointed out in the other comment, this is too implementation-specific.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -80,113 +83,383 @@ config:
...
 
-The aggregated number of requests is reported to the RLQS server at configured
-intervals. The report action is triggered by the Report Timers. RLQS Client
-manages a gRPC stream to the RLQS server. It's used by the filter state to send
-periodic bucket usage reports, and to receive new rate limit quota assignments
-to the buckets.
+When a report timer fires, the RLQS Filter State retrieves all buckets with the
+corresponding reporting interval from the RLQS Bucket Map. For each bucket, the
+filter snapshots the current usage counters, and resets them. The filter then
+sends the snapshot to RLQS server using RLQS Client.
```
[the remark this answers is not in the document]
sergiitk: Very good point.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -80,113 +83,383 @@ config:
...
+##### RLQS Buckets and Multithreading
+
+There are several mutex-synchronized operations on RLQS buckets that are
+executed during latency-sensitive data plane RPC processing:
+
+1. Inserting/reading a bucket from the RLQS Bucket Map using `bucket_id`. Note
+   that `bucket_id` is represented as a `Map<String, String>`, which may
+   introduce complexities in efficient cache sharding for certain programming
+   languages.
+2. Incrementing `num_request_allowed`/`num_request_denied` bucket counters.
```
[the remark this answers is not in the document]
sergiitk: Noted. I've discussed cpp implementation approaches for counter increments with @yashykt a while back, and we concluded that is doable and can be done in efficient manner. With Java, @ejona86 and I came up with a few efficient options too. I'll check out the outlier detection to see if we can lift any some applicable wording directly from it, or just add a link as a "suggested approach".

A77-xds-rate-limiting-rlqs.md
sergiitk: oops

A77-xds-rate-limiting-rlqs.md
sergiitk: Normally I'd agree, but it's surprisingly difficult to find and identify these - just because approaches are really different between languages. IMO it'll save the implementers some time. 

Do you think there's a way to keep these, but without making them look as the prescribed implementation? Maybe under a different section?

A77-xds-rate-limiting-rlqs.md
sergiitk: I don't want to introduce a legacy dependency we have to support forever without a real need for it. 

Note that CEL libraries will keep supporting the legacy format, but they're being migrated to the canonical protos, f.e. https://github.com/google/cel-cpp/commit/aff86967fc811ed4605f099ee2f502d331d5bbe7.  This means canonical protos will be used internally, and legacy CEL operations will incur a performance overhead. 

As for java, the initial release of CEL-java only supported canonical CEL in the first place, however now it's possible to convert the legacy protos  to the canonical one, but this leads to adding separate dependencies (CEL Java V1alpha1 Utility, CEL Java Protobuf Adapter).

AFAIK Envoy will be upgrading its CEL intermediate lib to the canonical protos too.

sergiitk: Also - we can always add the support for CEL if necessary. But we won't be able to remove it once the genie is out of the bottle.

A77-xds-rate-limiting-rlqs.md
sergiitk: `HttpRequestHeaderMatchInput` can be used with the built-in `valtype.matcher.v3.StringMatcher value_match`. I didn't specify it's explicitly because it's not an extension, as the line above implies:

> In this iteration the following Unified Mather extensions will be supported:

HiramSilvey:
Hi @sergiitk, apologies for posting this on the PR itself but I couldn't find a better way to get in contact. I just wanted to reach out to ask if this proposal will likely be merged or if there is still a chance it may not be accepted. Thanks so much.

sergiitk:
@HiramSilvey I expect it to be merged. High-level design has been approved some time ago.
In the past few weeks we needed to reconsider some details and reach the consensus on the filter state retention per-language implementation nuances. I'm planning to resume working on this PR sometime next week.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,618 @@
...
+malicious Control Planes. This could lead to potential exploits such as leaking
+customer's Application Default Credentials OAuth token.
+
+To prevent that, we'll introduce an allow-list to the bootstrap file introduced
+in [gRFC A27]. This allow-list will be a map from a fully-qualified server
+target URI to an object containing channel credentials to use for that target.
+
+```json5
+// The allowlist of Control Planes allowed to be configured via xDS.
+"allowed_grpc_services": {
```
markdroth: I think we should use this field only if the `trusted_xds_server` server feature is not set in the bootstrap config.  See [gRFC A81](https://github.com/grpc/proposal/blob/master/A81-xds-authority-rewriting.md#server-feature-in-bootstrap-config) for details.  If that server feature is present, then we can use the target URI and credentials specified by the control plane.

We will wind up using this same mechanism for ext_proc and ext_authz as well.

A77-xds-rate-limiting-rlqs.md
```diff
@@ -0,0 +1,618 @@
...
+  // The key is fully-qualified server URI.
+  "dns:///xds.example.org:443": {
+    // The value is an object containing "channel_creds".
+    "channel_creds": [
```
markdroth: We should also support setting call credentials here, using the same infrastructure introduced in [gRFC A97](https://github.com/grpc/proposal/blob/master/A97-xds-jwt-call-creds.md#configuring-call-credentials-in-xds-bootstrap-file).  We will need this for an internal use-case.

A77-xds-rate-limiting-rlqs.md
markdroth: Note that this means that we need to flesh out how we'll handle the channel creds and call creds configuration in the `GoogleGrpc` proto.

@ejona86 @dfawley

markdroth: After a bunch of discussion, I wound up creating a separate gRFC just to describe how this will work, which is A102 (https://github.com/grpc/proposal/pull/510).

I think we can remove the entire "Security Considerations" section from this gRFC and instead just reference A102.

A77-xds-rate-limiting-rlqs.md
markdroth: One particular thing of note in the context of config validation: When validating unified matcher trees, we will need to impose some max recursion depth limit, to avoid OOMing due to a control plane sending us too deep of a tree.

A77-xds-rate-limiting-rlqs.md
sergiitk: I assume this goes away since we're replacing the whole thing with a reference to A102 (#510).

A77-xds-rate-limiting-rlqs.md
sergiitk: Done

A77-xds-rate-limiting-rlqs.md
sergiitk: I've misunderstood the question. This was about supporting `envoy.config.common.matcher.v3.Matcher` and had nothing to do with CEL. 

Yes, we can support `envoy.config.common.matcher.v3.Matcher`. Would it make sense to note that we should unpack it as `xds.type.matcher.v3.Matcher`?

A77-xds-rate-limiting-rlqs.md
sergiitk: This is done.

A77-xds-rate-limiting-rlqs.md
sergiitk: This is done.

A77-xds-rate-limiting-rlqs.md
sergiitk: See https://github.com/sergiitk/grfc-proposal/blob/a77-rlqs/A77-xds-rate-limiting-rlqs.md#on-sending-usage-reports

A77-xds-rate-limiting-rlqs.md
sergiitk: See https://github.com/sergiitk/grfc-proposal/blob/a77-rlqs/A77-xds-rate-limiting-rlqs.md#on-sending-usage-reports

A77-xds-rate-limiting-rlqs.md
sergiitk: Wrong link. 

- The filter config and validation is in https://github.com/sergiitk/grfc-proposal/blob/a77-rlqs/A77-xds-rate-limiting-rlqs.md#filter-configuration
- Unified matcher parsing and validation, including the tree depth restrictions are in https://github.com/sergiitk/grfc-proposal/blob/a77-rlqs/A77-xds-rate-limiting-rlqs.md#unified-matcher-matcher

A77-xds-rate-limiting-rlqs.md
sergiitk: Covered in https://github.com/sergiitk/grfc-proposal/blob/a77-rlqs/A77-xds-rate-limiting-rlqs.md#unified-matcher-api-support

PR #418, opened 2024-02-09
L114: Node Server Connection Injection

murgatroid99:


wenbozhu (approved):

ItalyPaleAle:
Thank you so much for implementing this!

Do you have an example of how the new APIs are meant to be used?

murgatroid99:
The implementation PR grpc/grpc-node#2675 has a test that shows the usage of the main APIs `createConnectionInjector` and `injectConnection`. I think the other two should be pretty self-explanatory.

PR #419, opened 2024-02-13
A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient

markdroth:


A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,148 @@
+A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient
```
atollena, line 1: I have a high level question about this design: existing metrics, such as those defined in A66, are implemented using the call tracer plugin mechanism, called out in A66, such as `StreamTracer` in java and `stats.Handler` in Go. Those allow co-existence of opencensus and opentelemetry, an example that is given in A66. I don't see it specified in this gRFC to have callbacks for each one of those new metrics rather than direct use of opentelemetry instrumentation. And I am asking because we make heavy use of a statsd based implementation internally, and ideally those new instrumentations are not gated behind a migration to opentelemetry. We also have Datadog customers that might prefer continuing with their statsd based approach rather than moving to OpenTelemetry for their existing services, and still make use of those new metrics.

Otherwise those metrics that are not per-call are going to be very useful, thank you for adding them!

markdroth: @yashykt is working on another gRFC that will be a companion to this one, which will spell out the framework for supporting non-per-call metrics.  That other gRFC is currently reffered to in this doc as "A??"; that link will be updated once the PR for that gRFC is created and the number assigned.  These two gRFCs will move forward in lock step, since they are part of the same effort; the only reason for splitting this up into two gRFCs is that some people are interested in the metrics we're defining but not in the underlying architecture.

I think the right place for this discussion will be that other gRFC, which will be created soon.  However, as a quick high-level answer to your question, the framework we're building will indeed provide a layer of abstraction between the stats plugin (e.g., OpenCensus or OpenTelemetry) and the metrics exported from within gRPC.  However, at least initially, the API to implement new stats plugins will not be public, and the only implementation we will support will use OpenTelemetry.  But as the framework matures, we may decide to expose it as a public API.

atollena: OK, thank you. Sounds like this other gRFC is what I'll want to have a closer look at.

Feel free to resolve.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,148 @@
...
+### Pick First LB Policy
+
+The Pick First LB policy predates the gRFC process but was updated in
+[A62].  We propose to add the following metrics to it.
+
+All Pick First metrics will have the following label:
+
+| Name        | Description |
+| ----------- | ----------- |
+| grpc.target | Indicates the target of the gRPC channel in which WRR is used.  (Same as the attribute defined in [A66].) |
```
atollena, line 96: The description should be about pick first rather than WRR.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,148 @@
...
+The XdsClient component was originally described in [A27].  Note that in
+[A71], we are moving from a single global XdsClient instance to a
+separate global XdsClient instance for each channel target.  The
+proposed metric schema here reflects that change.
+
+The following labels may be used for XdsClient metrics:
+
+| Name        | Description |
+| ----------- | ----------- |
+| grpc.target | For clients, indicates the target of the gRPC channel in which WRR is used (i.e., the same as the attribute defined in [A66]). For servers, will be the string "#server". |
```
atollena, line 117: This is about the gRPC channel in which xDS is used, rather than WRR.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,148 @@
...
+| grpc.xds_resource_type | Indicates an xDS resource type, such as "envoy.config.listener.v3.Listener". |
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.xds_client.xds_server.connected | Gauge | {bool} | grpc.target, grpc.xds_server | Whether or not the xDS client currently has a working ADS stream to the xDS server.  For a given server, this will be set to 0 when we have a connectivity failure or when the ADS stream fails without seeing a response message, as per [A57].  It will be set to 1 when we receive the first response on an ADS stream. |
+| grpc.xds_client.xds_server.updates | Counter | {updates} | grpc.target, grpc.xds_server, grpc.xds_resource_type | A counter of resource updates from the xDS server.  Note that this is a count of resources, not response messages; if a response message contains two resources, then we will increment the counter twice. |
```
atollena, line 128: Will this count cases where the resource has not changed? I think this should be specified.

atollena, line 128: Those two do not have the grpc.xds_authority, is that an omission? I don't see why it would not apply.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,148 @@
...
+| grpc.xds_client.xds_server.resources | Gauge | {resources} | grpc.target, grpc.xds_server, grpc.xds_authority, grpc.xds_cache_state, grpc.xds_resource_type | Number of xDS resources. |
+
+### Temporary environment variable protection
+
+This proposal does not include any features enabled via exteranl I/O, so
```
atollena, line 133:
```suggestion
This proposal does not include any features enabled via external I/O, so
```

A78-grpc-metrics-wrr-pf-xds.md
atollena: OK, that makes sense.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,141 @@
...
+
+The `weighted_round_robin` LB policy is described in [A58].  We propose to
+add the following metrics to it.
+
+WRR metrics will have the following labels:
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.target | required | Indicates the target of the gRPC channel in which WRR is used.  (Same as the attribute defined in [A66].) |
+| grpc.lb.locality | optional | The locality to which the traffic is being sent. This will be based on the resolver attribute passed down from the `weighted_target` policy. |
```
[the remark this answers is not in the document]
markdroth: Thanks, added.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,154 @@
...
+[A71], we are moving from a single global XdsClient instance to a
+separate global XdsClient instance for each channel target.  The
+proposed metric schema here reflects that change.
+
+XdsClient metrics will have the following labels:
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.target | required | For clients, indicates the target of the gRPC channel in which the XdsClient is used (i.e., the same as the attribute defined in [A66]). For servers, will be the string "#server". |
+| grpc.xds.server | required | The name of the xDS server with which the XdsClient is communicating. |
```
ejona86, line 123: What is "the name"?

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,154 @@
...
+| grpc.xds.resource_type | required | Indicates an xDS resource type, such as "envoy.config.listener.v3.Listener". |
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.xds_client.connected | Gauge | {bool} | grpc.target, grpc.xds.server | Whether or not the xDS client currently has a working ADS stream to the xDS server.  For a given server, this will be set to 1 when the stream is initially created.  It will be set to 0 when we have a connectivity failure or when the ADS stream fails without seeing a response message, as per [A57].  Once set to 0, it will be reset to 1 when we receive the first response on an ADS stream. |
+| grpc.xds_client.resource_updates_valid | Counter | {resource} | grpc.target, grpc.xds.server, grpc.xds.resource_type | A counter of resources received that were considered valid.  The counter will be incremented even for resources that have not changed. |
```
ejona86, line 133: I'm still wary of the "resources that have not changed" because of compatibility and impact if we swap to delta protocol.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,154 @@
...
+| grpc.xds_client.resource_updates_invalid | Counter | {resource} | grpc.target, grpc.xds.server, grpc.xds.resource_type | A counter of resources received that were considered invalid. |
+| grpc.xds_client.resources | Gauge | {resource} | grpc.target, grpc.xds.authority, grpc.xds.cache_state, grpc.xds.resource_type | Number of xDS resources. |
+
+### Temporary environment variable protection
+
+This proposal does not include any features enabled via external I/O, so
```
ejona86, line 139: Well, but we also don't want them on before they are ready as it exposes stability/compatibility concerns for the exported metrics. We were going to have them default-off initially?

ejona86 (commented):
Am I right to believe that none of these need to be updated on the data plane? I didn't miss any, right?

A78-grpc-metrics-wrr-pf-xds.md
dfawley: Yes, no environment variable is fine, but if the plan is to have them off by default initially (enabled via an API) then that SGTM.

dfawley: What about when we decide to stabilize them?  Will they be on by default or off by default at that time?  If they'll be on by default, then there will be a behavior change.  Will there be a gRFC for that change or should this gRFC cover how to de-experimentalize them?

markdroth: I've been thinking about de-experimentalizing metrics as being roughly analogous to how we handle removing env var guards for new features: the gRFC really specifies only how the feature will work when it's enabled by default, even though the expectation is that the feature will start disabled by default.  Just as we don't need a separate gRFC to remove env var guards, I wouldn't think we'd need one to de-experimentalize metrics.

While it is true that gRFCs for features do say something about the env var guards, I think that's necessary because (a) we need to specify the env var name so that we use the same name cross-language and (b) we need to specify exactly which part of the code is guarded by the env var.  Neither of those things are issues with metrics, so I don't think this needs to be called out in the gRFC.  But if either of you would feel more comfortable with some boilerplate that says "all metrics start as experimental and will be de-experimentalized later", I'm fine with that.

The only other difference I see between the two is that the criteria for de-experimentalizing is different: in the case of features, it happens when interop tests pass, whereas for metrics, it will probably happen more when we decide that we want to fully commit to the metric definitions.  But I'm not sure how much that matters.

I'm open to alternative approaches here, but I would prefer to define a process once in A79 and not need a lot of boilerplate in each individual gRFC that defines new metrics.

dfawley: Maybe I'm missing something, though, because I don't see where this gRFC says which of these metrics would be enabled-by-default or disabled-by-default.  This seems like something that should be included.

Or maybe you're saying it doesn't matter?  Like we might take optional metrics and change them to be enabled by default at will, or vice-versa, whenever we want?

IMO for each metric we should call out how we intend to treat it, according to the list here:

https://github.com/grpc/proposal/pull/421/files#diff-58a6eedc2e7a57c026455f8674357915d0943d732e594574cd62bd83dcdd1985R330

markdroth: What I'm saying is that the A79 draft defines what I think is the normal progression for all metrics:

"""
New metrics should start as experimental. Once they are stable, they can optionally be made on-by-default.
"""

I think we should assume that all metrics follow this path without having to repeat what the path is in every individual gRFC that defines new metrics.

I think that we should be able to move a metric from experimental to stable at any time and from off-by-default to on-by-default at any time.  The reverse moves are not necessarily okay, though.

dfawley: So who makes the decision to make a metric on or off by default?  Each language owner?

markdroth: I think we should assume that all metrics are intended to eventually be on by default unless a gRFC explicitly states otherwise.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,154 @@
...
+| grpc.xds.resource_type | required | Indicates an xDS resource type, such as "envoy.config.listener.v3.Listener". |
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.xds_client.connected | Gauge | {bool} | grpc.target, grpc.xds.server | Whether or not the xDS client currently has a working ADS stream to the xDS server.  For a given server, this will be set to 1 when the stream is initially created.  It will be set to 0 when we have a connectivity failure or when the ADS stream fails without seeing a response message, as per [A57].  Once set to 0, it will be reset to 1 when we receive the first response on an ADS stream. |
```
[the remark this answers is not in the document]
markdroth: I think we will be able to tell that by seeing the backup server listed in this metric, so it doesn't seem necessary to add an additional one.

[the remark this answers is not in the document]
yousukseung: I think a counter type could provide a more reliable signal on the control plane availability, it may be worth it to add separately.

IMO gauge is good to watch trends etc but is unreliable to track occurrences of an event since it loses what happened between snapshots collected. I expect monitoring client TD fail over is one common use case, then a gauge can be ineffective in monitoring brief control plane unavailability events.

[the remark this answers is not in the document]
markdroth: How likely is it that we'll be using the fallback server for a shorter period of time than the metric collection interval?  If we think that will be a common enough occuurance, then I agree that a counter may be useful.  But if not, the counter won't really add much value.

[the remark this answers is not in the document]
yousukseung: I think an outage is more likely last longer than the collection period but not always, and we may never know the extend of short outages if they happen. In prod the time granularity is configurable, AFAIK 1 minute is the shortest and the default.

One example we have is micro outages we detected from a swarm of Flume clients when we started TD migration. The gRPC Chubby fallback counter was sensitive enough to detect and alert this which we tentatively concluded TD overload/contention leading to timeouts, so the failure was gray. I think a gauge could have shown something eventually but certainly would have been much less sensitive in this case as they were brief and gray.

yousukseung, line 132: Unless the client may connect to multiple xds servers per {target, resource_type}, isn't a string gauge that records the xds server address (or none) better? I think this is closer to how gauge counters are typically used.

A78-grpc-metrics-wrr-pf-xds.md
ejona86: > it will probably happen more when we decide that we want to fully commit to the metric definitions

How would someone implementing the gRFC know if we have committed? As far as I'm concerned, the "Temporary environment variable protection" section should really say when the protection can be removed, which is generally "when interop tests pass." That's even more important here if this is a cross-language decision.

Even if we have a normal process, I do think it is appropriate to mention that things will start experimental and reference that process in the gRFC. Messing this up messes up all implementations, which is to me is why we have the section about the environment variable protection. (I don't care much about the standardized variable name; that is convenient but non-essential. The benefit was "don't pollute new features with broken implementations, which then require design workarounds".)

> I think we should assume that all metrics are intended to eventually be on by default unless a gRFC explicitly states otherwise.

That is definitely not part of A79.

>> Once they are stable, they can optionally be made on-by-default.

That language tells me we should document in the gRFC whether it will be on-by-default.

A78-grpc-metrics-wrr-pf-xds.md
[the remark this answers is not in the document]
markdroth: Can't we detect these cases by looking at server-side statistics on the fallback server?  If any client is talking to the fallback server, we know it wasn't able to use the primary.

I'm not strongly opposed to adding a counter if we really need it.  I just want to be convinced that we can't actually get what we need from the existing data.

markdroth: In the general case, it's entirely possible for a client to talk to multiple servers for a given target, due to federation.

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -0,0 +1,154 @@
...
+
+When xDS is used, it is desirable for some metrics to include an optional
+label indicating which xDS locality the metrics are associated with.
+We want to provide this optional label for the metrics in both the
+existing per-call metrics defined in [A66] and in the new metrics for
+the WRR LB policy, described below.
+
+#### Per-Call Metrics
+
+To support the locality label in the per-call metrics, we will provide
```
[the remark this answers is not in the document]
markdroth: I think if the label is enabled, we should always add it.  I'd fine setting it to the empty string.

A78-grpc-metrics-wrr-pf-xds.md
markdroth: Okay, I've added a section on metric stability.  My expectation is that we will wind up cutting-and-pasting this exact text into all future gRFCs that define new metrics.

A78-grpc-metrics-wrr-pf-xds.md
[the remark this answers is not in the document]
yashykt: I think either "" or "unknown" is fine. Let's add that language?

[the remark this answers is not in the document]
yashykt: Also, I think we need to specify which metrics this will show up on - 
* grpc.client.attempt.duration
* grpc.client.attempt.sent_total_compressed_message_size
* grpc.client.attempt.rcvd_total_compressed_message_size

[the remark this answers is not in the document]
markdroth: Added those details, as well as addressing your suggestion from #36049.

A78-grpc-metrics-wrr-pf-xds.md
ejona86: Okay, given the updated text, for this gRFC we _aren't_ making stable metrics (opposite of what you were saying earlier). That's fine. I'm very glad we added that, because I don't think we'd have done similar things cross-language.

dfawley (approved):

A78-grpc-metrics-wrr-pf-xds.md
markdroth: Summarizing our offline conversation for suture reference:

The main use-case is knowing when the server is spamming the client with tons of resource updates.  For this purpose, we really want to show the updates even when the resources haven't changed.  It's true that the value of this metric will change if/when we switch to delta, but it will still reflect the underlying reality of how many updates we're processing from the control plane.

The one reason it might be useful to report only when the resources change is to know when things like EDS assignments change.  However, it's very hard to know when an assignment change is meaningul -- e.g., changing the order of endpoints within a locality is technically a change but matters only if the client is using pick_first -- and it's also hard to reliably detect these differences in a way that we can commit to keep working.  So we think the ROI is not favorable to try to address that use-case.

ejona86 (approved):

A78-grpc-metrics-wrr-pf-xds.md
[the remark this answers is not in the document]
yousukseung: Likely for the primary failure but not for the fallback failure. From what we saw so far , I expect backup failure can happen if the control plane server fails to keep up with the growth. When it happens I think only the client side counter type can really tell us the actual magnitude of the impact.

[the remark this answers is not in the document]
markdroth: A fallback counter wouldn't help to catch cases where the fallback server is unavailable, because there's nothing to fall back to from there.  It sounds like what you actually want here is a counter of the number of times the channel for a given server goes from healthy to unhealthy.  Is that what you want?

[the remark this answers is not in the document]
markdroth: I've added a proposed counter for this.  PTAL.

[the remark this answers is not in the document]
yousukseung: Does it count as a server failure when the currently active subchannel fails but it could switch to another successfully? If not, this LGTM.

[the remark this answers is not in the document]
markdroth: No, it won't.  As per A57, we consider it failing only when either (a) the channel reports TRANSIENT_FAILURE or (b) the ADS stream fails without the client having seen any response messages.

yousukseung (approved):

PR #421, opened 2024-02-23
A79: Non-per-call Metrics Architecture

yashykt:


A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+recorded.
+
+When registering an instrument, a hint is provided as to whether stats plugins
+should consider that instrument on-by-default or off-by-default. The rationale
+being that component/plugin writers would have the best judgment on whether a
+metric should be enabled by default or not, especially if collecting the
+measurement is expensive.
+
+Instrument descriptors are expected to be registered at static initialization
+time (initialization before main). Implementations can optionally provide the
```
markdroth, line 93: I don't think we should say that implementations can support dynamic registration, because we haven't designed that yet.  Instead, let's say that we may decide to add this capability in the future, but it's not part of the design for now.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+
+(To be filled.)
+
+##### Go
+
+(To be filled.)
+
+#### Global Stats Plugin Registry
+
+A global stats plugin registry is added. This registry keeps track of all stats
```
markdroth, line 174: Didn't we say that Java and Go aren't going to have a global registry, or maybe they would have a global registry that can set only a single stats plugin?

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+    third-party plugins to gRPC (custom load balancing policies/resolvers) might
+    want to use the same architecture to record metrics as well.
+*   Channel-scoped stats plugins - Applications should be able to register stats
+    plugins such that metrics for only certain channels are recorded on those
+    plugins. (Current implementations only support global registration of stats
+    plugins.) There can be multiple stats plugins registered as well.
+*   gRPC/Third-party plugins control the metric instrument definition - As gRPC
+    library/component writers, we want to be in control of the metric definition
+    that gRPC provides, irrespective of the stats plugin used. OpenTelemetry is
+    the only supported stats plugin as of now, but that might change in the
```
markdroth, line 47: I think we should explicitly state somewhere that we do not currently plan to provide a public API for third-party stats plugin implementations, but that we might make such an API public in the future.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+the channel's target and the authority is provided to the stats plugin to make
+this decision.
+
+![](A79_graphics/stats-plugin-scoping.png)
+
+In the second approach, stats plugins that are to be scoped to a channel are
+registered with the corresponding channel directly, instead of the global stats
+plugin registry. On channel creation time, stats plugins from the registry are
+combined with the list of stats plugins registered directly on the channel to
+form the complete list of stats plugins for this channel. There is a known issue
```
markdroth, line 203: I would start a new paragraph with the sentence starting with "There is a known issue", since this is really a separate concept from what the first couple of sentences are talking about.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+with this approach. There are certain components like XdsClient instances that
+are not 1:1 to the channel but instead to the channel's target. (Multiple
+channels with the same xDS target will use the same XdsClient instance.) If
```
markdroth, line 206: I think this should mention that the change from a singleton XdsClient to XdsClient per target is described in gRFC A71, which should also be listed in the "Related proposals" list.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+    `grpc.resolver.dns.`.
+*   Metrics from transports should be nested under `grpc.transport.`, e.g.,
+    `grpc.transport.http2.`
+
+### Metric Stability Guidelines and Guarantees
+
+From a stability perspective, metrics exported by gRPC will be in one of the
+following categories -
+
+*   On by default
```
markdroth, line 307: I think these options would be a bit clearer if presented as a flat list:

- Experimental, off by default
- Stable but off by default
- Stable and on by default

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+to record the additional optional labels. One suggested approach of implementing
+this is for the plugin instance to store a bit set for each instrument where
+each bit denotes whether the plugin should record the label in the corresponding
+position of the set of optional labels.
+
+API to enable/disable metrics - Users should be able to override the default
+behavior.
+
+Channel Scoped OpenTelemetry Plugins - API should be provided based on the
+approach chosen from the `Scoping Stats Plugins for Channels` section.
```
markdroth, line 336: This can actually be a link to the relevant section.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+[A66: OpenTelemetry Metrics](A66-otel-stats.md).
+
+```c++
+class OpenTelemetryPluginBuilder {
+  // Records \a optional_label_key on all metrics that provide it.
+  void AddOptionalLabel(absl::string_view optional_label_key);
+  // Set scope filter to choose which channels are recorded by this plugin.
+  // Server-side recording remains unaffected.
+  OpenTelemetryPluginBuilder& SetChannelScope(
+      absl::AnyInvocable<bool(ChannelScope /*scope*/) const> scope_filter);
```
markdroth, line 350: I think this will need to be `StatsPlugin::ChannelScope`.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+want to record, components would instead have call-outs to the `ChannelTracer`
+on various interesting events. This approach has two major issues.
+
+1.  The stats plugin defines the metrics schema and every metric would need to
+    be implemented in every supported stats plugin. The stats plugin
+    implementation of the `ChannelTracer` would also need to know about events
+    from all possible gRPC components so that it can record metrics for those
+    components. This would not work for third-party plugins. (Note that one
+    major reason the `CallTracer` architecture was chosen for per-attempt/call
+    metrics was for legacy reasons. It was already implemented in various
```
markdroth, line 384: I don't think the legacy reasons are the main issue here; I think the main reason is performance.  If it weren't for the performance issues, I think we would have switched the per-call metrics in A66 to also be implemented using the non-per-call framework described in this gRFC, since that would have kept gRPC (as opposed to the stats plugin impl) in control of the schema.

markdroth (commented):
This looks really good!  Comments are mostly cosmetic things.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+Describe a cross-language architecture for collecting non-per-call metrics.
+
+## Background
+
+[A66: OpenTelemetry Metrics](A66-otel-stats.md) is a spec for gRPC to collect
+client/server per-attempt/call metrics for OpenTelemetry. It details an
+architecture that uses a CallTracer approach to record various per-call metrics
+like latency and message sizes. This approach works for per-attempt/call metrics
+but does not work so well for non-per-call metrics such as xDS metrics, load
+balancing policy metrics, resolver metrics and transport metrics which are not
```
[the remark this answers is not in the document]
zasweq: Wow that's an interesting link!

A79-non-per-call-metrics-architecture.md
zasweq: For Go no global registry, just the ability to set a global dial option (and eventually server option).

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+    [A78: gRPC OTel Metrics for WRR, Pick First and XdsClient](https://github.com/grpc/proposal/pull/419)
+    defines metrics for WRR, Pick-First and XdsClient components. In the future,
+    we would want to record metrics for other components as well. Additionally,
+    third-party plugins to gRPC (custom load balancing policies/resolvers) might
+    want to use the same architecture to record metrics as well.
+*   Channel-scoped stats plugins - Applications should be able to register stats
+    plugins such that metrics for only certain channels are recorded on those
+    plugins. (Current implementations only support global registration of stats
+    plugins.) There can be multiple stats plugins registered as well.
+*   gRPC/Third-party plugins control the metric instrument definition - As gRPC
```
markdroth, line 44: We should also explicitly state somewhere that the "grpc." namespace is reserved for gRPC, and metrics provided by third-party plugins should use their own namespaces.

A79-non-per-call-metrics-architecture.md
yashykt: added a using statement

A79-non-per-call-metrics-architecture.md
yashykt: For our opentelemetry implementation, we don't have batching support for opentelemetry anyway, and thinking about it, if we wanted to we could batch multiple measurements with the non-per-call metrics architecture as well.

The legacy reasons which resulted in different metrics between the different implementations seems to be the major issue to me.

A79-non-per-call-metrics-architecture.md
yashykt: done

A79-non-per-call-metrics-architecture.md
yashykt: done

A79-non-per-call-metrics-architecture.md
yashykt: done

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
+## Title
```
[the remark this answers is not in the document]
yashykt: done

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+
+*   Author(s): Yash Tibrewal (@yashykt)
+*   Approver: Mark Roth (@markdroth)
+*   Status: In Review
+*   Implemented in: <language, ...>
+*   Last updated: Feb 23, 2024
```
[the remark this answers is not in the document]
yashykt: done

A79-non-per-call-metrics-architecture.md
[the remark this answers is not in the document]
yashykt: :D

A79-non-per-call-metrics-architecture.md
yashykt: Added the comment in the first requirement.

A79-non-per-call-metrics-architecture.md
yashykt: done

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+registered with the corresponding channel directly, instead of the global stats
+plugin registry. On channel creation time, stats plugins from the registry are
+combined with the list of stats plugins registered directly on the channel to
+form the complete list of stats plugins for this channel. There is a known issue
+with this approach. There are certain components like XdsClient instances that
+are not 1:1 to the channel but instead to the channel's target. (Multiple
+channels with the same xDS target will use the same XdsClient instance.) If
+there are multiple channels to the same target (with potentially different stats
+plugins), the XdsClient will use the stats plugins provided by the first channel
+created with that target.
```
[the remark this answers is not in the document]
yashykt: done

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,418 @@
...
+#### C++
+
+The following API will be added to `OpenTelemetryPluginBuilder` introduced in
+[A66: OpenTelemetry Metrics](A66-otel-stats.md).
+
+```c++
+class OpenTelemetryPluginBuilder {
+  using ChannelScope = grpc_core::StatsPlugin::ChannelScope;
+  // Records \a optional_label_key on all metrics that provide it.
+  void AddOptionalLabel(absl::string_view optional_label_key);
```
yijiem, line 359: Where would the values for this `optional_label_key` be provided?

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,418 @@
...
+off-by-default metrics allows gRPC to support metrics that may be desirable in
+some cases but too expensive to collect by default.
+
+### Changes to gRPC OpenTelemetry Plugin
+
+Optional Labels - Users should be able to configure gRPC OpenTelemetry plugins
+to record the additional optional labels. One suggested approach of implementing
+this is for the plugin instance to store a bit set for each instrument where
+each bit denotes whether the plugin should record the label in the corresponding
+position of the set of optional labels.
```
yijiem, line 340: A few questions:

1. Are "the additional optional labels" be configured here the optional label keys registered with the `GlobalInstrumentsRegistry` API or it's the `optional_label_key` added through the `OpenTelemetryPluginBuilder` API?
2. "to store a bit set for each instrument", it sounds like this is to configure the optional label keys registered with the `GlobalInstrumentsRegistry` API since each instrument has different set of optional label keys?
3. If 2 is the case, then what would this API look like? I'm guessing the user of the `OpenTelemetryPluginBuilder` API needs to set this bitset for the instrument that they are interested in configuring, but would they have access to the `GlobalInstrumentHandle` at the point of configuration or they can use the `GlobalInstrumentsRegistry::ForEach` API and use the index to map to the bitset (I guess that probably would work)?

yijiem (commented):
Thanks Yash!

A79-non-per-call-metrics-architecture.md
yashykt: resolved offline

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,405 @@
...
+    library/component writers, we want to be in control of the metric definition
+    that gRPC provides, irrespective of the stats plugin used. OpenTelemetry is
+    the only supported stats plugin as of now, but that might change in the
+    future. Third-party plugins to gRPC that choose to use this architecture
+    would be responsible for the definitions of metrics exported by their
+    plugins.
+*   Stats plugins should be able to enable/disable recorded metrics - gRPC has
+    many components that would want to record metrics. Not all of those metrics
+    will be interesting to stats plugins. Stats plugin writers should allow
+    users to configure which metrics and enabled/disabled.
```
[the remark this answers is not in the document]
yashykt: thanks!

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,418 @@
...
+
+class GlobalStatsPluginRegistry {
+public:
+ class StatsPluginGroup {
+  public:
+   // Use the stats plugin group object to record metrics
+   template <typename T>
+   void CounterAdd(GlobalUInt64CounterHandle handle, T value,
+                   RefCountedPtr<LabelValueSet> label_values,
+                   RefCountedPtr<LabelValueSet> optional_values);
```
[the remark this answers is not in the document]
yashykt: thanks!

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,418 @@
...
+   template <typename T>
+   void HistogramRecord(GlobalDoubleHistogramHandle handle, T value,
+                        RefCountedPtr<LabelValueSet> label_values,
+                        RefCountedPtr<LabelValueSet> optional_values);
+
+   // Use the stats plugin to get a representation of label values that can be
+   // saved for multiple uses later.
+   RefCountedPtr<LabelValueSet> MakeLabelValueSet(
+       absl::Span<absl::string_view> label_values);
```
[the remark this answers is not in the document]
yashykt: thanks!

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,418 @@
...
+ };
+
+ void RegisterStatsPlugin(std::shared_ptr<StatsPlugin> plugin);
```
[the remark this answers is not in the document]
yashykt: thanks!

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,418 @@
...
+
+ // Invoked to get a stats plugin list for a specified scope.
+ static StatsPluginGroup GetStatsPluginsForChannel(ChannelScope scope);
+
+private:
+ std::vector<std::shared_ptr<StatsPlugin>> plugins_;
```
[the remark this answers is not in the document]
yashykt: done

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,426 @@
...
+#### C++
+
+The following API will be added to `OpenTelemetryPluginBuilder` introduced in
+[A66: OpenTelemetry Metrics](A66-otel-stats.md).
+
+```c++
+class OpenTelemetryPluginBuilder {
+  using ChannelScope = grpc_core::StatsPlugin::ChannelScope;
+  // Records \a optional_label_key on all metrics that provide it.
+  void AddOptionalLabel(absl::string_view optional_label_key);
```
yijiem, line 367: This should probably be `OpenTelemetryPluginBuilderImpl& AddOptionalLabel(absl::string_view optional_label_key);` right?

yijiem: Or if we want to make it easier for the user to specify multiple optional label keys, maybe `OpenTelemetryPluginBuilderImpl& AddOptionalLabels(absl::Span<const absl::string_view> optional_label_keys);`?

markdroth (commented):
Only one substantive comment remaining.  The rest are cosmetic.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,426 @@
...
+    for consistency.
+*   Type - Implementation defined instrument type, e.g. Long Counter, Double
+    Histogram.
+*   Set of Label Keys (set of strings) - These determine the set of labels that
+    will be present with each recording/measurement.
+*   Set of Optional Label Keys (set of strings) - This is an additional set of
+    labels that stats plugins can choose to either record or throw away based on
+    user configuration. This is useful in cases where certain labels might be
+    deemed not useful for the majority of the cases, and hence shouldn't be
+    recorded by default. (OpenTelemetry allows users to create custom views that
```
DNVindhya, line 83: I think adding OpenTelemetry documentation on custom views would help here.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,426 @@
...
+##### Scoping Stats Plugins for Channels
+
+To allow stats plugins to scope themselves to a certain set of channels,
+implementations can choose one of two approaches.
+
+In the first approach, stats plugins provide a scope filter that applications
+can configure. Based on the channel's configuration, a stats plugin can decide
+whether it wants to record metrics for the channel. At channel creation time,
+the channel's target and the authority is provided to the stats plugin to make
+this decision.
```
DNVindhya, line 203: The following diagram has Stats Plugin group and the description here does not refer to the concept. Can you add it in the descroption?

A79-non-per-call-metrics-architecture.md
yashykt: Amended the text.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,426 @@
...
+
+### Metric Instrument Naming Conventions
+
+gRPC metrics follow the same general
+[guidelines](https://opentelemetry.io/docs/specs/semconv/general/metrics/#general-guidelines)
+as OpenTelemetry. The following additional guidelines apply -
+
+*   Metrics defined by gRPC should go under the `grpc.` namespace.
+*   Third-party plugins to gRPC should NOT use the `grpc.` namespace.
+*   Newer metrics should follow the same style as existing metrics. * Metrics
```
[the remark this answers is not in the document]
yashykt: Thanks!

A79-non-per-call-metrics-architecture.md
```diff
@@ -200,20 +208,23 @@ In the second approach, stats plugins that are to be scoped to a channel are
...
+
+There is a known issue with this approach. There are certain components like
+XdsClient instances that are not 1:1 to the channel but instead to the channel's
+target. (As per [A71: xDS Fallback](A71-xds-fallback.md) Multiple channels with
+the same xDS target will use the same XdsClient instance.) If there are multiple
+xDS enabled channels to the same target (with potentially different stats
 plugins), the XdsClient will use the stats plugins provided by the first channel
-created with that target.
+created with that target while ignoring the stats plugins passed to the second
+client.
```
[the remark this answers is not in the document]
yashykt: Done

A79-non-per-call-metrics-architecture.md
```diff
@@ -200,20 +208,23 @@ In the second approach, stats plugins that are to be scoped to a channel are
...
-form the complete list of stats plugins for this channel. There is a known issue
-with this approach. There are certain components like XdsClient instances that
-are not 1:1 to the channel but instead to the channel's target. (Multiple
-channels with the same xDS target will use the same XdsClient instance.) If
-there are multiple channels to the same target (with potentially different stats
+form the complete list of stats plugins for this channel.
+
+There is a known issue with this approach. There are certain components like
+XdsClient instances that are not 1:1 to the channel but instead to the channel's
+target. (As per [A71: xDS Fallback](A71-xds-fallback.md) Multiple channels with
```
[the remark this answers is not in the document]
yashykt: Done

A79-non-per-call-metrics-architecture.md
yashykt: Done

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,432 @@
...
+
+*   Metrics defined by gRPC should go under the `grpc.` namespace.
+*   Third-party plugins to gRPC should NOT use the `grpc.` namespace.
+*   Newer metrics should follow the same style as existing metrics.
+*   Metrics from load-balancing policies should be nested under `grpc.lb.`, e.g.
+    `grpc.lb.pick_first.`.
+*   Metrics from resolvers should be nested under `grpc.resolver.`, e.g.
+    `grpc.resolver.dns.`.
+*   Metrics from transports should be nested under `grpc.transport.`, e.g.,
+    `grpc.transport.http2.`
```
dfawley, line 321: Clearly by this you mean "Metrics from first-party gRPC ___s".  Any guidance for third-party plugins for these components?

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,432 @@
...
+*   Set of Optional Label Keys (set of strings) - This is an additional set of
+    labels that stats plugins can choose to either record or throw away based on
+    user configuration. This is useful in cases where certain labels might be
+    deemed not useful for the majority of the cases, and hence shouldn't be
+    recorded by default. (OpenTelemetry allows users to create
+    [custom views](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#view)
+    that can filter labels, but in the absence of custom views, the view takes
+    the entire set of labels provided by the instrumentation library.)
+
+On registering an instrument a `GlobalInstrumentHandle` is returned. The
```
dfawley, line 88: Nit: please add a comma for clarity: "On registering an instrument, a `...`"

dfawley (approved):

markdroth (approved):
Looks great!

A79-non-per-call-metrics-architecture.md
yashykt: The only recommendation from gRPC for third-party plugins is to NOT use the `grpc.` namespace. In general, the same recommendations from OpenTelemetry (https://opentelemetry.io/docs/specs/semconv/general/metrics/#general-guidelines) should be followed

A79-non-per-call-metrics-architecture.md
yashykt: Done

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,598 @@
...
+change or remove these metrics. Documentation for the given metric determines
+whether the metric is experimental or non-experimental. An alternative approach
+of using an `experimental` namespace was considered, but not preferred since
+de-experimentalization of the metric would result in disruption of the
+corresponding metrics data (timeseries).
+
+New metrics should start as experimental. Once they are stable, they can
+optionally be made on-by-default. The distinction between on-by-default and
+off-by-default metrics allows gRPC to support metrics that may be desirable in
+some cases but too expensive to collect by default.
```
ejona86, line 476: I assume "too expensive to collect" is about CPU cost. But there is a storage cost argument as well, after it is . Are we only concerned with CPU cost here? (There could be memory costs, but in general I assume we won't have metrics that are too expensive in memory.)

ejona86 (commented):
Here's what I have, before I get on a flight.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,598 @@
...
+    /** 
+     * Adds optionalLabelKey to all the metrics that can provide value for the 
+     * optionalLabelKey. */
+    public void addOptionalLabel(String optionalLabelKey);
+
+    /** 
+     * Enables metrics specified in the set along with metrics that are enabled 
+     * by default.
+     */
+    public Builder enableMetrics(Set<String> enableMetrics);
```
ejona86, line 543: @DNVindhya, users won't have a set for this already. We'll probably want to use `Collection` or `Iterable` instead.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,598 @@
...
+    
+    /** Disable metrics specified in the set. */
+    public Builder disableMetrics(Set<String> disableMetrics);
+    
+    /** Disable all metrics that are enabled by default. */
```
ejona86, line 548: @DNVindhya, Does this _only_ disable metrics disabled by default, or if someone had called enabledMetrics, would it remove those as well?

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,598 @@
...
+  // Register and return a double histogram metric descriptor for recording 
+  // double histogram values.
+  public DoubleHistogramDescriptor registerDoubleHistogram(
+      String name, String description, String unit, List<Double> bucketBoundaries,
+      List<String> labelKeys, List<String> optionalLabelKeys, 
+      boolean isEnabledByDefault);
+}
+
+interface MetricDescriptor {
+  long index;
```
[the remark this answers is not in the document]
DNVindhya: My bad. Overlooked the part of data member being final and static. 
How about I add common methods for descriptors like `getIndex()`, `getName()` in  an interface and common data members like name, description and unit in an abstract class. Then `LongCounterDescriptor` can implement the interface and extend the abstract class?

A79-non-per-call-metrics-architecture.md
DNVindhya: I just checked C++, this will disable all metrics even the ones added by `enableMetrics`. I will update the comment.

A79-non-per-call-metrics-architecture.md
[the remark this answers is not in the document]
ejona86: Having an interface with the methods is fine. We can still have an abstract class, but it can be an implementation detail (not `public`).

A79-non-per-call-metrics-architecture.md
yashykt: yeah, we are mostly worried about collection costs here which I assume would be mostly CPU, maybe memory as well, if, for example, the metric now requires to store some information for each RPC.

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,598 @@
...
+
+##### Scoping Stats Plugins for Channels
+
+To allow stats plugins to scope themselves to a certain set of channels,
+implementations can choose one of two approaches.
+
+In the first approach, stats plugins provide a scope filter that applications
+can configure. Based on the channel's configuration, a stats plugin can decide
+whether it is interested in the channel. At channel creation time, the channel's
+target and the authority is provided to the stats plugin to make this decision.
```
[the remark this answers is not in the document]
yashykt: Yeah, it's the canonical target.

For the authority, it's the default authority for the channel. If the user overrides the default one for the channel, that will be used, otherwise it'll be derived from the target.

ejona86 (approved):

A79-non-per-call-metrics-architecture.md
```diff
@@ -253,9 +258,10 @@ implementations can choose one of two approaches.
 In the first approach, stats plugins provide a scope filter that applications
 can configure. Based on the channel's configuration, a stats plugin can decide
 whether it is interested in the channel. At channel creation time, the channel's
-target and the authority is provided to the stats plugin to make this decision.
-Various gRPC components use this list of interested stats plugins
-(`StatsPluginGroup`) to record metrics.
+canonical target and the default authority is provided to the stats plugin to
+make this decision. (The default authority provided is derived from the target
+unless an override was specified on the channel) Various gRPC components use
```
markdroth, line 263: Nit: Need a period before the close-paren.

yashykt: hawkeye :)

DNVindhya (approved):

A79-non-per-call-metrics-architecture.md
```diff
@@ -0,0 +1,605 @@
...
+unless an override was specified on the channel.) Various gRPC components use
+this list of interested stats plugins (`StatsPluginGroup`) to record metrics.
+
+![](A79_graphics/stats-plugin-scoping.png)
+
+In the second approach, stats plugins that are to be scoped to a channel are
+registered with the corresponding channel directly, instead of the global stats
+plugin registry. On channel creation time, stats plugins from the registry are
+combined with the list of stats plugins registered directly on the channel to
+form the complete list of stats plugins for this channel.
```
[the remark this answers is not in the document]
yashykt: It is possible that we will support it in the future.

XuanWang-Amos (approved):

PR #423, opened 2024-04-11
A68: Random subsetting with rendezvous hashing LB policy

s-matyukevich:
Replaces https://github.com/grpc/proposal/pull/383

Related to https://github.com/grpc/proposal/pull/430  which describes an LB policy that could be used in combination with random subsetting to correct the resulting server-side load imbalance.


A68-random-subsetting.md
```diff
@@ -0,0 +1,117 @@
...
+
+Introduce a new LB policy, `random_subsetting`. This policy selects a subset of addresses and passes them to the child LB policy. It maintains 2 important properties:
+* The policy tries to distribute connections among servers as equally as possible. The higher `(N_clients*subset_size)/N_servers` ratio is, the closer the resulting server connection distribution is to uniform.
+* The policy minimizes the amount of connection churn generated during server scale-ups by using [rendezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing)
+
+### Subsetting algorithm
+
+* The policy receives a single configuration parameter: `subset_size`, which must be configured by the user.
+* When the lb policy is initialized it also creates a random 32-byte long `salt` string. 
+* After every resolver update the policy picks a new subset. It does this by implementing `rendezvous hashing` algorithm:
```
atollena, line 31: I think it would help to define the algorithm in pseudo code. See https://github.com/grpc/proposal/blob/master/A55-xds-stateful-session-affinity.md#lb-policy-for-stateful-session-affinity or https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md#aggregated-connectivity-state for examples of gRFCs that do this.

A68-random-subsetting.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### Low connection churn during server rollouts
+
+The graphs provided in the previous section prove this is the case in practice (we rollout all servers in the middle of every test, and there is no visible increase in the number of connections per server) Low connection churn during server rollouts is the primary motivation why rendezvous hashing was used as the subsetting algorithm: it guaranties that if a single server is either added or removed to the IP address list, every client will update at most 1 entry in its subset. This is because the hashes for all unaffected servers remain the same, which guarantees that the order of the servers after sorting also remains stable. The same logic applies to the situation when multiple servers got updated. 
+
+### LB Policy Config and Parameters
+
+The `random_subsetting` LB policy config will be as follows.
+
+```
```
atollena, line 64: You can use the proto language (this tip was given to me by another maintainer).

```suggestion
```proto
```

A68-random-subsetting.md
```diff
@@ -0,0 +1,117 @@
...
+
+    // The config for the child policy.
+    repeated LoadBalancingConfig child_policy = 2;
+}
+```
+
+
+### Handling Parent/Resolver Updates
+
+When the resolver updates the list of addresses, or the LB config changes, Random subsetting LB will run the subsetting algorithm, described above, to filter the endpoint list. Then it will create a new resolver state with the filtered list of the addresses and pass it to the child LB. Attributes and service config from the old resolver state will be copied to the new one.
```
atollena, line 83: I think you should replace addresses with endpoints to take [A61](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md) into consideration.

A68-random-subsetting.md
```diff
@@ -0,0 +1,117 @@
...
+## Background
+
+Currently, gRPC is lacking a way to select a subset of endpoints available from the resolver and load-balance requests between them. Out of the box, users have the choice between two extremes: `pick_first` which sends all requests to one random backend, and `round_robin` which sends requests to all available backends. `pick_first` has poor connection balancing when the number of client is not much higher than the number of servers. The problem is exacerbated during rollouts because `pick_first` does not change endpoint on resolver updates if the current subchannel remains `READY`. `round_robin` results in every server having as many connections open as there are clients, which is unnecessarily costly when there are many clients, and makes local load balancing decisions (such as outlier detection) less precise.
+
+### Related Proposals: 
+* [gRFC A27: A52: gRPC xDS Custom Load Balancer Configuration](https://github.com/grpc/proposal/blob/master/A52-xds-custom-lb-policies.md)
+
+## Proposal
+
+Introduce a new LB policy, `random_subsetting`. This policy selects a subset of addresses and passes them to the child LB policy. It maintains 2 important properties:
```
atollena, line 23: I think you need to replace `addresses` with `endpoints` to account for https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md, where each endpoint may have multiple addresses.

A68-random-subsetting.md
```diff
@@ -0,0 +1,117 @@
...
+
+### xDS Integration
+
+Random subsetting LB won't depend on xDS in any way. People may choose to initialize it by directly providing service config. We will only provide a corresponding xDS policy wrapper to allow configuring this LB via xDS.
+
+#### Changes to xDS API
+
+`random_subsetting` will be added as a new LB policy.
+
+```textproto
```
atollena, line 97: This should be 

```suggestion
```proto
```

textproto is used to define the content of a instance of a protobuf message.

A68-random-subsetting.md
```diff
@@ -0,0 +1,117 @@
...
+Introduce a new LB policy, `random_subsetting`. This policy selects a subset of addresses and passes them to the child LB policy. It maintains 2 important properties:
+* The policy tries to distribute connections among servers as equally as possible. The higher `(N_clients*subset_size)/N_servers` ratio is, the closer the resulting server connection distribution is to uniform.
+* The policy minimizes the amount of connection churn generated during server scale-ups by using [rendezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing)
+
+### Subsetting algorithm
+
+* The policy receives a single configuration parameter: `subset_size`, which must be configured by the user.
+* When the lb policy is initialized it also creates a random 32-byte long `salt` string. 
+* After every resolver update the policy picks a new subset. It does this by implementing `rendezvous hashing` algorithm:
+  * Concatenate `salt` to each address in the list.
```
atollena, line 32: You'll have to decide which address, in case the endpoint has more than one (I think you can use the first address?).

A68-random-subsetting.md
```diff
@@ -0,0 +1,117 @@
...
+}
+```
+
+As you can see, the fields in this policy match exactly the fields in the random subsetting LB service config.
+
+#### Integration with xDS LB Policy Registry
+As described in [gRFC A52](https://github.com/grpc/proposal/blob/master/A52-xds-custom-lb-policies.md), gRPC has an LB policy registry, which maintains a list of converters. Every converter translates xDS LB policy to the corresponding service config. In order to allow using the Random subsetting LB policy via xDS, the only thing that needs to be done is providing a corresponding converter function. The function implementation will be trivial as the fields in the xDS LB policy will match exactly the fields in the service config.
+
+## Rationale
+### Alternatives Considered: Deterministic subsetting
```
atollena, line 111: You should probably discuss the trade offs of doing this kind of subsetting in the control plane, since it was discussed in the original proposal.

s-matyukevich: Yeah, but I posted a link to the tl;dr; of the discussion, so you think this is not enough?

atollena: I'm thinking of the option of doing random subsetting in the control plane by sending different EDS responses (with different subsets) to each dataplane, or the equivalent with other resolvers. It is simple to implement with xDS and works for Envoy and gRPC. IIRC the main argument for not going that route is the need to have an xDS infrastructure (this is a big barrier for our orgs, and probably others), and existing limitations of https://github.com/envoyproxy/go-control-plane.

This was discussed in https://github.com/grpc/proposal/pull/383/files#r1308024474.

In order to understand this proposal, I think users will need to understand the trade off of doing it as a balancer in each data plane rather than directly in service discovery.

A68-random-subsetting.md
```diff
@@ -0,0 +1,117 @@
...
+* The policy tries to distribute connections among servers as equally as possible. The higher `(N_clients*subset_size)/N_servers` ratio is, the closer the resulting server connection distribution is to uniform.
+* The policy minimizes the amount of connection churn generated during server scale-ups by using [rendezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing)
+
+### Subsetting algorithm
+
+* The policy receives a single configuration parameter: `subset_size`, which must be configured by the user.
+* When the lb policy is initialized it also creates a random 32-byte long `salt` string. 
+* After every resolver update the policy picks a new subset. It does this by implementing `rendezvous hashing` algorithm:
+  * Concatenate `salt` to each address in the list.
+  * For every resulting entity compute [MurmurHash3](https://en.wikipedia.org/wiki/MurmurHash) hash, which produces 128-byte output.
```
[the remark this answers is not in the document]
s-matyukevich: updated to use xxhash, This changes the algorithms slightly as we can use random pre-generated seed instead of concatenating salt to each address. The new version is even simpler.

A68-random-subsetting.md
s-matyukevich: Done.

A68-random-subsetting.md
s-matyukevich: Done.

A68-random-subsetting.md
s-matyukevich: Updated the doc to use the first address.

A68-random-subsetting.md
s-matyukevich: Done.

s-matyukevich:
Bump on this. It has been almost a month since the proposal was submitted and no one from gRPC maintainers commented on it yet. cc @markdroth and @ejona86 since you both reviewed previous version of the proposal and have full context.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+
+```proto
+message LoadBalancingConfig {
+    oneof policy {
+        RandomSubsettingLbConfig random_subsetting = 21 [json_name = "random_subsetting"];
+    }
+}
+message RandomSubsettingLbConfig {
+    // subset_size indicates how many backends every client will be connected to.
+    // Default is 20.
```
markdroth, line 96: Do we want this field to be optional or required?  If optional, is 20 the right default?

My gut reaction (but I'm certainly open to arguments here) is that we don't really know what default would be appropriate for all environments, so we might want to just make this field required.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+
+    // The config for the child policy.
+    repeated LoadBalancingConfig child_policy = 2;
+}
+```
+
+
+### Handling Parent/Resolver Updates
+
+When the resolver updates the list of endpoints, or the LB config changes, Random subsetting LB will run the subsetting algorithm, described above, to filter the endpoint list. Then it will create a new resolver state with the filtered list of the endpoints and pass it to the child LB. Attributes and service config from the old resolver state will be copied to the new one.
```
markdroth, line 107: "Attributes and service config from the old resolver state will be copied to the new one."

I'm not sure what that means.  In general, all of the attributes should be passed down with each resolver update, so I don't think we want to retain attributes from the previous update if they were removed in the new update.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### Changes to xDS API
+
+`random_subsetting` will be added as a new LB policy.
+
+```proto
+package envoy.extensions.load_balancing_policies.random_subsetting.v3;
+message RandomSubsetting {
+  google.protobuf.UInt32Value subset_size = 1;
+  repeated LoadBalancingConfig child_policy = 2;
```
markdroth, line 125: I think this will want to be:

```
config.cluster.v3.LoadBalancingPolicy child_policy = 2;
```

Note that it does *not* need to be `repeated`, because in xDS the list of options is inside of the `LoadBalancingPolicy` message.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+Random subsetting LB won't depend on xDS in any way. People may choose to initialize it by directly providing service config. We will only provide a corresponding xDS policy wrapper to allow configuring this LB via xDS.
+
+#### Changes to xDS API
+
+`random_subsetting` will be added as a new LB policy.
+
+```proto
+package envoy.extensions.load_balancing_policies.random_subsetting.v3;
+message RandomSubsetting {
+  google.protobuf.UInt32Value subset_size = 1;
```
markdroth, line 124: See comments above about whether this is required or optional, what the field type is, etc.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
+A68: Random subsetting with rendezvous hashing LB policy.
+----
+* Author(s): @s-matyukevich
+* Approver:
```
markdroth, line 4: You can list me as the approver.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+* N_clients = 100, N_servers = 100, subset_size = 25
+![State Diagram](A68_graphics/subsetting100-100-25.png)
+* N_clients = 100, N_servers = 10, subset_size = 5
+![State Diagram](A68_graphics/subsetting100-10-5.png)
+* N_clients = 500, N_servers = 10, subset_size = 5
+![State Diagram](A68_graphics/subsetting500-10-5.png)
+* N_clients = 2000, N_servers = 10, subset_size = 5
+![State Diagram](A68_graphics/subsetting2000-10-5.png)
+
+#### Low connection churn during server rollouts
```
markdroth, line 80: This is a very nice property!

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Handling Subchannel Connectivity State Notifications
+
+Random subsetting LB will simply redirect all requests to the child LB without doing any additional processing. This also applies to all other callbacks in the LB interface, besides the one that handles resolver and config updates (which is described in the previous section). This is possible because random subsetting LB doesn't store or manage sub-connections - it acts as a simple filter on the resolver state, and that's why it can redirect all actual work to the child LB. 
+
+### xDS Integration
+
+Random subsetting LB won't depend on xDS in any way. People may choose to initialize it by directly providing service config. We will only provide a corresponding xDS policy wrapper to allow configuring this LB via xDS.
+
+#### Changes to xDS API
```
markdroth, line 117: It's probably worth chatting with @wbpcode to make sure this design would also work for Envoy.  I don't know of any reason it wouldn't, but might make sense to check just in case.

markdroth (commented):
Sorry for the delay reviewing this!  Overall, it looks very good -- my comments here are mostly minor.

One of the reasons for the delay was that we were having a bunch of dicussions internally about ways to potentially reduce the number of unnecessary idle connections with policies other than pick_first, and I wanted to make sure that where we landed on that didn't conflict with this proposal.  Just as an FYI, let me sketch out what we have in mind to eventually do for that, so that you know where we're heading.

The general problem that we're trying to solve is that LB policies like RR and WRR proactively maintain connections to all endpoints and therefore waste a lot of resources on idle connections when the client is idle.  To deal with this, we will eventually want to add an LB policy that tracks the number of simultaneous RPCs in flight on the channel over the last N minutes and scale the number of connections accordingly.  This new policy would operate similarly to the subsetting policy described in this gRFC: it would filter the set of addresses passed down to the child policy.

One of the things we were discussing was how this theoretical new policy would interact with the subsetting policy described in this gRFC.  We wound up deciding that we would probably just put this theoretical new policy underneath the subsetting policy, so that the two policies can basically be independent of each other.  Obviously, this will probably make the actual connection distribution a bit less even, but it will do so only in the face of idle clients, and it will not actually break the distribution imposed by the subsetting policy -- it would just eliminate some of the connections that would have been chosen by the subsetting policy but would have been idle.

Anyway, as I mentioned, this gRFC looks very good!  I'd like to get reviews from @ejona86 and @dfawley as well, but I don't foresee any significant blockers.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+## Abstract
+
+Add support for the `random_subsetting` load balancing policy.
+
+## Background
+
+Currently, gRPC is lacking a way to select a subset of endpoints available from the resolver and load-balance requests between them. Out of the box, users have the choice between two extremes: `pick_first` which sends all requests to one random backend, and `round_robin` which sends requests to all available backends. `pick_first` has poor connection balancing when the number of client is not much higher than the number of servers. The problem is exacerbated during rollouts because `pick_first` does not change endpoint on resolver updates if the current subchannel remains `READY`. `round_robin` results in every server having as many connections open as there are clients, which is unnecessarily costly when there are many clients, and makes local load balancing decisions (such as outlier detection) less precise.
+
+### Related Proposals: 
+* [gRFC A27: A52: gRPC xDS Custom Load Balancer Configuration](https://github.com/grpc/proposal/blob/master/A52-xds-custom-lb-policies.md)
```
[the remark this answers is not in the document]
s-matyukevich: done.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Proposal
+
+Introduce a new LB policy, `random_subsetting`. This policy selects a subset of endpoints and passes them to the child LB policy. It maintains 2 important properties:
+* The policy tries to distribute connections among servers as equally as possible. The higher `(N_clients*subset_size)/N_servers` ratio is, the closer the resulting server connection distribution is to uniform.
+* The policy minimizes the amount of connection churn generated during server scale-ups by using [rendezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing)
+
+### Subsetting algorithm
+
+* The policy receives a single configuration parameter: `subset_size`, which must be configured by the user.
```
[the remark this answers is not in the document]
s-matyukevich: The field should be required, I removed the default value. I moved the config section accordingly to your suggestion, but kept the bullet point for now because It describes the origin of `subset_size` variable used in the pseudocode below.  Can remove it if you still think it is unnecessary.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+* N_clients = 500, N_servers = 10, subset_size = 5
+![State Diagram](A68_graphics/subsetting500-10-5.png)
+* N_clients = 2000, N_servers = 10, subset_size = 5
+![State Diagram](A68_graphics/subsetting2000-10-5.png)
+
+#### Low connection churn during server rollouts
+
+The graphs provided in the previous section prove this is the case in practice (we rollout all servers in the middle of every test, and there is no visible increase in the number of connections per server) Low connection churn during server rollouts is the primary motivation why rendezvous hashing was used as the subsetting algorithm: it guaranties that if a single server is either added or removed to the endpoints list, every client will update at most 1 entry in its subset. This is because the hashes for all unaffected servers remain the same, which guarantees that the order of the servers after sorting also remains stable. The same logic applies to the situation when multiple servers got updated. 
+
+### LB Policy Config and Parameters
```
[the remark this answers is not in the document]
s-matyukevich: done.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+
+The `random_subsetting` LB policy config will be as follows.
+
+```proto
+message LoadBalancingConfig {
```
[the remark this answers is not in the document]
s-matyukevich: Sure, I'll do that.

A68-random-subsetting.md
```diff
@@ -0,0 +1,141 @@
...
+    oneof policy {
+        RandomSubsettingLbConfig random_subsetting = 21 [json_name = "random_subsetting"];
+    }
+}
+message RandomSubsettingLbConfig {
+    // subset_size indicates how many backends every client will be connected to.
+    // Default is 20.
+    google.protobuf.UInt32Value subset_size = 1;
```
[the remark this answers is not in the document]
s-matyukevich: Updated the value to uint32 and added the comment indicating that value must be >= 1.

A68-random-subsetting.md
s-matyukevich: Yes, this was a mistake. I removed the default value.

A68-random-subsetting.md
s-matyukevich: That is exactly what I meant, but I phrased it poorly, updated to `Attributes and service config from the parent resolver state will be passed down with each resolver update`

A68-random-subsetting.md
s-matyukevich: This was a bad copy-paste. Updated.  
Should the corresponding field in the LB policy config be repeated? I removed repeated keyword for now as in this case the child policy is mandatory.

A68-random-subsetting.md
s-matyukevich: Changed subset_size type from google.protobuf.UInt32Value to uint32 and made it mandatory. Also copy-pasted corresponding comments from the policy config definition.

A68-random-subsetting.md
s-matyukevich: Done

A68-random-subsetting.md
markdroth: The reason for `repeated` here is to allow new policies to be configured without breaking older clients.  The client will iterate through the list of child policies, stopping at the first one that it supports.  So let's say that you start with a bunch of clients that support only policy A, and now you want to move to a new policy B, but not all of your clients support B yet.  The control plane can send the list [B, A], which will allow new clients to use B while still allowing older clients to use A.

In the gRPC service config, the `LoadBalancingConfig` message represents a single policy, so we need to make the field `repeated` wherever it gets used.  In xDS, the `config.cluster.v3.LoadBalancingPolicy` message itself contains [a `repeated` list of policies](https://github.com/envoyproxy/envoy/blob/7f7234be42dd7f90c612c193187a65775c119607/api/envoy/config/cluster/v3/cluster.proto#L1235), so there's no need to make it repeated.  (In hindsight, I wish we'd made the same kind of intermediate message in the gRPC service config, so that we didn't always have to remember to make the field `repeated` wherever we use it.  At the time we were concerned with trying to minimize the length of the JSON representation to allow it to squeeze into DNS TXT records, which is why we chose the current approach.)

So yes, please do add `repeated` in the gRPC service config representation.

ejona86 (commented):
> To deal with this, we will eventually want to add an LB policy that tracks the number of simultaneous RPCs in flight on the channel over the last N minutes and scale the number of connections accordingly.

One thing for me was how similar the subsetting would be between the two load balancer policies. But the subsetting algorithm here could be used directly by the other LB policy as well. (The other LB policy could maybe just delegate to this, changing its subset_size over time; or we extend this to support dynamic subset size. All the options remain on the table with this gRFC.)

A68-random-subsetting.md
```diff
@@ -0,0 +1,147 @@
...
+    LoadBalancingConfig child_policy = 2;
+}
+```
+
+### Subsetting algorithm
+
+* The policy receives a single configuration parameter: `subset_size`, which must be configured by the user.
+* When the lb policy is initialized it also creates a random integer `seed`. 
+* After every resolver update the policy picks a new subset. It does this by implementing `rendezvous hashing` algorithm:
+  * For every endpoint in the list compute 64-bit hash using `XXH64` function as deined in https://github.com/Cyan4973/xxHash with previosly generated seed.
```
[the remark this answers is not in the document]
s-matyukevich: Replaced with:

A68-random-subsetting.md
```diff
@@ -0,0 +1,147 @@
...
+* Last updated: 2024-04-15
+* Discussion at: https://groups.google.com/g/grpc-io/c/oxNJT1GgjEg
+
+## Abstract
+
+Add support for the `random_subsetting` load balancing policy.
+
+## Background
+
+Currently, gRPC is lacking a way to select a subset of endpoints available from the resolver and load-balance requests between them. Out of the box, users have the choice between two extremes: `pick_first` which sends all requests to one random backend, and `round_robin` which sends requests to all available backends. `pick_first` has poor connection balancing when the number of client is not much higher than the number of servers. The problem is exacerbated during rollouts because `pick_first` does not change endpoint on resolver updates if the current subchannel remains `READY`. `round_robin` results in every server having as many connections open as there are clients, which is unnecessarily costly when there are many clients, and makes local load balancing decisions (such as outlier detection) less precise.
```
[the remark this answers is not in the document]
s-matyukevich: yes, done.

A68-random-subsetting.md
s-matyukevich: Added `repeated` back to RandomSubsettingLbConfig.

A68-random-subsetting.md
```diff
@@ -0,0 +1,148 @@
...
+    repeated LoadBalancingConfig child_policy = 2;
+}
+```
+
+### Subsetting algorithm
+
+* The policy receives a single configuration parameter: `subset_size`, which must be configured by the user.
+* When the lb policy is initialized it also creates a random integer `seed`. 
+* After every resolver update the policy picks a new subset. It does this by implementing `rendezvous hashing` algorithm:
+  * For every endpoint in the list compute a hash with previosly generated seed. Any uniform hash function should work, but we suggest using `XXH64` function as deined in https://github.com/Cyan4973/xxHash since it is already used in [gRFC A42](https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md).
```
ejona86, line 54: nit: s/previosly/previously/

Would it be improved by adding a comma after "list" in the first sentence?

s-matyukevich: fixed.

ejona86 (approved):

atollena:
> The general problem that we're trying to solve is that LB policies like RR and WRR proactively maintain connections to all endpoints and therefore waste a lot of resources on idle connections when the client is idle. To deal with this, we will eventually want to add an LB policy that tracks the number of simultaneous RPCs in flight on the channel over the last N minutes and scale the number of connections accordingly. This new policy would operate similarly to the subsetting policy described in this gRFC: it would filter the set of addresses passed down to the child policy.

This is also the general problem that we are trying to solve with this gRFC and https://github.com/grpc/proposal/pull/430. And making the subset size dynamic would be ideal, but seems much harder, as the subset size impacts load balancing, and server load must be reasonably predictable. I don't know of any published work on making the number of connections dynamic (no fixed subset size) while also keeping the server load reasonably balanced.

ejona86:
> I don't know of any published work on making the number of connections dynamic (no fixed subset size) while also keeping the server load reasonably balanced.

Basically the theory is you would size a minimum subset like you are doing here to a degree. But if the client causes a lot of load, then you scale up, potentially to all backends. IIRC, we were thinking of defining load as "concurrent RPCs." Most of the discussion was spent avoiding wildly scaling up/down, like with bursty clients, causing damage.

markdroth (approved):

markdroth:
> This is also the general problem that we are trying to solve with this gRFC and #430. And making the subset size dynamic would be ideal, but seems much harder, as the subset size impacts load balancing, and server load must be reasonably predictable. I don't know of any published work on making the number of connections dynamic (no fixed subset size) while also keeping the server load reasonably balanced. So I'm very curious where you are at with those discussion and if you have rough ideas on how you would address load balancing.

I think the approach that we have in mind would probably result in slightly less ideally balanced server load than the results you've seen, specifically because the connections will no longer be perfectly balanced.  But in cases where the RPC rate from different clients diverge wildly, or where RPC rates are very bursty, maintaining unnecessary idle connections can cost significant amounts of memory and some additional CPU.  I think there are situations where the costs of that unnecessary overhead is expensive enough that it's worth taking steps to reduce it, even if it results in slightly less optimally balanced server load.

Like anything else in this space, it's a bit of a balancing act (pun intended).  There will definitely be cases where the approach that we have in mind won't work well, but there will be cases where it will.  But we think it will be a useful tool to have in our toolbox at some point.

A68-random-subsetting.md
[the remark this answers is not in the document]
s-matyukevich: Here is the PR https://github.com/grpc/grpc-proto/pull/157

s-matyukevich:
What are the next steps here? Can someone with the right permissions merge this?

ejona86:
@dfawley, Mark is the listed approver currently, but I'm fine merging this as it seems things are in order. Since the first implementation is in Go, you may want to take a look, but you agree the approvers seems ready?

A68-random-subsetting.md
```diff
@@ -0,0 +1,148 @@
...
+  * Sort all endpoints by hash.
+  * Pick first `subset_size` values from the list.
+* Pass the resulting subset to the child LB policy.
+* If the number of endpoints is less than `subset_size` always use all available endpoints.
+
+Here is the implementation of this algorithm in pseudocode.
+
+```
+func filter_endpoints(endpoints, subset_size, seed)
+    endpoints_with_hash = []
```
dfawley, line 65: Maybe add `if subset_size >= len(endpoints) { return endpoints }` at the top?

A68-random-subsetting.md
```diff
@@ -0,0 +1,148 @@
...
+    // The value must be greater than 0.
+    uint32 subset_size = 1;
+
+    // The config for the child policy.
+    // The value is required.
+    config.cluster.v3.LoadBalancingPolicy child_policy = 2;
+}
+```
+
+As you can see, the fields in this policy match exactly the fields in the random subsetting LB service config.
```
dfawley, line 136: Can we somehow guarantee this is always the case?  If not, what happens if envoy adds a field here without telling anyone, e.g.?  Does gRPC only support the fields listed here, and we ignore anything else?

dfawley:
> @dfawley, Mark is the listed approver currently, but I'm fine merging this as it seems things are in order. Since the first implementation is in Go, you may want to take a look, but you agree the approvers seems ready?

This LGTM overall, but I'm a little concerned about this comment from Mark and its impact on this gRFC: https://github.com/grpc/proposal/pull/423/files#r1676227354

Have you had this discussion yet, @s-matyukevich?  Otherwise, should we wait to merge this until that's been finalized (more of a Q for @ejona86)?  @markdroth did approve, so maybe he didn't see it as a blocker.

A68-random-subsetting.md
s-matyukevich: added

A68-random-subsetting.md
s-matyukevich: Not sure about this. IMO this problem applies to any gRFC that defines matching xds and gRPC LB policies. @wbpcode is currently the [codeowner](https://github.com/envoyproxy/envoy/blob/main/CODEOWNERS#L317-L324) for all xds lb policies so I redirect this question to him.

markdroth:
I don't really anticipate any problems on the Envoy side, but it would be good to get a sanity check from @wbpcode before merging this, just to be safe.  I'll ping him offline to request his input.

A68-random-subsetting.md
markdroth: We can't guarantee that new fields won't be added to xDS (although as an xDS API shepherd, I'd see the change go by if it did), but that's fine -- it's just like any other feature getting implemented in one data plane before another.

The important thing is just that the fields we add to xDS won't get *removed*, and the backward compatibility requirements for xDS guarantee that.

So I think we're fine here.

A68-random-subsetting.md
wbpcode: It is not hard to implement that in envoy. So, it's fine.

wbpcode (approved):

markdroth:
Thanks for confirmation, @wbpcode!

I think that's all the open questions here, so I'll go ahead and merge.  Thanks!

PR #424, opened 2024-04-12
Rename all files for better sorting and consistency

dfawley:
Sorry for the disruption, but I feel this makes our repo _considerably_ easier to read.  Thanks!


ejona86 (commented):
I don't disagree this should have been considered originally, but this breaks every link to existing gRFCs which I would argue is the primary way they are consumed.

PR #426, opened 2024-04-17
L115: Refactor generic service and generic stub

yousukseung:


atollena:
The PR description should be L115 rather than A115.

yousukseung:
> The PR description should be L115 rather than A115.

Thanks, fixed.

ctiller (approved):

PR #427, opened 2024-04-18
A36 update: edit for clarity, fix some inconsistencies

murgatroid99:
From my perspective working on implementing this design, there are three major parts:

 - How the server interacts with the xDS API.
 - How the server behaves as a result of the configurations from xDS.
 - The changes to xDS resource validation rules.

I rearranged some of the information to structure the design that way. I also made a few other changes that improve clarity, in my opinion:

 - Removed a couple of statements that contradict other gRFCs
 - Decreased focus on some potential use cases that have not yet materialized
 - Added a couple of callouts using GitHub markdown's recent [alerts](https://github.com/orgs/community/discussions/16925) feature.
 - Added more subheadings to break up longer sections.

### DIFF
--- a/A36-xds-for-servers.md
+++ b/A36-xds-for-servers.md
@@ -54,8 +54,8 @@ Each language will create an "XdsServer" API that will wrap the normal "Server"
 API. This will allow the implementation to see the listening ports, start event,
 stop event, and also allow it access to the server to change behavior (e.g., via
 an interceptor). Users will use the XdsServer API to construct the server
-instead of the existing API. XdsServer will not support ephemeral ports as part
-of this gRFC but may be enhanced in the future.
+instead of the existing API. XdsServer will not support ephemeral ports (binding
+port 0) as part of this gRFC but may be enhanced in the future.
 
 Credential configuration will be managed separately by providing
 XdsServerCredentials, similar to client-side. The user must pass the
@@ -127,6 +127,8 @@ of issues preventing startup progress.
 
 ### xDS Protocol
 
+#### Bootstrap Changes
+
 The `GRPC_XDS_BOOTSTRAP` file will be enhanced to have a new field:
 ```
 {
@@ -139,9 +141,9 @@ The `GRPC_XDS_BOOTSTRAP` file will be enhanced to have a new field:
 }
 ```
 
+#### Listener Resource Handling
+
 XdsServer will use the normal `XdsClient` to communicate with the xDS server.
-xDS v3 support is required. xDS v2 support is optional and may have lower
-quality standards when adhering to the specifics of v2 when they differ from v3.
 There is no default value for `server_listener_resource_name_template` so if it
 is not present in the bootstrap then server creation or start will fail or the
 XdsServer will become "not serving". XdsServer will perform the `%s` replacement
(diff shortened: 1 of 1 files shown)


A36-xds-for-servers.md
```diff
@@ -237,22 +185,79 @@ prevents replacing the Listener when nothing changes. The grace period should be
...
+
+ 1. If the state of the server is "not serving", close the connection.
+ 2. Find the most-specifically-matching `FilterChain` for the connection, as
+    described in [the FilterChainMatch section](#filterchainmatch). If nothing
+    matches, use `Listener.default_filter_chain` instead, or if there is no
+    `default_filter_chain`, close the connection.
+ 3. If TLS is supported (see [gRFC A29: xDS TLS Security][A29]), apply the
+    corresponding TLS configuration from the `FilterChain` to the connection.
+
+When the server receives a request on an active connection, it should do the
```
markdroth, line 234: The first two bullets below say "If RouteConfiguration is supported", and the last bullet says "If any server-side HTTP filters are supported".  However, I think those basically all describe the same case, because if HTTP filters are supported, then RouteConfiguration also needs to be supported.

Given that, I suggest checking this sentence to start with "If the implementation supports HTTP filters", and then removing the conditionals from the individual bullets below.

A36-xds-for-servers.md
```diff
@@ -237,22 +185,79 @@ prevents replacing the Listener when nothing changes. The grace period should be
...
+following (implementations may reorder these operations):
+
+ 1. If `RouteConfiguration` is supported and if `HttpConnectionManager.rds`
+    references a NACKed resource without a previous good version, an
+    unavailable resource because of communication failures with control plane
+    or a triggered loading timeout, or a non-existent resource, fail the RPC
+    with the UNAVAILABLE status.
```
markdroth, line 241: Do we want to say something about what information should be in the status message here?  IIRC, we discussed not making this too specific for security reasons.

Same thing for the next bullet.

A36-xds-for-servers.md
```diff
@@ -334,6 +339,44 @@ connection, it must still be checked when verifying that the most-specific
...
+   - The `filter_chain_match` field must be normalized as described in the
+     [FilterChainMatch section](#filterchainmatch) and there must not be
+     duplication between any normalized matchers in the `Listener`.
+ - For the `default_filter_chain` field and every entry in the `filter_chains`
+   field:
+   - The `filters` field must have exactly one entry. Inside of it:
+     - The `type_url` field must have the value
+       `envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager`
+     - Inside the `value` field, decoded as a `HttpConnectionManager` message:
+       - The `http_filters` field must be validated as specified in
```
markdroth, line 358: As per my comment elsewhere, I think that these two bullets have the same conditions: if the implementation supports HTTP filters, then it must do both; if it does not, then it can do neither.  So I suggest saying something like this:

- The `value` field must contain a valid `HttpConnectionManager` message.  If the implementation supports HTTP filters, then the following must be validated:
  - The `http_filters` field must be validated as per A39.
  - Either the `route_config` field or the `rds` field must be set:
    - If `route_config` is set, it must be validated according to the `RouteConfiguration` validation rules below.
    - If `rds` is set, an RDS watch must be started for the specified resource.

A36-xds-for-servers.md
```diff
@@ -334,6 +339,44 @@ connection, it must still be checked when verifying that the most-specific
...
+> `udpa.type.v1.TypedStruct`, then its [`type_url`][TypedStruct.type_url] is used
+> instead. The `HttpConnectionManager` filter must always be present in the
+> last entry in each `filters` list.
+
+The `RouteConfiguration` proto validation rules (predomenently defined in [gRFC A28][A28-validation]) are modified as follows:
+
+ - Every entry of `virtual_hosts` is validated, not just the most specifically matching one.
+ - For each entry of `virtual_hosts`:
+   - For each entry of `routes`:
+     - The `action` field can now have the value `non_forwarding_action` in addition to other accepted values.
```
markdroth, line 378: I think we should explicitly call out that this means that a parsed route can now have multiple types of actions: RouteAction (usable for clients), NonForwardingAction (usable for servers), or unknown (an action type not currently supported by gRPC).  On clients, any action type other than RouteAction will cause RPCs to fail with UNAVAILABLE; on servers, any action type other than NonForwardingAction will cause RPCs to fail with UNAVAILABLE.  And as per my comment above, we should also say something about what information should be included in the RPC failure status message.

markdroth (commented):
Thanks for doing this!  Overall, this looks like a good improvement to the spec; my comments are mostly minor.

Please also update the "last modified date" near the top of the file.

A36-xds-for-servers.md
murgatroid99: I added something.

A36-xds-for-servers.md
```diff
@@ -334,6 +339,44 @@ connection, it must still be checked when verifying that the most-specific
...
+       - If any `RouteConfiguration` features are supported:
+         - Either the `route_config` field or the `rds` field must be set
+         - If the `route_config` field is set, its value must be validated
+           according to the `RouteConfiguration` validation rules.
+
+> [!NOTE]
+> Once other `Fitler` types are supported, the validation logic will need to
+> be expanded to accept other `Filter` types. `Filter` types are the type contained in the
+> [`typed_config`][Filter.typed_config] Any. If the type is
+> `udpa.type.v1.TypedStruct`, then its [`type_url`][TypedStruct.type_url] is used
```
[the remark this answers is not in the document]
murgatroid99: done.

PR #428, opened 2024-04-22
Create A80-grpc-metrics-for-tcp-connection

nanahpang:


yousukseung (commented):
LGTM to me as discussed offline, I'll leave others to approve.

A80-grpc-metrics-for-tcp-connection
```diff
@@ -0,0 +1,62 @@
...
+- `grpc.tcp.delivery_rate`
+- `grpc.tcp.packets_sent`
+- `grpc.tcp.packets_retransmitted`
+- `grpc.tcp.packets_spurious_retransmitted`
+
+The metrics will have label:
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.tcp.remote_peer_address | optional | Store the peer address info in the format as `ip:port`. |
```
ctiller, line 40: @markdroth @yashykt I feel like there's an equivalent label already in use?

A80-grpc-metrics-for-tcp-connection
```diff
@@ -0,0 +1,62 @@
...
+
+The metrics will be exported as:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.tcp.min_rtt | Distribution | s | grpc.tcp.remote_peer_string | Reports TCP's current estimate of minimum round trip time (RTT), typically used as an indication of the network health between two endpoints. |
```
ctiller, line 46: @yashykt how well does seconds here align with other metrics we have?

A80-grpc-metrics-for-tcp-connection
```diff
@@ -0,0 +1,62 @@
...
+| grpc.tcp.delivery_rate | Distribution | bit/s | grpc.tcp.remote_peer_string | Records the most recent non-app-limited throughput at the time that Fathom samples the connection statistics. |
+| grpc.tcp.packets_sent | Counter | {packet} | grpc.tcp.remote_peer_string | Records total packets TCP sends in the calculation period. |
+| grpc.tcp.packets_retransmitted | Counter | {packet} | grpc.tcp.remote_peer_string | Records total packets lost in the calculation period, including lost or spuriously retransmitted packets. |
+| grpc.tcp.packets_spurious_retransmitted | Counter | {packet} | grpc.tcp.remote_peer_string | Records total packets spuriously retransmitted packets in the calculation period. |
```
ctiller, line 50: I'm not sure what spurious means in this context... maybe we could link to some write-up, or include some more text defining these things?

A80-grpc-metrics-for-tcp-connection
ctiller: This could be a relatively expensive fan-out for a busy server

A80-grpc-metrics-for-tcp-connection
yousukseung: I could not find an authoritative source we can link, but I think we can include what's on go/fathom-metrics i.e. "These are retransmissions that TCP later discovered unnecessary.". The definition is public.

A80-grpc-metrics-for-tcp-connection
yousukseung: Agreed, we discussed about this offline. Is there a way to support an opaque user-defined peer info string? In prod, by default we only record the cell information due to the fan-out concern.

A80-grpc-metrics-for-tcp-connection
nanahpang: Sounds good, I will add that in the description.

A80-grpc-metrics-for-tcp-connection
atollena, line 46: A couple of remarks on this:
1. Do you know how portable this is going to be? It would help to explain how you plan to collect that data. Do you plan to use `getsockopt` + `TCP_INFO` that yields a min_rtt in some cases? Even on linux, I'm not sure all TCP algorithms measure the minimum RTT.
2. What is a Distribution type? Do you mean a histogram?
3. Since this is a property of the transport, this probably belongs under `grpc.transport.tcp`.

A80-grpc-metrics-for-tcp-connection
```diff
@@ -0,0 +1,62 @@
...
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.tcp.remote_peer_address | optional | Store the peer address info in the format as `ip:port`. |
+
+The metrics will be exported as:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.tcp.min_rtt | Distribution | s | grpc.tcp.remote_peer_string | Reports TCP's current estimate of minimum round trip time (RTT), typically used as an indication of the network health between two endpoints. |
+| grpc.tcp.delivery_rate | Distribution | bit/s | grpc.tcp.remote_peer_string | Records the most recent non-app-limited throughput at the time that Fathom samples the connection statistics. |
```
atollena, line 47: What is Fathom? A quick internet search yields https://dl.acm.org/doi/pdf/10.1145/3603269.3604815 which sounds related (but I have not read it)... Do you think anything that uses it is going to useful to anyone outside Google?

yousukseung: No, and app-limitedness is also not a general property either. I think we should update to something like "Latest throughput measured of the TCP connection."

A80-grpc-metrics-for-tcp-connection
yousukseung: I think min_rtt is general enough that we can include, I presume most congestion control tracks RTT or equivalent to keep track of the estimated BDP.

And I think this gRFC is supposed to cover the definition of the metric and not about the collection method. @yashykt  Yes, in Linux we can get it via either `getsockopt(TCP_INFO)` or `TCP_NLA_MIN_RTT` from `cmsg` when timestamping is enabled.

A80-grpc-metrics-for-tcp-connection
nanahpang: Sounds good, updated.

A80-grpc-metrics-for-tcp-connection
atollena: > I think min_rtt is general enough that we can include, I presume most congestion control tracks RTT or equivalent to keep track of the estimated BDP.

OK, and even if it weren't the case, there is probably no problem with omitting it.

> And I think this gRFC is supposed to cover the definition of the metric and not about the collection method.

OK, yes. My point is more that the definition needs to be precise enough so that different implementations measure the same thing.

A80-grpc-metrics-for-tcp-connection
```diff
@@ -0,0 +1,62 @@
+A80: gRPC Metrics for TCP connection
```
yashykt, line 1: Can you rename this file with a .md extension. I think that's why we aren't seeing the md formatting

A80-grpc-metrics-for-tcp-connection
yashykt: Not yet

A80-grpc-metrics-for-tcp-connection
```diff
@@ -0,0 +1,62 @@
...
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.tcp.remote_peer_address | optional | Store the peer address info in the format as `ip:port`. |
+
+The metrics will be exported as:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.tcp.min_rtt | Distribution | s | grpc.tcp.remote_peer_string | Records TCP's current estimate of minimum round trip time (RTT), typically used as an indication of the network health between two endpoints. |
```
yashykt, line 46: I think maybe we should add the remote_peer_string and maybe the local_peer_string as optional labels on these metrics

A80-grpc-metrics-for-tcp-connection
yashykt: We are following OpenTelemetry's conventions mostly which advocates for seconds 's' as the unit for duration

yashykt: I think if it's important for the correctness of the metrics, it makes sense to give more information on this metric. Maybe you could say that a valid implementation of this is through `getsockopt`?

yashykt (approved):

yashykt (changes requested):
(approved by mistake)

A80-grpc-metrics-for-tcp-connection
markdroth: I propose URI form, such as `ipv4:1.2.3.4:567`.  @ejona86 @dfawley, would that be a problem for Java and Go if we ever added these metrics in your languages?

A80-grpc-metrics-for-tcp-connection
markdroth: +1

A80-grpc-metrics-for-tcp-connection
```diff
@@ -0,0 +1,62 @@
...
+- `grpc.tcp.delivery_rate`
+- `grpc.tcp.packets_sent`
+- `grpc.tcp.packets_retransmitted`
+- `grpc.tcp.packets_spurious_retransmitted`
+
+The metrics will have label:
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.tcp.remote_peer_address | optional | Store the peer address info in the format as `ip:port`. |
```
markdroth, line 40: Let's call this `grpc.tcp.peer_address`.  And if we add a label for the local address, it should be called `grpc.tcp.local_address`.

A80-grpc-metrics-for-tcp-connection
markdroth, line 46: For consistency, please use the term "Histogram" instead of "Disrtibution".

A80-grpc-metrics-for-tcp-connection
```diff
@@ -0,0 +1,62 @@
...
+## Background
+
+To improve the network debugging capabilities for gRPC users, we propose adding per-connection TCP metrics in gRPC. The metrics will utilize the metrics framework outlined in  [A79].
+
+### Related Proposals: 
+* [A79]: gRPC Non-Per-Call Metrics Framework (pending)
+
+[A79]: https://github.com/grpc/proposal/pull/421
+
+## Proposal
```
markdroth, line 23: Please add an "Implementation" section as per the gRFC template that says that we plan to implement this in C-core but currently have no plans to implement in other languages.

nanahpang: Sure, done.

A80-grpc-metrics-for-tcp-connection
nanahpang: Sure, done.

A80-grpc-metrics-for-tcp-connection
nanahpang: Sure, done.

nanahpang: Sounds good, updated.

A80-grpc-metrics-for-tcp-connection
nanahpang: Sure, done.

A80-grpc-metrics-for-tcp-connection
nanahpang: I added a summary of how the metrics are collected and linked some references. Let me know if anything is unclear.

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -0,0 +1,74 @@
...
+## Abstract
+
+This document proposes adding new TCP connection metrics to gRPC for improved network analysis and debugging.
+
+## Background
+
+To improve the network debugging capabilities for gRPC users, we propose adding per-connection TCP metrics in gRPC. The metrics will utilize the metrics framework outlined in  [A79].
+
+### Related Proposals: 
+* [A79]: gRPC Non-Per-Call Metrics Framework (pending)
```
markdroth, line 19: This gRFC has been merged, so please remove the "(pending)" and change the lin to point to A79-non-per-call-metrics-architecture.md.

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -0,0 +1,74 @@
...
+| grpc.tcp.peer_address | optional | Store the peer address info in URI format such as `ipv4:1.2.3.4:567`. |
+| grpc.tcp.local_address | optional | Store the local address info in URI format such as `ipv4:1.2.3.4:567`. |
+
+The metrics will be exported as:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.tcp.min_rtt | Histogram (double) | s | grpc.tcp.peer_address, grpc.tcp.local_address | Records TCP's current estimate of minimum round trip time (RTT), typically used as an indication of the network health between two endpoints.  |
+| grpc.tcp.delivery_rate | Histogram (double) | bit/s | grpc.tcp.peer_address, grpc.tcp.local_address | Records latest throughput measured of the TCP connection. |
+| grpc.tcp.packets_sent | Counter (int64) | {packet} | grpc.tcp.peer_address, grpc.tcp.local_address | Records total packets TCP sends in the calculation period. |
```
markdroth, line 49: @yashykt What types should we be defining here?

markdroth (commented):
Need review from @ejona86 and @dfawley to make sure these definitions are still okay in terms of potential future implementation in other languages.

A80-grpc-metrics-for-tcp-connection.md
nanahpang: Sure, updated.

A80-grpc-metrics-for-tcp-connection.md
nanahpang: I updated the Counter type to uint64 based on the register methods in metrics.h. 
@yashykt Feel free to modify if it will use other types.

yashykt (approved):

A80-grpc-metrics-for-tcp-connection.md
markdroth: Still need to update the link to point directly to the merged file rather than the PR.

nanahpang: Thanks for catching this, updated.

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -18,7 +18,7 @@ To improve the network debugging capabilities for gRPC users, we propose adding
 ### Related Proposals: 
 * [A79]: gRPC Non-Per-Call Metrics Framework
 
-[A79]: https://github.com/grpc/proposal/pull/421
+[A79]: https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md
```
markdroth, line 21: This can just link to "A79-non-per-call-metrics-architecture.md".  It doesn't need the full URI.

nanahpang: Done, thanks for the suggestion.

yousukseung (approved):

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -0,0 +1,73 @@
...
+### Related Proposals: 
+* [A79]: gRPC Non-Per-Call Metrics Framework
+
+[A79]: A79-non-per-call-metrics-architecture.md
+
+## Proposal
+
+This document proposes changes to the following gRPC components.
+
+#### Per-Connection TCP Metrics
```
atollena, line 27: nit: this can be `###` instead of `####`.

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -0,0 +1,73 @@
...
+- `grpc.tcp.delivery_rate`
+- `grpc.tcp.packets_sent`
+- `grpc.tcp.packets_retransmitted`
+- `grpc.tcp.packets_spurious_retransmitted`
+
+The metrics will be exported as:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.tcp.min_rtt | Histogram (double) | s | None | Records TCP's current estimate of minimum round trip time (RTT), typically used as an indication of the network health between two endpoints.<br /> RTT = packet acked timestamp - packet sent timestamp. |
```
atollena, line 40: Buckets for histogram are explicitly specified for request latency in https://github.com/grpc/proposal/blob/master/A66-otel-stats.md. Do you plan to reuse the same buckets? It might be worth specifying.

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+#### Metric Collection Design
+
+A high-level approach to collecting TCP metrics (on Linux) is as follows:
+1) **Enable Network Timestamps for Metric Calculation:** Enable the `SO_TIMESTAMPING` option in the kernel's TCP stack through the `setsocketopt(fd, SOL_SOCKET, SO_TIMESTAMPING, &val, sizeof(val))` system call. This enables the kernel to capture packet timestamps during transmission.
+2) **Calculate Metrics from Timestamps:**  Linux kernel calculates TCP connection metrics based on the captured packet timestamps. These metrics can be retrieved using the `getsockopt(TCP_INFO)` system call. For example, the delivery_rate metric estimates the goodput—the rate of useful data transmitted—for the most recent group of outbound data packets within a single flow ([code](https://elixir.bootlin.com/linux/v5.11.1/source/net/ipv4/tcp.c#L391)).
+3) **Periodically Collect Statistics:** At a specified time interval (e.g., every 5 minutes), gRPC aggregates the calculated metrics and updates the corresponding statistics records.
```
atollena, line 52: That could be a bit more precise. I have 3 questions:

1. Do you plan to have this interval be specified by users somehow? Or should it be fixed to 5 minutes? 5 minutes seems long to me for updating metrics, ideally it should be something close to the collection period (which iiuc is more typically in the 1 minute).
2. How should the aggregation logic look like? I imagine you have something like this in mind:
- for counters, iterate over sockets and sum up packets_sent, packets_restransmitted of all open sockets.
- for histograms, for each socket, record the value (delivery_rate & min_rtt) in the histogram for the corresponding metric.

That's the only thing that really makes sense given the definition of the metrics above, so perhaps it's fine to leave it implicit.

3. What happens to sockets that have been closed in between the interval? Do we just not collect statistics for those and loose the data? I'm not sure that there is an alternative that works, such as calling `getsockopts` and updating statistics just before closing the socket.

nanahpang: I think in C++ the interval is not fixed, and the default value is 5 minutes in Fathom. The goal of this high-level plan aims to offer a general understanding of the existed collection process in C++ (through Fathom), while granting flexibility to other languages. The Fathom documentation likely contains detailed answers to your questions. To maintain the clarity and focus of this proposal, I'll keep it concise and high-level. Please feel free to refer to the Fathom documentation for further details on implementation specifics. Thanks.

@yousukseung, please feel free to chime in with any additional details or clarification on Fathom implementation questions. Thanks.

A80-grpc-metrics-for-tcp-connection.md
yashykt: @nanahpang If you do not specify buckets, we will get the default OpenTelemetry buckets which is { 0, 5, 10, 25, 50, 75,100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000}. Are there better defaults that we can suggest or do we leave it to the user to figure it out?

yashykt (approved):

A80-grpc-metrics-for-tcp-connection.md
yashykt: `int64` and `double` are very C/C++ specific. I prefer `integer` and `floating-point` to keep it generic.

A80-grpc-metrics-for-tcp-connection.md
nanahpang: Regarding OpenTelemetry buckets, if you're referring to a plugin used for exporting metrics, that would be a separate concern from the metric types we're defining here.  However, I agree that specifying a smaller range for the min_rtt metric buckets is the right approach, especially since we're using seconds as the unit (to align with the open-source standard) while the underlying measurements are usually in microseconds.

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -0,0 +1,73 @@
...
```
ejona86, line 51: "At a specified time interval" means the interval will be configurable. How is it configured?

A80-grpc-metrics-for-tcp-connection.md
nanahpang: Done, thanks for the suggestion.

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -0,0 +1,73 @@
...
+- `grpc.tcp.packets_retransmitted`
+- `grpc.tcp.packets_spurious_retransmitted`
+
+The metrics will be exported as:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.tcp.min_rtt | Histogram (double) | s | None | Records TCP's current estimate of minimum round trip time (RTT), typically used as an indication of the network health between two endpoints.<br /> RTT = packet acked timestamp - packet sent timestamp. |
+| grpc.tcp.delivery_rate | Histogram (double) | bit/s | None | Records latest goodput measured of the TCP connection. <br /> Elapsed time = packet acked timestamp - last packet acked timestamp. <br /> Delivery rate = packet acked bytes / elapsed time. |
+| grpc.tcp.packets_sent | Counter (uint64) | {packet} | None | Records total packets TCP sends in the calculation period. |
```
ejona86: What is a "calculation period"?

A80-grpc-metrics-for-tcp-connection.md
```diff
@@ -0,0 +1,73 @@
...
+| grpc.tcp.packets_spurious_retransmitted | Counter (uint64) | {packet} | None | Records total packets spuriously retransmitted packets in the calculation period. These are retransmissions that TCP later discovered unnecessary.|
+
+#### Metric Collection Design
+
+A high-level approach to collecting TCP metrics (on Linux) is as follows:
+1) **Enable Network Timestamps for Metric Calculation:** Enable the `SO_TIMESTAMPING` option in the kernel's TCP stack through the `setsocketopt(fd, SOL_SOCKET, SO_TIMESTAMPING, &val, sizeof(val))` system call. This enables the kernel to capture packet timestamps during transmission.
+2) **Calculate Metrics from Timestamps:**  Linux kernel calculates TCP connection metrics based on the captured packet timestamps. These metrics can be retrieved using the `getsockopt(TCP_INFO)` system call. For example, the delivery_rate metric estimates the goodput—the rate of useful data transmitted—for the most recent group of outbound data packets within a single flow ([code](https://elixir.bootlin.com/linux/v5.11.1/source/net/ipv4/tcp.c#L391)).
```
ejona86, line 50: I'd still like to know what the mapping is for each metric. The first few are easy because the names here mirror tcp_info (I assume). But then it gets less obvious, and tcp_info isn't documented very well.

A80-grpc-metrics-for-tcp-connection
dfawley: I think this should be clarified more.  For instance, is `ipv4:///1.2.3.4:567` also OK?  What is the format for `ipv6`?  What other schemes do we support?  @ejona86 is concerned that `ipv4:1.2.3.4:567` is not a URI we should support in the first place.  So maybe this should just be specified fully as strictly:

```
- ipv4:<ipv4-address>:<port-number>
OR
- ipv6:[<ipv6-address>]:<port-number>
```

yashykt: Outdated since we removed all these labels. We will probably need some optional labels in the future. It's not clear what at the moment, we'll need to revisit.

A80-grpc-metrics-for-tcp-connection.md
yashykt: Added

A80-grpc-metrics-for-tcp-connection.md
yashykt: outdated

A80-grpc-metrics-for-tcp-connection.md
yashykt: Added details on the suggested algorithm

A80-grpc-metrics-for-tcp-connection.md
yashykt: I've added more language regarding the suggested algorithm to collect this data. 

> How should the aggregation logic look like? I imagine you have something like this in mind:
for counters, iterate over sockets and sum up packets_sent, packets_restransmitted of all open sockets.
for histograms, for each socket, record the value (delivery_rate & min_rtt) in the histogram for the corresponding metric.

The aggregation is handled by OpenTelemetry.

> What happens to sockets that have been closed in between the interval? Do we just not collect statistics for those and loose the data? I'm not sure that there is an alternative that works, such as calling getsockopts and updating statistics just before closing the socket.

That is exactly what we want :) Added it in the language

A80-grpc-metrics-for-tcp-connection
yashykt: Added a big write-up

ctiller (approved):

PR #430, opened 2024-04-26
PID LB policy

s-matyukevich:
Follow up on https://github.com/grpc/proposal/pull/383 and https://github.com/grpc/proposal/pull/423

This proposal was implemented and tested with a few real apps in production environment, here are the results for one of the apps:
Screenshot 2024-04-02 at 3 44 04 PM


A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+  totalUtilization = 0
+  count = len(data.utilizationPerSubchannel)
+  for _, utilization in data.utilizationPerSubchannel {
+    totalUtilization += utilization
+  }
+  data.meanUtilization = totalUtilization / count
+}
+```
+
+The proposal is to make `wrrCallbacks` public. This has a number of significant benefits. Besides PID, there are other cases where one might need to extend `wrr`. For example, Spotify [demonstrates](https://www.youtube.com/watch?v=8E5zVdEfwi0) a gRPC load balancer to reduce cross-zone traffic – this can be implemented nicely in terms of `wrr` weights. We are also considering the same and incorporating things like latency into our load balancing decisions. Existing ORCA extension points don't cover these use cases. We leverage ORCA for custom server utilization metrics, but we also need the ability to combine server and client metrics to generate the resulting weight. The alternative is to write our own balancer with custom EDF scheduler and handle details related to subchannel management and interactions with resolvers. With this new API, use cases like this can be covered naturally, users have full control over the end-to-end definition of weights.
```
s-matyukevich, line 230: cc @yannickepstein would something like the proposed wrrCallbacks interface  work for your use-case, so you don't have to copy-paste the whole wrr?

A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+message LoadBalancingConfig {
+  oneof policy {
+    PIDLbConfig pid = 20 [json_name = "pid"];
+  }
+}
+
+message PIDLbConfig {
+  // Configuration for the WRR load balancer as defined in [gRFC A58][A58].
+  // The PID balancer is an extension of WRR and all settings applicable to WRR also apply to PID identically.
+  WeightedRoundRobinLbConfig wrr_config = 1;
```
markdroth, line 48: I don't think we want to include the WRR config here.  Instead, let's just duplicate the fields from that proto in this one, so that the two are independent.

A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+
+  // Threshold beyond which the balancer starts considering the ErrorUtilizationPenalty.
+  // This helps avoid oscillations in cases where the server experiences a very high and spiky error rate.
+  // We avoid eliminating the error_utilization_penalty entirely to prevent redirecting all traffic to an instance
```
markdroth, line 52: If this is the only thing that you need the error utilization penalty for, an alternative would be to set the penalty to zero and instead use outlier detection (see [gRFC A50](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md) to avoid sending traffic to such a backend.  Then we presumably wouldn't need this knob.

A80-pid.md
markdroth, line 230: I am hesitant to make this a public API.  This seems like a valuable internal thing to allow us to reuse LB policy implementations, but making it public implies a stability commitment that I'm not super comfortable with.

It's worth noting that the example of using latency would require that this API surface be even broader, because we'd need a way for this to trigger the LB policy to measure the latency on each RPC.  (We do have that mechanism internally, but we'd need to expose it via this API in addition to the existing mechanism as part of the LB policy API itself.)

I'd like to hear thoughts from @ejona86 and @dfawley on this.

A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+
+By understanding and addressing these factors, the `pid` balancer can be more effectively tuned to manage load balancing across different environments and usage scenarios, minimizing the risks associated with oscillations.
+
+### Moving Average Window for Load Reporting
+
+As outlined in the previous section, smoothing the utilization measurements in server load reports is essential for the `pid` balancer to achieve convergence on spiky workloads. To address this, we propose integrating a moving average window mechanism into the `MetricRecorder` component, as described in [gRFC A51][A51]. This involves adding a `MovingAverageWindowSize` parameter to the component. Instead of storing a single value per metric, `MetricRecorder` will now maintain the last MovingAverageWindowSize reported values in a circular buffer. The process is detailed in the following pseudo-code:
```
markdroth, line 250: It's not clear to me that this is something we need to build directly into gRPC APIs.  Can't the application do this smoothing itself before it reports the data to the gRPC `MetricRecorder`?

markdroth (commented):
This looks very interesting -- and the graphs definitely show impressive results!

My main concern here is the comment about making the WRR policy extensible as a public API.  I think that needs some careful thought.

I'd like @ejona86 and @dfawley to review as well.

A80-pid.md
s-matyukevich: Yes, that's what we are doing now. However, this is not a trivial amount of code and PID almost certainly requires it. If we provide the balancer but don't provide the smoothing implementation the UX won't be ideal, as users most likely will start using it without any smoothing and soon come to the conclusion that it doesn't work. 
My plan was to document that smoothing is required for PID and then mention "Use property X of the MetricsResorder to configure smoothing" IMO this is a lot better than say "implement smoothing yourself".

A80-pid.md
s-matyukevich: Good point, we can use outlier detection with zero utilization penalty for servers with very high and spiky error rates. I can remove this knob.

A80-pid.md
markdroth: If this is really critical to the behavior of the PID policy, wouldn't it be better to build this directly into the PID policy, so that it's not possible to have a misconfiguration between the client and the server?  In other words, why not do this smoothing on the client side?

A80-pid.md
s-matyukevich: I expected this answer but decided to give it a try anyway. Making this interface private works for us as well - next time we need to extend PID of WRR we'll come back with more gRFCs. I assume that it is much easier to make an interface public than otherwise, so maybe we can revisit this decision in the future when we have more data and better defined use-cases where such interface might be useful. 

For now I can remove this paragraph and replace it with a note that the interface will be private. @markdroth does this work for you?

A80-pid.md
s-matyukevich: We considered this but decided to implement it on the server for the following reasons:
* How much we should smooth is mostly a server property (it depends on how spiky is the workload, which in turn depends on how expensive are individual requests and how requests are distributed in time) It easier to reason about this from the perspective of service owner rather than the client. 
* Because of the first point, it doesn't really make sense to have different clients with different smoothing parameters.
* It protects the server from misconfigured clients: if some clients don't use enough smoothing this could easily result in oscillations, which could hurt the server.
* Most importantly smoothing on the server is more accurate as it takes into account actual load and is not delayed in time. If we do smoothing on the client the result might be less accurate as the client only have data sampled at random points.  If there was a big CPU spike between 2 consecutive requests sent by a particular client this information will be missed  by client-side smoothing, so different clients may have different resulting view of the server load. If we do smoothing on the server we don't have such problems as we can sample CPU as frequently as a few milliseconds and use monotonically increasing cgrpup CPU counter, so we never miss any CPU spikes or drops.

ejona86 (commented):
Mark, Doug, Craig, and I discussed this on Friday. These are essentially our notes from the meeting with some minor extra review items from me. I tried to say "we" for the meeting and "I" for the other things. But I'm also going to follow this with a review just from myself.

A80-pid.md
s-matyukevich: Trying to brainstorm new ideas here: what do you think if we change a bit the structure of load reports and make them additionally and optionally report the timestamp at the time of measurement and the CPU time as monotonically increased counter. Basically we take the same interface as kernel provide us via [getrusage](https://man7.org/linux/man-pages/man2/getrusage.2.html) syscall and make the server report the raw values reported by this syscall and don't do any aggregations. Now aggregation can be done on the client and we can use any type of smoothing we want and still get perfect accuracy. PID balancer should try to use this new load report data and if it not present it fallbacks to using the default CPU utilization. 

I am not 100% sure what would be the best way to handle AppUtilization in this case, but we have a couple of options:
* Make the new load report field more generic (e.g. call it `UtilizationCounter`) make it take precedence over both CpuUtilization and AppUtilization.
* Make existing AppUtilization field take precedence over the new field. In this case we can also either apply or skip client-side smoothing for AppUtilization.

ejona86 (commented):
This one is the "personal" review that I promised. Consider it food for thought. Y'all have practical experience with the PID, and I'm not trying to override that as I'm more of a newbie. But I'm also trying to avoid gotchas. See what resonates with you.

(I've picked this up several times since I mentioned it, so there may be cases that are fragmented/have strange flow.)

ejona86:
Oh, and I know the biggest issue with some of my suggestions is they can partly/wholly invalidate the testing you've done. The most important part of this is to have confidence in the result.

A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+  google.protobuf.FloatValue proportional_gain = 2;
+
+  // Adjusts the smoothness of the PID controller convergence. Higher values enhance smoothness but can decelerate convergence.
+  // Default is 1.
+  google.protobuf.FloatValue derivative_gain = 4;
+
+  // Maximum allowable weight. Weights proposed by the PID controller exceeding this value will be capped.
+  // This prevents infinite weight growth, which could occur if only a subset of clients uses PID and increasing weights
+  // no longer effectively corrects the imbalance. Default is 10.
+  google.protobuf.FloatValue max_weight = 5;
```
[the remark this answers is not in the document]
s-matyukevich: This is a very good point,  I verified that the drift actually happens in practice. It is not causing issues and we didn't notice this before, since weights are re-centered around 1 after every client redeployment and, as you said, drift stops once a significant portion of the clients reach the limit. 

Anyway, I'll update gRFC to include your proposal.

A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+    controlSignal = this.proportionalGain * this.controlError +
+                    this.derivativeGain * controlErrorDerivative
+    return controlSignal
+  }
+}
+```
+
+The `update` method is expected to be called on a regular basis, with `samplingInterval` being the duration since the last update. The return value is the control signal which, if applied to the system, should minimize the control error. In the next section, we'll discuss how this control signal is converted to `wrr` weight.
+
+The `proportionalGain` and `derivativeGain` parameters are taken from the LB config. `proportionalGain` should be additionally scaled by the `WeightUpdatePeriod` value. This is necessary because derivative error is calculated like `controlErrorDerivative = (this.controlError - previousError) / samplingInterval.Seconds()` and dividing by a very small `samplingInterval` value makes the result too big. `WeightUpdatePeriod` is roughly equal to `samplingInterval` as we will be updating the PID state once per `WeightUpdatePeriod`.
```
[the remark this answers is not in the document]
s-matyukevich: The proposal was to scale only proportionalGain, not derivativeGain. Also I'll update the doc to explicitly mention multiplication instead of scaling.

A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+  // but the PID controller keeps growing the weights and it may easily pass the balancing point.
+  if time.Since(lastApplied) < conf.WeightUpdatePeriod {
+    return -1
+  }
+
+  // use value calculated in the onEDFSchedulerUpdate method
+  meanUtilization = data.meanUtilization
+
+  // call the PID controller to get the value of the control signal.
+  controlSignal = data.pidController.update({
```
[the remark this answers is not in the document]
s-matyukevich: I like this idea, but I think it also depends on the decision we make about my proposal in this thread https://github.com/grpc/proposal/pull/430/files#r1761294507 If we update load reports to give us monotonically increasing CPU counter this should solve the problem with low QPS clients. 

Also I am fine with disabling/discouraging OOB. That's what we do as well in our infrastructure. Since we pair this solution with heavy subsetting it also helps us to ensure every client has fresh data about server utilization, even if the client has low QPS.

A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+    referenceSignal:  meanUtilization,
+    actualSignal:     utilization,
+    samplingInterval: time.Since(lastApplied),
+  })
+
+  // Normalize the signal.
+  // If meanUtilization ~= 0 the signal will be ~= 0 as well, and convergence will become painfully slow.
```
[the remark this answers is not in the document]
s-matyukevich: Ok, this make sense, I'll update the doc to divide PID input signal by meanUtilization

A80-pid.md
```diff
@@ -0,0 +1,290 @@
...
+  if useErrPenalty {
+    utilization += errorRate * conf.ErrorUtilizationPenalty
+  }
+
+  // Ensure at least WeightUpdatePeriod has passed since the last update.
+  // Prevents corruption of PID controller's internal state, which could happen in the following cases:
+  // * If 2 updates are very close to each other in time, samplingInterval ~= 0 and signal ~= infinity.
+  // * If multiple updates happened during a single WeightUpdatePeriod, the actual weights are not applied,
+  // but the PID controller keeps growing the weights and it may easily pass the balancing point.
+  if time.Since(lastApplied) < conf.WeightUpdatePeriod {
```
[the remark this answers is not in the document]
s-matyukevich: In the implementation we have we indeed added a lock. I am leaning toward implementing your proposal and tracking only most recent utilization on the data plane, and I think it is the same discussion as https://github.com/grpc/proposal/pull/430/files#r1760128032

A80-pid.md
markdroth: I'm not sure that I'd want exactly this interface in C-core, even as a private API.  I think it would be better to avoid being proscriptive about how the implementation needs to be structured, so that each implementation can make its own choice.

That having been said, I do agree that it's useful to point out that there's so much overlap between this new policy and the existing WRR policy and to suggest that implementations consider how to avoid code duplication.  And we can even describe this specific approach as one possible way to do that.  Let's just leave the final decision up to the implementation.

FWIW, I'm going to be putting together a gRFC soon to allow WRR to configure which ORCA utilization metrics it looks at, to make it a bit more flexible.  I don't think that actually affects this design, though, because it still doesn't make the weight calculation function pluggable.

A80-pid.md
markdroth: Reporting just the monotically increasing CPU time seems fundamentally different from reporting CPU *utilization*.  CPU time is essentially a counter of the total amount of resources consumed, without taking into account any upper limit.  CPU utilization does take that limit into account: it's essentially CPU time divided by the upper limit to produce a ratio indicating what percentage of the limit is being used.  Using that ratio is particularly important if each endpoint has a different limit or has different types of CPUs that perform differently.

I don't mind the idea of sending timestamps in ORCA.  But I think one challenge of doing this smoothing on the client side is that we'd need to store the circular buffer for each endpoint in each client, which would significantly increase memory usage.  We'd probably want some knob to limit the amount of data we are keeping here, to provide some reasonable cap on memory usage.

s-matyukevich: >  CPU time is essentially a counter of the total amount of resources consumed, without taking into account any upper limit

The idea was to report normalized CPU time (total CPU time divided by number of available CPU cores)

> But I think one challenge of doing this smoothing on the client side is that we'd need to store the circular buffer for each endpoint in each client

Yes, but we don't need to store all endpoints: we can skip endpoints that a very close to each other in time. Let me demonstrate how this could work with an example:

```
struct LoadReport {
  Timestamp int // timestamp with millisecond resolution since the time when we start measurement
  CPUTime float // total amount of CPU time used by the process since the time when we start measurement 
                           // divided by number of available CPU cores, measured in milliseconds
}
```
Suppose the client received 5 load reports
```
r1 = LoadReport{Timestamp: 10, CPUTime: 7.5}
r2 = LoadReport{Timestamp: 15, CPUTime: 9.0}
r3 = LoadReport{Timestamp: 17, CPUTime: 9.3}
r4 = LoadReport{Timestamp: 23, CPUTime: 14.3}
r5 = LoadReport{Timestamp: 24, CPUTime: 15.1}
```
In order to calculate server avg utilization we just need any 2 load reports, for example, for r1 and r2 the math would look like:
```
avgUtilization = (r2.CPUTime-r1.CPUTime) / (r2.Timestamp - r1.Timestamp) // (9.0 - 7.5) / (15 - 10) = 0.3
```
And the result here is 100% accurate.

When we store reports in circular buffer we can keep only reports that are farther than 4 ms from each other in time to save some space, so we will store only r1, r2 and r4.  

Now suppose that we received a new report r6 with Timestamp = 33, and also suppose that our desired smoothing interval is 20ms. At this point we delete r1 from our circular buffer (because r6.Timestamp - r1.Timestampt > 20) and use r6 and r2 to calculate the average utilization. So the window that we use to calculate the average will depend on timestamps that we have and it will never match exactly the desired value, but this is ok since we have perfectly accurate measurements otherwise. 

Using this approach we can find the balance between the accuracy of smoothing and the amount of memory we need to store past load reports.

markdroth: > The idea was to report normalized CPU time (total CPU time divided by number of available CPU cores)

Isn't this the same as CPU utilization -- i.e., it's the same value already encoded in the ORCA `cpu_utilization` field?

> Yes, but we don't need to store all endpoints: we can skip endpoints that a very close to each other in time.

That does sound like a good idea to avoid storing unnecessary data points.  But the total number of data points still has no strict upper bound here, since it will depend entirely on how close together we keep data points (the 4ms in your example) and how much total history we keep for each endpoint (e.g., "keep 10s worth of history for each endpoint").  So we'd need to figure out what knobs to provide to control this.

I'm thinking that we'd probably need two knobs here: the max number of data points to keep for each endpoint, and the amount of history to keep for each endpoint.  From those two, we can presumably dynamically determine how close together we want to keep data points -- e.g., if we are keeping 10s worth of history and have a max of 5 data points, then we don't keep data points closer than 2s apart.

ejona86: Each weight update period we save the most recent utilization. And we keep X amount of time in the circular buffer, and X already had to be specified by the user.

s-matyukevich: > Isn't this the same as CPU utilization -- i.e., it's the same value already encoded in the ORCA cpu_utilization field?

If we use the current ORCA cpu_utilization field we can miss some server CPU spikes if server doesn't receive any requests from a particular client during the spike, so the client can get wrong avg CPU. Or to put it differently, the downsides of the current approach are the fact that avg utilization calculated by the client depend on RPS and on the way server does CPU stat aggregation to calculate cpu_utilization field (As @ejona86 pointed out in another [comment](https://github.com/grpc/proposal/pull/430/files#r1759567463), server must do some aggregation, as otherwise there is no instantaneous value for CPU utilization (other than a weak boolean (running/not running) per core)) Those 2 things makes it very difficult to reason about CPU smoothing on the client and the proposed approach should help with both.

ejona86: I had purposefully not weighed in on the "communicate CPU time" approach too much, so I could hear from others. I think such a design is relatively straight-forward. My only concern is, "is it worth it?"

I think if we were starting over with ORCA it could make some sense, but I don't think the system has been shown to be bad enough to warrant the infrastructure for it and the limitations. (It was shown to be bad, but there's options.) The approach is quite different than the rest of ORCA, which feels like we'd fragment the ORCA "ecosystem."

I'm also concerned it'd be limited to cpu utilization. You could record memory as MB-seconds (and then normalize), or a queue as entry-seconds (and then normalize), but I suspect ordinary developers would think such things bizarre and obtuse and use that as a reason to: 1) stick with just cpu or 2) use plain utilization.

I think the strongest argument for communicating CPU time is that it is "one less knob," especially on the server. (Although the server probably would still have a knob for the existing ORCA utilization fields.) The approach affords lots of options. We could use it _just_ between window update periods and then do exponential moving average.

But simply averaging utilization over each weight update period would help a lot and is exactly how LRS works. And we can use an exponential moving average across update periods (or a circular buffer, if you were so inclined). Those are trivial to implement and would greatly reduce the noise of the utilization vs what was initially tested (and shown to be lacking).

(One interesting part of communicating cpu time is you need at least two RPCs to measure utilization. The `blackout_period` means this wouldn't be noticed during start-up. But it also means you always get "the average since the last measurement." If you are deep into the `weight_expiration_period` (3 minute default) and then do an RPC, you'll be seeing average utilization for the past minutes as opposed to the recent utilization. I don't see that as a problem; it is a edge case and by definition we were fine using up to a 3 minute old weight. Realize though that the old utilization would significantly expand the time period of the circular buffer for a while.)

markdroth: I still don't fully understand what's being proposed here in terms of what data will actually be encoded in the ORCA load report.  On one hand, you're saying that you want to report cumulative CPU time so that we don't miss any spikes, which means that you'd be reporting counter data.  But on the other hand, you're also saying that we would still divide the reported CPU time by the amount of available CPU, thus reporting a gauge instead of a counter.  I don't understand how we can do both in a single number.

Hmm... unless the idea here is that for each CPU time value we get, we just divide it by the amount of available CPU before we accumulate it?  I guess that would basically still be a counter; it's just that the values we're accumulating are utilization values instead of CPU time.

How do we handle cases where the amount of available CPU time changes?  If every data point is computed by dividing the total CPU time used since the start of the process by the *current* amount of available CPU, then we're going to retroactively change the interpretation of the CPU time that was accrued before that change.  How will that affect the math here?

Are there going to be any issues with precision loss when the accumulated CPU time gets very large?  With integers, we'd need to be prepared to handle wrap-around, but doubles, IIRC we'd start losing precision after a certain point.  Do we believe in all cases that that point is high enough that we won't ever run into it in practice?

I think it would really help me if you could provide a concrete example in the form of a timeline of events, showing what the server's CPU utilization is at each point and what values would be sent in the ORCA load reports at each point.

In general, I am not worried about fracturing the ORCA ecosystem, because if we're needing a different value here than ORCA's existing CPU utilization, then we just need to encode the value in a different field.  Yes, that's something that would need to be configured on the server, but the gRPC server today doesn't provide CPU utilization out of the box anyway.  No matter what we do, I think it's still a requirement that the server provides the data that the client's LB policy depends on.

s-matyukevich: > Hmm... unless the idea here is that for each CPU time value we get, we just divide it by the amount of available CPU before we accumulate it? I guess that would basically still be a counter; it's just that the values we're accumulating are utilization values instead of CPU time.

Yes! This is exactly the idea. 

> How do we handle cases where the amount of available CPU time changes? If every data point is computed by dividing the total CPU time used since the start of the process by the current amount of available CPU, then we're going to retroactively change the interpretation of the CPU time that was accrued before that change. How will that affect the math here?

This is a good point and I admit I didn't think about it before. I think we can tell service owners to not do any special processing here and simply divide the number of CPU time by the number of CPU cores that is **currently** available to the process. When a new CPU is added, the counter will drop significantly, and the client can easily detect this by comparing current utilization counter with previous one. In this case the client will keep using the last calculated avg utilization until the circular buffer is filled with values reported after the change happened. Similarly we can detect cases when num of available CPU decreases - in this case the counter increases sharply, which means that the diff between 2 consecutive utilizations will be large than the time difference between them. 

> Are there going to be any issues with precision loss when the accumulated CPU time gets very large? With integers, we'd need to be prepared to handle wrap-around, but doubles, IIRC we'd start losing precision after a certain point. Do we believe in all cases that that point is high enough that we won't ever run into it in practice?

I can do the exact math here later, but I am pretty certain that we can fit reasonably large interval into double protobuf type if we use millisecond precision. If not, we can always report value modulo some large hardcoded value (like number of milliseconds in 1 day) and handle overflow on the client. 

> I think it would really help me if you could provide a concrete example in the form of a timeline of events, showing what the server's CPU utilization is at each point and what values would be sent in the ORCA load reports at each point.

I was trying to do it in [this](https://github.com/grpc/proposal/pull/430#discussion_r1843216088) comment, In this example Timestamp and CPUTime fields will be taken directly from load report sent by the server. Let me know if this doesn't work and you want a more specific example.

markdroth: Thanks, @s-matyukevich, I think I understand this idea much better now.

I think this is doable, but I share the concerns that @ejona86 mentioned above in terms of whether this is worth the complexity.  I'd like to understand whether the results would be substantially better with this complexity vs. the alternatives that Eric suggested above.

I am also still concerned about bounding the memory usage in a reasonable way.  Eric said this:
> Each weight update period we save the most recent utilization. And we keep X amount of time in the circular buffer, and X already had to be specified by the user (for this sort of approach.)

I don't quite follow the idea here.  The data points are accumulated when we get each load report, so the number of data points is primarily dependent on the rate of data plane RPCs.  The weight update period controls how often we update the scheduler, so there can be an unbounded number of data points since the last time we updated the scheduler.

---

Stepping back a bit, it's also worth noting that one of the reasons for exploring configuring this on the client side was to avoid cases where the client and server configuration needed to be tightly coupled.  But if we are requiring the server to provide this kind of custom load report, then we haven't really solved that problem, and it might be better to just do the smoothing on the server side after all.  But I still don't think that gRPC should own the logic of generating this data, since this seems likely to be a sufficiently niche thing that most applications won't need it, and I don't want us to have to maintain code for every single metric anyone ever thinks up.  That's the whole reason why we provide APIs for the server application to report whatever metrics it wants -- generating the data isn't our problem.  (I do think that we should eventually provide code to generate data for very common metrics like CPU utilization, but that's only for very common cases, and I'm not sure the proposal here fits into that category.)  But you could still distribute a separate library to generate this data that server applications can use if they want to use the PID LB policy.

Taking all of the above into account, I guess my position here is that we could probably make this work if we really had to, but I'd like to be convinced that the complexity is really worthwhile.  I think this would be a lot easier to sell if we can make it simpler somehow.

ejona86: >> Each weight update period we save the most recent utilization. And we keep X amount of time in the circular buffer, and X already had to be specified by the user (for this sort of approach.)
>
> I don't quite follow the idea here. The data points are accumulated when we get each load report, so the number of data points is primarily dependent on the rate of data plane RPCs. The weight update period controls how often we update the scheduler, so there can be an unbounded number of data points since the last time we updated the scheduler.

Only the most recent load report in each update period would actually be used. That's the way it is already, and changing how the utilization is reported doesn't change that. Data plane keeps track of the most recent it sees, then each update period we grab the current value, add it to the circular buffer, and calculate the average utilization over the buffer's period.

The "average utilization" is `newest - oldest / time`. "newest" is clearly the same as the data plane's most recent. "oldest" _could_ be the _first_ report in a period, but that's essentially identical (and in several definitions, actually identical) to the _last_ report one period earlier.

> one of the reasons for exploring configuring this on the client side was to avoid cases where the client and server configuration needed to be tightly coupled. But if we are requiring the server to provide this kind of custom load report, then we haven't really solved that problem,

It is still decoupling the components. Right now the server needs a magic number for how long to average its utilization over. And that magic number will impact the magic numbers to use for the PID. Yes, you have to plumb things initially, but after that point we could have a single config that would could be updated atomically with all the magic numbers in agreement.

s-matyukevich: +1 to all what Eric said.

> I'd like to be convinced that the complexity is really worthwhile. I think this would be a lot easier to sell if we can make it simpler somehow.

So here is the use-case where perfectly accurate smoothing becomes very important: Suppose you have 100 servers, 10 000 clients, all clients are using random subsetting with subset size = 10 and every individual client generates very low QPS (1 req per 10 sec on average) The server is running a loop that measures its CPU usage every 1 sec, and average CPU utilization during the last second is reported via ORCA. This is actually a very common setup for servers that are not directly involved in end user request processing, but are doing some side things (like generating auth tokens, issuing certificates, updating sharing schema, etc.) In this setup every client will send a request to each server in its own subset every QPS*subset_size = 10*10 = 100 sec. Ideally we want PID to eventually converge to stable weights, where weights are inversely proportional to the number of connections each server receives (this number will be different between servers because of the random subsetting) I can argue that convergence will never happen because of the very low QRS per server as measured per each individual client. Because of the very low QPS every client's view of the server utilization in its own subset will be effectively random, as we are taking into account information about utilization during 1 second and ignoring utilization during remaining 99 seconds since the last request. 

One way to deal with this would be to increase the interval that configures how often every server measures it own utilization. Now we are back to the problem we are trying to solve here, as the config will be spread across clients and servers. Additionally, measuring utilization every 100 sec on the server brings another problem: if utilization permanently increases the server need to wait another 100 sec until it can communicate this to the clients, which significantly slow down load convergence. Usage of circular buffer (either on the client or on the server) helps with this particular problem. Using circular buffer on the server helps with overall memory usage (since we are storing the list of utilization measurements once and sharing it with all clients) Storing the circular buffer on the client is probably not a big deal either, as we can store utilization every weight-update period, like Eric suggested, or even with lower frequency (e. g. during every 10th weight update) We could also do EWMA instead of circular buffer on the server, but I can't see how EWMA measured on the client can work for very low QPS clients.

Another note here is that we can't simply say that if a client has very low QPS we can simply ignore such clients. In this specific example the total number of clients is huge, so overall QPS per server is still high and we still want clients to balance requests in the best way they can.

ejona86: > Suppose you have 100 servers, 10 000 clients, all clients are using random subsetting with subset size = 10 and every individual client generates very low QPS (1 req per 10 sec on average)

For starters, the weight update period should be at least subset size / qps = 100 seconds for clients that (in sum) contribute significant load, and I'd argue for multiple times of that. Updating weights without any updates to utilization steers clients in the wrong direction, so you at least want those contributing the most load to have updated data.

Each server is getting 10000 clients * .1 qps / 100 servers = 10 qps. A server getting 10 qps is going to have noisy actual utilization, no matter the strategy. Each RPC causes 10% of its utilization for a second.

> Additionally, measuring utilization every 100 sec on the server brings another problem: if utilization permanently increases the server need to wait another 100 sec until it can communicate this to the clients, which significantly slow down load convergence

I'm not sure you need it that high. But honestly, that doesn't sound bad. And on average it would take 50 sec, not the full 100. It sounds like a big number, but only because each client talks to each server only once every 100 seconds (assuming uniform weights). Even if the client was told as soon as the load was added and the weights were immediately recalculated, it would take 100 seconds before the client _doesn't_ send to this overloaded backend.

But I think that's irrelevant. We were comparing to the current ORCA protocol. Circular buffers and EWMA are both possible on the server with current ORCA. We don't have to change the architecture for that.

-----

Looking more deeply at this, I think noise only matters within the most recent weight update period. proportional_gain is low, so it is mostly derivative_gain impacted by noise. Averaging utilization from prior periods does nothing that modifying derivative_gain couldn't do. Let's say you have a update period of 1 second and you want to average utilization for 10 seconds. Each update period contributes 10% of the average utilization, so we can multiply derivative_gain by 10% instead of averaging.

I think the focus so much on averaging is simply because derivative_gain of 1 is too big. That will echo any noise back into the system, assuming a weight_update_period of 1 second and weights near 1.

I think we should consider not having derivative_gain (use `0.0`). I don't think we use the predictive nature of it, and instead benefit from it counter-acting changes like large-clients coming and going suddenly. But tuning for that will be hard, and if needed a larger proportional_gain can compensate with fewer risks.

s-matyukevich: > I think we should consider not having derivative_gain (use 0.0) 

I like this idea, we can simplify things a lot if we rely only on smoothing and proportional gain.

---

We recently bumped into issue with a service that was using plain WRR (not PID) with round-robin and without any subsetting. The server was measuring CPU every 1 sec, and this is what we get after we added server-side smoothing with 1 min sliding window. 

Screenshot 2024-12-12 at 8 24 13 AM

I still don't fully understand what is going on here (why load per server pod was stable, but deviation between pods was so high) and whether we could achieve the same result using just longer measuring interval on the server.  I want to reproduce this in a synthetic benchmark and will report back when I have more data.

ejona86: It's not immediately obvious to me precisely what happened, but WRR assumes utilization and rps are for the same rough time period. If you smooth one without smoothing the other, I wouldn't be surprised to see weird results.

s-matyukevich: In both cases (before and after smoothing was added) utilization and rps were measured for the same period.

ejona86: Oh, you're pointing out how the banding went away, not the initial spikes. Which is interesting because we wouldn't expect banding at all. If that was with DNS, I would question if clients aren't polling for address updates, so the high outliers would be older servers from before a scale-up. To get a hint as to the cause, I suggest watching that dashboard and seeing if the bands re-form after a period without restarting all backends.

PR #432, opened 2024-05-12
Add value type (integer/floating-point) to metrics

yashykt:


A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -1,34 +1,35 @@
-A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient
-----
-* Author(s): @markdroth
-* Approver: @ejona86, @dfawley
-* Status: {Draft, In Review, Ready for Implementation, Implemented}
-* Implemented in: <language, ...>
-* Last updated: 2024-04-08
-* Discussion at: https://groups.google.com/g/grpc-io/c/A2Mqz8OMDys
+## A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient
```
markdroth, line 1: Can you please undo the reformatting here?  This makes it really hard to see what's actually changed.

A66-otel-stats.md
```diff
@@ -406,22 +407,22 @@ the data.
 *   **grpc.client.attempt.started** <br>
     The total number of RPC attempts started, including those that have not completed. <br>
     *Attributes*: grpc.method, grpc.target <br>
-    *Type*: Counter <br>
+    *Type*: Counter (integer) <br>
```
markdroth, line 410: I think we might want to say something more precise than just "integer" vs. "floating point".  For example, would it be okay for a gRPC implementation to use a 32-bit integer here?  I assume not, in which case we need language here saying so.

PR #435, opened 2024-05-17
A81: xDS Authority Rewriting

markdroth:


atollena:
A few comments:

- I think it would help to provide a very brief explanation of when this feature is expected to be useful, if only for gRPC maintainers' own culture. As far as I know it is necessary when the server is using the authority header to make decisions. This is common in reverse proxies where each deployment serve multiple domains. Servers can then take routing decisions based on it (forward the request to different backends). I think it is fairly uncommon to make use of the authority header for direct gRPC client -> gRPC servers, but I could be wrong on this.

- It would help to specify whether you expect implementations to use the rewritten authority for the authentication handshake so that servers can provide different TLS certificates based on the requested domain (SNI), rather than just for the authority header. I think Envoy supports this, although I'm not sure how this is configured.

- grpc-go provides the `WithAuthority` channel option (https://pkg.go.dev/google.golang.org/grpc#WithAuthority), so I think it would also be useful to specify which one takes precedence, in case both this option and the management server provides an authority.

markdroth:
Thanks for those comments!

> * I think it would help to provide a very brief explanation of when this feature is expected to be useful, if only for gRPC maintainers' own culture. As far as I know it is necessary when the server is using the authority header to make decisions. This is common in reverse proxies where each deployment serve multiple domains. Servers can then take routing decisions based on it (forward the request to different backends). I think it is fairly uncommon to make use of the authority header for direct gRPC client -> gRPC servers, but I could be wrong on this.

I've added a sentence about this.

> * It would help to specify whether you expect implementations to use the rewritten authority for the authentication handshake so that servers can provide different TLS certificates based on the requested domain (SNI), rather than just for the authority header. I think Envoy supports this, although I'm not sure how this is configured.

This feature is purely about rewriting the authority header on data plane RPCs, which happens independently of the TLS handshake.  The same is true in Envoy.  I've added a sentence about this.

(Both Envoy and gRPC have other features that allow controlling the TLS handshake, but that's unrelated to this gRFC.)

> * grpc-go provides the `WithAuthority` channel option (https://pkg.go.dev/google.golang.org/grpc#WithAuthority), so I think it would also be useful to specify which one takes precedence, in case both this option and the management server provides an authority.

Good question!  We should figure out the right precedence rules here.

All languages have an option similar to the one you mentioned for setting the default authority for the channel as a whole.  C++ also has options to set the authority on a per-call basis, although I don't think other languages have that.

For the channel-level option, my inclination is that the xDS config should take precedence over the channel-level default, especially since there will be a knob in the bootstrap file to disable that if it's not desirable for some reason.

For the C++ option to set the authority on a per-RPC basis, I think that the principled answer is that that API should take precedence over the xDS config.  However, in practice, I suspect no one will ever try to use these two options together, and it's less work to allow the xDS config to take precedence, so I'm leaning in that direction.

@ejona86 @dfawley @ctiller Thoughts on that?

dfawley:
> For the channel-level option, my inclination is that the xDS config should take precedence over the channel-level default, especially since there will be a knob in the bootstrap file to disable that if it's not desirable for some reason.

> For the C++ option to set the authority on a per-RPC basis, I think that the principled answer is that that API should take precedence over the xDS config. However, in practice, I suspect no one will ever try to use these two options together, and it's less work to allow the xDS config to take precedence, so I'm leaning in that direction.

That all SGTM.

markdroth:
> For the C++ option to set the authority on a per-RPC basis, I think that the principled answer is that that API should take precedence over the xDS config. However, in practice, I suspect no one will ever try to use these two options together, and it's less work to allow the xDS config to take precedence, so I'm leaning in that direction.

It turns out that the easy thing here is actually the principled answer: if the application explicitly sets the authority on a per-RPC basis, that will take precedence over the xDS config.

A81-xds-authority-rewriting.md
```diff
@@ -0,0 +1,171 @@
...
+attribute in the subchannel wrapper when its child policy creates a
+subchannel.  In the xds_cluster_impl picker, if the `auto_host_rewrite`
+option is enabled in the route (passed from the ConfigSelector as
+described above) and the `hostname` attribute on the subchannel wrapper
+is non-empty, the picker will set the `:authority` header to the value of
+the `hostname` attribute.
+
+To support this, we will add an API to allow the LB picker to explicitly
+set the `:authority` header for the RPC.  Note that this authority
+rewriting will modify the default authority for the subchannel, which
```
dfawley, line 119: Technically: s/modify/override/ since it doesn't alter the default itself.

dfawley (approved):
Just a few wording nits otherwise LGTM!

A81-xds-authority-rewriting.md
```diff
@@ -0,0 +1,171 @@
...
+- Bootstrap config change: We will add a new server feature in the bootstrap
+  config to indicate that we should allow authority rewriting.
+- xDS resource validation: We will process authority rewriting fields in RDS
+  resources and the hostname field in EDS resources, and return that
+  information to the XdsClient watchers as part of the parsed resources.
+- xDS ConfigSelector changes: We will change the xDS ConfigSelector to enable
+  authority rewriting based on the chosen route.
+- xds_cluster_impl LB policy changes: We will change the xds_cluster_impl LB
+  policy to actually perform the authority rewriting was enabled by the
+  ConfigSelector.
```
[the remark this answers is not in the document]
markdroth: Done.

A81-xds-authority-rewriting.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Server Feature in Bootstrap Config
+
+In order to address use-cases where authority rewriting may not be
+acceptable from a security perspective, we will add a new server feature
+to the bootstrap config.  The server feature will be specfied via the
+`server_features` field described in [gRFC A30][A30].  The feature will
+be the string `trusted_xds_server`.  (Note that the name is intentionally
+fairly general, since it may be used to trigger other functionality in
```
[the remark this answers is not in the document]
markdroth: Done.

A81-xds-authority-rewriting.md
markdroth: Done.

A81-xds-authority-rewriting.md
```diff
@@ -0,0 +1,171 @@
...
+field.  The boolean value of this field will be included in the parsed
+resource struct that is passed to the XdsClient watcher.
+
+When validating an EDS resource, we will look at the [`Endpoint.hostname`
+field](https://github.com/envoyproxy/envoy/blob/b65de1f56850326e1c6b74aa72cb1c9777441065/api/envoy/config/endpoint/v3/endpoint_components.proto#L89).
+The string value of this field will be included in a per-endpoint resolver
+`hostname` attribute in the parsed resource struct that is passed to the
+XdsClient watcher.  Note that the resolver attribute used here should
+be a general-purpose one, not something specific to EDS; for example, in
+the future, we may want to use this same attribute to expose per-endpoint
```
ejona86, line 91: What is this about? If we're going to be looking at this for general things, I don't really understand the semantics. Right now this is happening independent of `trusted_xds_server` (is that right?) and `auto_host_rewrite`, so this is just a random string at this point from the perspective of the system. I'm very wary calling this attribute "hostname," even when trusted_xds_server=true, because it is conditional on other things. If it were only for xds, maybe I could live with it, but if we are having it be for random other things too... we need to very clearly specify what the value means/is. I have no idea what semantics the value has when trusted_xds_server=false.

A81-xds-authority-rewriting.md
```diff
@@ -0,0 +1,171 @@
...
+the `hostname` attribute.
+
+To support this, we will add an API to allow the LB picker to explicitly
+set the `:authority` header for the RPC.  Note that this authority
+rewriting will override the default authority for the subchannel, which
+would otherwise be set from the resolver factory, a channel option,
+or a per-address attribute returned by the resolver (i.e., the xDS
+authority rewriting will take precedence over any of those settings).
+However, if the implementation allows the application to explicitly
+set the authority on a per-RPC basis (currently only C-core does this),
```
ejona86, line 124: @kannanjgithub, note that Java _partly_ has this (`CallOptions.withAuthority`). But very importantly it is missing the next paragraph. Whether we need to just start enforcing the hostname checking in the current API or whether we should make a new API and deprecate+remove the old API is TBD.

A81-xds-authority-rewriting.md
markdroth: The idea is that this is a general-purpose attribute (not xDS-specific) to indicate a human-readable name associated with the individual address.  Any resolver can populate this attribute, and if it's set, we can display this information in channelz.

This is something we started working on a long time ago for internal purposes but never really drove to completion.  In C-core, we added the attribute in https://github.com/grpc/grpc/pull/24542, and a bunch of internal work was done to include this info in our internal resolver in b/130755297 and b/130754689, but the remaining work of actually plumbing it into channelz hasn't been done yet (b/130755264).

ejona86: There's still some awkwardness and the semantics are a bit strange (generally I'd write, "this value is untrusted; do not use it for security" but here we're immediately going to use it for security). But that's all fine; we don't paint ourselves into any corner here. We can "do something" here and see where things go.

ejona86 (approved):

PR #436, opened 2024-05-17
A82: xDS System Root Certificates

markdroth:


ejona86 (approved):

A82-xds-system-root-certs.md
```diff
@@ -0,0 +1,81 @@
...
+
+[A29]: A29-xds-tls-security.md
+
+## Proposal
+
+We have added a [`system_root_certs`
+field](https://github.com/envoyproxy/envoy/blob/84d8fdd11e78013cd50596fa3b704e152512455e/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L399)
+to the xDS `CertificateValidationContext` message (see
+envoyproxy/envoy#34235).  If this field is present and the
+`ca_certificate_provider_instance` field is unset in the gRPC client,
```
ejona86, line 39: "in the gRPC client" is awkward. "for a" would be better. Dunno what is best.

dfawley (approved):

A82-xds-system-root-certs.md
markdroth: I've attempted to clarify the wording here.

A82-xds-system-root-certs.md
```diff
@@ -0,0 +1,83 @@
...
+will configure XdsCreds to use system root certs.
+
+### XdsCredentials Changes
+
+The XdsCredentials code will be modified such that if it is configured
+to use system root certs, it will configure the TlsCreds code to do that.
+
+### Temporary environment variable protection
+
+Use of the `use_system_root_certs` field in CDS and LDS will be guarded
```
ejona86, line 69: I'm not sure I noticed that we were going to have different validation for client-side even though this is in a shared message. Up above it says "Note that LDS validation will be unchanged" which appears to disagree with this line. Only supporting it on client-side might be a bit harder to support, since this is in CommonTlsContext. Is there a reason we're only supporting it on client-side?

Also: s/use_system_root_certs/system_root_certs/

CC @kannanjgithub

markdroth: We did talk about this question specifically before finalizing this gRFC.  I had originally intended to support this option on both the client side and server side, just for consistency, but when I went to implement this in C-core, it turned out to be non-trivial, because we don't already have server-side code for using system root certs, so we decided to exclude it.

You're right about the typo.  I'll send a separate PR to fix that.

markdroth: Sent #451 to fix this.

PR #438, opened 2024-05-21
A83: xDS GCP Authentication Filter

markdroth:


A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,218 @@
...
+
+For now, the only registered metadata type we support is
+[`extensions.filters.http.gcp_authn.v3.Audience`](https://github.com/envoyproxy/envoy/blob/c16faca3619fb44c24b12d15aad8a797b9e210ab/api/envoy/extensions/filters/http/gcp_authn/v3/gcp_authn.proto#L66).
+In this message, the `url` field must be non-empty; if empty, the
+resource will be NACKed.  The parsed representation of this message can
+be a simple string.
+
+Note that in Envoy, the metadata keys must exactly match the legacy
+filter name (e.g., "envoy.filters.http.gcp_authn").  However, as per
+envoyproxy/envoy#34251, it is desirable to instead use the HTTP filter
```
ejona86, line 146: We already support Metadata keys that aren't tied to any filter. So this is only related to filters with specific cluster configuration?

A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,218 @@
...
+RPCs sent on insecure connections.  However, the GCP Authentication
+filter will run before load balancing has chosen a connection, so the
+filter cannot directly add the token to the RPC.  Instead, it must add
+the call credential to the RPC, and the call credential will do the work
+of adding the token to the RPC later, after load balancing has chosen a
+connection.
+
+### Temporary environment variable protection
+
+Support for the GCP Authentication filter in LDS will be guarded by the
```
ejona86, line 193: It should also protect extensions.filters.http.gcp_authn.v3.Audience from being in the metadata registry.

A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,218 @@
...
+guard will be removed once the feature passes interop tests.
+
+## Rationale
+
+It is not our intention to support this mechanism for GCP only; in
+principle, it should be possible to support JWT identity tokens for any
+cloud provider.  However, at present, the existing xDS HTTP filter
+supports only GCP, so that's what we're initially focusing on, for
+compatibility with Envoy.  We would be open to future contributions from
+the OSS communicate to provide similar functionality for other cloud
```
ejona86, line 204: s/communicate/community/

A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,218 @@
...
+providers, in both gRPC and Envoy.
+
+Note that having the filter cache
+GcpServiceAccountIdentityCallCredentials instances instead of directly
+caching JWT tokens means that the cache semantics will be slightly
+different than Envoy: cache expiration will be based solely on
```
ejona86, line 210: "cache" here is imprecise, as there's two caches: one in the filter, one in the cred. And you most recently talked about "caching JWT tokens." But this is also sorta wrong: there are two cache expirations (thus "cache expiration _is not_ based solely on LRU") whereas in Envoy there's only one. I think if you specify which cache, and describe both caches, in order to contrast it.

markdroth: Good point.  Rewrote this to improve clarity.

A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,218 @@
...
+[A74]: A74-xds-config-tears.md
+[RFC-7519]: https://datatracker.ietf.org/doc/html/rfc7519
+
+## Proposal
+
+We will support the GCP Authentication xDS HTTP filter in the gRPC client.
+
+### Call Credentials
+
+gRPC should support a GcpServiceAccountIdentityCallCredentials call
```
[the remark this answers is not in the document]
ejona86: Ah, yeah, that's appropriate.

markdroth:
I made one more small change here related to how we plumb the CDS data into the filters.  I think that can just reuse the attribute added in A74 (which was already a prereq for this design) rather than adding a new attribute.

A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,330 @@
...
+
+### xDS HTTP Filter Configuration
+
+The xDS HTTP filter will be configured via the
+[`extensions.filters.http.gcp_authn.v3.GcpAuthnFilterConfig`
+message](https://github.com/envoyproxy/envoy/blob/c16faca3619fb44c24b12d15aad8a797b9e210ab/api/envoy/extensions/filters/http/gcp_authn/v3/gcp_authn.proto#L27).
+The fields will be interpretted as follows:
+- `cache_config`: Optional.  Within this message:
+  - `cache_size`: Optional.  If set, must be greater than 0 and less
+    than `INT64_MAX`.  Defaults to 10.
```
ejona86, line 125: In other similar cases hasn't there been an upper limit to this sort of thing (even if implementation-defined)? Not that I think we need it for our security model or such, since this is xds-specific, but Envoy itself commonly had limits. I'm fine without, but it seemed strange/irregular.

ejona86 (commented):
I'm not approving yet to resolve the "RPC during backoff delay" behavior. The rest looks fine/easy.

A83-xds-gcp-authn-filter.md
ejona86: So this is a uint64 field that is limited to int64-1 range. Java will be limited to int32-1 range.

Is there really any benefit to NACKing too large of a value here? It seems silly; we should just clamp to our implementation limit. It seems the current limit is there because of an Envoy implementation limit.

ejona86: Oh, and Envoy explicitly supports `0` but we don't. There's a lot of strangeness here.

A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,330 @@
...
+
+To handle potential clock skew issues and to account for processing time
+on the server, the credential will set the cache expiration time to be
+30 seconds before the expiration time encoded in the token.  All logic
+in the call credential code will use this modified expiration time
+instead of the expiration time encoded in the token.
+
+When the credential is asked for a token for a data
+plane RPC, if the token is not yet cached or the cached
+token will expire within some fixed refresh interval
```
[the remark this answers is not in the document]
markdroth: Done.

A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,330 @@
...
+token is invalid), all queued data plane RPCs will be failed with status
+`UNAUTHENTICATED`.  Otherwise, the cache is updated, and the returned
+token is added to all queued data plane RPCs, which may then continue.
+
+If the HTTP request does not result in the cache being updated (i.e.,
+if the HTTP request fails or if it returns an invalid JWT token),
+[backoff](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md)
+must be applied before the next attempt may be started.  If a data
+plane RPC is started when there is no cached token available and while
+in backoff delay, it will be queued until the completion of the next
```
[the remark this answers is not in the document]
markdroth: Hmm.  That's a good point, and I think I agree that we should ideally fail RPCs in this case.

The reason we did it this way was so that we have some notion of when to reset the backoff state.  When the backoff expires, if there are queued RPCs, then we immediately trigger another HTTP fetch, but if there are *no* queued RPCs, then we clear the backoff state.  If we didn't do that, then there could be situations where we have a couple of failures and get the backoff delay up, but then a day goes by without another RPC, and then the next RPC that sees a failure causes us to reuse the day-old backoff state.

But thinking about this now, I think there are better alternatives here.  The obvious one is that we could just accept that we're using day-old backoff state in that scenario.  If the server has actually been down the whole time, then that's actually not inappropriate.  And even if it came back up while we weren't trying but has now gone down again, that's a sufficiently coincidental case that it's probably fine to treat it as if it had been down the whole time.

Or, if we really want to avoid this, we could have some fixed timeout for resetting the backoff state if we haven't made another attempt in a long enough period of time.  I'm thinking maybe something like 30 minutes.

Thoughts...?

A83-xds-gcp-authn-filter.md
markdroth: Yeah, disabling the cache doesn't seem like a good idea to me (much less as a default!), since that could result in hammering the HTTP server.  It's also going to be more expensive memory-wise for us than for Envoy, since we're caching an entire call creds instance, not just the data for the token.  So I figured we'd use default 10 and not allow 0.

The upper bound was copied from Envoy; I didn't see any reason to change that.  I agree that it's odd that its implementation has that constraint, but I didn't have a good reason to do anything else in this case, so I opted for consistency.  But I agree that clamping to our max seems fine rather than NACKing in this case, and I've made that change.

Why would Java be limited to int32-1?  I know you don't have unsigned types, but you do have 64-bit types, right?  I would have expected you to have the same limitation as Envoy in that regard.

A83-xds-gcp-authn-filter.md
```diff
@@ -0,0 +1,218 @@
...
+two maps; the value from `filter_config` will be used only if the
+key is not present or is of an unknown protobuf message type in
+`typed_filter_config`.  In the resulting map in the parsed cluster
+resource, the map value will be either JSON or the appropriate internal
+form, depending on which of the two maps the entry came from.
+
+To match Envoy, the logic to validate cluster metadata will look something
+like this (pseudo-code):
+
+```
```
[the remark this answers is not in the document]
markdroth: Okay, done.

A83-xds-gcp-authn-filter.md
```diff
@@ -203,14 +203,34 @@ selected cluster.  This data is available via the `XdsConfig` attribute
...
+GcpServiceAccountIdentityCallCredentials instances, one for each audience,
+along with a last-used list that tracks how recently the entries
+have been used.  As an entry is used, it is moved to the front of the
+last-used list.  The maximum number of entries in the cache is bounded
+by the config field `cache_config.cache_size`; if the cache exceeds that
+size, then entries will be removed starting from the end of the
+last-used list.
+
+Note that the `cache_config.cache_size` parameter in the filter config
+is a truly global parameters, not settable per-route, and we want the
```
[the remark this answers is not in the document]
markdroth: Good point, clarified wording.

A83-xds-gcp-authn-filter.md
ejona86: Java is essentially 32 bit at its core. Arrays are limited to 31 bits of length. We'd need to use a two-tier list to handle larger values. The `-1` is for us to tell when we've exceeded the limit.

@shivaspeaks, note that this changed.

A83-xds-gcp-authn-filter.md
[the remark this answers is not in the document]
ejona86: > but then a day goes by without another RPC, and then the next RPC that sees a failure causes us to reuse the day-old backoff state.

Not actually? That RPC will cause us to fetch a token, starting immediately. If that succeeds, then the overly large backoff is never used. If it fails, then using a large backoff is still fine.

(Compared to reconnect handling, this will have blips of increased RPC latency as (fail to) fetch the token. But that's acceptable and core to the approach taken here.)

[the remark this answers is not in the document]
markdroth: Right, sorry, I meant the RPC *after* the one that occurs after a day of being idle.  The one that occurs after a day of being idle will definitely be dispatched immediately, but if we retain backup state from the previous day, then the delay after that failure will be longer than maybe it should be.

But it sounds like we're converging on that being okay.  I've updated the doc accordingly, and I will also update the C-core implementation.

A83-xds-gcp-authn-filter.md
markdroth: Interesting, I didn't realize that.  In this case, the cache should be a map, not an array, but I don't know if that makes any difference.

I've changed the wording here to be a bit more generic.

ejona86: Hash maps are implemented using arrays. And the LRU is implemented using a hash map with a linked list between entries. Sure, trees are available, but we'd still need to do some custom work to have a 64 bit size and manage the LRU. I thought it was obvious there is no point in doing extra work to supporting things that size.

ejona86 (approved):

dfawley (approved):
LGTM

A83-xds-gcp-authn-filter.md
shivaspeaks: Okay, so now it is generalized and it includes our case to cap the cache to int32-1. So I guess we don't need to change our LruCache? @ejona86 

> cache_size: Optional. If set, must be greater than 0. Defaults to 10. Implementations that cannot support caches as large as UINT64_MAX may cap this value at their maximum supported size.

ejona86: @shivaspeaks, let's discuss more in your PR, but the main thing is we no longer need to return-an-error/NAK if the value is too large; we'd only NAK on zero. Other cases we just clamp to our maximum.

PR #440, opened 2024-05-29
Modern proposal for exposing client TLS cert in server-side call context in gRPC-JS

davidfiala:
Modern proposal for exposing client TLS cert in server-side call context. Discussion on https://github.com/grpc/grpc-node/issues/2730 suggested we start here.

### DIFF
--- a/L35-node-getAuthContext.md
+++ b/L35-node-getAuthContext.md
@@ -1,56 +1,65 @@
-Exposing the per-call authentication context data in Node
+Exposing the Peer Certificate in gRPC-JS Server Call Context
+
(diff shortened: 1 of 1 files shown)


murgatroid99:
When I talked about editing this proposal, I was thinking about an addendum that would expand the `sslPeerCertificate` field in the grpc-js implementation. It's not appropriate to edit an existing proposal to propose a complete different API and implementation. Instead it would be better to have a new proposal for a change this substantial. Then after that we could edit this proposal to indicate that it has been superseded by the new one.

PR #445, opened 2024-07-13
A74 update: xds resolver needs to handle re-resolution

markdroth:


ejona86 (approved):
This is clearly better than without refreshes, but I am wary that one cluster will cause some other cluster to "work" and get updates. This breaks predictability compared to pre-A74.

markdroth:
I agree that it's a sub-optimal change, although I think it's unlikely to be a problem in practice, because (a) there probably aren't too many targets that have more than one logical DNS cluster in their config, and (b) polling-based resolvers like DNS should impose a min time between queries anyway.  Also, if/when this actually becomes a problem, I suspect we can come up with ways to address it.

PR #450, opened 2024-08-14
L117 update: Add 3 new absl wrapper functions

tanvi-jagtap:
L117 update: Add 3 new absl wrapper functions .
These are needed for using from the `C` code because C code cannot use abl directly.


L117-core-replace-gpr-logging-with-abseil-logging.md
```diff
@@ -80,6 +80,30 @@ void SomeInitFunctionCalledByGrpcInit() {
...
 
+### Newly Added Functions
+
+* The following functions will be added. These are very simple wrappers around absl logging.
+  * These are needed because PHP and RUBY cannot use absl directly. Absl logging cannot be used from `C` code. These simple wrapper functions will be exported so that they can be used from `C` code.
+  * We want to avoid usage of format specifiers that RUBY and PHP are currently using.
+
+```
+// Equivalent to ABSL_LOG(severity) << message_str;
+GPRAPI void grpc_absl_log(const char* file, int line,
```
yashykt, line 91: I am not a fan of these functions.  Are we going to need `grpc_absl_log_float`, `grpc_absl_log_int_int`, `grpc_absl_log_str_int` and variations in the future?

Is it possible to get PHP and Ruby to use C++ functions directly?

yashykt: @apolcyn @stanley-cheung can you comment on this please? what's the future here?

tanvi-jagtap: @apolcyn , @stanley-cheung : Please let us know.

apolcyn: > I am not a fan of these functions. Are we going to need grpc_absl_log_float, grpc_absl_log_int_int, grpc_absl_log_str_int and variations in the future?

I don't think we will need variations in the future.

The proposed APIs are not the most *ergonomic* to work with, but practically speaking they are "good enough" for any use from ruby I can imagine.

If we need to log multiple things, we can just use multiple log statements, and correlate the resulting logs through their thread IDs during analysis.

While `float` *might* come up, there are no current, past, or foreseen uses to log a `float` from ruby. If a use does come up in the future, I'd be surprised if we couldn't break it down into a smaller unit and log an integer multiple of it (e.g. logging an integer multiple of microseconds rather than a fractional number of seconds).

> Is it possible to get PHP and Ruby to use C++ functions directly?

Possibly, yes. We'd just need to figure out how to compile ruby and php as C++. This isn't impossible, but would take some time to figure out. We could try to fit this in e.g. for ruby but I'm not sure I'll have cycles within say Q4.

tanvi-jagtap: Thanks apolcyn for the inputs.

@stanley-cheung : Please let us know so that we can proceed with this.

yashykt: I think I want to use language that says this is temporary API while Ruby/PHP transition to C++. We can do that actual transition later or whenever it makes sense, but I'd like the intention to be clear.

stanley-cheung: I have no issue with these APIs. PHP is already compiling C++ source codes so I think PHP should be fine.

yashykt (approved):

PR #452, opened 2024-08-23
A6 update: change specification of backoff behavior to make sense

markdroth:
See discussion in https://github.com/grpc/grpc-go/issues/7514.

### DIFF
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -108,6 +108,8 @@ gRPC's call deadline applies across all attempts for a given RPC. For example, i
 
 The `initialBackoff`, `maxBackoff`, and `backoffMultiplier` parameters tune the exponential delay before retry attempts.
 
+Jitter of plus or minus 0.2 is applied to the backoff delay to avoid hammering servers at the same time from a large number of clients.  Note that this means that the backoff delay may actually be slightly lower than `initialBackoff` or slightly higher than `maxBackoff`.
+
 The initial retry attempt will occur after `initialBackoff * random(0.8, 1.2)`.  After that, the `n`-th attempt will occur after `min(initialBackoff*backoffMultiplier**(n-1), maxBackoff) * random(0.8, 1.2))`.
 
 ##### Retryable Status Codes
--- a/A6-client-retries.md
+++ b/A6-client-retries.md
@@ -4,7 +4,7 @@ gRPC Retry Design
 * Approver: a11r
 * Status: Implemented
 * Implemented in: Java, .NET, Node, Go except hedging, and C-Core except hedging
-* Last updated: 2022-11-17
+* Last updated: 2024-08-23
 * Discussion at: https://groups.google.com/forum/#!topic/grpc-io/zzHIICbwTZE
 
 Table of Contents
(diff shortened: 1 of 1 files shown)


A6-client-retries.md
```diff
@@ -106,9 +106,9 @@ gRPC's call deadline applies across all attempts for a given RPC. For example, i
 
 ##### Exponential Backoff
 
-The `initialBackoff`, `maxBackoff`, and `backoffMultiplier` parameters determine the randomized delay before retry attempts.
+The `initialBackoff`, `maxBackoff`, and `backoffMultiplier` parameters tune the exponential delay before retry attempts.
 
-The initial retry attempt will occur at `random(0, initialBackoff)`. In general, the `n`-th attempt will occur at `random(0, min(initialBackoff*backoffMultiplier**(n-1), maxBackoff))`.
+The initial retry attempt will occur after `initialBackoff`.  After that, the `n`-th attempt will occur after `min(initialBackoff*backoffMultiplier**(n-1), maxBackoff) * random(0.8, 1.2))`.
```
ejona86, line 111: The initial attempt should also have the jitter. Ditto in the schema documentation.

ejona86, line 111: Subtle: the jitter is applied to the maxBackoff. I expect you thought about that when writing it. I think the idea is "Yes, the backoff can exceed maxBackoff, but that confusing part is better than not having the jitter." I'm willing to agree, but wanted to call out the detail. (And I don't think we need to discuss the option of doing 0.8-1x the maxBackoff.)

ejona86 (approved):

A6-client-retries.md
markdroth: Good point!  Fixed.

It looks like C-core's backoff implementation isn't actually doing this, so I'll fix that too. :)

markdroth: That hadn't actually occurred to me, but I agree with you that it's the right behavior.  Even when we get to max backoff, we want the jitter, and I don't see any reason to cap it that strictly.

This also matches the [connection backoff spec](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md), which keeps everything consistent.

A6-client-retries.md
```diff
@@ -106,9 +106,9 @@ gRPC's call deadline applies across all attempts for a given RPC. For example, i
 
 ##### Exponential Backoff
 
-The `initialBackoff`, `maxBackoff`, and `backoffMultiplier` parameters determine the randomized delay before retry attempts.
+The `initialBackoff`, `maxBackoff`, and `backoffMultiplier` parameters tune the exponential delay before retry attempts.
 
-The initial retry attempt will occur at `random(0, initialBackoff)`. In general, the `n`-th attempt will occur at `random(0, min(initialBackoff*backoffMultiplier**(n-1), maxBackoff))`.
+The initial retry attempt will occur after `initialBackoff * random(0.8, 1.2)`.  After that, the `n`-th attempt will occur after `min(initialBackoff*backoffMultiplier**(n-1), maxBackoff) * random(0.8, 1.2))`.
```
dfawley, line 111: This means the backoff can be both:

1. less than `initialBackoff`, and
1. more than `maxBackoff`

(1) is probably OK but (2) seems to subvert a reasonable user's assumption of what `maxBackoff` would mean.

markdroth: Eric commented on this above.  I think it's fine for the jitter to make it exceed max backoff.  Note that this is consistent with the [connection backoff spec](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md).

dfawley: Since we independently (oops) noticed the same thing, maybe we should explicitly have a note that says "Note that the backoff used may exceed the configured maxBackoff by up to 20 percent"?

markdroth: Good idea.  Done.

dfawley (approved):

ejona86:
We need to notify grpc-io about this change, and update the [user guide](https://grpc.io/docs/guides/retry/).  At least in one place it says "The actual initial backoff delay is a random time period within initialBackoff value configured above (between 0-100ms)."

markdroth:
Good catch.  I've sent https://github.com/grpc/grpc.io/pull/1356 to update the wording in the retry user guide.

I don't feel like we really need a post to grpc.io about this, but I can do one if you feel strongly.

PR #453, opened 2024-09-18
add environmental variable for  java otel tracing: GRPC_EXPERIMENTAL_…

YifeiZhuang:
…ENABLE_OTEL_TRACING


ejona86 (approved):
And for reference, this is Java-only because the other languages require the user to call an API/set a field to enable tracing. Java was the only one that had no API surface changes.

PR #456, opened 2024-09-24
A78 update: add note about PF metrics not being in the subchannel

markdroth:


dfawley (approved):

ejona86 (approved):
The paragraph is not particularly convincing. Even if you have 100 addresses in PF and someone fixes the unplugged ethernet cable, the chance for the race is low as events are unlikely to align. And if you do hit the race, with low frequency (sub-10% is probably acceptable), then it seems there's no harm. But whatever. I don't really need it and doing it in the subchannel does have to make assumptions about what PF is doing.

PR #460, opened 2024-10-14
L120: Requiring C++17

veblush:


L120-requiring-cpp17.md
```diff
@@ -0,0 +1,29 @@
...
+* Discussion at: TBD
+
+## Abstract
+
+gRPC starts requiring C++17.
+
+## Background
+
+To leverage the advancements in C++ standards, gRPC has progressively updated its requirements.  
+Initially, it adopted C++11 in 2017 (as per [Allow C++ in gRPC Core Library](L6-core-allow-cpp.md)).
```
markdroth, line 17: Please add "L6:" to the start of the text for this link, to make it clear that it references a gRFC.

L120-requiring-cpp17.md
```diff
@@ -0,0 +1,29 @@
...
+* Author(s): veblush
+* Approver: markdroth
+* Status: Draft
+* Implemented in: n/a
+* Last updated: Oct 14, 2024
+* Discussion at: TBD
+
+## Abstract
+
+gRPC starts requiring C++17.
```
markdroth, line 12: Do we want to have to keep publishing new gRFCs every time we add support for a new version of C++, or do we just want to publish a gRFC that says "from now on, we'll follow the OSS foundational C++ support policy"?

markdroth (commented):
Thanks for doing this!

L120-requiring-cpp17.md
veblush: Done.

L120-requiring-cpp17.md
veblush: It's a good point. Our plan is to keep up with the OSS foundational C++ support policy so we can have this last gRFC making it clear and skip new gRFC for the next bump.

eugeneo:
Are we going to allow, e.g., to `std::string_view` from `absl::string_view`? Should this gRFC mention this?

L120-requiring-cpp17.md
veblush: I mentioned the future plan at the bottom. PTAL.

veblush:
> Are we going to allow, e.g., to `std::string_view` from `absl::string_view`? Should this gRFC mention this?

This topic is beyond the scope of this doc but that should be communicated well within the maintainers. Abseil by default makes `absl::string_view` an alias of `std::string_view` under C++17 but it still provides an opt-in build option to use their own implementation. Hence, to support this build option, gRPC will have to stick to `absl::string_view` until we stop supporting this option.

veblush:
> > Are we going to allow, e.g., to `std::string_view` from `absl::string_view`? Should this gRFC mention this?
> 
> This topic is beyond the scope of this doc but that should be communicated well within the maintainers. Abseil by default makes `absl::string_view` an alias of `std::string_view` under C++17 but it still provides an opt-in build option to use their own implementation. Hence, to support this build option, gRPC will have to stick to `absl::string_view` until we stop supporting this option.

https://github.com/grpc/grpc/pull/37930 just got merged to prevent accidental use of `std::string_view` in the future.

L120-requiring-cpp17.md
markdroth: If this is actually our plan going forward, then I think we should re-title this gRFC -- it's less about C++17 specifically than it is about the fact that we're now going to follow the OSS foundational C++ support guidelines.

Maybe this warrants a brief discussion at our Tuesday design review meeting?

veblush:
We've decided to limit this proposal to the C++17 transition only. We'll repeat this process for future transitions.

markdroth (approved):

PR #462, opened 2024-10-30
A87: mTLS SPIFFE Support

markdroth:


A87-mtls-spiffe-support.md
```diff
@@ -0,0 +1,200 @@
...
+3. Verify the peer certificate using the default security library using
+   the SPIFFE trust bundle certificates as roots.
+
+SAN matching will be performed as configured via xDS, just as it is with
+X509 verification.
+
+gRPC currently supports only one certificate provider implementation,
+which is the `file_watcher` provider.  We will add a new field to the
+configuration of this provider implementation configuration called
+`spiffe_trust_bundle_map_file`.  This field will specify the path to
```
dfawley, line 70: Can we shorten the name to `spiffe_trust_bundle_file` (drop the `map`)?

dfawley: OK I see it's being called a "bundle map" so I guess it's needed. Sorry for the spam.

A87-mtls-spiffe-support.md
```diff
@@ -0,0 +1,188 @@
...
+essentially the same semantics as in xDS.  We will add the ability for
+the `FileWatcherCertificateProvider` to be instantiated with a SPIFFE
+trust bundle map file instead of a CA certificate file.
+
+TODO: @erm-g to fill in API details.  We'll probably need to introduce
+an options struct for FileWatcherCertificateProvider.
+
+#### Java
+
+In Java, we need new API to work with [SPIFFE] and [SPIFFE bundle
```
ejona86, line 122: This is _internal_ API, not public. This section documents how the xDS part is implemented, but it _doesn't_ describe any changes to TlsCredentials for use outside of xDS.

A87-mtls-spiffe-support.md
```diff
@@ -0,0 +1,188 @@
...
+`CertificateProvider` to support the new SPIFFE trust bundle map.
+
+#### Go
+
+TODO: @erm-g to fill this in.
+
+### Temporary environment variable protection
+
+The xDS functionality will be guarded via the
+`GRPC_EXPERIMENTAL_XDS_MTLS_SPIFFE` environment variable.  The new
```
ejona86, line 167: Ugh, `GRPC_EXPERIMENTAL_SPIFFE_TRUST_BUNDLE_MAP` was used in Java. I think that may be because the initial implementation pre-dates this gRFC.

A87-mtls-spiffe-support.md
markdroth: Okay.  Do we not plan to support this functionality with TlsCreds in Java?

A87-mtls-spiffe-support.md
markdroth: I'd prefer to stick with the same env var naming format that we've used for other features.

Can we fix the Java impl?  I don't think we have the interop tests done yet, so this shouldn't break anything to change.

A87-mtls-spiffe-support.md
ejona86: @gtcooke94, we'll want to flesh this out a bit.

A87-mtls-spiffe-support.md
ejona86: Yeah, we can change it for Java.

gtcooke94: @ejona86 do you want to just make this change or should I handle it?

A87-mtls-spiffe-support.md
gtcooke94: How would you like to see this fleshed out with regards to the work Andrey did?

A87-mtls-spiffe-support.md
```diff
@@ -0,0 +1,260 @@
...
+public void updateTrustCredentials(File trustCertFile) throws IOException,
+   GeneralSecurityException
+
+public Closeable updateTrustCredentials(File trustCertFile, long period, TimeUnit unit,
+   ScheduledExecutorService executor) throws IOException, GeneralSecurityException
+```
+
+We will add three similar new functions for updating the SPIIFE root of trust
+```
+public void updateSpiffeTrustBundle(SpiffeBundle spiffeBundle) throws IOException,
```
ejona86, line 205: `SpiffeBundle` is internal API, so we shouldn't expose it here. Unless you were thinking we would make it public, in which case, I'd probably still argue against it since then we'd be signing up to support a full Spiffe parsing API.

I think instead we can just drop this method and not have it at all, or have an InputStream based version.

A87-mtls-spiffe-support.md
ejona86: I don't mind making it, but it'll be less likely to be lost if you do it.

A87-mtls-spiffe-support.md
ejona86: Sounds good.

A87-mtls-spiffe-support.md
gtcooke94: Our team will take it

gtcooke94 (approved):

ejona86 (approved):

A87-mtls-spiffe-support.md
```diff
@@ -0,0 +1,266 @@
...
+We will add support for SPIFFE certificate verification in mTLS.  This
+will be supported both with TlsCredentials and via xDS.
+
+## Background
+
+[SPIFFE] is a standardized way of encoding workload identity in mTLS
+certificates.  Specifically, we want to support using the [SPIFFE bundle
+format] in place of a single CA certificate for peer certificate
+verification.  The bundle map format is defined in
+https://github.com/spiffe/spiffe/pull/304.
```
dfawley, line 21: This is a PR, and the spec is already referenced from the previous sentence.  Delete this?  Or change to "is defined in" to "was added by" - but it seems irrelevant to me.

A87-mtls-spiffe-support.md
```diff
@@ -0,0 +1,266 @@
...
+
+[gRFC A29]: A29-xds-tls-security.md
+[SPIFFE]: https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE.md
+[SPIFFE ID format]: https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md#2-spiffe-identity
+[SPIFFE bundle format]: https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE_Trust_Domain_and_Bundle.md#4-spiffe-bundle-format
+[Publishing SPIFFE bundle format]: https://github.com/spiffe/spiffe/blob/main/standards/X509-SVID.md#61-publishing-spiffe-bundle-elements
+
+## Proposal
+
+There are two main parts to this proposal, support for SPIFFE
```
dfawley, line 34: Nit:

```suggestion
There are two main parts to this proposal: support for SPIFFE
```

A87-mtls-spiffe-support.md
```diff
@@ -0,0 +1,266 @@
...
+package](https://pkg.go.dev/github.com/spiffe/go-spiffe/v2) for parsing and
+object abstractions.  Currently, similarly to C++ above, Go already supports
+[file watcher credential
+providers](https://github.com/grpc/grpc-go/blob/e0d191d8adcdd73aad084154769404dd2f6b0fc6/credentials/tls/certprovider/pemfile/watcher.go#L92C4-L99)
+with xDS. We can further modify these providers, specifically the file watcher,
+to allow them to be configured with and return SPIFFE bundles instead of only
+certificates.
+Specifically, the Provider is designed to return a
+[KeyMaterial](https://github.com/grpc/grpc-go/blob/e0d191d8adcdd73aad084154769404dd2f6b0fc6/credentials/tls/certprovider/provider.go#L91-L97)
+struct where we could easily add a trust bundle.
```
dfawley, line 235: Can you be more specific about the actual change proposed to the cert provider?  Are we simply adding a field to `KeyMaterial`?  What is its name/type?

gtcooke94: Done, PTAL

dfawley (approved):

PR #465, opened 2024-11-11
A83 update: add details about filter state retention/sharing mechanism

markdroth:
CC @sergiitk

### DIFF
--- a/A83-xds-gcp-authn-filter.md
+++ b/A83-xds-gcp-authn-filter.md
@@ -4,7 +4,7 @@ A83: xDS GCP Authentication Filter
 * Approver: @ejona86, @dfawley
 * Status: {Draft, In Review, Ready for Implementation, Implemented}
 * Implemented in: <language, ...>
-* Last updated: 2024-10-25
+* Last updated: 2024-11-11
 * Discussion at: https://groups.google.com/g/grpc-io/c/76a0zWJChX4
 
 ## Abstract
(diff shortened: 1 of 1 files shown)


A83-xds-gcp-authn-filter.md
```diff
@@ -229,10 +236,57 @@ RouteConfiguration update, so that we don't wind up needlessly refetching
 tokens after the update.  Implementations should provide a mechanism for
 new instances of the filter to retain the cache from previous instances.
 
-If an LDS update changes the cache size, the filter must apply that
-change to the cache.  If the cache currently has more entries in it than
-the new cache size, then the least recently used entries will be removed
-to make the cache adhere to the new size limit.
+To address these requirements, we will implement a general-purpose
+mechanism for xDS HTTP filters to share state and retain that state
```
ejona86, line 240: We're not looking at anything for filters to share state in Java.

ejona86 (approved):

A83-xds-gcp-authn-filter.md
markdroth: Hmm, I guess there are some nuances here.  If I understand the plan for Java correctly, I think it would allow sharing state between different routes for the same filter instance, but not sharing state between filter instances, right?

This makes me wonder whether there will be cases in the future where we might want to share state across filter instances.  That's not a requirement right now, but the C-core mechanism would easily allow it if we needed it for some reason.  If we do encounter such a requirement in the future, how hard do you think it would be to do something like that in Java?  Are there changes we could make to the Java approach to do this, like maybe storing the state in a channel-level object instead of in each filter?  And would that change be any easier to do now, or would it be just as easy to change later, in which case there's no benefit in changing anything now?

markdroth: Okay, I'm content to leave sharing between filters as a possible future thing.  I've updated the wording accordingly.

ejona86 (approved):

dfawley (approved):

PR #466, opened 2024-11-22
A88: xDS Data Error Handling

markdroth:


A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,451 @@
...
+- `OnResourceChanged(StatusOr<ResourceType> resource)`: Will be invoked
+  to notify the watcher of a new version of the resource received from
+  the xDS server or an error indicating the reason why the resource cannot
+  be obtained.  Note that if an error is passed to this method after the
+  watcher has previously seen a valid resource, the watcher is expected
+  to stop using that previously delivered resource.  In this case, the
+  XdsClient will remove the resource from its cache, so that CSDS will
+  not report a resource that the client is not actually using.
+
+- `OnAmbientError(Status status)`: Will be invoked to notify the watcher
```
larry-safran, line 185: Naming the argument `error` would make it clearer.

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,451 @@
...
+
+When we receive an error for a given resource, we will do the following:
+1. Cancel the resource timer, if any.
+2. Set the cache entry's status to `RECEIVED_ERROR` and record the error,
+   so that that information can be reported via CSDS.
+3. If the status code is NOT_FOUND or PERMISSION_DENIED and the
+   "fail_on_data_errors" server feature is present in the bootstrap
+   config, delete the existing cached resource, if any.
+4. If there is an existing cached resource, call the watchers'
+   `OnAmbientError()` method with the error.  Otherwise, call th
```
larry-safran, line 253: s/th/the/

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,451 @@
...
+unconditionally use them if they are present.
+
+When we receive an error for a given resource, we will do the following:
+1. Cancel the resource timer, if any.
+2. Set the cache entry's status to `RECEIVED_ERROR` and record the error,
+   so that that information can be reported via CSDS.
+3. If the status code is NOT_FOUND or PERMISSION_DENIED and the
+   "fail_on_data_errors" server feature is present in the bootstrap
+   config, delete the existing cached resource, if any.
+4. If there is an existing cached resource, call the watchers'
```
larry-safran, line 252: For (3) you mention "fail_on_data_errors", but that isn't mentioned in (4), even though it would also affect whether or not to use `OnAmbientError`

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,451 @@
...
+`grpc.xds.cache_state` label, which is used for the
+`grpc.xds_client.resources` metric.  In keeping with the new CSDS client
+state introduced in [xRFC TP3], we will add two new possible values for
+this label:
+- `received_error`: The server sent an error for this resource.
+- `received_error_but_cached`: There is a version of this resource
+  cached, but the server subsequently sent an error for this resource.
+
+By default, the second value will be seen; if the `fail_on_data_errors`
+server feature is present, then the second value will be seen.
```
larry-safran, line 349: Both default and condition mention `second value`, I believe you intended default to be the first value.

A88-xds-data-error-handling.md
markdroth: Fixed.

A88-xds-data-error-handling.md
markdroth: We don't actually need to check it here.  I've updated the wording to make this clear.

A88-xds-data-error-handling.md
markdroth: Good catch, fixed.

A88-xds-data-error-handling.md
markdroth: I think it's clear that a status is an error, so I don't feel the need to change that.

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,529 @@
...
+LDS or CDS resource deletion from server | true | false | `OnResourceDoesNotExist()`, Drop existing resource and fail RPCs (watcher notification skipped if "ignore_resource_deletion" server feature is present) | `OnAmbientError(Status(NOT_FOUND))`, Ignore
+LDS or CDS resource deletion from server | true | true | `OnResourceDoesNotExist()`, Drop existing resources and fail RPCs (watcher notification skipped if "ignore_resource_deletion" server feature is present) | `OnResourceChanged(Status(NOT_FOUND))`, Drop existing resource and fail RPCs
+[xRFC TP3] error with status NOT_FOUND or PERMISSION_DENIED | false | (any) | N/A | `OnResourceChanged(status)`, Fail data plane RPCs
+[xRFC TP3] error with status NOT_FOUND or PERMISSION_DENIED | true | false | N/A | `OnAmbientError(status)`, Ignore
+[xRFC TP3] error with status NOT_FOUND or PERMISSION_DENIED | true | true | N/A | `OnResourceChanged(status)`, Drop existing resource and fail RPCs
+[xRFC TP3] error with other status | false | (any) | N/A | `OnResourceChanged(status)`, Fail data plane RPCs
+[xRFC TP3] error with other status | true | (any) | N/A | `OnAmbientError(status)`, Ignore
```
dfawley, line 390: These are N/A because we aren't currently setting `supports-resource-error-unwrapping` (which seems to be mentioned as almost an afterthought in TP3, in the "Rationale" section, oddly)?  This document should probably contain that string somewhere, right?

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,529 @@
...
+
+### Handling Ambient Errors
```
dfawley, line 392: Hmm, the organization of this doc feels a little bit off.

It feels like there are two big sections:

1. Tweaking the current behavior (watcher API changes & related)
2. Supporting xRFC TP3.

I wonder if we should add those as the `###` headers and make all these be `####`?

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,529 @@
...
+
+To support this, we will change pick_first and the petiole policies (see
+[gRFC A61]) to include the `resolution_note` field in all error messages
+generated by those policies.  The xds resolver will set the resolution
+note in the `XdsConfig` (see [gRFC A74]) to include ambient errors for
+all resources (LDS, RDS, CDS, and EDS).  If none of the resources
+have ambient errors, the xds resolver will set the resolution note to
+indicate the xDS node ID, which can be useful for debugging.
+
+### Propagating Status to Data Plane RPCs
```
dfawley, line 412: Maybe forward-ref this section above in at least one place where you talk about failing data plane RPCs?

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,529 @@
...
+is expected that UNAVAILABLE is an appropriate code for data plane RPCs.
+
+### Temporary environment variable protection
+
+The new functionality will be guarded by the
+`GRPC_EXPERIMENTAL_XDS_DATA_ERROR_HANDLING` env var until it passes
+interop testing.  This env var will guard using the new response fields
+from [xRFC TP3] and the new server features "fail_on_data_errors" and
+"resource_timer_is_transient_error".  It will also guard whether we
+default to ignoring resource deletion.
```
dfawley, line 430: The response fields won't be set unless we set `supports-resource-error-unwrapping`, though? So this would gate setting that client feature, not reading the new response fields?

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,529 @@
...
+As noted in [gRFC A54], the set of status codes that gRPC should generate
+for RPCs is intentionally limited.  In particular, INVALID_ARGUMENT and
+NOT_FOUND are two of the codes that are reserved for application use
+and should never be generated by gRPC itself.  Implementors should not
+blindly propagate status codes from the XdsClient watcher to data plane
+RPCs; instead, they should be careful to set status codes carefully on
+data plane RPCs on both the client and server side.  In most cases, it
+is expected that UNAVAILABLE is an appropriate code for data plane RPCs.
+
+### Temporary environment variable protection
```
dfawley, line 423: Should we have a section on the behavior of new watchers for resources that already have watchers?

E.g. do we replay ambient errors to them?

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,529 @@
...
+to consider the xDS server connection to have failed if we haven't gotten
+back a response in a certain amount of time, but we cannot do that while
+we are depending on the timer to detect individual resources not existing.
+
+As future work, we should consider whether there are ways to safely
+trigger xDS fallback if the primary server is responding too slowly.
+There are a number of challenges here:
+
+- Given the number of issues we've seen where the control plane has been
+  slow to respond, it's not clear to me that we would want all of those
```
dfawley, line 473: s/me/us/?  (You say "we" a lot, above)

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,529 @@
...
+  cases to trigger fallback, since we'd likely hit fallback way too often.
+  I suspect that we would first need the control plane to monitor and provide
```
dfawley, line 475: s/I/We/?   And one instance at the bottom, too.

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,529 @@
...
+  trigger fallback in this case.
+
+- As a middle ground, we could consider only having a timeout on the very
+  first response on the ADS stream to trigger fallback, or even the
+  very first response for each resource type.  However, this would
+  still leave us blind to cases where the control plane is slow when new
+  subscriptions are added later in the stream's lifetime, which we have
+  seen in the wild.
+
+- Even if we could identify the right timeout criteria to trigger fallback,
```
dfawley, line 505: Or on second thought, maybe this RFC isn't the right place for all of the details of these complications.  We could leave them in our discussion notes & in the PR description instead?

A88-xds-data-error-handling.md
markdroth: I agree that the organization isn't as smooth as I'd like.  Unfortunately, this is one of those designs where the various parts are somewhat intertwined, and there's a lot of cross-references between sections that makes it hard to figure out a fixed order for the implementation.  Just as one example, it doesn't really make sense to explain the "resource_timer_is_transient_error" server feature until after we describe the xRFC TP3 changes.

I'd welcome specific suggestions on how to improve the organization, but I don't think the two broad categories you mentioned are going to work.

A88-xds-data-error-handling.md
markdroth: I like this idea, but I don't see an obvious place where it makes sense to do so where it wouldn't seem cumbersome.

FWIW, this proposal isn't really changing anything from A54 at all here; the only reason I added this section was to make sure implementors keep this in mind and don't accidentally break it when implementing the new watcher API.

A88-xds-data-error-handling.md
markdroth: Good idea.  I added a short section above.

A88-xds-data-error-handling.md
markdroth: As mentioned above, xRFC TP3 does not actually require that client feature.

A88-xds-data-error-handling.md
markdroth: Good catch, fixed.

A88-xds-data-error-handling.md
markdroth: Done.

A88-xds-data-error-handling.md
dfawley: Aha, I see, sorry.  So it's fine for the management server to always send the errors back, and it's N/A here because we are not yet reading the new field.

markdroth: Yes, exactly.

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,538 @@
...
+operators when these errors occur.  To configure this, we will introduce a
+new server feature in the bootstrap config called "fail_on_data_errors",
+which indicates that we cannot depend on the control plane to properly
+alert human operators.
+
+This server feature affects all data errors that can occur when a
+resource is already cached, which are the following cases:
+- NACKs
+- resource deletions in LDS and CDS
+- [xRFC TP3] errors with status NOT_FOUND or PERMISSION_DENIED
```
dfawley, line 220: This (the paragraph below) is the main area that I was hoping to include a reference to the "Propagating status" section.  `NOT_FOUND` is mentioned here explicitly as a status that could result in us deleting the resource, which should pretty much have the potential to fail RPCs.  So a special call-out that the `NOT_FOUND` status should _not_ be propagated directly, and to see the below section for proper handling of the errors seems useful.  Maybe at the very end of that paragraph after it says to call the `OnResourceChanged` method?

markdroth: Done.

yannickepstein:
I don't have any constructive feedback, but as users of gRPC's xDS capabilities who rely on the `ignore_resource_deletion` flag for reliability, we welcome these changes. I particularly appreciate the flexibility to make the data plane resilient against control plane issues given that we prefer to detect data plane outages caused by service discovery within the control plane. This level of control is invaluable in a large-scale organization, where the majority of engineers rarely interact with xDS or service discovery directly.

ejona86 (approved):

A88-xds-data-error-handling.md
```diff
@@ -0,0 +1,565 @@
...
+[xRFC TP3]: https://github.com/cncf/xds/blob/main/proposals/TP3-xds-error-propagation.md
+
+## Proposal
+
+This proposal has the following parts:
+- changes in the XdsClient watcher APIs
+- handling data errors and transient errors
+- support for errors provided by the xDS server as per [xRFC TP3]
+- changes to the resource timer behavior
+- deprecating the "ignore_resource_deletion" server feature
```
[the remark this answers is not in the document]
markdroth: Changed "deprecating" to "removing".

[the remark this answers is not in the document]
ejona86: The header below still says deprecated... I guess it doesn't matter enough to fix.

PR #468, opened 2024-12-17
A90: Add List Method to gRPC Health Service

marcoshuck:
# Context
This proposal introduces a new `List` RPC endpoint for the Health service, allowing clients to retrieve the statuses of all monitored services. This feature simplifies integration with status-reporting dashboards and enhances observability for microservices.

# Change
This PR submits the proposal document for review and potential approval by the gRPC maintainers.

# Additional information
I started the implementation of the health service (including just the gRPC service endpoint spec) in this PR: https://github.com/grpc/grpc-proto/pull/143


A##-list-health.md
```diff
@@ -0,0 +1,59 @@
...
+* **Discussion at:** TBD
+
+
+## Abstract
+
+This proposal introduces a new `List` RPC endpoint for the Health service, allowing clients to retrieve the statuses of all monitored services or a filtered list of specific services. This feature simplifies integration with status-reporting dashboards and enhances observability for microservices.
+
+## Background
+
+The existing Health service provides basic health check functionality but lacks a mechanism to retrieve a comprehensive list of all monitored servicesand their statuses.
```
marcoshuck, line 17:
```suggestion
The existing Health service provides basic health check functionality but lacks a mechanism to retrieve a comprehensive list of all monitored services and their statuses. 
```

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
...
+
+This proposal introduces a new `List` RPC endpoint for the Health service, allowing clients to retrieve the statuses of all monitored services or a filtered list of specific services. This feature simplifies integration with status-reporting dashboards and enhances observability for microservices.
+
+## Background
+
+The existing Health service provides basic health check functionality but lacks a mechanism to retrieve a comprehensive list of all monitored servicesand their statuses. 
+
+This limitation makes it challenging for clients to aggregate health information across multiple services, particularly for use cases like publishing service statuses to dashboards such as [Cachet](https://cachethq.io/) or [Statuspage](https://www.atlassian.com/software/statuspage).
+
+Kubernetes provides a similar capability with its `/readyz?verbose` endpoint, which lists the status of all components.This proposal aims to bring analogous functionality to the Health service, providing a unified view of service health.
```
marcoshuck, line 21:
```suggestion
Kubernetes provides a similar capability with its `/readyz?verbose` endpoint, which lists the status of all components. This proposal aims to bring analogous functionality to the Health service, providing a unified view of service health.
```

marcoshuck:
@a11r care to give it a review whenever you can? :)

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
+# Add List endpoint to gRPC Health service
+
+* **Author(s):** Marcos Huck.
```
Falco20019, line 3: Very minor NIT
```suggestion
* **Author(s):** Marcos Huck
```

marcoshuck: Appreciate it :)

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Abstract
+
+This proposal introduces a new `List` RPC endpoint for the Health service, allowing clients to retrieve the statuses of all monitored services or a filtered list of specific services. This feature simplifies integration with status-reporting dashboards and enhances observability for microservices.
+
+## Background
+
+The existing Health service provides basic health check functionality but lacks a mechanism to retrieve a comprehensive list of all monitored services and their statuses. 
+
+This limitation makes it challenging for clients to aggregate health information across multiple services, particularly for use cases like publishing service statuses to dashboards such as [Cachet](https://cachethq.io/) or [Statuspage](https://www.atlassian.com/software/statuspage).
```
markdroth, line 19: Can you help me understand the use-case here?  Both of those publishing services seem to be aimed at communicating overall service health to end-users.  However, the gRPC health service is really about exposing the health of an individual endpoint, not of the service as a whole.  In what context would these be used together?

marcoshuck: Thanks for the review. 

I imagine the following use case: 

We need to aggregate these endpoint statuses to get overall service health. 

We'd build an aggregator service that polls these endpoints, applies logic to determine service status, and then feeds that consolidated data to tools like `Cachet` or `Statuspage` for end-user visibility, and also for internal monitoring. So, while `gRPC` is endpoint-focused, the Health service is a key data source for a broader service health system. 

Adding a new `List` endpoint would allow towards that end.

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
+# Add List endpoint to gRPC Health service
```
markdroth, line 1: It looks like the next unassigned number is A90.  Please rename this file to `A90-health-service-list-method.md` and change this line to the following:

```
A90: Add List Method to gRPC Health Service
----
```

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
...
+
+* **Author(s):** Marcos Huck
```
markdroth, line 3: Please list this as a github login, preceded by an `@`.

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
...
+* **Discussion at:** TBD
+
+
+## Abstract
+
+This proposal introduces a new `List` RPC endpoint for the Health service, allowing clients to retrieve the statuses of all monitored services or a filtered list of specific services. This feature simplifies integration with status-reporting dashboards and enhances observability for microservices.
+
+## Background
+
+The existing Health service provides basic health check functionality but lacks a mechanism to retrieve a comprehensive list of all monitored services and their statuses.
```
markdroth, line 17: One thing to keep in mind here is that the list of health services exposed by an application can change over the lifetime of the process.  New services can be added at any point, which means that the set of service names returned by the List method may change over the lifetime of the process.

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
...
+
+Adding the List RPC endpoint to the Health service strikes a balance between functionality and simplicity.
+
+### Alternative approaches
+- Separate Service: Extract the new functionality into a standalone service. This avoids altering the existing Health service API but increases complexity by introducing another service for clients to interact with.
+
+## Implementation
+
+1. Add the `List` RPC endpoint and the associated request/response messages in the Health service .proto file.
+2. Update client libraries to support the `List` endpoint. Provide examples demonstrating how to call the endpoint and handle its response.
```
markdroth, line 53: I think there's another important step before this one, which is to implement the new method in the health service implementations in each gRPC language.

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
...
+* **Approver:** TBD
+* **Status:** Draft  
+* **Implemented in:** TBD
+* **Last updated:** 2024-12-17  
+* **Discussion at:** TBD
+
+
+## Abstract
+
+This proposal introduces a new `List` RPC endpoint for the Health service, allowing clients to retrieve the statuses of all monitored services or a filtered list of specific services. This feature simplifies integration with status-reporting dashboards and enhances observability for microservices.
```
markdroth, line 13: What we're talking about adding here is a method, not an endpoint.  Please update terminology throughout.

A##-list-health.md
marcoshuck: That seems reasonable, might be worth adding a note.

A##-list-health.md
marcoshuck: Would you mind pointing me in the right direction? I'm happy to write code in pretty much all the languages supported by gRPC, but I just want to make sure what are those places I need to change.

A##-list-health.md
```diff
@@ -0,0 +1,59 @@
+# Add List endpoint to gRPC Health service
+
+* **Author(s):** Marcos Huck
+* **Approver:** TBD
+* **Status:** Draft  
+* **Implemented in:** TBD
+* **Last updated:** 2024-12-17  
+* **Discussion at:** TBD
```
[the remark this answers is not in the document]
marcoshuck: Done, will wait until it gets posted to update with the link.

A90-health-service-list-method.md
```diff
@@ -0,0 +1,84 @@
...
+Adding the List RPC method to the Health service strikes a balance between functionality and simplicity.
+
+### Alternative approaches
+- Separate Service: Extract the new functionality into a standalone service. This avoids altering the existing Health service API but increases complexity by introducing another service for clients to interact with.
+
+## Implementation
+
+1. Add the `List` RPC method and the associated request/response messages in the Health service `.proto` file.
+2. Update client libraries to support the `List` method. Provide examples demonstrating how to call the method and handle its response.
+3. Implement `List` RPC method in the respective gRPC languages.
```
ejona86, line 78: Not too important: You probably want to implement it on the server (step 3) before talking about client-side (step 2). So you may reorder these two. You may also remove step 4, as that should really be done in steps 2 and 3.

ejona86 (commented):
Seems fair. I could bikeshed, but I don't have any real concerns. :shipit:

A90-health-service-list-method.md
```diff
@@ -0,0 +1,84 @@
...
+  repeated string services = 1; // Contains the name of the services that should be obtained. If empty, must return all.
+}
+
+message HealthListResponse {
+  map<string, HealthCheckResponse> statuses = 1; // Contains all the services and their respective status.
+}
+
+
+service Health {
+  // List gets the health of all the available services. A filtered list can
```
ejona86, line 50: We should say something about it not guaranteeing an atomic snapshot. Tearing shouldn't cause trouble for clients, so I don't see any reason to avoid it and allow more efficient implementations. For example, grpc-java uses a ConcurrentHashMap, which can be iterated over and read without locking. But adds/removes while iterating may or may not show up in the iteration, and there would be no snapshotting of statuses between services.

A90-health-service-list-method.md
marcoshuck: Good call, let me swap the order and remove step 4.

A90-health-service-list-method.md
marcoshuck: What about this?
```suggestion
  // List provides a non-atomic snapshot with the health of all the available services. A filtered list can
```

ejona86: SGTM

A##-list-health.md
murgatroid99: In Node the health service is implemented here: https://github.com/grpc/grpc-node/blob/master/packages/grpc-health-check/src/health.ts#L82. The `List` method would need to be added. It should mostly just be straightforwardly iterating through the `statusMap`.

A90-health-service-list-method.md
```diff
@@ -0,0 +1,82 @@
...
+
+- Retrieve the health statuses of all monitored services or a filtered list of specified services.
+- Ensure the implementation is idempotent and side-effect free.
+- Provide a clear schema for the request and response to facilitate integration with external tools.
+
+### Proposed API Changes
+
+```protobuf
+message HealthListRequest {
+  repeated string services = 1; // Contains the name of the services that should be obtained. If empty, must return all.
```
dfawley, line 41: This field doesn't feel very `List`y - it's basically a bulk `Check` operation when this is set.  How important is that use case, vs. calling `Check` repeatedly?

A90-health-service-list-method.md
```diff
@@ -0,0 +1,82 @@
...
+}
+
+message HealthListResponse {
+  map<string, HealthCheckResponse> statuses = 1; // Contains all the services and their respective status.
```
dfawley, line 45: Should there be any concern about the length of this in the result?  I know we don't expect it to be long.  But it seems at the very least, a generic gRPC-provided implementation of the service should have some kind of safeguard here.  E.g. restrict it to 100 entries, error if there are too many, etc.

I suppose we can always add such things later if it turns out to be a problem, but I thought I'd call it out now.

markdroth: We talked about this internally (although I think you may have been out that day).  In general, we expect the number of services to be fairly small here, so it doesn't seem worth adding the complexity of supporting pagination here.  If we ever run into a problem where the size is too large here, we can discuss adding something like that at that point.

A90-health-service-list-method.md
```diff
@@ -0,0 +1,82 @@
...
+
+This proposal introduces a new `List` RPC method for the Health service, allowing clients to retrieve the statuses of all monitored services or a filtered list of specific services. This feature simplifies integration with status-reporting dashboards and enhances observability for microservices.
+
+## Background
+
+The existing Health service provides basic health check functionality but lacks a mechanism to retrieve a comprehensive list of all monitored services and their statuses.
+
+This limitation makes it challenging for clients to aggregate health information across multiple services, particularly for use cases like publishing service statuses to dashboards such as [Cachet](https://cachethq.io/) or [Statuspage](https://www.atlassian.com/software/statuspage).
+
+It's important to keep in mind that the list of health services exposed by an application can change over the lifetime of the process.
```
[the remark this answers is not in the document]
markdroth: I agree, it would probably be good to note this in a comment on the new method.  It's not really an implementation concern, but it is something that callers need to know: they are not guaranteed that the same set of services will be returned on subsequent calls to the method.

A90-health-service-list-method.md
marcoshuck: You're right on all accounts, the same result can be achieved by either calling `Check` repeatedly or simply sending an array of services to check in this method. I'm more than happy to remove this field or keep it and end up implementing it as some sort of `filter`.

A##-list-health.md
marcoshuck: Thank you all for the guidance.

A90-health-service-list-method.md
```diff
@@ -0,0 +1,82 @@
...
+* **Implemented in:** TBD
+* **Last updated:** 2025-02-22  
+* **Discussion at:** https://groups.google.com/g/grpc-io/c/tEI5G9sX0zc
+
+
+## Abstract
+
+This proposal introduces a new `List` RPC method for the Health service, allowing clients to retrieve the statuses of all monitored services or a filtered list of specific services. This feature simplifies integration with status-reporting dashboards and enhances observability for microservices.
+
+## Background
```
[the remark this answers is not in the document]
marcoshuck: Excellent, thanks for raising this. Let's add both references.

A90-health-service-list-method.md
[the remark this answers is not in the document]
marcoshuck: Good call, let's add it as a comment in the new method.

A90-health-service-list-method.md
marcoshuck: I'd be more than happy to help adding pagination if we ever face a problem like this.

A90-health-service-list-method.md
dfawley: If you don't need it, I think I'd prefer to remove it.  Or rename to `filter` if you do need it.

A90-health-service-list-method.md
dfawley: I think if I'm reviewing an implementation of the service, I'd expect us to have some kind of safeguard in there.  E.g. if there are more than 100 entries in the service, it would either return an error or truncate the response.

I don't feel too strongly about exactly what the behavior should be, but I do think there should be something in place, ideally specified or at least mentioned somewhere in this gRFC.

I disagree with the philosophy of "we don't think it will be a problem, so we'll just let it be unbounded until someone OOMs, and then we'll fix it".  I totally agree that it's not likely to be a problem for anyone, ever.  But adding a limiter now only takes 5 extra minutes just in case.

A90-health-service-list-method.md
marcoshuck: We can add it later on if we find a specific use case, I was just doing some premature optimization by adding filter capabilities.

A90-health-service-list-method.md
larry-safran: If you are going to put in a limit, 1000 would be much better than 100.  Even though 100 seems like an unreasonably large number, someone going crazy with microservices might hit it.  If the concern is OOM, then even a million would be okay.

dfawley: A million starts to get questionable.  I'd be fine with a limit field and allowing it to go up to 10k or so.

I'd also be fine with hard-coding a limit of something like 10k for now to make it simple.  We can add a limit field later if we need it.

Like I said, I don't feel strongly about how we do it, just that we should have _something_ in place.  I'd _prefer_ it be something very lightweight and easy to implement, like a hard-coded 10k limit, but I won't fight if others prefer something more complicated.

dfawley: We met and discussed this and agreed that erroring with RESOURCE_EXHAUSED if there are more than 100 results sounds like the most reasonable.  We can add a limit field, etc in the future if the need ever arises, though nobody seems to think it will.

A90-health-service-list-method.md
```diff
@@ -1,9 +1,9 @@
-# A90: Add List Method to gRPC Health Service
+A90: Add List Method to gRPC Health Service
 ----
 * **Author(s):** @marcoshuck
 * **Approver:** @markdroth
-* **Status:** In Review  
-* **Implemented in:** TBD
+* **Status:** Final
+* **Implemented in:** C++, Go, Java and Node.
```
markdroth, line 6: It isn't actually implemented in these languages yet, so they shouldn't be listed here.  For now, you can leave this field empty.

A90-health-service-list-method.md
```diff
@@ -24,21 +24,20 @@ Kubernetes provides a similar capability with its `/readyz?verbose` endpoint, wh
 
 ### Related Proposals
 
-N/A
+- [A17 - Client-Side Health Checking](https://github.com/grpc/proposal/blob/master/A17-client-side-health-checking.md)
```
markdroth, line 27: This can be a relative link, since it's the same directory as the thing it's pointing to -- i.e., the link target just needs to be "A17-client-side-health-checking.md".

markdroth (commented):
Just a couple of minor nits remaining!

A90-health-service-list-method.md
marcoshuck: You're right on all accounts, I forgot about relative references in markdown for some reason, thank you!

A90-health-service-list-method.md
marcoshuck: Ah, hm, yes, that's exactly what I missed from above the guidelines:

> If there is consensus as deemed by the APPROVER during the comment period, the APPROVER will mark the proposal as final and assign it a gRFC number. Once this is assigned (as part of the closure of discussion), the OWNER will update the state of the PR as final and submit the PR. Commits must not be squashed; the commit history serves as a log of changes made to the proposal.

Let's remove the language list.

marcoshuck:
```suggestion
* **Implemented in:** -
```

A90-health-service-list-method.md
marcoshuck:
```suggestion
- [A17 - Client-Side Health Checking](A17-client-side-health-checking.md)
```

A90-health-service-list-method.md
```diff
@@ -0,0 +1,85 @@
+A90: Add List Method to gRPC Health Service
+----
+* **Author(s):** @marcoshuck
+* **Approver:** @markdroth
+* **Status:** Final
+* **Implemented in:** -
+* **Last updated:** 2025-02-22
```
marcoshuck, line 7:
```suggestion
* **Last updated:** 2025-03-10
```

A90-health-service-list-method.md
```diff
@@ -0,0 +1,85 @@
...
+  // error.
+  //
+  // Clients should set a deadline when calling List, and can declare the
+  // server unhealthy if they do not receive a timely response.
+  //
+  // Clients should keep in mind that the list of health services exposed by an application 
+  // can change over the lifetime of the process.
+  //
+  // List implementations should be idempotent and side effect free.
+  rpc List(HealthListRequest) returns (HealthListResponse);
```
dfawley, line 61: Nit: please fix the line wrapping to be consistent at 80 columns.

dfawley (approved):

PR #469, opened 2025-01-03
Add gRPC fork support proposal

eugeneo:


drfloob (commented):
There are no public API or platform changes being made here, the fork support commitment is unchanged. I don't think a gRFC is the right forum for discussion on this implementation.

PR #471, opened 2025-01-10
A89: Backend Service Metric Label

ejona86:


A89-xds-cluster-metric-label.md
```diff
@@ -0,0 +1,75 @@
...
+## Proposal
+
+Each pick in the `xds_clunter_impl` policy, `xds_cluster_impl` will add the
+optional label `grpc.xds.cluster` to the call attempt tracer. The value will be
+copied from `xds_clunter_impl`'s service config `cluster` key. This is done
+regardless of the pick's result. It is possible for later picks for the same RPC
+to have a different value. This is the case for locality as well, and the last
+pick's value should be used.
+
+The `grpc.xds.cluster` label will be available on the following per-call
```
ejona86, line 48: @yashykt, [Java hard-codes](https://github.com/grpc/grpc-java/blob/d65d3942e6078c75500b034155a58b3fa6d516e8/opentelemetry/src/main/java/io/grpc/opentelemetry/OpenTelemetryMetricsModule.java#L209) support for locality today. I remember there being a conversation about supporting any optional labels, and I think we had kicked that can down the road.

I'm envisioning just adding another hard-coded case to the otel module for this.

A89-xds-cluster-metric-label.md
```diff
@@ -0,0 +1,75 @@
...
+### Related Proposals:
+* [gRFC A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient][gRFC A78]
+* [gRFC A79: Non-per-call Metrics Architecture][gRFC A79]
+
+[gRFC A78]: A78-grpc-metrics-wrr-pf-xds.md#per-call-metrics
+[gRFC A79]: A79-non-per-call-metrics-architecture.md
+
+## Proposal
+
+Each pick in the `xds_clunter_impl` policy, `xds_cluster_impl` will add the
```
yashykt, line 41: s/xds_clunter_impl/xds_cluster_impl

A89-xds-cluster-metric-label.md
```diff
@@ -0,0 +1,75 @@
...
+optional label `grpc.xds.cluster` to the call attempt tracer. The value will be
+copied from `xds_clunter_impl`'s service config `cluster` key. This is done
```
yashykt, line 43: s/xds_clunter_impl/xds_cluster_impl

A89-xds-cluster-metric-label.md
```diff
@@ -0,0 +1,75 @@
...
+API added in [gRFC A79][gRFC A79].
+
+Similar to how locality metrics are useful for analyzing _where_ traffic is
+being routed, the xDS cluster is useful for knowing _to whom_ it is being
+routed. `grpc.target` is generally all that's necessary to know which service is
+receiving traffic, but non-deterministic routing in xDS like weighted clusters,
+aggregate clusters, and cluster specifier plugins mean different clusters (and
+thus potentially different services or service versions) would comingle metrics
+unless the selected cluster is added as a label. It can also be helpful to know
+the selected cluster to confirm that deterministic routing like path matching is
```
larry-safran, line 29: Should have commas around "like path matching"

yashykt (approved):

A89-xds-cluster-metric-label.md
```diff
@@ -0,0 +1,75 @@
...
+clusters as multiple clusters are involved. For placing in a label, there are
+two potential choices: the top-level aggregate cluster and the leaf cluster.
+Using the leaf cluster seems to provide the most insight when using aggregate
+clusters as failing over to a different priority would be significant. If the
+top-level cluster is needed in the future, it can be added as well.
+
+## Implementation
+
+@ejona86 will immediately implement in gRPC Java. Other languages will follow as
+able. The implementation is very quick.
```
larry-safran, line 75: "The implementation is very quick" is more rationale than blueprint.

markdroth (approved):
5 stars, would read again. :)

A89-xds-cluster-metric-label.md
dfawley: I'm not sure that kind of justification is needed for this document.  It feels a little out of place in a specification intended for long-term use.  But I wouldn't block the PR on it if you feel like it's needed.

dfawley (approved):
Don't forget to wait for the requisite two week comment period before merging, and update "Status:".

A89-xds-cluster-metric-label.md
```diff
@@ -61,6 +61,8 @@ variable protection is unnecessary.
 gRFC A78 added `grpc.lb.locality` to per-call and WRR metrics, while this is
 only adding the new label to per-call. Cluster is an xDS-specific concept, so
 it is more awkward to add to WRR and that is left as potential future work.
+FIXME: the name was changed so is no longer an xds-specific concept and there is
```
markdroth, line 64: I'd be fine also adding this to the WRR metrics while we're at it.

Probably not worth adding it to the PF metrics, since @yashykt will presumably be working on a gRFC to replace the PF metrics with the new subchannel metrics that we discussed today.

ejona86: Yeah, I saw that locality was only WRR, and then I was wondering why locality wasn't part of pick-first metrics and vice-versa for this for WRR. I'll work with Yash to get the dust to settle, but overall I'm not worried.

markdroth: There wasn't any real reason not to add the locality label to PF metrics; it was just that no one had asked for it.

larry-safran (approved):

A89-backend-service-metric-label.md
```diff
@@ -0,0 +1,113 @@
...
+Label definition:
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.lb.backend_service | optional | The backend service to which the traffic is being sent. This is relevant when a single channel target can be sent to different sets of servers. When using xDS, this will be the cluster name. When not relevant, the value will be the empty string. |
+
+The value will be communicated to the gRPC OpenTelemetry module by the call
+attempt tracer. When an LB policy provides the label value to the tracer it
+will do so each pick that the information is available, regardless of the pick's
+result. This allows DEADLINE_EXCEEDED and UNAVAILABLE failures to include a
```
markdroth, line 59: I assume this is talking solely about including this information in metrics, right?  Do we already include the locality label for DEADLINE_EXCEEDED and UNAVAILABLE?  @yashykt

ejona86: Yes, I was meaning in metrics. The current locality implementation _can't_ include this information in erring metrics because xds_cluster_impl is what does the `addOptionalLabel()` and it only knows xds_wrr_locality's locality decision by looking at the subchannel of a successful pick.

I had some discussion with Yash yesterday about some edge cases and improving things post-A75, just to see how he felt. I was going to bring them up in our Friday meeting. (I just invited Yash to it.)

(Preview: Post-A75, have this logic in the leaf cds policy (the first policy under priority) so we get labels for outlier detection failures. Similarly, we could move the locality handling to weighted_target, or a new policy between xds_wrr_locality/weighted_target and the endpoint-picking policy, to add the locality label for failed picks. The question is mostly "do we care.")

A89-backend-service-metric-label.md
```diff
@@ -0,0 +1,113 @@
...
+relevant backend service. It is possible for later picks for the same RPC to
+have a different value. This is the case for locality as well, and the last
+pick's value should be used by the OpenTelemetry module.
+
+The only LB policy with a backend service concept at this time is
+`xds_cluster_impl`. It will notify the call attempt tracer of the
```
ejona86, line 65: TODO: Move to cds post-A75

ejona86:
@markdroth, I'd appreciate a quick look at the A75-mentioning commit, just to keep me honest.

A89-backend-service-metric-label.md
```diff
@@ -93,6 +94,14 @@ isn't the best of names, but all other names seemed to be obviously worse (e.g.,
 cluster) or no better. It is more important that the name not be misunderstood
 than for the name to be meaningful.
 
+Prior to A75 the only LB policy with a backend service concept is
```
markdroth, line 97: I'd prefer the language to be a little stronger here to make sure that we don't accidentally cause more work for implementors.

If this gRFC gets implemented before A75, then the code will have to be moved from the xds_cluster_impl policy to the cds policy anyway in order to implement A91, so it doesn't really matter whether that move is described here or in A91.  However, if A75 gets implemented first, then someone could read this without knowing about A91, and they may think that there's no reason not to put it in the xds_cluster_impl policy, which would result in them later having to move it to the cds policy.

To avoid that, I suggest saying something like this:

"""
If this gRFC is implemented prior to A75, then the `grpc.lb.backend_service` label will be added in the `xds_cluster_impl` policy.  However, if A75 is implemented first, then the label will be added in the `cds` LB policy instead.  This will pave the way for supporting this label on the outlier detection metrics described in A91.
"""

Note that the current text in A91 doesn't say anything about moving this functionality.  It just contains the following description for this label:

"""
The backend service to which the traffic is being sent, as described in A89. Note that this label will be supported only if A75 has already been implemented.
"""

ejona86: Some things:
1. This is in rationale section. I'm saying why the above was chosen. Up above is where I actually said it will always be done a singular way.
2. That's not what we had agreed in our meeting. We had agreed that it didn't really matter one way or another.
3. I'm very surprised you are recommending different languages do it differently
4. Outlier detection metrics in A91 are unrelated to the label here, because they aren't per-RPC. The plumbing is different for them. They are not helped nor harmed by the plumbing here

For (4), cluster_resover_impl is copying the value from `cluster` in its service config. We'll need to develop a new attribute to pass down or some such, because I don't think outlier detection knows the backend service today.

markdroth: My recollection is that we did agree on what I'm describing.  It looks like we didn't write down anything in our internal notes doc, but I wrote the following comment on the PR for A91 on the day we discussed it (March 7th):

https://github.com/grpc/proposal/pull/478#discussion_r1985862621

I'm not sure whether we miscommunicated or whether you're remembering wrong, but I can at least rule out the possibility that I'm misremembering here. :)  But in any case, regardless of what we discussed in the past, let's focus on figuring out how to move forward from first principles.

I think the end state that we want in all languages is for the outlier detection metrics to support the backend service label.  This gRFC already proposes a non-xDS-specific attribute to be passed down through the LB policy tree to indicate the value, and it says that we will use that attribute to add this optional label to the WRR metrics.  I think we want to use that same attribute in the same way for the outlier detection metrics -- i.e., the OD policy will interact with the attribute exactly the same way as the WRR policy does.  The only catch is that we need to set the attribute above the OD policy instead of below it.

The OD policy is above the xds_cluster_impl policy, so setting the attribute in xds_cluster_impl won't work for the OD metrics; instead, we'll need to set it in the cds policy, which is above the OD policy.  However, setting it in the cds policy won't work prior to A75, because in the aggregate cluster case, a single cds policy may have a separate OD child policy for each underlying cluster, and there's no reasonable way to set the attribute separately for each OD child policy.  This means that we can't support the backend metric label on the OD metrics prior to A75, but it also means that even with A75, we need to set the backend metric attribute in the cds policy instead of the xds_cluster_impl policy.

I'm not recommending that different languages do this differently in the long run; I think the goal is to eventually get all languages to the same place, which is that A75, this gRFC, and A91 are all implemented, and the backend metric label is set in the cds policy instead of the xds_cluster_impl policy.  I think the only question here is how to get to that end state in each language while minimizing the wasted work.

If a language implements this gRFC before A75 (which seems likely to be the case in Java), then they have no choice but to initially set the backend service label in the xds_cluster_impl policy, and then they will need to do additional work later to move that code from the xds_cluster_impl policy to the cds policy in order to implement A91.  However, if a language implements A75 before this gRFC (which is the case in C-core), then when we implement this gRFC, we should prefer to set the attribute in the cds policy to begin with, so that we don't have to move it later.

In other words, the thing that I'm trying to avoid here is that we hand this gRFC to a junior engineer in C-core, and they wind up implementing it in the xds_cluster_impl policy instead of the cds policy, thus requiring us to do more work later to move it to the cds policy.  We should instead just do it in the right place to begin with, and to ensure that that happens, this gRFC should say that.

I agree that the change I'm suggesting here is not just restricted to the rationale section.  What I'm suggesting is that this gRFC should not say that the label will be set in the xds_cluster_impl policy; instead, it should say that the label will be set in the xds_cluster_impl LB policy prior to A75 and in the cds policy after A75.

ejona86: > This gRFC already proposes a non-xDS-specific attribute to be passed down through the LB policy tree to indicate the value, and it says that we will use that attribute to add this optional label to the WRR metrics.

I forgot that change happened, and that's not been implemented anywhere. I was thinking about just the per-RPC metric plumbing "The LB policy `xds_cluster_impl` will notify the call attempt tracer of the `grpc.lb.backend_service` label. The value will be copied from `xds_cluster_impl`'s service config `cluster` key."

Okay, then that does need to move up to cds. But it should move up to cds when A75 is implemented or when A91 is implemented. Not dependent on "whichever A75 and A89 is implemented first, until A91 is implemented." The only pain with moving it is tests; the production code should be ~2 lines.

There was no hope for me to read that comment of yours on A91; it was lost in 19 comments and marked resolved.

I'll update this to say the implementation depends on A75, and add a note to A75 to say that this change needs to be done if A89 has been implemented.

A89-backend-service-metric-label.md
```diff
@@ -0,0 +1,131 @@
...
+being routed, the xDS cluster is useful for knowing _to whom_ it is being
+routed. `grpc.target` is generally all that's necessary to know which service is
+receiving traffic, but non-deterministic routing in xDS like weighted clusters,
+aggregate clusters, and cluster specifier plugins mean different clusters (and
+thus potentially different services or service versions) would comingle metrics
+unless the selected cluster is added as a label. It can also be helpful to know
+the selected cluster to confirm that deterministic routing, like path matching,
+is behaving as expected.
+
+### Related Proposals:
```
markdroth, line 33: We should probably also reference A66 here, since that's where the per-call metrics are defined.

A89-backend-service-metric-label.md
```diff
@@ -0,0 +1,131 @@
...
+[gRFC A78]: A78-grpc-metrics-wrr-pf-xds.md#per-call-metrics
+[gRFC A79]: A79-non-per-call-metrics-architecture.md
+[gRFC A75]: A75-xds-aggregate-cluster-behavior-fixes.md
+
+## Proposal
+
+### Add grpc.lb.backend_service to per-call metrics
+
+We will add the `grpc.lb.backend_service` optional label to the following
+per-call metrics:
```
markdroth, line 47: Suggest saying "defined in A66" here.

A89-backend-service-metric-label.md
```diff
@@ -0,0 +1,131 @@
...
+be copied from `xds_cluster_impl`'s service config `cluster` key.
+
+If A75 has been implemented, the LB policy `cds` for non-aggregate clusters will
+notify the call attempt tracer of the `grpc.lb.backend_service` label. The value
+will be copied from `cds`'s service config `cluster` key.
+
+### Add grpc.lb.backend_service to WRR metrics
+
+We will add the `grpc.lb.backend_service` optional label, with the same
+definition as for per-RPC metrics, to all existing WRR metrics:
```
markdroth, line 77: Suggest adding "defined in A78".

markdroth (approved):
This looks great!

Remaining comments are purely cosmetic; feel free to merge after addressing.

PR #472, opened 2025-02-03
[fallback] Remove env var mention

eugeneo:


ejona86 (commented):
NACK. All new implementations should use the environment variable until they are properly tested.

You should say that the environment variable is "until interop tests pass" or some such.

ejona86 (changes requested):
Making the NACK official.

ejona86 (approved):
If you look at the template, it says to define the testing criteria for the env variable removal ("until they have passed some testing criteria, which should also be detailed here"). But this is more correct than what was there before (it'll be env protected forever!).

PR #474, opened 2025-02-06
A72 update: various clarifications and corrections

yashykt:
Changes - 
* Add link to OpenCensus's binary format for grpc-trace-bin
* Add names for parent call span, attempt span and server span based on existing OpenCensus implementation.
* Clarify `previous-rpc-attempts` does not include transparent retries as per https://github.com/grpc/proposal/blob/master/A45-retry-stats.md
* Add missing `sequence-number` attribute from events in examples.
* Remove superfluous and confusing `sent` and `received` words from `Inbound message received` and `Outbound message sent`. Reasoning - `Inbound` and `Outbound` already convey the direction. The way it is currently structured, you could have `Outbound message sent` followed by `Outbound message compressed` , and `Inbound compressed message` followed by `Inbound message received`. In the future, we might want additional events (especially for C++) for when we know that bytes have actually been sent on the wire.

### DIFF
--- a/A72-open-telemetry-tracing.md
+++ b/A72-open-telemetry-tracing.md
@@ -34,9 +34,9 @@ design considerations about creating a child span for each individual call attem
 A gRPC server uses span ID from the incoming request header as a parent span to 
 maintain the parent child span relationship with the gRPC client. To propagate 
 span context over the wire, gRPC uses metadata (header name: `grpc-trace-bin`) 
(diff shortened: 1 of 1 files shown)


yashykt:
cc @markdroth @dfawley

YifeiZhuang (approved):
LGTM

dfawley (approved):

PR #475, opened 2025-02-14
A61 update: fix ring_hash proactive connection attempt logic

markdroth:
Fix logic for ring_hash proactive connection attempts to trigger even if the state update was not triggered by an endpoint entering TRANSIENT_FAILURE state.  This addresses two edge cases spotted by @arjan-bal:

1. There are four endpoints in the following states: TF, TF, READY, and IDLE. If the READY endpoint fails, it transitions to IDLE, resulting in the new states: TF, TF, IDLE, IDLE.
2. There are four endpoints in the following states: TF, TF, CONNECTING, and IDLE. If the CONNECTING endpoint is removed, the new states become: TF, TF, IDLE.

### DIFF
--- a/A61-IPv4-IPv6-dualstack-backends.md
+++ b/A61-IPv4-IPv6-dualstack-backends.md
@@ -4,7 +4,7 @@ A61: IPv4 and IPv6 Dualstack Backend Support
 * Approver: @ejona86, @dfawley
 * Status: Ready for Implementation
 * Implemented in: C-core
-* Last updated: 2023-12-15
+* Last updated: 2025-02-13
 * Discussion at: https://groups.google.com/g/grpc-io/c/VjORlKP97cE/m/ihqyN32TAQAJ
 
 ## Abstract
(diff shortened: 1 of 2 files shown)


A61-IPv4-IPv6-dualstack-backends.md
```diff
@@ -549,25 +549,57 @@ for (i = 0; i < ring.size(); ++i) {
...
+a small number of them are, and it would never try any of the others,
+thus never recovering from that incorrect assumption.  To work around
+this, when the aggregated connectivity state is either TRANSIENT_FAILURE
+or CONNECTING, the ring_hash policy proactively triggers connection
+attempts across all of the subchannels, even without seeing any picks.
+This ensures that if any of the subchannels become reachable, ring_hash
+will eventually report READY, and the priority policy will switch back
+to using the ring_hash child.
+
+The current logic for these connection attempts tries very hard to avoid
```
dfawley, line 573:
```suggestion
The logic proposed in [A42] for these connection attempts tries very hard to avoid
```

dfawley (approved):

A61-IPv4-IPv6-dualstack-backends.md
markdroth: Done, with slightly different wording.

ejona86 (approved):
I do think talking so much about priority makes things overly complicated. CONNECTING has always implied "the LB policy is connecting." And since we adopted sticky-TF everywhere, TF also implies "the LB policy is connecting (with backoff)". But I understand why priority is being discussed; it just is unfortunate it is being discussed so much.

PR #477, opened 2025-03-01
A61 update: remove special case for LOGICAL_DNS clusters

markdroth:


A61-IPv4-IPv6-dualstack-backends.md
```diff
@@ -791,6 +792,19 @@ def Pick(pick_args):
...
 ```
 
+#### Changes to Logical DNS Clusters
+
+Currently, as per [gRFC A37][A37], we handle LOGICAL_DNS clusters by
+hard-coding the cluster's LB policy to pick_first, so that we get the
+proper semantics of connecting to only one address for the cluster at a
+time.  However, now that we have the ability to encode multiple
+addresses per endpoint, we can do away with that special case.  Instead,
+we will encode all addresses returned by the DNS resolver as a single
```
ejona86, line 802: "we will encode". Can you specify which component will do that? Since the override was previously done in cluster_resolver, were you thinking that'll be done in the same place? Thus cluster_resolver pre-A74? Post-A74 cluster_resolver logic moves to the xds dependency manager and cds lb, and technically it could be done either place. I imagine xds dependency manager is nicer, though.

markdroth: Yes, it would be done in xds_cluster_resolver pre-A74, and I've updated it to say that.

This gRFC predates A74, and A74 basically says that everything that was done in xds_cluster_resolver will move to XdsDependencyManager, so I think that should be clear implicitly.

ejona86 (approved):
Note for posterity: it is safe to combine DNS results into a single endpoint because their attributes will be the same. Even with SRV records, since A26, DNS doesn't return the grpclb addresses mixed with the regular addresses.

dfawley (approved):

markdroth:
@murgatroid99 FYI.

PR #478, opened 2025-03-04
A87-grpc-metrics-xds-outlier-detection.md

PardhuKonakanchi:
gRFC for gRPC metrics for xDS Outlier Detection

### DIFF
--- a/A87-grpc-metrics-xds-outlier-detection.md
+++ b/A87-grpc-metrics-xds-outlier-detection.md
@@ -29,7 +29,7 @@ Outlier Detection metrics will have the following labels:
 
 | Name        | Disposition | Description |
 | ----------- | ----------- | ----------- |
-| grpc.target | required | Indicates the target we are outlier detection on. (Same as the attribute defined in A66.) |
+| grpc.target | required | Indicates the target we are running outlier detection on. (Same as the attribute defined in [A66](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md).) |
 
 The following metrics will be exported:
 
--- a/A87-grpc-metrics-xds-outlier-detection.md
+++ b/A87-grpc-metrics-xds-outlier-detection.md
@@ -1,0 +1,64 @@
+A87: gRPC OTel Metrics for xDS Outlier Detection
+---
+* Author(s): Pardhu Konakanchi, Sai Teja
+* Approver: a11r
+* Status: Draft
+* Implemented in: Core, Go
+* Last updated: 03-04-25
+
+## Abstract
+
+This document proposes some new metrics that will be added in gRPC for xDS Client Outlier Detection
+
+## Background
+
+A50: gRPC xDS Outlier Detection Support is a spec for gRPC to support xDS Outlier Detection. The current implementation only offers debug and trace logging in terms of visibility, which can be insufficient for understanding and diagnosing decision making in large-scale production systems. 
+
+Using A79: Non-per-call Metrics Architecture, it is possible to add granular metrics to make visibility into outlier detection easy for service owners utilizing gRPC.
+
+### Related proposals: 
+* A50: gRPC xDS Outlier Detection Support
+* A66: OpenTelemetry Metrics
+* A79: Non-per-call Metrics Architecture
+
+## Proposal
+
+A79’s non-per-call metrics architecture fits perfectly into the metrics reporting solution. We propose to aim for parity where appropriate with Envoy’s xDS implementation for metrics to be collected in gRPC
+
+Outlier Detection metrics will have the following labels:
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.target | required | Indicates the target we are outlier detection on. (Same as the attribute defined in A66.) |
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+|  grpc.xds_client.outlier_detection.ejections_enforced_total | Counter | {ejection} | 	grpc.target	Total enforced ejections due to any outlier type |
+|  grpc.xds_client.outlier_detection.ejections_active | Gauge |	{ejection} |	grpc.target |	Number of currently ejected hosts |
+|  grpc.xds_client.outlier_detection.ejections_overflow |	Counter |	{ejection} |	grpc.target |	Number of ejections aborted due to max ejection percentage |
+|  grpc.xds_client.outlier_detection.ejections_enforced_success_rate |	Counter |	{ejection} |	grpc.target |	Enforced success rate outlier ejections |
+|  grpc.xds_client.outlier_detection.ejections_detected_success_rate |	Counter |	{ejection} |	grpc.target |	Detected (but unenforced) success rate outlier ejections |
+|  grpc.xds_client.outlier_detection.ejections_enforced_failure_percentage |	Counter |	{ejection} |	grpc.target |	Enforced failure percentage outlier ejections |
+|  grpc.xds_client.outlier_detection.ejections_detected_failure_percentage |	Counter |	{ejection} |	grpc.target |	Detected (but unenforced) failure percentage outlier ejections |
+
+These metrics will all be available via global OpenTelemetry recorder. On any ejection/unejection, these metrics will be accordingly updated in the current outlier detection.
+
+### Metric Stability
+
+All metrics added in this proposal will start as experimental and therefore off by default. The long term goal will be to de-experimentalize them and have them be on by default, but the exact criteria for that change are TBD.
+
+### Temporary environment variable protection
+
+This proposal does not include any features enabled via external I/O, so it does not need environment variable protection.
+
+## Rationale
+
+The metrics defined here are generally a trade-off between the usefulness
+of the metric and the cost of reporting it. As the design goal is offering parity to envoy metrics,
+we decided to include any metric that was appropriate to gRPC outlier detection.
+
+## Implementation
+
+Dropbox is able to contribute towards a Core and Go implementation, in that order. Implementation of remaining gRPC languages is left for respective gRPC team or other contributors.
--- a/A87-grpc-metrics-xds-outlier-detection.md
+++ b/A87-grpc-metrics-xds-outlier-detection.md
@@ -1,7 +1,7 @@
 A87: gRPC OTel Metrics for xDS Outlier Detection
 ---
 * Author(s): Pardhu Konakanchi, Sai Teja
-* Approver: a11r
+* Approver:
 * Status: Draft
 * Implemented in: Core, Go
 * Last updated: 03-04-25
--- a/A87-grpc-metrics-xds-outlier-detection.md
+++ b/A87-grpc-metrics-xds-outlier-detection.md
@@ -12,18 +12,18 @@ This document proposes some new metrics that will be added in gRPC for xDS Clien
 
 ## Background
 
-A50: gRPC xDS Outlier Detection Support is a spec for gRPC to support xDS Outlier Detection. The current implementation only offers debug and trace logging in terms of visibility, which can be insufficient for understanding and diagnosing decision making in large-scale production systems. 
+[A50: gRPC xDS Outlier Detection Support](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md) is a spec for gRPC to support xDS Outlier Detection. The current implementation only offers debug and trace logging in terms of visibility, which can be insufficient for understanding and diagnosing decision making in large-scale production systems. 
 
 Using A79: Non-per-call Metrics Architecture, it is possible to add granular metrics to make visibility into outlier detection easy for service owners utilizing gRPC.
 
 ### Related proposals: 
-* A50: gRPC xDS Outlier Detection Support
-* A66: OpenTelemetry Metrics
-* A79: Non-per-call Metrics Architecture
+* [A50: gRPC xDS Outlier Detection Support](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md)
+* [A66: OpenTelemetry Metrics](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md)
+* [A79: Non-per-call Metrics Architecture](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)
 
 ## Proposal
 
-A79’s non-per-call metrics architecture fits perfectly into the metrics reporting solution. We propose to aim for parity where appropriate with Envoy’s xDS implementation for metrics to be collected in gRPC
+[A79’s](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md) non-per-call metrics architecture fits perfectly into the metrics reporting solution. We propose to aim for parity where appropriate with [Envoy’s Outlier Detection metric collection](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats#outlier-detection-statistics) to be collected in gRPC
 
 Outlier Detection metrics will have the following labels:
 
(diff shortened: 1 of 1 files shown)


A87-grpc-metrics-xds-outlier-detection.md
```diff
@@ -0,0 +1,64 @@
...
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.target | required | Indicates the target we are running outlier detection on. (Same as the attribute defined in [A66](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md).) |
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+|  grpc.xds_client.outlier_detection.ejections_enforced_total | Counter | {ejection} | 	grpc.target |	Total enforced ejections due to any outlier type |
```
markdroth, line 38: The prefix `grpc.xds_client.` is not appropriate here, since these metrics are going to be generated by the outlier detection policy, not by the XdsClient.

I think these should use the prefix `grpc.lb.` instead.

A87-grpc-metrics-xds-outlier-detection.md
```diff
@@ -0,0 +1,64 @@
+A87: gRPC OTel Metrics for xDS Outlier Detection
```
markdroth, line 1: I don't think there's actually anything OT-specific in this proposal.  These metrics will use the non-per-call metric APIs, which can in principle be used by any gRPC stats plugin.  Currently, the only such plugin to exist is the OT plugin, but there could be others in the future, and these metrics would work fine with that plugin when that happens, since there will be nothing OT-specific in the implementation of these metrics.

Given that, I think we should remove most of the references to OT in this proposal.

A87-grpc-metrics-xds-outlier-detection.md
markdroth, line 38: Do we want to also support the `grpc.lb.backend_service` label on these metrics, which would indicate the xDS cluster?  See pending gRFC A89 (https://github.com/grpc/proposal/pull/471) for the current proposal around that label.  It seems like it would be equally useful for the metrics described here.

Note that the current proposal in A89 is to set this attribute in the xds_cluster_impl policy, but the outlier_detection policy is above the xds_cluster_impl policy in the tree, so we'd probably need to set this attribute in the cds policy instead.  And I think this would depend on first implementing A75, which is not yet implemented in Go -- I don't see a clean way to set this label before that change, because before that, the cds policy will have multiple leaf clusters in different priorities, with no obvious way to pass down the right attributes to each child of the priority policy.  But we could have this gRFC say that the label will be supported only once A75 is implemented.

@ejona86 and @dfawley, thoughts on this?

A87-grpc-metrics-xds-outlier-detection.md
```diff
@@ -0,0 +1,64 @@
...
+---
+* Author(s): Pardhu Konakanchi, Sai Teja
+* Approver:
+* Status: Draft
+* Implemented in: Core, Go
+* Last updated: 03-04-25
```
markdroth, line 7: Also, please start a thread for this gRFC on the grpc-io mailing list, and then add a link to that thread here, as per the gRFC process.

A87-grpc-metrics-xds-outlier-detection.md
markdroth, line 1: Please rename the file to `A91-outlier-detection-metrics.md`.

A87-grpc-metrics-xds-outlier-detection.md
```diff
@@ -0,0 +1,65 @@
...
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+|  grpc.lb.outlier_detection.ejections_enforced_total | Counter | {ejection} | 	grpc.target |	Total enforced ejections due to any outlier type |
+|  grpc.lb.outlier_detection.ejections_active | Gauge |	{ejection} |	grpc.target |	Number of currently ejected hosts |
+|  grpc.lb.outlier_detection.ejections_overflow |	Counter |	{ejection} |	grpc.target |	Number of ejections aborted due to max ejection percentage |
+|  grpc.lb.outlier_detection.ejections_enforced_success_rate |	Counter |	{ejection} |	grpc.target |	Enforced success rate outlier ejections |
+|  grpc.lb.outlier_detection.ejections_detected_success_rate |	Counter |	{ejection} |	grpc.target |	Detected (but unenforced) success rate outlier ejections |
```
murgatroid99, line 43: The name of this metric seems to include all detected success rate ejections, including the ones that were not enforced, but the description seems to indicate that it only includes unenforced ejections. Which of those is intended? `grpc.lb.outlier_detection.ejections_detected_failure_percentage` also has the same issue.

A87-grpc-metrics-xds-outlier-detection.md
markdroth: We chatted about this internally today.  We agree that adding this optional label is useful, but it does require A75.

@ejona86 is going to change his pending A89 to indicate that post-A75, we'll set the attribute in the cds policy instead of the xds_cluster_impl policy, which will make it available in the outlier_detection policy.

For this gRFC, please add this optional label, but note that implementations will not support that optional label until they implement A75.

PardhuKonakanchi: @markdroth sorry for the delay. Added the optional label with the note

A91-outlier-detection-metrics.md
```diff
@@ -0,0 +1,74 @@
...
+| ----------- | ----------- | ----------- |
+| grpc.target | required | Indicates the target we are running outlier detection on, as described in [A66]. |
+| grpc.lb.backend_service | optional | The backend service to which the traffic is being sent, as described in [A89]. Note that this label will be supported only if [A75] has already been implemented. |
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+|  grpc.lb.outlier_detection.ejections_enforced_total | Counter | {ejection} | 	grpc.target, grpc.lb.backend_service |	Total enforced ejections due to any outlier type |
+|  grpc.lb.outlier_detection.ejections_active | Gauge |	{ejection} |	grpc.target, grpc.lb.backend_service |	Number of currently ejected hosts |
```
markdroth, line 49: I just noticed one problem here, which is that there will be a conflict for this metric whenever the application creates more than one channel to the same target.

This isn't a problem for counter metrics, because each channel can increment the same counter metric whenever an event occurs, and the resulting data is still valid and correct.  But it's a problem for gauges, because each channel would report its own total value to the same metric, but there can be only one value for the metric.

Unfortunately, I don't think that we have a good solution for this.  The only alternative I can think of is to add an additional label to this metric containing a UUID for each instance of the outlier_detection LB policy, just to avoid the conflict.  If that doesn't work, then we should probably just remove this metric from the gRFC as something that we can't support right now. :(

PardhuKonakanchi: @markdroth An interesting problem.

We spent some time thinking about this and also couldn't really think of a good solution. I think adding a UUID may increase cardinality of the metric too much (at least for users who are limited in cardinality). For now, let's remove the metric.

If someone is able to think of a solution in the future, perhaps we can update the grfc at that point? But removed it for now

markdroth:
This looks great to me!

Waiting on reviews from @ejona86, @dfawley, and @yashykt.

A91-outlier-detection-metrics.md
```diff
@@ -0,0 +1,73 @@
...
+|  grpc.lb.outlier_detection.ejections_enforced_total | Counter | {ejection} | 	grpc.target, grpc.lb.backend_service |	Total enforced ejections due to any outlier type |
+|  grpc.lb.outlier_detection.ejections_overflow |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Number of ejections aborted due to max ejection percentage |
+|  grpc.lb.outlier_detection.ejections_enforced_success_rate |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Enforced success rate outlier ejections |
+|  grpc.lb.outlier_detection.ejections_detected_success_rate |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Detected (even if unenforced) success rate outlier ejections |
+|  grpc.lb.outlier_detection.ejections_enforced_failure_percentage |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Enforced failure percentage outlier ejections |
+|  grpc.lb.outlier_detection.ejections_detected_failure_percentage |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Detected (even if unenforced) failure percentage outlier ejections |
+
+On any ejection/unejection, these metrics will be accordingly updated using globally available metric recorder.
```
ejona86, line 55: The metric recorder isn't global. It is per-channel.

A91-outlier-detection-metrics.md
```diff
@@ -0,0 +1,73 @@
...
+## Proposal
+
+[A79]’s non-per-call metrics architecture fits perfectly into the metrics reporting solution. We propose to aim for parity where appropriate with [Envoy’s Outlier Detection metric collection](https://www.envoyproxy.io/docs/envoy/latest/configuration/upstream/cluster_manager/cluster_stats#outlier-detection-statistics) to be collected in gRPC.
+
+Outlier Detection metrics will have the following labels:
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.target | required | Indicates the target we are running outlier detection on, as described in [A66]. |
+| grpc.lb.backend_service | optional | The backend service to which the traffic is being sent, as described in [A89]. Note that this label will be supported only if [A75] has already been implemented. |
```
ejona86, line 42: `backend_service` isn't available to Outlier Detection, even with A75 and A89. A89 only passes it to per-RPC metrics. In this gRFC, I think we'll have to make a resolver attribute to pass down from non-aggregate cluster `cds` policies, similar to [the attribute passed from weighted_target](https://github.com/grpc/proposal/blob/413b966783781e98b69ae0d8025cff84b5f5962a/A78-grpc-metrics-wrr-pf-xds.md?plain=1#L77).

@markdroth, @dfawley, thoughts?

markdroth: A89 also passes down an attribute to be used in the WRR metrics, which are also not per-RPC.  I think we can use that same attribute for the OD metrics.  However, I agree that in order to make that attribute visible to the OD policy, we need to set the attribute in the cds policy instead of the xds_cluster_impl policy -- but we can't do that until A75 is implemented.

For more details, see my reply in https://github.com/grpc/proposal/pull/471#discussion_r2025304376.

markdroth: Sounds good.

@PardhuKonakanchi, please add a note at the end of the "Proposal" section saying that the implementation will populate this label based on the resolver attribute passed down from the cds policy, as described in A89.  Thanks!

A91-outlier-detection-metrics.md
PardhuKonakanchi: I removed that detail regarding recorder as I don't think relevant here

A91-outlier-detection-metrics.md
```diff
@@ -0,0 +1,75 @@
...
+* [A66: OpenTelemetry Metrics][A66]
+* [A79: Non-per-call Metrics Architecture][A79]
+* [A75: xDS Aggregate Cluster Behavior Fixes][A75]
+* [A89: Backend Service Metric Label][A89] (pending)
+
+[A50]: A50-xds-outlier-detection.md
+[A66]: A66-otel-stats.md
+[A75]: A75-xds-aggregate-cluster-behavior-fixes.md
+[A79]: A79-non-per-call-metrics-architecture.md
+[A89]: https://github.com/grpc/proposal/pull/471
```
markdroth, line 31: A89 has now been merged, so this link can be changed to point to `A89-backend-service-metric-label.md`.

A91-outlier-detection-metrics.md
```diff
@@ -0,0 +1,75 @@
+A91: gRPC Metrics for Outlier Detection
+---
+* Author(s): @pardhukonakanchi, @huntsman90
+* Approver: @markdroth
+* Status: In Review
+* Implemented in: Core, Go
```
markdroth, line 6: It's not actually implemented anywhere yet, so let's leave this field blank.

markdroth (commented):
This looks really good!

@dfawley, waiting for your review.

A91-outlier-detection-metrics.md
```diff
@@ -0,0 +1,75 @@
...
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+|  grpc.lb.outlier_detection.ejections_enforced_total | Counter | {ejection} | 	grpc.target, grpc.lb.backend_service |	Total enforced ejections due to any outlier type |
+|  grpc.lb.outlier_detection.ejections_overflow |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Number of ejections aborted due to max ejection percentage |
+|  grpc.lb.outlier_detection.ejections_enforced_success_rate |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Enforced success rate outlier ejections |
+|  grpc.lb.outlier_detection.ejections_detected_success_rate |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Detected (even if unenforced) success rate outlier ejections |
+|  grpc.lb.outlier_detection.ejections_enforced_failure_percentage |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Enforced failure percentage outlier ejections |
```
yashykt, line 54: I'll just mention that the alternative way to state this would be to have an attribute/label for whether an ejection is enforced or not. With that, we'll only be incrementing one counter in the case where we have an ejection that is enforced, as opposed to two with the current proposal.

PardhuKonakanchi: @yashykt This is a good point. I followed the approach Envoy took, but I'm open to updating this to `ejections_enforced` and `ejections_unenforced`. This may also be easier for users who are searching for one or the other, rather than total detected.

One thing is we should probably specify reasons an outlier is unenforced. The two I can think of are doesn't pass `enforcement percentage` or we are already above `max_ejection_percent`

PardhuKonakanchi: @yashykt @markdroth I went ahead and updated the metric to be unenforced instead of detected and included the Rationale. If we want to keep as-is, can revert, but I do think it makes sense this way as a user

PardhuKonakanchi: @dfawley Great idea! Just made an update accordingly. Let me know what you think

A91-outlier-detection-metrics.md
```diff
@@ -0,0 +1,79 @@
...
+| grpc.lb.backend_service | optional | The backend service to which the traffic is being sent, as described in [A89]. Note that this label will be supported only if [A75] has already been implemented |
+
+The `grpc.lb.backend_service` label will be populated based on the resolver attribute passed down from the cds policy, as described in A89.
+
+The following metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+|  grpc.lb.outlier_detection.ejections_enforced | Counter | {ejection} | 	grpc.target, grpc.lb.backend_service |	Total enforced ejections due to any outlier type |
+|  grpc.lb.outlier_detection.ejections_overflow |	Counter |	{ejection} |	grpc.target, grpc.lb.backend_service |	Number of ejections aborted due to max ejection percentage |
```
dfawley: @PardhuKonakanchi I'm assuming you wanted to delete these two lines?

```suggestion
```

PardhuKonakanchi: @dfawley Yes, missed those earlier. Thanks!

A91-outlier-detection-metrics.md
```diff
@@ -39,20 +39,18 @@ Outlier Detection metrics will have the following labels:
 | Name        | Disposition | Description |
 | ----------- | ----------- | ----------- |
 | grpc.target | required | Indicates the target we are running outlier detection on, as described in [A66]. |
-| grpc.lb.backend_service | optional | The backend service to which the traffic is being sent, as described in [A89]. Note that this label will be supported only if [A75] has already been implemented. |
+| grpc.lb.outlier_detection.detection_method | required | Indicates the method with which we detected outlier. Currently one of {"success_rate", "failure_percentage"}
+| grpc.lb.outlier_detection.unejection_reason | required | Indicates the reason we did not eject a detected outlier. Currently one of {"enforcement_percentage", "max_ejection_overflow"}
```
markdroth: Shouldn't this be called `unenforced_reason` rather than `unejection_reason`?  Un-ejecting is what happens after an ejection expires, and that's not what's happening here.

dfawley (approved):
LGTM module Mark's comment.

A91-outlier-detection-metrics.md
PardhuKonakanchi: @markdroth Good catch! Fixed

markdroth (approved):
Looks great!

markdroth:
> @markdroth Thanks for all the help reviewing this! Anything blocking us from merging the gRFC now?

Nothing other than the fact that I forgot to hit the button. :)

Thanks for the contribution!

PR #482, opened 2025-03-11
A74 update: improve XdsDependencyManager watcher API

markdroth:
The original intent here was that the xds resolver would handle `OnError()` the same way that it did prior to A74: it would return a result to the channel that had an error for the service config, which would cause the channel to ignore the update if it had already previously seen a valid result.  This poses a problem for cases like receiving a socket listener instead of an API listener or receiving a route config that does not contain a virtual host matching the channel's default authority, because in those cases we want to stop using the previously seen resources and start failing data plane RPCs.

This change fixes that by specifying that the xds resolver will see errors only in cases where we want to stop using the previously received resources, if any.  As a result, we are able to collapse all three methods on the watcher into one.

### DIFF
--- a/A74-xds-config-tears.md
+++ b/A74-xds-config-tears.md
@@ -150,13 +150,10 @@ class Watcher {
  public:
   virtual ~Watcher() = default;
 
-  // Called when a new xDS config is available.
-  virtual void OnUpdate(XdsConfig config) = 0;
-
-  // Called when there is an error on LDS or RDS only.  The context will
-  // be a human-readable string indicating the scope in which the error
-  // occurred (e.g., the resource type and name).
-  virtual void OnError(absl::string_view context, absl::Status status) = 0;
+  // Called when a new xDS config is available or when there is an error
+  // for the LDS or RDS resource for which we want to stop using the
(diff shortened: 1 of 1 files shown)


ejona86:
CC @larry-safran

ejona86 (approved):

A74-xds-config-tears.md
```diff
@@ -203,17 +198,32 @@ struct XdsConfig {
...
+resources, if any.  Specifically, it will be invoked with an error in
+the following cases:
+- When the listener or route configuration watchers see
+  OnResourceDoesNotExist().
+- When the listener or route configuration watchers see OnError(), if
+  they did not previously see OnResourceChanged().  (If the watchers
+  *did* previously see OnResourceChanged(), then the OnError() event is
+  ignored by the XdsDependencyManager and not passed on to the xds
+  resolver.)
+- When the received listener resource is a socket listner instead of an
```
dfawley, line 213: listEner*

markdroth: Good catch, thanks.  Fixed.

PR #484, opened 2025-03-13
A93: xDS ExtProc Support

markdroth:


A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+the channel across updates.
+
+TODO: stats plugin propagation?
+
+#### Events on the ext_proc Stream
+
+On the ext_proc stream, the events sent and received must be in the same
+order as on the data plane.  For client-to-server events, the order must
+be headers, followed by zero or more messages, followed by a half-close.
+For server-to-client events, the order must be headers, followed by zero
```
dfawley, line 95: I assume server-to-client headers are optional in trailers-only?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+
+#### Messages Sent To the ext_proc Server
+
+The [`ProcessingRequest`
+message](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L62)
+sent to the server will be populated as follows:
+- [request_headers](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L76)
+  and
+  [response_headers](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L81).
+  Populated when sending client headers or server headers, respectively.
```
dfawley, line 582: I know we talked about this, but what's the story around what headers should be included vs. not? Are we going to define which headers specifically?  It would be good to call out exactly whether method/scheme/path/te/user-agent/message-type/etc, etc are supposed to be included or not.  In Go many of these things are added by the transport on the way out or are removed by the transport on the way in.  It would be great if we can specify: "only the things set by the application [plus X, Y, and Z, which some libraries may need to manually synthesize before sending to the ext_proc server]"

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+  - [body](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L235):
+    Contains the serialized message.  Will be empty if the client sends
+    a half-close when there is no message to send (i.e., if the client
+    never sent any message on the stream, or if the half-close is sent
+    after the filter has already sent the last message to the ext_proc
+    server).
+  - [end_of_stream](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L239):
+    For client messages, may be true if the client sent a half-close at
+    the same time as the last message.  For server messages, will always
+    be false.
```
dfawley, line 603: I don't think Go will be able to observe this scenario on the server.  It has a `RecvMsg()` method on the `ServerStream` that only returns a message or an `io.EOF` error at stream end. There is no way to observe a message _and_ an end-stream together.  Instead Go will need to send the request message and the end stream separately.  (Or we'll have to design a new interception API and add in the required plumbing between our transport and core grpc layers.)

Is this distinction a must-have feature for the in-progress Rust design?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+      the filter will cancel the ext_proc stream and treat it as if it
+      failed with a non-OK status (see above for details).
+    - [body_mutation](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L317):
+      - [streamed_response](https://github.com/envoyproxy/envoy/blob/564612e32eafc10a7a7fd490cdb5cc7149e5802b/api/envoy/service/ext_proc/v3/external_processor.proto#L447):
+        Replaces the original serialized message.  Within it:
+        - [body](https://github.com/envoyproxy/envoy/blob/564612e32eafc10a7a7fd490cdb5cc7149e5802b/api/envoy/service/ext_proc/v3/external_processor.proto#L419):
+          The serialized message body.
+        - [end_of_stream](https://github.com/envoyproxy/envoy/blob/564612e32eafc10a7a7fd490cdb5cc7149e5802b/api/envoy/service/ext_proc/v3/external_processor.proto#L424C8-L424C21):
+          If true, indicates that a half-close should be sent after the
+          message.  Honored only on client-to-server messages.
```
dfawley: This puts the stream into an interesting state where the client application may continue sending messages but the filter needs to ignore them.  In Go, on the client, it can simply ignore calls to `SendMsg()`, but on the server (in all languages with a blocking API, so maybe just Go/Rust?), it means we need to spawn a job to read from the client send stream and discard the results so the client doesn't block on flow control.

Actually in Rust we may implement it a bit differently, at least with my current design -- we would be giving the handler an owned RecvStream to poll and the gRPC library will already have to handle the situation where the application drops it early before returning from the handler -- in that case, we probably want to do this behavior of discarding any received messages vs. the alternative of letting the client application back up on flow control.

(I assume it's better to drain it, otherwise it could lead to deadlocks.)

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.client_ext_proc.client_headers_duration | Histogram | s | grpc.target, grpc.lb.backend_service | Time between when the ext_proc filter sees the client's headers and when it allows those headers to continue on to the next filter. |
+| grpc.client_ext_proc.client_half_close_duration | Histogram | s | grpc.target, grpc.lb.backend_service | Time between when the ext_proc filter sees the client's half-close and when it allows that half-close to continue on to the next filter. |
+| grpc.client_ext_proc.server_headers_duration | Histogram | s | grpc.target, grpc.lb.backend_service | Time between when the ext_proc filter sees the server's headers and when it allows those headers to continue on to the next filter. |
+| grpc.client_ext_proc.server_trailers_duration | Histogram | s | grpc.target, grpc.lb.backend_service | Time between when the ext_proc filter sees the server's trailers and when it allows those trailers to continue on to the next filter. |
+
+#### Server-Side Metrics
+
+The following server-side metrics will be exported:
```
dfawley, line 861: Are there no labels that could be used on the server?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+sub-optimal approach, for the following reasons:
+- It would still not work for streaming cases, where the ext_proc server
+  needs to be able to modify individual messages as they are sent,
+  rather than only modifying the entire stream as a single payload.
+- It would waste bandwidth sending data that really isn't useful.
+- It would not avoid the need for the ext_proc server to handle the
+  gRPC framing.
+- It would further spread use of the gRPC framing to ext_proc servers,
+  which will make it awkward for gRPC to support other transports with
+  different framing in the future.
```
dfawley, line 896: It also fails to represent the real data on the stream, as the "is_compressed" flag wouldn't be set correctly, and the exact compressed bytes, if applicable, can't really be knowable.

A93-xds-ext-proc.md
markdroth: Yes.  Clarified.

A93-xds-ext-proc.md
markdroth: If the implementation can't observe these two things at the same time, then it won't ever set this bit when sending a message, which is fine.  It just means that it will need to send a separate message to the ext_proc server indicating the EOS, which is a little less efficient but will still work.

Now that you mention it, C-core is likely to have a similar issue in the v3 API.  We should probably think about whether this is something we want to fix, but I don't think it's a prereq for this design.

A93-xds-ext-proc.md
markdroth: Yes, I think we need to drain it for flow control reasons.

This seems like mostly a comment about the implementation, not the gRFC, so let's discuss further offline if needed.

A93-xds-ext-proc.md
markdroth: Not that I can think of.  If you have any suggestions of labels that you think might be useful here, I'm open to them.

A93-xds-ext-proc.md
markdroth: Good point.  Added a bullet about this.

A93-xds-ext-proc.md
markdroth: Yeah, we need to add this for both ext_authz and ext_proc.  I think @easwars was compiling a list of what headers we should document.

A93-xds-ext-proc.md
dfawley: This is going to apply to all implementations, though, so I think it's important to mention it here -- half-close is normally initiated by the client, but in this case the server is simulating it with no way to signal it is doing so to the client.  So servers need to make sure they continue to read from the request stream and discard anything received.

A93-xds-ext-proc.md
dfawley: Aren't these actually per-call/attempt metrics, so should they have all the same labels as the default per-call metrics?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+best-effort and not guaranteed.
+
+To implement this, the filter will store the processing mode separately
+for each data plane RPC.  The processing mode for an RPC will be
+initialized based on the filter's config, but it may be modified later
+by subsequent overrides.
+
+### Metrics
+
+The ext_authz filter will export metrics using the non-per-call metrics
```
dfawley: `ext_proc`?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+the server.
+
+#### Client-Side Metrics
+
+The client-side metrics will have the following labels:
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.target | required | The target of the gRPC channel in which ext_authz is used, as the defined in [A66]. |
+| grpc.lb.backend_service | optional | The backend service to which the traffic is being sent, as defined in [A89].  This will be populated from the xDS cluster name, which will be passed to the ext_authz filter as described in [A60]. |
```
dfawley: `ext_proc` x2

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+
+To implement this, the filter will store the processing mode separately
+for each data plane RPC.  The processing mode for an RPC will be
+initialized based on the filter's config, but it may be modified later
+by subsequent overrides.
+
+### Metrics
+
+The ext_authz filter will export metrics using the non-per-call metrics
+architecture defined in [A79].  There will be a separate set of metrics
```
dfawley: Oh maybe the implication is that it will get all of those labels already? If so we should call that out, probably?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+means that messages sent to the ext_proc server will always be
+uncompressed: for sent messages, compression will not happen until after
+the ext_proc filter runs, and for received messages, decompression will
+happen before the ext_proc filter runs.  However, in Envoy, the messages
+may be either compressed or uncompressed, depending on whether the
+sender compressed them before sending them to Envoy.  To expose this
+difference to the ext_proc server, we have added a new field called
+`grpc_message_compressed` in both the ext_proc request and response
+message (see https://github.com/envoyproxy/envoy/pull/38753).  Envoy
+will set this bit in the ext_authz request based on the corresponding
```
dfawley: ext_proc?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,744 @@
...
+  If unset, the delay is 5 seconds.  If present, the value must obey the
+  restrictions specified in the [`google.protobuf.Duration`
+  documentation](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration),
+  and it must have a positive value.
+
+The following fields will be ignored by gRPC:
+- message_timeout and max_message_timeout: Message timeouts do not make
+  sense in GRPC body send mode.
+- http_service: It doesn't make sense for gRPC to support non-gRPC
+  mechanisms for contacting the ext_authz server.
```
dfawley: ext_proc

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,751 @@
...
+  - [headers](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L216):
+    Contains the headers.
+  - [end_of_stream](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L227):
+    This will always be false for client headers.  For server headers,
+    it will be true when the server sends a Trailers-Only response.
+- [request_body](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L85)
+  and
+  [response_body](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L89).
+  Populated when sending a client message or server message, respectively.
+  Inside of them:
```
dfawley, line 593: Have we talked about the performance implications here?  If Go uses its current interceptor API, then we'll have only deserialized messages available, meaning we'll have to re-serialize when sending to the ext_proc server.  So in theory we're burning extra CPU than if we could just take the payload straight from the bytes and set them in the body field.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,751 @@
...
+
+We will support the following fields in the [`ExternalProcessor`
+proto](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L101):
+- [grpc_service](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L129):
+  This field must be present.  It will be validated as described in [A102].
+- [failure_mode_allow](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L177):
+  By default, if the RPC to the ext_proc server fails with a non-OK
+  status, the data plane RPC will be failed with status UNAVAILABLE.  If
+  this field is set to true, then the data plane RPC will instead be
+  allowed to continue, with no further action taken by the ext_proc filter.
```
rishesh007: If the ext_proc server returns a non-OK status, should we forward the original headers/message that were sent to the ext_proc server, or should the request be dropped?

markdroth: If we fail the RPC, then we immediately report status UNAVAILABLE and stop processing all events on the RPC.

A93-xds-ext-proc.md
markdroth: We definitely don't want the overhead of duplicate serialization/deserialization if we can avoid it.  I've added a paragraph about that in the "Payload Handling" section above.

A93-xds-ext-proc.md
markdroth: Okay, added a sentance about that.

A93-xds-ext-proc.md
markdroth: We have the following labels on per-call metrics today:
- method: This seems reasonable to add if we think it will be useful.
- target (client only): This also seems reasonable to add if it will be useful (client only).
- status: I don't think this one makes sense, because we don't know the status of the data plane RPC at the point at which we're updating these metrics.
- the new per-RPC label being introduced in A108: I'd be okay with this one too.

I'm open to adding all but status, as long we think they'll be useful and are not worried about performance issues in Java or Go.

A93-xds-ext-proc.md
markdroth: Not sure what you mean by "all of those labels".  A79 doesn't define any labels that are intended to be applied to all metrics.

Maybe you meant to refer to the per-call metric labels?  If so, I've answered that below.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,763 @@
...
+messages, so each body chunk sent to or from the ext_proc server will
+be a complete, deframed gRPC message.
+
+The new `GRPC` mode will be the only processing mode supported in gRPC.
+It is be desirable for Envoy to implement the same mode, so that users
+can switch back and forth between proxy and proxyless data planes without
+breaking their ext_proc servers.
+
+gRPC implementations should structure their xDS HTTP filter APIs such that
+the filter has access to the serialized bytes of each message rather than
```
eshitachandwani, line 232: Correct me if I am wrong , but from what I know is that a decision has been made to bear the double serialization/deserialization cost as of now , at least in Go. Do you think it is worth mentioning that here?

markdroth: As far as this spec is concerned, implementations should not do redundant serialization/deserialization.  Go may choose to initially implement it that way in order to get the functionality done faster, but that really isn't a good state to be in for the long term, and I expect that people will complain about the performance.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,834 @@
...
+to attribute value.
+
+The list of attributes to include is specified by the
+[`request_attributes`](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L188)
+config field, for client-to-server events, or the
+[`response_attributes`](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L195)
+config field, for server-to-client events.
+
+The set of supported attribute names is the same as what we support for
+any CEL expression in xDS, as defined in [A106] and [A103].
```
eshitachandwani: A103 link seems to be broken.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,834 @@
...
+on the stream.  Note that the number of messages produced by the ext_proc
+server may not match the number of messages originally sent to it.
+
+A client half-close is represented to the ext_proc server as an
+end-of-stream indicator on a header or message event, rather than being
+a discrete event.
+
+Events on the data plane RPC should be sent on the ext_proc stream as
+they occur, even if the filter has not yet received a response from the
+ext_proc server for a previous event.  For example, if the filter is
```
eshitachandwani, line 131: Curious about the example here, from what I understand, sending and getting the modified headers should be a blocking operation and only after modifying the headers should we use them to create the dataplane stream. The Send and Receive functions should come after the stream is created. The example seems to suggest that Send can happen before getting the header modification. Just want to understand the sequence of events.

markdroth: In the C-core API, the application does not need for sending the headers to complete before it sends the first message on the stream, which is why this can happen.

I am guessing from your comment that the grpc-go API does not allow the application to send the first message until it gets the stream object, which is not returned until the headers have gone out on the wire?  If so, then I think there are two possible ways to deal with it:

1. Keep things the way they are, which means this specific case can't happen in Go.
2. Tell the application that the stream has been created as soon as headers are sent to the ext_proc server, without waiting for the headers to be sent to the data plane server.  This would increase the amount of memory used for buffering headers, but it would allow better pipelining.

I think either option is probably fine.  @easwars and/or @dfawley may have thoughts on which one makes more sense for Go.

Note that even if this specific example can't happen in Go, there are other cases that will happen. For example, in the server-to-client direction, the ext_proc filter does not need to wait until it receives the response headers from the ext_proc server before it sends the first response message.

@ejona86 Not sure if this is an issue for Java.

A93-xds-ext-proc.md
markdroth: It will work once this is merged into master.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,834 @@
...
+
+In Response to Data Plane RPC Event | Mode Override | Supported? | May Miss Data?
+----------------------------------- | ------------- | ---------- | --------------
+client headers | client body | `NONE` to `GRPC`, but not the reverse | yes
+client headers | server headers | yes | no
+client headers | server body | yes | no
+client headers | server trailers | yes | no
+client body | server headers | yes | yes
+client body | server body | `NONE` to `GRPC`, but not the reverse | yes
+client body | server trailers | yes | yes
```
eshitachandwani: Can response of client body not specify client body override?

markdroth: No, because if we're getting a client body response, that means that we're already in `GRPC` mode, and the only possible mode we could switch to would be `NONE`.  However, as described above, that transition is not supported, because we don't have a safe way to stop streaming the body without terminating the entire stream.

eshitachandwani: Ahh okay !! Right!

A93-xds-ext-proc.md
ejona86: IIRC, in Go the client headers block until a stream is allocated. So it could make sense to wait for those to be processed by ext_proc before unblocking the client, to still wait for the stream to be allocated. That doesn't sound _required_ though. And it would add significant latency, especially for unary RPCs.

Java doesn't have flow control for unary RPCs, so it doesn't impact that. We do typically delay saying the RPC is ready for data until we allocate the stream, but we can trivially implement this either way. (I think it'd only impact a single if.)

easwars: From what I can see, when an application attempts to create a stream:
- the grpc layer creates an instance of the client stream interface and as part of doing this, it creates an `attempt`
- the attempt retrieves the transport by doing a `Pick` and creates a stream on the transport
- this results in an instance of the client stream interface being created at the transport layer
  - this allocates the stream ID and checks for write quota (which should exist, the default is 64K)
  - creates the header frame and queues it in the control buf
  - returns

At this point, the application should have a stream and should be able to send messages on it. They won't get sent out on the wire until the headers frame is sent out. I don't see the transport code even waiting to get a headers frame before sending out data (as long as there is write quota).

@eshitachandwani : If you think otherwise, let's talk and go over the code together.

markdroth: @easwars The ext_proc filter would run between the first and second bullet in your list -- i.e., after the client stream instance is created but before the LB pick.  This means that the workflow will go like this:

1. gRPC layer creates client stream interface, which creates an attempt.
2. Headers are sent to the ext_proc filter.  The ext_proc filter sends them to the ext_proc server and waits for the response.
3. The ext_proc filter receives the headers from the ext_proc server.  Now the ext_proc filter sends the headers on.
4. We do an LB pick, resulting in creating the client stream at the transport layer.  This gets returned to the application, which can then send the first message.

The problem here is that there is a round-trip to the ext_proc server in steps 2 and 3, and the application will need to wait for that to finish before it can send the first message on the stream.  This will hurt latency.

The alternative is to do something like this:

1. gRPC layer creates client stream interface, which creates an attempt.
2. Headers are sent to the ext_proc filter.  The ext_proc filter sends them to the ext_proc server and waits for the response.
3. While waiting for the ext_proc response, return the client stream interface to the application, so that it can start sending messages on the stream.
4. The client application sends a message on a stream.  This message gets down to the ext_proc filter, which sends the message to the ext_proc server.
5. The ext_proc filter receives the headers from the ext_proc server.  Now the ext_proc filter sends the headers on.
6. We do an LB pick, resulting in creating the client stream at the transport layer.

This approach would improve latency by allowing the application to start sending messages on the stream without waiting to get the headers back from the ext_proc server.

eshitachandwani: Understood. I think the alternative should be doable in Go too.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,834 @@
...
+  control on both the ext_proc stream and on the data plane stream.
+- In Go, a write is blocking and doesn't return until the write passes
+  flow control.  So observability mode doesn't need to do anything
+  special to handle flow control; it will simply not wait for the
+  ext_proc response before sending the message on the data plane stream.
+
+### Filter Configuration
+
+The filter supports both a top-level configuration and an override
+config.
```
kannanjgithub, line 439: Should the override config altogether replace the top level config config for the filter, or should individual fields be merged? If it is the latter, should it only be merging at level 1 (top level fields) (so it doesn't create a GrpcService config with target from top level config and channel creds from override config, for example)?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,834 @@
...
+  Headers to remove.
+
+All mutations (additions, modifications, and removals) are subject to the
+rules in the `mutation_rules` field in the filter config.  This field is
+an `envoy.config.common.mutation_rules.v3.HeaderMutationRules` message,
+which will be handled as described in [A92].
+
+#### Processing Mode Override
+
+The `mode_override` field in the ext_proc response allows the ext_proc
```
eshitachandwani: The docstring of the `mode_override` field states [here](https://github.com/envoyproxy/envoy/blob/73f90132ecbad2b2d24bee8e3555c43ed8eb3310/api/envoy/service/ext_proc/v3/external_processor.proto#L227-L234) that this field is only considered when set in response to header request. If set in response to body or trailers request , the dataplane should ignore it.  Are we changing that behavior for gRPC?

A93-xds-ext-proc.md
kannanjgithub: However the ProcessingMode override config is _meant_ to be considered at each individual field level, so this field needs to be merged with the top level ProcessingMode config.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,834 @@
...
+the trace context from the data plane RPC to the ext_proc RPC, so that
+the ext_proc RPC appears as a child span on the data plane RPC's trace.
+
+#### ExtProc Side Channel
+
+The ext_proc filter will create a gRPC channel to the ext_proc server.
+It will use the mechanism described in [A102] to determine the server to
+talk to and the channel credentials to use.
+
+We do not want to recreate this channel on LDS updates, unless the
```
kannanjgithub: LDS and RDS updates since override filter config for the route and weighted cluster/plugin can come via RDS.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,834 @@
...
+target URI or channel credentials changes.  The ext_proc filter will
+use the filter state retention mechanism described in [A83] to retain
```
kannanjgithub: Making the filter resource lifecycle dependent on only xds updates has a problem when the filter resource is a channel. An rpc could be scheduled for  future in the FaultFilter and if the FaultFilter wraps the ext-proc filter, and an xds update shuts down the channel, but the rpc now reaches the ext-proc filter, it will fail. We could apply a grace time for the channel shutdown and it could help when the scheduled execution is earlier than this shutdown period but not when it is not. To keep the channel alive in this case we would need to do more than rely only on the filter state mechanism of A83. When routing a rpc through a route we would need some reference count mechanism for all the resources managed by the interceptors in that route, and this reference count should keep the channel resource alive when if the channel is no longer referenced by any route after a xds update. This came out of [this](https://docs.google.com/document/d/1qFaP5BXFOtlbxmr5VbcpggC-Qi5ftIkHO7T85rInqS4/edit?resourcekey=0-RKa4-uRZ5tgeP902Ujw0fA&disco=AAAB3W2dQ38) discussion.

A93-xds-ext-proc.md
kannanjgithub: The attributes list while being same as those supported in any CEL expression, since we are not actually supporting expressions in the ext-proc filter, we can directly set their values without using the CEL runtime evaluation system used by the Rbac or Unified matcher filter?

kannanjgithub: Figured that especially for headers we will need to support attributes like request.headers['some-header'] or request.headers.some-header which will use of the CEL evaluation engine.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,834 @@
...
+the RPC.  For example, the ext_proc server may decide after seeing the
+client headers event that it does not actually need to see the client
+message events.
+
+If the `allow_mode_override` config field is set to false, then the
+`mode_override` field will be ignored.
+
+If the `allowed_override_modes` config field is non-empty and
+the value of the `mode_override` field does not match any entry in
+`allowed_override_modes`, the mode override will be ignored.  Note that
```
kannanjgithub: In envoy it applies the override first to arrive at an effective processing mode and then checks this resultant mode against the list of allowed_mode_overrides: https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/ext_proc/ext_proc.cc#L1820-L1824

kannanjgithub: Another thing is when comparing mode override against the list of allowed mode overrides, for headers send modes, since a mode override of DEFAULT means to leave the top level config specified mode unchanged, when doing the comparison for DEFAULT it seems logical that we should consider whether the overriden mode said SKIP / SEND but is still same as the top level config's even if didn't use the word DEFAULT, and consider the override mode as allowed.
However Envoy doesn't consider top level config for this comparison and does a raw enum value based comparison https://github.com/envoyproxy/envoy/blob/c83d0a1012719f76e3ea7bc703ea693261fab9c6/source/extensions/filters/http/ext_proc/allowed_override_modes_set.h#L42

What should we do in gRPC?

eshitachandwani: > In envoy it applies the override first to arrive at an effective processing mode and then checks this resultant mode against the list of allowed_mode_overrides: [envoyproxy/envoy@`main`/source/extensions/filters/http/ext_proc/ext_proc.cc#L1820-L1824](https://github.com/envoyproxy/envoy/blob/main/source/extensions/filters/http/ext_proc/ext_proc.cc#L1820-L1824)

The docstring only says that `AllowModeOverride` is to be considered when override is specified in the `ProcessingResponse`. Do we need to consider it for overriding using override config too?

A93-xds-ext-proc.md
markdroth: Fixed.

A93-xds-ext-proc.md
markdroth: Yes, that sounds right.  I've added a note about additional reference counting needed here.

A93-xds-ext-proc.md
markdroth: Good catch!  I hadn't noticed that.  I've updated this section accordingly.

A93-xds-ext-proc.md
markdroth: Thanks for calling out these details.  We should do the same thing as Envoy.  I've updated the text here to describe this.

A93-xds-ext-proc.md
markdroth: Each of the fields in the override config replaces the corresponding field from the top-level config as a whole.  I don't think the processing mode is any different here.

A93-xds-ext-proc.md
kannanjgithub: The [envoy attributes](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#attributes) list here doesn't look like it needs CEL support however, so once again reopening this comment. The gRFC should explicitly call out if we need to use the CEL evaluation or not for evaluating the attribute values to send to ext-proc.

markdroth: I've added a note indicating that this functionality does not necessarily need to be provided via CEL.

markdroth:
@kannanjgithub @eshitachandwani @rishesh007 FYI, it looks like we don't actually have a use-case for the mode override feature, and the logic for that is a little ugly, so I've removed it from the design.  We can consider adding it in the future if/when we encounter a use-case for it.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,773 @@
...
+    and
+    [response_body_mode](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto#L127):
+    The only modes we support here are
+    [`NONE`](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto#L66)
+    and `GRPC` (to be added).
+  - We ignore the request_trailer_mode field, since gRPC never sends
+    request trailers.
+- [request_attributes](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L188)
+  and
+  [response_attributes](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L195):
```
kannanjgithub, line 472: Should we add more details on how to determine the value of some of these response attributes? For example client stream tracer would help getting the response.size and response.backend_latency but response.header_size poses a problem because the Netty/Ok http don't surface the raw header size on the wire to the tracer callback, but only the parsed headers.

kannanjgithub: Also A106 doesn't talk about response attributes yet, so should we support them in ext_proc filter or not?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,773 @@
...
+
+TODO: stats plugin propagation?
+
+#### Events on the ext_proc Stream
+
+On the ext_proc stream, the events sent and received must be in the same
+order as on the data plane.  For client-to-server events, the order must
+be headers, followed by zero or more messages, followed by a half-close.
+For server-to-client events, the order must be headers (skipped for
+Trailers-Only), followed by zero or more messages, followed by trailers.
```
kannanjgithub: It is not skipped for trailers-only. It is sent as HttpHeaders with end_of_stream set to true.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,773 @@
...
+- We ignore the dynamic_metadata field, since it is not relevant to gRPC.
+- We ignore the mode_override field, since we are not supporting this
+  feature.
+
+Note that the responses from the ext_proc server must come back in the
+same order that the events were sent by the filter.  For example, if the
+client sends a client headers event and a client message event and the
+ext_proc server responds to the client message event first, that is
+considered a protocol error.  The filter will treat that as if the
+ext_proc stream failed with a non-OK status.
```
kannanjgithub, line 787: That implies we return a status code UNAVAILABLE. Doesn't INTERNAL seem more appropriate here? That's what envoy does (HTTP status code 500). Same question for invalid header mutation.

A93-xds-ext-proc.md
kannanjgithub: Processing mode override is special and the merge applies at the individual fields in it, reading the description of HeaderSendMode [here](https://github.com/envoyproxy/envoy/blob/099a9d71ebfd8aa9f823e1738b34138cb634a07b/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto#L30).

A93-xds-ext-proc.md
kannanjgithub, line 787: We are saying to use UNAVAILABLE everywhere except in the case of immediate response that specifies the status code to use. For protocol errors like this and in other places would it not be more appropriate to use the INTERNAL status? Envoy uses 500 HTTP status code in these cases.

A93-xds-ext-proc.md
eshitachandwani: Are you referring to the `DEFAULT` mode behavior? The description only covers overrides coming from the proc server response (which is now out of scope), not from the override config. Because of that, I think we can simply override the complete top-level field.

kannanjgithub: You are right, thanks.

A93-xds-ext-proc.md
eshitachandwani: @markdroth I realised there might possibly also be a downside here. In the case where the ext proc stream ends with OK status before it can deliver the initial header modifications, we would want to use the original headers to create the dataplane stream. But it could so happen that in the meantime , we have sent some request body messages to the proc server, so they will be lost. Or does the guaruntee that that proc server must emit all the messages it has received before ending stream hold here as well and and we will either get the header response and all the sent messages back?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,773 @@
...
+    a half-close when there is no message to send (i.e., if the client
+    never sent any message on the stream, or if the half-close is sent
+    after the filter has already sent the last message to the ext_proc
+    server).
+  - [end_of_stream](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L239):
+    For client messages, may be true if the client sent a half-close at
+    the same time as the last message.  For server messages, will always
+    be false.
+  - end_of_stream_without_message (new field being added in
+    https://github.com/envoyproxy/envoy/pull/38753): Will be set to true
```
rishesh007, line 597: Do we need to send `end_of_stream` or `end_of_stream_without_message` to indicate half close even if the processing mode for c2s message is not enabled?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,773 @@
...
+[`observability_mode`](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L283)
+config field is set to true, the contents of the data plane RPC events
+are sent to the ext_proc server, but the ext_proc server is not expected
+to make any modifications to the data plane RPC, so the data plane
+RPC proceeds without waiting for a response from the ext_proc server.
+Note that in this mode, all messages on the ext_proc stream will have the
+[`observability_mode`](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L126)
+field set.
+
+#### Flow Control
```
rishesh007, line 278: Let’s say the gRPC server has already started sending S2C messages, and those messages are being forwarded to the proc server. However, before any response is received from the proc server for those S2C messages, the server trailers arrive from the upstream server.

In this scenario, when the trailer `ProcessingMode` is set to `SKIP`, what should be the expected behavior of the filter? Should the filter immediately close the processing stream and continue forwarding the trailers downstream, or is there another recommended way to handle this case?

Edit: Same scenario for ServerInitialMetadata instead of S2C messages

A93-xds-ext-proc.md
markdroth: Good catch!  In the context of the ext_proc protocol, it does make sense that they'd represent Trailers-Only as headers rather than trailers.

I've updated the text here accordingly.

A93-xds-ext-proc.md
markdroth: If the ext_proc filter is configured to send request or response messages to the ext_proc server, then the ext_proc server must use the drain procedure described in the next section before closing the stream.  I've updated the text in the next section to attempt to clarify this.

A93-xds-ext-proc.md
markdroth: Thanks for calling this out!

This made me notice that if the response body send mode is `GRPC`, then the response trailer send mode cannot be `SKIP`, because otherwise the ext_proc server cannot indicate when it's finished sending messages.  (Note that Envoy has this [same requirement for `FULL_DUPLEX_STREAMED` mode](https://github.com/envoyproxy/envoy/blob/839f03a22c61d59f8b7f19902e6971cfd1ba5707/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto#L103), which is the mode that `GRPC` mode is based on.)  I've noted this restriction in the configuration section.

I don't think this is a problem for server initial metadata, because (as @kannanjgithub pointed out in a comment above), the ext_proc protocol represents Trailers-Only as headers rather than trailers, and that's the only case where we'd see trailers before headers.

A93-xds-ext-proc.md
markdroth: I don't think we need to support response attributes right now.  We can consider adding them later if we encounter a use case that needs them.

A93-xds-ext-proc.md
markdroth: No, if we're not configured to send a given event on the ext_proc stream, then we would not send that event just to indicate an EOS.  The EOS is essentially an attribute of the event, and if the ext_proc server isn't seeing that event, it doesn't need to know about the EOS.

A93-xds-ext-proc.md
markdroth: Good catch.  I've changed it to use INTERNAL instead of UNAVAILABLE.

markdroth: Changed to use INTERNAL instead of UNAVAILABLE.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,776 @@
...
+ext_proc stream failed with a non-OK status.
+
+#### Header Mutations
+
+The ext_proc filter will leverage the same header mutation types as
+defined in the ext_authz filter in [A92].
+
+When responding to a client headers, server headers, or server trailers
+event, the ext_proc server can return a
+[`HeaderMutation`](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L369)
```
rishesh007, line 796: If a header named 'xyz' is marked as immutable (header mutation is not allowed), but its `HeaderValueOption.AppendAction` is set to `kAddIfAbsent`, should the RPC fail if the header is already present in the metadata? Since the header is already there, no actual mutation would take place.

A93-xds-ext-proc.md
eshitachandwani: The gRFC now says that : `Note that if response_body_mode is set to GRPC, then response_header_mode must be set to SEND.` but @markdroth your comment and the [FULL_DUPLEX_STREAMED mode](https://github.com/envoyproxy/envoy/blob/839f03a22c61d59f8b7f19902e6971cfd1ba5707/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto#L103) says that  **response trailer** send mode cannot be `SKIP` ?  Am I missing something?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,784 @@
...
+    [`NONE`](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/processing_mode.proto#L66)
+    and `GRPC` (to be added).  Note that if response_body_mode is set to
+    `GRPC`, then response_header_mode must be set to `SEND`.
+  - We ignore the request_trailer_mode field, since gRPC never sends
+    request trailers.
+- [request_attributes](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L188)
+  and
+  [response_attributes](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto#L195):
+  Attributes to be sent to ext_proc server along with client-to-server
+  and server-to-client events, respectively.  The set of supported
```
kannanjgithub: Envoy sends them only on the first request for the respective client-server or server-client message to ext_proc.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,879 @@
...
+set a default for that server instance, so the only alternative here would
+be to have the ext_proc server instance set this on a per-stream basis,
+which would require an additional round-trip and therefore hurt latency.
+This unfortunately means that the ext_proc server instance has a bit less
+control: as soon as it receives these initial values, it can immediately
+send a window update that reduces the window, but it must be prepared
+to handle any data that the sender has already sent.
+
+As the receiver reads data, it must send back a window update telling
+the sender that it can send more data.  The window update can be
```
rishesh007: Will this result in an immediate call? It could potentially double the number of calls (2x).

A93-xds-ext-proc.md
```diff
@@ -277,41 +277,106 @@ field set.
...
+positive or negative.  The sender must immediately add this value to
+its available flow control window for sending.
+
+The ext_proc filter will use this mechanism to handle push-back on each
+path by not considering reads complete (and therefore releasing the flow
+control back to the sender) until its corresponding write has passed
+flow control.  For example, for path (1) above, when reading client
+messages from downstream, the filter will not release flow control back
+to the downstream until its write to the ext_proc sidestream has cleared
```
eshitachandwani: What does `until its write to the ext_proc sidestream has cleared flow control.` mean? Does it mean until we have written that message to side stream or does it mean until we have received window update for that message?

A93-xds-ext-proc.md
```diff
@@ -277,41 +277,106 @@ field set.
...
+that message exceeds the available flow control window.  At that point,
+the flow control window will be negative and the sender must not send
+the next message until it becomes positive.  This means that the
+receiver can't completely cap the amount of bytes it will receive, but
+it is necessary to deal with the fact that GRPC body send mode requires
+sending complete messages at a time, rather than splitting them up into
+DATA frames as HTTP/2 does, which would allow making incremental
+progress.
+
+The initial window sizes for all four paths are set by the filter in
```
kannanjgithub, line 342: Where are FlowControlInit initial window sizes to be set by the filter taken from? Should we add them to `GrpcService` config?

A93-xds-ext-proc.md
markdroth: Good catch!  That was a typo.  Fixed.

A93-xds-ext-proc.md
markdroth: I think each data plane implementation needs to figure this out on its own.  A simple implementation would be to just use hard-coded values.  But data planes may ultimately want some mechanism to dynamically tune these values when they are under memory pressure.  In C-core, we plan to eventually use our ResourceQuota mechanism to do that.

I don't think it's appropriate to set these values from the control plane, because the control plane cannot know how much memory pressure a given data plane is under.

A93-xds-ext-proc.md
markdroth: This isn't changing the number of calls at all, but it can change the number of ProcessingResponse messages sent on the stream, which may be what you meant here.

Note that in the common case where the ext_proc server passes messages through with minimal or no changes, I would generally expect that as it processes each message, it will need to send that message back to the ext_proc filter.  In that case, it won't increase the number of ProcessingResponse messages sent on the stream at all, because the ext_proc server can send the window update in the same ProcessingResponse that it's sending anyway to send back the message for the data plane stream.

A93-xds-ext-proc.md
markdroth: We don't actually have to have it fully written to the sidestream, but we do have to have allocated enough flow control to send it at both the ext_proc layer and the HTTP/2 layer.

I've attempted to clarify the wording here.

A93-xds-ext-proc.md
markdroth: Good catch.  Updated.

A93-xds-ext-proc.md
markdroth: I've clarified this behavior in https://github.com/grpc/proposal/pull/510.

A93-xds-ext-proc.md
rishesh007: In the original design document, `ProcessingResponse` was planned as a `oneof` message. That's why I asked the question, if it were a `oneof`, the number of messages will increase. We may need to update the document if that change hasn't already been reflected.

That said, I agree with your explanation, and it addresses my concern.

markdroth: The `ProcessingResponse` message does include a `oneof`, but the window update is not part of that `oneof`.  See https://github.com/envoyproxy/envoy/pull/45509 for the proposed API change.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,884 @@
...
+ext_proc server instance, but there is no connection-level handshake to
+set a default for that server instance, so the only alternative here would
+be to have the ext_proc server instance set this on a per-stream basis,
+which would require an additional round-trip and therefore hurt latency.
+This unfortunately means that the ext_proc server instance has a bit less
+control: as soon as it receives these initial values, it can immediately
+send a window update that reduces the window, but it must be prepared
+to handle any data that the sender has already sent.
+
+As the receiver reads data, it must send back a window update telling
```
eshitachandwani: The design doc says : `The receiver SHOULD reset the GRPC stream with a protocol error otherwise.` Should we add that here too? Also what does it mean to reset the stream? Is it equivalent to the stream failing with non-OK status or something else?

markdroth: Thanks for calling this out.

Resetting the stream just means cancelling the stream.

I've added a paragraph about this.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,896 @@
...
+    Contains the serialized message.  Will be empty if the client sends
+    a half-close when there is no message to send (i.e., if the client
+    never sent any message on the stream, or if the half-close is sent
+    after the filter has already sent the last message to the ext_proc
+    server).
+  - [end_of_stream](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/service/ext_proc/v3/external_processor.proto#L239):
+    For client messages, may be true if the client sent a half-close at
+    the same time as the last message.  For server messages, will always
+    be false.
+  - end_of_stream_without_message (new field being added in
```
eshitachandwani, line 596: Is there a restriction here too similar to response received by the proc server that EOS without message should be set to true along with EOS and EOS should not be false if EOS without message is true ?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,896 @@
...
+
+The ext_proc filter will use this mechanism to handle push-back on each
+path by not considering reads complete (and therefore releasing the flow
+control back to the sender) until its corresponding write has passed
+flow control at both the ext_proc and HTTP/2 layers.  For example, for
+path (1) above, when reading client messages from downstream, the filter
+will first wait until it has allocated enough ext_proc flow control to
+send the message on the sidestream and then wait for the write to the
+sidestream to clear flow control at the HTTP/2 layer for the sidestream,
+and only then will it release flow control back to the downstream.
```
kannanjgithub, line 368: For path (1) it is clear how the flow control for the downstream application should be held back until the write to the side stream has completed. What about path (4)? There seems to be 2 ways of doing this - either at the transport level by accumulating `request(int)` meant for ext_proc until there are no outstanding `ProcessingResponse`s for response bodies yet to be delivered to the downstream application, or by deferring sending the `ClientWindowUpdate` for path (4) to ext-proc. Which one should we implement? The first approach has deadlock risk.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,896 @@
...
+- `flow_control_init` (new field being added in
+  https://github.com/envoyproxy/envoy/pull/45509): Will be populated on
+  the first message sent on the ext_proc stream when not in
+  observability mode.  All four initial window size fields must be set.
+  A good default initial value is 65536, but implementations can raise
+  or lower that based on their own memory management requirements.
+- `client_window_update` (new field being added in
+  https://github.com/envoyproxy/envoy/pull/45509): Normally, whenever the
+  filter reads a request body chunk from the ext_proc side-stream, it will
+  send a window update setting the
```
ejona86: We probably want to accumulate window updates and only send eagerly when at least half the initial window can be returned. That allows batching of the window update and greatly extends the amount of time available to piggy-back on an already-being-sent processing request.

A93-xds-ext-proc.md
eshitachandwani: Hey @markdroth gentle ping on this !

A93-xds-ext-proc.md
kannanjgithub: I think it is clear that we should go with the `ClientWindowUpdate` withholding for pushback on path (4).

A93-xds-ext-proc.md
kannanjgithub: We should also send client window updates eagerly if the sidesidestream's window just became 0 or negative, to avoid deadlocks.

A93-xds-ext-proc.md
markdroth: I'm not sure I fully understand your question.  However, the intent here is that `end_of_stream_without_message` is relevant only if `end_of_stream` is true; if `end_of_stream` is false, then the ext_proc server shouldn't look at `end_of_stream_without_message` at all.

A93-xds-ext-proc.md
markdroth: I've added a paragraph to the "Flow Control" section talking about options that implementations should consider, and I'm referring back to that here.

markdroth:
As per offline discussion, I've removed the backend_service label from the metrics, because the ext_proc filter will run above load balancing, which means that we don't actually know the leaf cluster in a reliable way.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,906 @@
...
+
+| Name        | Disposition | Description |
+| ----------- | ----------- | ----------- |
+| grpc.target | required | The target of the gRPC channel in which ext_proc is used, as the defined in [A66]. |
+
+The following client-side metrics will be exported:
+
+| Name          | Type  | Unit  | Labels  | Description |
+| ------------- | ----- | ----- | ------- | ----------- |
+| grpc.client_ext_proc.client_headers_duration | Histogram | s | grpc.target, grpc.lb.backend_service | Time between when the ext_proc filter sees the client's headers and when it allows those headers to continue on to the next filter. |
```
eshitachandwani: this table still has `grpc.lb.backend_service`

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,906 @@
...
+control back to the sender) until its corresponding write has passed
+flow control at both the ext_proc and HTTP/2 layers.  For example, for
+path (1) above, when reading client messages from downstream, the filter
+will first wait until it has allocated enough ext_proc flow control to
+send the message on the sidestream and then wait for the write to the
+sidestream to clear flow control at the HTTP/2 layer for the sidestream,
+and only then will it release flow control back to the downstream.
+
+If the ext_proc filter receives more data than allowed by flow control
+(i.e., if it receives more data once the available flow control window
```
ejona86: This is not precise enough if a negative window update was sent. Do we want to just say ext_proc filter won't use negative window updates at this time?

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,906 @@
...
+ext_proc server instance, but there is no connection-level handshake to
+set a default for that server instance, so the only alternative here would
+be to have the ext_proc server instance set this on a per-stream basis,
+which would require an additional round-trip and therefore hurt latency.
+This unfortunately means that the ext_proc server instance has a bit less
+control: as soon as it receives these initial values, it can immediately
+send a window update that reduces the window, but it must be prepared
+to handle any data that the sender has already sent.
+
+As the receiver reads data, it must send back a window update telling
```
ejona86: s/reads/consumes/ ? This "read" language shows up multiple places, but the important part is that it has been _processed_. Yes, I see the next paragraph says how we won't consider a "read complete" until it passes flow control, but that is nuanced as you aren't actually talking about "reading the message from the side stream" as that _will_ complete and we're going to even read _more_ messages from the side stream even without the "read completing". It's instead some higher-level conceptual model and it isn't clear whether each "read" is the actual read or the conceptual model. I had decided against making this comment, but then I read "One simple implementation would be that as the filter reads each message from the side-stream" and again I thought "people are going to get the wrong idea."

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,906 @@
...
+from the side-stream, it will send back a window update refilling the
+number of bytes it just read.  However, implementations are also free
+to determine window updates more dynamically based on memory usage; for
+example, the C-core will likely use `ResourceQuota` to dynamically
+resize flow control windows.  Note that care must be taken to avoid
+impacting performance or causing deadlocks.  For example, to avoid
+chattiness on the wire, it may be desirable to wait (up to at least
+some limit) for the next message being sent anyway rather than sending
+a message containing only a window update, but at the same time,
+delaying a window update may cause unnecessary delays on the sender
```
ejona86: Delaying returning half of the window does _not_ cause delays, unless the window is already too small. We should not shy away from delaying half the window.

kannanjgithub: @markdroth A gentle reminder about this comment.

A93-xds-ext-proc.md
markdroth: I've tweaked the wording a bit to try to address this.

A93-xds-ext-proc.md
markdroth: I've removed this paragraph.  Given the fact that window updates can be negative, I don't think we can actually tell when the peer is behaving badly.  This is unfortunate, but I think it's an inherent part of the trade-offs we've made in the flow control design.

A93-xds-ext-proc.md
markdroth: @ejona86 I don't understand that comment.  I think delaying a window update *can* cause delays.  Consider the case where the sender needs to send three messages of 10 bytes each and the available flow control window is 15.  The sender will be able to send the first two messages, after which the available window will be -5, so it will need to wait before sending the third message.  If the receiver sends an immediate window update of 10 after it processes the first message, then the sender can immediately send the third message.  But if the sender waits to send the update until after it processes the second message -- or even longer, if it wasn't going to send another message for some other reason anyway -- then I think that will cause a delay.

A93-xds-ext-proc.md
markdroth: Thanks for catching this!  Removed.

A93-xds-ext-proc.md
ejona86: You are assuming infinite throughput.

If the window is smaller than the BDP, then you can see a latency increase. But that is because the window is too small and throughput is artificially limited and not because of the delay. Basically, if the window is too small you can burst briefly, but that's it. Delaying the window update has no impact to the throughput. The problem is the window is too small, not the delay.

If the window of 15 was appropriate, then that means it takes at least ½ RTT to send 7.5 bytes on-the-wire, in which case the sender will get a window update before the sender consumes all 15 bytes of window, modulo aliasing because we process things message-by-message (e.g., four 8 byte messages would be fine.)

Note that sending updates too frequently can reduce throughput and increase CPU usage because of inefficient processing. Each window update is highly likely to be its own syscall and packet, because it is small.

markdroth: Hmm.  How would we factor that kind of calculation into the ext_proc implementation?  Are you expecting that implementations will actually do some sort of BDP calculation, like we do in the H2 transport?  How would an implementation decide whether to delay a window update, and how long to delay it for?

ejona86: Definitely no BDP estimation now. As I said, this isn't really a "delayed window update" problem. In the future if people want increased throughput then we do optimizations and such, which could include BDP or just configuration. But it is really unrelated.

I'd suggest:
* Eagerly send back (using its own message) a window update if at least half the window is to be returned
* Always piggy-back available window on any sent message

That's it. No timers. It could be delayed forever, but we know the other side can still send.

markdroth: Updated as per our offline conversation.

A93-xds-ext-proc.md
kannanjgithub: Rather than let the window size be breached, can we devise a nonce-ACK for the window updates so that we can actually enforce it?

markdroth: I'm not sure what you mean by a "nonce-ACK", but I don't think we can prevent the window from going negative.  The problem is that we cannot guarantee that the receiver will ever give us enough window to send a single message, and we still need to be able to make forward progress if it never happens.  With HTTP/2-level flow control, this isn't an issue, because a smaller window still allows us to make incremental progress by sending individual DATA frames, even if those frames contain only a subset of a gRPC message.  However, with the ext_proc GRPC body send mode, we don't have a way to make incremental progress -- any given gRPC message must be sent all at once.  We deal with this by saying that as long as the available window is greater than zero, the sender may send a single message, even if that message makes the available window go negative.

This is definitely sub-optimal, but we'd need to add a lot of complexity to fix it, because we'd basically need some way for the sender to tell the receiver how much window they need to send the next message.  I don't think that level of complexity in the protocol is really worthwhile, especially since implementations can still deal with this kind of thing in a reactive way (e.g., like how gRPC C++'s `ResourceQuota` mechanism works).

kannanjgithub: My message was not accurate. We were already allowing the window to go negative to allow full gRPC messages to be sent, my question was not meant to be about that, but about the removal of treating it as a flow control violation if the window was already zero or negative and a message was still received. You removed it because we support negative window updates and there is a race condition in that the receiver doesn't know if the sender had received the window size reduction when it sent the message. To this problem I was suggesting a nonce/epoch to be sent by the receiver with every window update, and the sender to ACK it. So now the receiver can know from the sender's last ACK what window it was supposed to have adhered to, no matter any in-flight dynamic window update it hasn't yet received.

markdroth: Ah, okay, I understand what you're suggesting.  I agree that that would work, but I don't think it's really worth the complexity.  The flow control behavior is already adding complexity for ext_proc server implementations to support, and we don't want to make that problem worse.

A93-xds-ext-proc.md
```diff
@@ -0,0 +1,915 @@
...
+The new `GRPC` mode will be the only processing mode supported in gRPC.
+It is be desirable for Envoy to implement the same mode, so that users
+can switch back and forth between proxy and proxyless data planes without
+breaking their ext_proc servers.
+
+gRPC implementations should structure their xDS HTTP filter APIs such that
+the filter has access to the serialized bytes of each message rather than
+the deserialized message, to avoid the overhead of repeated
+serialization/deserialization on each message.  In other words, on the
+client side, the filters should run after the message is serialized, and
```
kannanjgithub, line 235: In gRPC-Java, `RetriableStream` supports retries by buffering the request message POJO and calling `method.streamRequest(msg)` to produce a fresh `InputStream` on each attempt. When [RawMessageClientInterceptor](https://github.com/grpc/grpc-java/blob/b8f745a362217ecccdae5a0bd5578445e237b64f/xds/src/main/java/io/grpc/xds/XdsNameResolver.java#L1132) (added for this purpose) converts the method to `MethodDescriptor<InputStream, InputStream>`, `RetriableStream` receives and buffers a raw `InputStream`, which is exhausted on attempt 1, causing retry attempts to send an empty 0-byte payload (issue [#13010](https://github.com/grpc/grpc-java/issues/13010)).

To support retries under `RawMessageClientInterceptor`, we would need to buffer the serialized payload into an in-memory `byte[]` or `ByteString` before passing it downstream. However, doing this unconditionally even when ext_proc isn't involved breaks gRPC-Java's zero-copy `Drainable` optimization (which writes protos directly into Netty buffers) and forces an extra heap allocation/copy on every request (both of which we will have to endure for ext_proc usecases anyway), severely impacting high-throughput users like GCS.

Can we ensure `RawMessageClientInterceptor` is only added to the interceptor chain when an `ext_proc` filter is actually configured on the route?

markdroth: It seems fine to do this only when it's actually needed, but I would suggest not tying that to the ext_proc filter specifically.  The ext_proc filter is *currently* the only filter that will look at the payloads, but in principle any xDS HTTP filter could do so, and it's quite possible that we will have other such filters in the future.  I don't think we want to wind up in a place where there's a list of specific filters instances that need this behavior.

Instead, consider something like having a method on the xDS HTTP filter factory to indicate whether the filter needs access to the payloads, so that you can change the behavior if any filter needs it.

PR #485, opened 2025-03-19
A94: OTel metrics for Subchannels

yashykt:
Closes https://github.com/grpc/grpc/issues/34886

### DIFF
--- a/A78-grpc-metrics-wrr-pf-xds.md
+++ b/A78-grpc-metrics-wrr-pf-xds.md
@@ -4,7 +4,7 @@ A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient
 * Approver: @ejona86, @dfawley
 * Status: {Draft, In Review, Ready for Implementation, Implemented}
 * Implemented in: <language, ...>
-* Last updated: 2024-09-24
+* Last updated: 2025-07-01
 * Discussion at: https://groups.google.com/g/grpc-io/c/A2Mqz8OMDys
 * Updated by: [A88: xDS Data Error Handling](A88-xds-data-error-handling.md), [A94: OTel metrics for Subchannels](A94-subchannel-otel-metrics.md)
 
@@ -103,7 +103,7 @@ The following metrics will be exported:
 | grpc.lb.wrr.endpoint_weight_stale | Counter | {endpoint} | grpc.target, grpc.lb.locality | Number of endpoints from each scheduler update whose latest weight is older than the expiration period. |
 | grpc.lb.wrr.endpoint_weights | Histogram | {weight} | grpc.target, grpc.lb.locality | Weight of each endpoint, recorded on every scheduler update. Endpoints without usable weights will be recorded as weight 0. |
 
-### Pick First LB Policy
+### [Outdated] Pick First LB Policy (Updated by [A94](A94-subchannel-otel-metrics.md))
 
 The Pick First LB policy predates the gRFC process but was updated in
 [A62].  We propose to add the following metrics to it.
--- a/A78-grpc-metrics-wrr-pf-xds.md
+++ b/A78-grpc-metrics-wrr-pf-xds.md
@@ -6,6 +6,7 @@ A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient
 * Implemented in: <language, ...>
 * Last updated: 2024-09-24
 * Discussion at: https://groups.google.com/g/grpc-io/c/A2Mqz8OMDys
+* Updated by: [A94: OTel metrics for Subchannels](A94-subchannel-otel-metrics.md)
 
 ## Abstract
 
--- a/A94-subchannel-otel-metrics.md
+++ b/A94-subchannel-otel-metrics.md
@@ -28,6 +28,7 @@ to client-side per-attempt metrics. This label has xds cluster information.
 *   [A18]: TCP User Timeout
 *   [A61]: IPv4 and IPv6 Dualstack Backend Support
 *   [A66]: OpenTelemetry Metrics
+*   [A74]: xDS Config Tears
 *   [A78]: gRPC OTel Metrics for WRR, Pick First, and XdsClient
 *   [A79]: Non-per-call Metrics Architecture
 *   [A89]: Backend Service Metric Label
@@ -37,6 +38,7 @@ to client-side per-attempt metrics. This label has xds cluster information.
 [A18]: A18-tcp-user-timeout.md
 [A61]: A61-IPv4-IPv6-dualstack-backends.md
 [A66]: A66-otel-stats.md
+[A74]: A74-xds-config-tears.md
 [A78]: A78-grpc-metrics-wrr-pf-xds.md
 [A79]: A79-non-per-call-metrics-architecture.md
 [A89]: A89-backend-service-metric-label.md
@@ -72,10 +74,12 @@ grpc.lb.locality        | Optional    | The locality to which the traffic is bei
 grpc.disconnect_error   | Optional    | Reason for disconnection.
 grpc.security_level     | Optional    | Denotes the security level of the connection. Allowed values - "none", "integrity_only" and "privacy_and_integrity".
 
-The resolver attributes for the `grpc.lb.backend_service` and `grpc.lb.locality`
-labels (defined in [A89] and [A78] respectively) will be passed into the
-subchannel. This implies that the subchannel will be recreated when these
-attributes change.
+The subchannel needs to be passed attributes with the values for the
+`grpc.lb.backend_service` and `grpc.lb.locality` labels (defined in [A89] and
+[A78] respectively). This implies that the subchannel will be recreated when
+these attributes change. Since currently, only xDS is using these labels, the
+attributes will be set for each endpoint by cds (post-[A74]) or
+xds_cluster_resolver (pre-[A74]) LB policies.
 
 List of allowed values for `grpc.disconnect_error` -
 
--- a/A94-subchannel-otel-metrics.md
+++ b/A94-subchannel-otel-metrics.md
@@ -1,10 +1,10 @@
 ## A94: OTel metrics for Subchannels
 
-*   Author(s): Yash Tibrewal (yashkt@)
+*   Author(s): Yash Tibrewal (yashykt@)
 *   Approver: Mark Roth (@markdroth)
 *   Status: In Review
 *   Implemented in:
-*   Last updated: Mar 18, 2025
+*   Last updated: 2025-04-02
 *   Discussion at: https://groups.google.com/g/grpc-io/c/iMdK7r4E5tU
 
 ## Abstract
@@ -20,21 +20,26 @@ disconnections for subchannels and connection attempts made for those
 subchannels. These metrics do not currently contain information on the reason
 for disconnection, the xds locality or the cluster information.
 
-[A89: Backend Service Metric Label](https://github.com/grpc/proposal/pull/471)
-is a proposal to introduce a new optional label `grpc.lb.backend_service` to
-client-side per-attempt metrics. This label has xds cluster information.
+[A89] is a proposal to introduce a new optional label `grpc.lb.backend_service`
+to client-side per-attempt metrics. This label has xds cluster information.
 
 ### Related Proposals:
 
-*   [A66: OpenTelemetry Metrics]
-*   [A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient]
-*   [A79: Non-per-call Metrics Architecture]
-*   [A89: Backend Service Metric Label]
+*   [A8]: Client-side Keepalive
+*   [A18]: TCP User Timeout
+*   [A61]: IPv4 and IPv6 Dualstack Backend Support
+*   [A66]: OpenTelemetry Metrics
+*   [A78]: gRPC OTel Metrics for WRR, Pick First, and XdsClient
+*   [A79]: Non-per-call Metrics Architecture
+*   [A89]: Backend Service Metric Label
 
-[A66: OpenTelemetry Metrics]: A66-otel-stats.md
-[A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient]: A78-grpc-metrics-wrr-pf-xds.md
-[A79: Non-per-call Metrics Architecture]: A79-non-per-call-metrics-architecture.md
-[A89: Backend Service Metric Label]: https://github.com/grpc/proposal/pull/471
+[A8]: A8-client-side-keepalive.md
+[A18]: A18-tcp-user-timeout.md
+[A61]: A61-IPv4-IPv6-dualstack-backends.md
+[A66]: A66-otel-stats.md
+[A78]: A78-grpc-metrics-wrr-pf-xds.md
+[A79]: A79-non-per-call-metrics-architecture.md
+[A89]: A89-backend-service-metric-label.md
 
 ## Proposal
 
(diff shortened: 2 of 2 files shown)


yashykt:
cc @joybestourous

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+enough time for users to transition to the new metrics. For example,
+implementations should report both the old pick-first metrics and the new
+subchannel metrics for 2 releases, and then remove the old pick-first metrics.
+
+Label Name              | Disposition | Description
+----------------------- | ----------- | -----------
+grpc.target             | Required    | Indicates the target of the gRPC channel (defined in [A66](A66-otel-stats.md).)
+grpc.lb.backend_service | Optional    | The backend service to which the RPC was routed (defined in [A89](https://github.com/grpc/proposal/pull/471))
+grpc.lb.locality        | Optional    | The locality to which the traffic is being sent. This will be set to the resolver attribute passed down from the weighted_target policy, or the empty string if the resolver attribute is unset. (defined in [A78](A78-grpc-metrics-wrr-pf-xds.md))
+grpc.disconnect_error   | Optional    | Reason for disconnection
```
joybestourous, line 67: we have an internal version of these metrics which also includes a `tls` label to track the security of the connection. Useful for discoverability of non-compliant services and for large migrations, though I do respect that in the general case it should be easy to check this from your DialOptions. curious for your thoughts on including this?

joybestourous (commented):
thanks for including the open_connection metric, this is super helpful for us.

A94-subchannel-otel-metrics.md
yashykt: Hey, thanks for taking a look! For security, I think there are a few questions that need to be answered. 
1) What security information do we want? (none/tls/mtls/other or none/integrity/privacy+integrity)
2) Do we want it per-channel or per-connection? (Given that with xDS, we might receive configuration from the control plane, per-connection makes more sense.)
3) Do we want to include it in these metrics or do we want a separate metric?
4) Is it feasible to collect it here?

If we do include the tls label to these metrics, I would want to add it as an optional label, so that -
1) it's not on by default
2) it does not block this proposal

I would also want @gtcooke94 to weigh in here from a security perspective.

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+PickFirst load-balancing policy were proposed that provide observability on
+disconnections for subchannels and connection attempts made for those
+subchannels. These metrics do not currently contain information on the reason
+for disconnection, the xds locality or the cluster information.
+
+[A89: Backend Service Metric Label](https://github.com/grpc/proposal/pull/471)
+is a proposal to introduce a new optional label `grpc.lb.backend_service` to
+client-side per-attempt metrics. This label has xds cluster information.
+
+### Related Proposals:
```
markdroth, line 27: Please include A8 and A18 in this list.

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+number of open connections by subtracting `grpc.subchannel.disconnections` from
+`grpc.subchannel.connection_attempts_succeeded`. This method does not work for
+exporters recording counters per period in a way that does not allow for a
+simple subtraction of the two counters
+(https://github.com/grpc/grpc/issues/34886).
+
+Adding an explicit metric that records the number of open connections avoids
+this.
+
+### Use “connection timed out” as disconnect error for gRPC Keepalive timeouts
```
markdroth, line 118: This section is a little unclear.  I think the title should be something like "Combining connection timeouts and keepalives into a single disconnection error", and we should explain that since we generally use the same timeouts for both keepalives and TCP_USER_TIMEOUT, it will essentially be a race as to which one fires first, so there's not really any point in splitting these into two different buckets.

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+
+List of allowed values for `grpc.disconnect_error` -
+
+Error string         | Description
+-------------------- | -----------
+GOAWAY <ERROR_CODE>  | HTTP2 GOAWAY frame with error code for example (“GOAWAY NO_ERROR”, “GOAWAY PROTOCOL_ERROR”, “GOAWAY ENHANCE_YOUR_CALM”). The list of error codes is available in [RFC 9113](https://www.rfc-editor.org/rfc/rfc9113.html#name-error-codes).
+subchannel shutdown  | The subchannel was shutdown. This can happen due to reasons such as the parent channel shutting down, channel becoming idle, the load balancing policy changing due to a resolver update or a change in list of endpoint addresses.
+connection reset     | Connection was reset (eg. ECONNRESET, WSAECONNERESET)
+connection timed out | Connection timed out, also includes connections closed due to gRPC keepalives.
+connection aborted   | Connection was aborted
```
markdroth, line 77: I don't remember discussing this one -- not sure if we didn't, or if I just forgot.

What exactly triggers this one?  How is it different from "connection reset"?

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+For a given connection, there can be multiple reasons reported to the subchannel
+for disconnection. For example, a connection could have seen a GOAWAY frame with
+`ENHANCE_YOUR_CALM` and then a socket error Broken Pipe. In such cases, the
+first seen reason should be chosen, `GOAWAY ENHANCE_YOUR_CALM` in this case.
+
+We might add more error cases to this in the future.
+
+## Rationale
+
+### Renaming pick-first metrics
```
markdroth, line 90: Did we discuss what we're going to do with the existing PF metrics?  I guess we can't stop supporting them, but maybe we should mark them as deprecated somehow?

We should also probably add an "Updated by: gRFC 94" header to the top of A78.

A94-subchannel-otel-metrics.md
markdroth: > 1. What security information do we want? (none/tls/mtls/other or none/integrity/privacy+integrity)

I would not want to encode this as "tls", because all TLS-specific knowledge is supposed to be in TlsCredentials, not in the core subchannel code, so encoding TLS-specific knowledge here would break an important abstraction.  But I might be open to a label that indicates the security level, as defined in [gRFC L62](https://github.com/grpc/proposal/blob/master/L62-core-call-credential-security-level.md).

> 2. Do we want it per-channel or per-connection? (Given that with xDS, we might receive configuration from the control plane, per-connection makes more sense.)

I think we would definitely need this to be per-connection, not per-channel.  Even without xDS, we have some cases internally where different connections on the same channel will wind up with different levels of encryption.  In the general case, I don't think we can say whether all of the connections are going to be encrypted based on the channel credentials used.

> 3. Do we want to include it in these metrics or do we want a separate metric?

I'm on the fence about that.  On one hand, separation of concerns suggests that it would be better to handle this separately, and I could easily imagine adding some TLS-specific metrics in the TlsCredentials code.  But on the other hand, if the goal is to detect when there are unecrypted connections happening, exposing that information centrally may make the most sense.

> 4. Is it feasible to collect it here?

Good question.  I think it wouldn't be too hard to get that information in C-core, but I'm not sure about the other languages.  I'd like input from @ejona86, @dfawley, and @murgatroid99 on that.

markdroth (commented):
Thanks for writing this up, Yash!

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
+## A94: OTel metrics for Subchannels
+
+*   Author(s): Yash Tibrewal (yashkt@)
```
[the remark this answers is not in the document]
yashykt: yes, thanks!

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+*   Approver: Mark Roth (@markdroth)
+*   Status: In Review
+*   Implemented in:
+*   Last updated: Mar 18, 2025
```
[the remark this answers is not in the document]
yashykt: done

A94-subchannel-otel-metrics.md
yashykt: done

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+client-side per-attempt metrics. This label has xds cluster information.
+
+### Related Proposals:
+
+*   [A66: OpenTelemetry Metrics]
+*   [A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient]
+*   [A79: Non-per-call Metrics Architecture]
+*   [A89: Backend Service Metric Label]
+
+[A66: OpenTelemetry Metrics]: A66-otel-stats.md
```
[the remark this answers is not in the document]
yashykt: done

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+grpc.lb.backend_service | Optional    | The backend service to which the RPC was routed (defined in [A89](https://github.com/grpc/proposal/pull/471))
+grpc.lb.locality        | Optional    | The locality to which the traffic is being sent. This will be set to the resolver attribute passed down from the weighted_target policy, or the empty string if the resolver attribute is unset. (defined in [A78](A78-grpc-metrics-wrr-pf-xds.md))
+grpc.disconnect_error   | Optional    | Reason for disconnection
+
+List of allowed values for `grpc.disconnect_error` -
+
+Error string         | Description
+-------------------- | -----------
+GOAWAY <ERROR_CODE>  | HTTP2 GOAWAY frame with error code for example (“GOAWAY NO_ERROR”, “GOAWAY PROTOCOL_ERROR”, “GOAWAY ENHANCE_YOUR_CALM”). The list of error codes is available in [RFC 9113](https://www.rfc-editor.org/rfc/rfc9113.html#name-error-codes).
+subchannel shutdown  | The subchannel was shutdown. This can happen due to reasons such as the parent channel shutting down, channel becoming idle, the load balancing policy changing due to a resolver update or a change in list of endpoint addresses.
```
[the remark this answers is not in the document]
yashykt: fixed :)

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,128 @@
...
+connection reset     | Connection was reset (eg. ECONNRESET, WSAECONNERESET)
+connection timed out | Connection timed out, also includes connections closed due to gRPC keepalives.
```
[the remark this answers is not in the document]
yashykt: done

A94-subchannel-otel-metrics.md
yashykt: amended

A94-subchannel-otel-metrics.md
yashykt: I added "connection timed out", "connection reset" and "connection aborted" as the most common socket errors and implemented across platforms. There are some differences across platforms as to the exact implementation but overall - 
"connection reset" - a TCP RST was received
"connection timed out" - keepalive or tcp user timeout
"connection aborted" - The connection was aborted/closed by the host itself (software initiated - application/OS). This can happen due to reasons like resource exhaustion, unexpected/internal error.

markdroth: In C-core, how are we going to differentiate between these cases?  We would presumably need to expose the difference between them in the EventEngine API, and then we need to plumb that data through the transport and up to the subchannel.  How exactly will this work?

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,147 @@
...
+Implementations that have already implemented the pick-first metrics should give
+enough time for users to transition to the new metrics. For example,
+implementations should report both the old pick-first metrics and the new
+subchannel metrics for 2 releases, and then remove the old pick-first metrics.
+
+| Label Name              | Disposition | Description                          |
+| ----------------------- | ----------- | ------------------------------------ |
+| grpc.target             | Required    | Indicates the target of the gRPC     |
+:                         :             : channel (defined in [A66].)          :
+| grpc.lb.backend_service | Optional    | The backend service to which the RPC |
```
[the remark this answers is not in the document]
markdroth: Yeah, agreed, we need a plan for this.  Unfortunately, due to pre-existing differences in attribute semantics across languages, the details will probably be a little different in each language.

In C-core, we'll just pass these through as channel args that are part of the subchannel key.  I think we actually already have most of this, but we'll need to modify it slightly to be non-xDS-specific.

A94-subchannel-otel-metrics.md
joybestourous: > encoding TLS-specific knowledge here would break an important abstraction. But I might be open to a label that indicates the security level, as defined in [gRFC L62](https://github.com/grpc/proposal/blob/master/L62-core-call-credential-security-level.md).

Yeah I understand the concern with maintaining abstractions. The security level label would be sufficient imo!

> I think we would definitely need this to be per-connection, not per-channel. Even without xDS, we have some cases internally where different connections on the same channel will wind up with different levels of encryption. In the general case, I don't think we can say whether all of the connections are going to be encrypted based on the channel credentials used.

Heavy +1 to this, per-connection is the right granularity here imo. 

>  If the goal is to detect when there are unencrypted connections happening, exposing that information centrally may make the most sense.

This is exactly how we use this tag now, and it's helpful to have it on the connection. I think it would lose some value if not related to the client - server(or `grpc.target`) connection in question

A94-subchannel-otel-metrics.md
yashykt: EventEngine callbacks have a status parameter that we can use. The endpoint will annotate the status object with the correct socket error message. 

For the subchannel, in Core, we have connectivity state watchers on the transport that also have a status object that is part of the callback, so the same status object that can be propagated up to the subchannel that can decode the error message.

A94-subchannel-otel-metrics.md
yashykt: Added

A94-subchannel-otel-metrics.md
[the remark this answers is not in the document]
ejona86: Let's just define it as part of the subchannel key. In Go that is in the address attributes. In Java that is in the endpoint attributes (which don't impact equality today).

And then to define where we set these details: we're storing locality in the endpoint attributes already in cds (post-A74)/cluster_resolver (pre-A74) for WrrLocalityLB, but it is xds-specific. We can set BACKEND_SERVICE at the same time.

A94-subchannel-otel-metrics.md
```diff
@@ -28,6 +27,7 @@ to client-side per-attempt metrics. This label has xds cluster information.
 *   [A8]: Client-side Keepalive
 *   [A18]: TCP User Timeout
 *   [A61]: IPv4 and IPv6 Dualstack Backend Support
+*   [A62]: gRPC security level negotiation between call credentials and channels
```
markdroth: This is L62, not A62.

Also, you didn't actually add the link anywhere.

A94-subchannel-otel-metrics.md
```diff
@@ -63,20 +63,13 @@ enough time for users to transition to the new metrics. For example,
...
-:                         :             : attribute is unset (defined in       :
-:                         :             : [A78].)                              :
-| grpc.disconnect_error   | Optional    | Reason for disconnection.            |
+Label Name              | Disposition | Description
+----------------------- | ----------- | -----------
+grpc.target             | Required    | Indicates the target of the gRPC channel (defined in [A66].)
+grpc.lb.backend_service | Optional    | The backend service to which the RPC was routed (defined in [A89].)
+grpc.lb.locality        | Optional    | The locality to which the traffic is being sent. This will be set to the resolver attribute passed down from the weighted_target policy, or the empty string if the resolver attribute is unset (defined in [A78].)
+grpc.disconnect_error   | Optional    | Reason for disconnection.
+grpc.security_level     | Optional    | Denotes the security level of the connection. Allowed values - "none", "integrity only" and "privacy and integrity".
```
markdroth: Please replace the spaces in these values with underscores.

A94-subchannel-otel-metrics.md
markdroth: We're trying to move away from adding payloads on `absl::Status` in general, since that requires allocation.  And I think this approach would also broaden the EventEngine API by establishing annotations that EE implementations would need to know about.

I'm not sure what the right answer is here.  I suggest that you and I and @ctiller sit down together to come up with a plan here.

A94-subchannel-otel-metrics.md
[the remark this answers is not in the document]
markdroth: I agree with the direction here, although we may need to bikeshed a bit on how to phrase it.

Do Java and Go already have a concept of "subchannel key"?  I didn't think that was a thing without a subchannel pool.

How about we just say something like this:

"""
The resolver attributes for the `grpc.lb.backend_service` and `grpc.lb.locality` labels (defined in [A89] and [A78] respectively) will be passed into the subchannel.
"""

A94-subchannel-otel-metrics.md
yashykt: done

A94-subchannel-otel-metrics.md
yashykt: thanks!

A94-subchannel-otel-metrics.md
yashykt: discussed on vc

A94-subchannel-otel-metrics.md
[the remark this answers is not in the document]
yashykt: done

A94-subchannel-otel-metrics.md
```diff
@@ -40,6 +40,7 @@ to client-side per-attempt metrics. This label has xds cluster information.
 [A78]: A78-grpc-metrics-wrr-pf-xds.md
 [A79]: A79-non-per-call-metrics-architecture.md
 [A89]: A89-backend-service-metric-label.md
+[L62]: L62-grpc-security-level-negotiation.md
```
markdroth: This link is broken.  The filename is L62-core-call-credential-security-level.md.

markdroth (approved):

A78-grpc-metrics-wrr-pf-xds.md
```diff
@@ -6,7 +6,7 @@ A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient
 * Implemented in: <language, ...>
 * Last updated: 2024-09-24
 * Discussion at: https://groups.google.com/g/grpc-io/c/A2Mqz8OMDys
-* Updated by: [A88: xDS Data Error Handling](A88-xds-data-error-handling.md)
+* Updated by: [A88: xDS Data Error Handling](A88-xds-data-error-handling.md), [A94: OTel metrics for Subchannels](A94-subchannel-otel-metrics.md)
```
dfawley, line 9: Could we add a prefix to the section with the metrics to indicate that that part is obsolete and to see the new gRFC instead?

yashykt: done

A94-subchannel-otel-metrics.md
```diff
@@ -0,0 +1,146 @@
...
+implementations should report both the old pick-first metrics and the new
+subchannel metrics for 2 releases, and then remove the old pick-first metrics.
+
+Label Name              | Disposition | Description
+----------------------- | ----------- | -----------
+grpc.target             | Required    | Indicates the target of the gRPC channel (defined in [A66].)
+grpc.lb.backend_service | Optional    | The backend service to which the RPC was routed (defined in [A89].)
+grpc.lb.locality        | Optional    | The locality to which the traffic is being sent. This will be set to the resolver attribute passed down from the weighted_target policy, or the empty string if the resolver attribute is unset (defined in [A78].)
+grpc.disconnect_error   | Optional    | Reason for disconnection.
+grpc.security_level     | Optional    | Denotes the security level of the connection. Allowed values - "none", "integrity_only" and "privacy_and_integrity".
```
AgraVator, line 76: Can we keep it "integrity" instead of "integrity_only" to match the existing enums ?

yashykt: https://github.com/grpc/proposal/blob/master/L62-core-call-credential-security-level.md#add-enum-grpc_security_level-to-includegrpcgrpc_security_constantsh it does match

AgraVator: I see, its a bit different in Java...

A94-subchannel-otel-metrics.md
```diff
@@ -72,6 +72,11 @@ grpc.lb.locality        | Optional    | The locality to which the traffic is bei
 grpc.disconnect_error   | Optional    | Reason for disconnection.
 grpc.security_level     | Optional    | Denotes the security level of the connection. Allowed values - "none", "integrity_only" and "privacy_and_integrity".
 
+The resolver attributes for the `grpc.lb.backend_service` and `grpc.lb.locality`
```
ejona86: This is pretty confusing for Java and Go, because resolver attributes aren't passed into subchannels and this doesn't define what will copy them to address attributes. Maybe something closer to:

> The subchannel will be passed attributes with the values for the `grpc.lb.backend_service` and `grpc.lb.locality` labels  (defined in [A89] and [A78] respectively). The subchannel will be recreated when those attributes change. Since only xDS is presently using these labels, the attributes will be set for each endpoint or address by cds (post-A74) or cluster_resolver (pre-A74) LB policies.

yashykt: adopted

ejona86: In Go it uses address attributes, not endpoint attributes. So I was actually serious when saying it will be for each endpoint or address, as it varies per language.

yashykt: ack :D

ejona86 (approved):

PR #486, opened 2025-03-21
A95: xDS LEDS and Fallback Improvements

markdroth:


A95-xds-endpoint-fallback.md
```diff
@@ -0,0 +1,148 @@
...
+here is that the EDS resource can use a different authority than the other
+resources, so that it can make use of the alternative fallback behavior.
+
+### LEDS List Collection Support
+
+The [LEDS] design was originally designed to address scalability
+concerns for large proxies.  The idea is to have the EDS resource
+contain only the locality assignments, but then have it refer to other
+resources for the endpoint assignments, where each endpoint is
+represented as a separate resoure of type
```
ejona86: r/resoure/resource/

A95-xds-endpoint-fallback.md
```diff
@@ -0,0 +1,148 @@
...
+don't need that level of scalability; all we actually need here is to be
+able to split up the locality assignment and endpoint assignment
+information into separate resources.  While we would eventually like to
+support the incremental protocol variant in gRPC, that is more work that
+we don't really need right now.  So instead of using a glob collection,
+we will use a list collection, which does not require the incremental
+protocol variant.  The list collection will be an `LbEndpointCollection`
+resource, introduced in https://github.com/envoyproxy/envoy/pull/38777.
+
+The validation rules for EDS as described in [A27] will change as
```
ejona86, line 100: How will this new data be encoded coming out of XdsDependencyManager in A74? We'll need to update XdsEndpointResource to support a reference to LbEndpointCollection, but do we
1. have XdsDependencyManager have a list of the localities with their own LbEndpointCollectionResource (closer to how we handle XdsRouteConfigResource) or
2. rewrite the XdsEndpointResource to remove the references and inline the endpoint information as if LEDS wasn't used?

markdroth: There are two different API layers here:

1. The `XdsConfig` representation generated by `XdsDependencyManager` and exposed to the resolver and LB policy code.
2. The individual resource type implementations exposed from `XdsClient` to `XdsDependencyManager`.

I think that for (1), the representation wouldn't change, since that representation already combines all resources into a single data structure, and that data structure is essentially structured in terms of how we need to use the data in the LB policy tree, independently of how it was represented in xDS resources.

For (2), there needs to be a separate object for each individual xDS resource, since each resource type that we plug into `XdsClient` needs to produce some type for the parsed resource.  But since the `LbEndpointCollection` can either be inlined in EDS or obtained from a separate LEDS resource, I expect that we'll handle this the same way that we do for the `RouteConfiguration`, which can either be inlined in LDS or obtained from a separate RDS resource.  In C-core, we handle this by having a [field in the parsed LDS resource](https://github.com/grpc/grpc/blob/53bc6c644078cca4c632a0706c88a49354c1b7ee/src/core/xds/grpc/xds_listener.h#L45) that either specifies the RDS resource name or inlines the parsed `RouteConfiguration`.  So I expect that what we'll do here is to add a new LEDS resource, and then change the existing parsed EDS representation to specify either the name of the LEDS resource or the parsed `LbEndpointCollection` itself.

Note that the actual parsing logic for the `LbEndpointCollection` is the same, regardless of whether it's inlined in EDS or sent in a separate LEDS resource.  So we should be able to just refactor the current EDS parsing code; we shouldn't need to duplicate it.  Today, when the `RouteConfiguration` is inlined into the LDS resource, the LDS resource parser essentially calls out to the RDS resource parser.  I expect that we'll do the same here, where the EDS parser will call out to the LEDS parser.

ejona86: (2) seems pretty clear to me; we've done that before, including code sharing.

For (1), A74 has:
```c++
    struct EndpointConfig {
      XdsEndpointResource endpoints;
      std::string resolution_note;
    };
```

Are you saying XdsEndpointResource is a _different_ object than produced by XdsClient? That's news to me. And that's surprising, as then there'd be no need for `route_config`, as that could be inlined into XdsListenerResource:
```c++
struct XdsConfig {
  // Listener resource.
  XdsListenerResource listener;
  // RouteConfig resource.  Will be populated even if RouteConfig is
  // inlined into the Listener resource.
  XdsRouteConfigResource route_config;
```

Right now I thought the resources presented in XdsConfig are identical to what was received. I agree we _could_ change them, but we already had an opportunity to rewrite one and we didn't take it. So I feel like we need to spell out what's expected here, since this is also the sort of thing you want everyone doing the same way.

markdroth: Sorry, I misunderstood your question.  You're right, we need to change the `XdsConfig` representation.

I've added some text to spell this out.

PR #487, opened 2025-04-02
A83 Update: Add Java implementation link

shivaspeaks:


A83-xds-gcp-authn-filter.md
```diff
@@ -375,13 +375,18 @@ credentials implementations in some languages.
 
 ## Implementation
 
-C-core implementation:
+#### C-core implementation:
```
markdroth, line 378: I don't think these need to be section headers.  It's fine for them to be just two bulletted lists in the same section.

markdroth (approved):
Thanks!

PR #488, opened 2025-04-03
A96: OTel Metrics for Retries

yashykt:


A96-retry-otel-stats.md
```diff
@@ -0,0 +1,79 @@
...
+
+[A45]: A45-retry-stats.md
+[A66]: A66-otel-stats.md
+[A79]: A79-non-per-call-metrics-architecture.md
+
+## Proposal
+
+Metric Name                          | Type      | Unit                | Labels                                         | Description
+------------------------------------ | --------- | ------------------- | ---------------------------------------------- | -----------
+grpc.client.call.retries             | Histogram | {retry}             | grpc.method (required), grpc.target (required) | Number of retry or hedging attempts excluding transparent retries made during the client call. The original attempt is not counted as a retry/hedging attempt.
```
ejona86, line 35: We'll need to define the buckets for the histograms, right?

ejona86 (approved):

A96-retry-otel-stats.md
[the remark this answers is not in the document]
markdroth: It might not be a bad idea to define a separate set of metrics for hedging.  In our current implementation, the client can't be configured for retries and hedging at the same time, so there's no conflict in sharing the same metrics between them.  But we may in the future have some cases where we want to configure both at the same time, in which case we'd need to separate the metrics for clarity.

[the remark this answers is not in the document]
yashykt: Added a label that differentiates the type of retry - "retry"/"hedge"/"transparent".

yashykt: added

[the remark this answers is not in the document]
markdroth: Why do this as a label, as opposed to making them different metric names?  The latter seems like it would be a bit clearer in terms of understanding what the metric indicates.

A96-retry-otel-stats.md
```diff
@@ -0,0 +1,95 @@
...
+
+[A45]: A45-retry-stats.md
+[A66]: A66-otel-stats.md
+[A79]: A79-non-per-call-metrics-architecture.md
+
+## Proposal
+
+Metric Name                  | Type      | Unit    | Labels                                                                     | Description
+---------------------------- | --------- | ------- | -------------------------------------------------------------------------- | -----------
+grpc.client.call.retries     | Histogram | {retry} | grpc.method (required), grpc.target (required), grpc.retry_type (required) | Number of retry attempts made during the client call. The original attempt is not counted as a retry attempt. Recommended histogram bucket boundaries are [0,1,2,3,4,5,10,100,1000].
```
dfawley, line 35: `grpc.client.call.attempts` and `grpc.attempt_type`?

Hedging is not a kind of retry.

But..do you _need_ it to be 0 for "normal" cases without any retry/hedging?  Because calling it `attempts` would mean it should be 1 in those cases.

dfawley, line 35: And what does it mean for the `retry_type` to be on the `retries` metric with unit `retry`?  What should be reported if there is one transparent retry attempt and two non-transparent retries?  What if they happen in a different order?

A96-retry-otel-stats.md
```diff
@@ -0,0 +1,95 @@
...
+grpc.client.call.retry_delay | Histogram | s       | grpc.method (required), grpc.target (required)                             | Total time of delay while there is no active attempt during the client call.
+
+The labels, `grpc.method` and `grpc.target` have been defined in [A66]. The
+label, `grpc.retry_type` can take one of three values - "retry", "hedge", or
+"transparent".
+
+The recommended histogram bucket boundary values for `grpc.client.call.retries`
+are [0,1,2,3,4].
```
dfawley, line 43: This doesn't match what you wrote above.  Delete either this paragraph or the values in the table?

A96-retry-otel-stats.md
```diff
@@ -0,0 +1,95 @@
...
+
+These metrics are recorded per call utilizing the `CallTracer` approach (also
```
dfawley, line 45: Are we reporting this at the end of the call based on how many retries occurred? Or when each retry attempt happens, it outputs an event for it?

A96-retry-otel-stats.md
yashykt: > Hedging is not a kind of retry.

https://github.com/grpc/proposal/blob/master/A45-retry-stats.md counted hedging stats under retries. I'm fine calling it as an attempt instead.

well, thinking about it, we could just make `grpc.attempt_type` an optional attribute on the existing metric `grpc.client.attempt.started`.

yashykt: Outdated. Now, we are just adding it as an attribute on `grpc.client.attempt.started`

A96-retry-otel-stats.md
yashykt: Thanks! removed

A96-retry-otel-stats.md
yashykt: You would continue to report `grpc.client.attempt.started` as it is today, while `grpc.client.call.retry_delay` would need to calculated for the entire call and reported at the end of it.

A96-retry-otel-stats.md
yashykt: ok, my bad, I clearly forgot that `grpc.client.attempt.started` is a counter and not a histogram. This change will not work. To avoid any confusion on hedges being counted as retries, I am going to just use three different metrics for hedges, retries and transparent retries.

A96-retry-otel-stats.md
yashykt: We report it at the end of the call.

A96-retry-otel-stats.md
yashykt: Now, we have separate metrics for retries, hedges and transparent retries.

yashykt: > https://github.com/grpc/proposal/blob/master/A45-retry-stats.md counted hedging stats under retries. I'm fine calling it as an attempt instead.

For reference, we decided to not call it an attempt since we do not want to count original attempts. Those can be figured out by `grpc.client.attempt.started` - `grpc.client.call.retries` - `grpc.client.call.transparent_retries` - `grpc.client.call.hedges`

A96-retry-otel-stats.md
```diff
@@ -30,20 +30,17 @@ metrics originally proposed in [A45].
...
+Metric Name                          | Type      | Unit                | Labels                                         | Description
+------------------------------------ | --------- | ------------------- | ---------------------------------------------- | -----------
+grpc.client.call.retries             | Histogram | {retry}             | grpc.method (required), grpc.target (required) | Number of retries during the client call. If there were no retries, 0 is not reported. Recommended histogram bucket boundaries are [1,2,3,4,5].
+grpc.client.call.transparent_retries | Histogram | {transparent_retry} | grpc.method (required), grpc.target (required) | Number of transparent retries during the client call. If there were no transparent retries, 0 is not reported. Recommended histogram bucket boundaries are [1,2,3,4,5,10].
+grpc.client.call.hedges              | Histogram | {hedge}             | grpc.method (required), grpc.target (required) | Number of hedges during the client call. If there were no hedges, 0 is not reported. Recommended histogram bucket boundaries are [1,2,3,4,5].
```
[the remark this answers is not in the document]
yashykt: It already does. Just transparent retries has an additional 10 right now, since the spec does not limit transparent retries at present.

A96-retry-otel-stats.md
```diff
@@ -30,20 +30,17 @@ metrics originally proposed in [A45].
...
+grpc.client.call.retry_delay         | Histogram | s                   | grpc.method (required), grpc.target (required) | Total time of delay while there is no active attempt during the client call.
```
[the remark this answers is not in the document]
yashykt: Added

A96-retry-otel-stats.md
```diff
@@ -30,20 +30,17 @@ metrics originally proposed in [A45].
...
----------------------------- | --------- | ------- | -------------------------------------------------------------------------- | -----------
-grpc.client.call.retries     | Histogram | {retry} | grpc.method (required), grpc.target (required), grpc.retry_type (required) | Number of retry attempts made during the client call. The original attempt is not counted as a retry attempt. Recommended histogram bucket boundaries are [0,1,2,3,4,5,10,100,1000].
-grpc.client.call.retry_delay | Histogram | s       | grpc.method (required), grpc.target (required)                             | Total time of delay while there is no active attempt during the client call.
+Metric Name                          | Type      | Unit                | Labels                                         | Description
+------------------------------------ | --------- | ------------------- | ---------------------------------------------- | -----------
+grpc.client.call.retries             | Histogram | {retry}             | grpc.method (required), grpc.target (required) | Number of retries during the client call. If there were no retries, 0 is not reported. Recommended histogram bucket boundaries are [1,2,3,4,5].
```
markdroth, line 36: We might need more than 5 as the upper bound here.  When we implement estubs retries, there are modes where there is no limit on the number of retry attempts.

A96-retry-otel-stats.md
```diff
@@ -30,20 +30,17 @@ metrics originally proposed in [A45].
...
+grpc.client.call.transparent_retries | Histogram | {transparent_retry} | grpc.method (required), grpc.target (required) | Number of transparent retries during the client call. If there were no transparent retries, 0 is not reported. Recommended histogram bucket boundaries are [1,2,3,4,5,10].
```
markdroth, line 37: In principle, there is no upper bound on the number of transparent retry attempts if each attempt never actually goes out on the wire.

A96-retry-otel-stats.md
markdroth, line 38: In estubs, the upper bound for the number of hedged requests can be quite large, so I think we'll want to support larger numbers here too.

A96-retry-otel-stats.md
```diff
@@ -30,20 +30,17 @@ metrics originally proposed in [A45].
...
+grpc.client.call.hedges              | Histogram | {hedge}             | grpc.method (required), grpc.target (required) | Number of hedges during the client call. If there were no hedges, 0 is not reported. Recommended histogram bucket boundaries are [1,2,3,4,5].
+grpc.client.call.retry_delay         | Histogram | s                   | grpc.method (required), grpc.target (required) | Total time of delay while there is no active attempt during the client call. Recommended to use exponential histograms if available, otherwise, use latency buckets defined in [A66].
 
-The labels, `grpc.method` and `grpc.target` have been defined in [A66]. The
-label, `grpc.retry_type` can take one of three values - "retry", "hedge", or
-"transparent".
+The labels, `grpc.method` and `grpc.target` have been defined in [A66].
```
markdroth: Grammar nit: Please remove the comma here.

A96-retry-otel-stats.md
markdroth: I think we've talked about this before, but I don't understand how expontential histograms can possibly work for aggregation.  And in any case, we don't support them in the non-per-call metrics APIs, so I don't see how we can recommend that here.

A96-retry-otel-stats.md
```diff
@@ -0,0 +1,95 @@
...
+[A45]: A45-retry-stats.md
+[A66]: A66-otel-stats.md
+[A79]: A79-non-per-call-metrics-architecture.md
+
+## Proposal
+
+Metric Name                  | Type      | Unit    | Labels                                                                     | Description
+---------------------------- | --------- | ------- | -------------------------------------------------------------------------- | -----------
+grpc.client.call.retries     | Histogram | {retry} | grpc.method (required), grpc.target (required), grpc.retry_type (required) | Number of retry attempts made during the client call. The original attempt is not counted as a retry attempt.
+grpc.client.call.retry_delay | Histogram | s       | grpc.method (required), grpc.target (required)                             | Total time of delay while there is no active attempt during the client call.
```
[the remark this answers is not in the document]
markdroth: Doug, I think it's fine for the CallTracer to record the time when there are no attempts ongoing, after the first attempt starts.  If it's slightly non-zero even when there's no real delay, that's probably fine.

A96-retry-otel-stats.md
yashykt: I think if we end up changing the limit on retries, we can update the advice here.

Otherwise, if we've already made enough progress on the new changes, please let me know what the better boundaries would be.

A96-retry-otel-stats.md
yashykt: I added a 10 taking that in consideration. I think this should be enough. I'm not sure it makes sense to distinguish between cases where the number if larger than 10. (I'm willing to change the boundaries here if there are strong opinions. These are just recommendations anyway.)

A96-retry-otel-stats.md
yashykt: I think if we end up changing the limit on retries, we can update the advice here.

Otherwise, if we've already made enough progress on the new changes, please let me know what the better boundaries would be.

A96-retry-otel-stats.md
yashykt: done

yashykt (commented):
> I think we've talked about this before, but I don't understand how expontential histograms can possibly work for aggregation. And in any case, we don't support them in the non-per-call metrics APIs, so I don't see how we can recommend that here.

Exponential Histograms are just a type of histograms. Our APIs do not need anything more to support exponential histograms. The OpenTelemetry SDK user can simply provide the buckets to use when creating the histogram view.

> expontential histograms can possibly work for aggregation

Aggregation for histograms, which in most cases is calculating percentiles (50% / 95% / 99 %) is always approximate since we no longer have the exact values.

A96-retry-otel-stats.md
yashykt: Exponential Histograms are just a type of histograms. Our APIs do not need anything more to support exponential histograms. The OpenTelemetry SDK user can simply provide the buckets to use when creating the histogram view.

Aggregation for histograms, which in most cases is calculating percentiles (50% / 95% / 99 %) is always approximate since we no longer have the exact values, but just the bucket ranges.

yashykt: recommending latency buckets as discussed on vc

A96-retry-otel-stats.md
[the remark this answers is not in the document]
dfawley: My concern is that the field is called `retry_delay`, and if it's non-zero when retry is not enabled, that would be confusing.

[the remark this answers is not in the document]
markdroth: I don't think we'd want it to be non-zero if retries are not enabled.  We want it to count the delay only after the initial attempt completes.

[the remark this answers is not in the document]
dfawley: How do you propose we do that, though?

The hedging configuration has two fields that are relevant here: `hedgingDelay` and `nonFatalStatusCodes`.  If the first attempt encounters a non-fatal code in less time than `hedgingDelay`, then I believe we will continue to wait out the full `hedgingDelay` before sending the next attempt.  In this case, counting the delay between attempts will be non-zero, even though retry is not enabled.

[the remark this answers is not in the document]
yashykt: That is incorrect. As per https://github.com/grpc/proposal/blob/master/A6-client-retries.md
```
If a non-fatal status code is received from a hedged request, then the next hedged request in line is sent immediately, shortcutting its hedging delay.
```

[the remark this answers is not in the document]
dfawley: OK I thought it was like that, but while chatting with Eric, he thought we'd wait out the delay.

Either way, nothing is _actually_ immediate.  So the first attempt will finish and _then_ the next attempt will begin, some number of nanoseconds or microseconds later.  There will be a delay no matter what, and we can't distinguish between these two scenarios without hacks (e.g. rounding as you suggested) or new plumbing.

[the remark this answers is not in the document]
yashykt: Note that we use a histogram for this, and the bucket boundaries (if we don't use exponential histograms) are latency buckets, where the first bucket captures data from 0 to 10 us. So, a minor delay between the previous attempt and a new attempt would get captured the same way as 0.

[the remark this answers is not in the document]
markdroth: I think it's fine to record a very small non-zero delay if we start a new attempt right after the previous attempt fails.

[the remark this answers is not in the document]
dfawley: I'm okay with that (given the alternative of needing to surface retry settings to tracers), but it might be worth calling out in the design that it is possible and OK.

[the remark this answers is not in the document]
yashykt: Done

dfawley (approved):

PR #491, opened 2025-04-24
A83 update: document Java's filter state retention.

sergiitk:
Document Java's implementation details for xDS HTTP Filter state retention.

Implementation PRs:
- grpc/grpc-java#11883
- grpc/grpc-java#11936


sergiitk:
FYI @ejona86 @dfawley

A83-xds-gcp-authn-filter.md
```diff
@@ -283,9 +283,90 @@ a cache size change will wind up affecting the old filter instance,
...
+config parsing, will be moved to `Filter.Provider`.
+
+We will implement `Filter.Provider` in each existing Filter class (as an
+inner static class). Filters that do not need to retain filter state
+may implement `newInstance` to keep returning a singleton instance of self.
+
+We will update `FilterRegistry` to register `Filter.Provider` instances instead
+of `Filter` instances.
+
+**Phase 2. Implement the lifecycle of Filter objects**
```
sergiitk, line 320: @dfawley FYI I kept this part a bit more detailed and technical because I think this may be useful as a reference for Go's implementation. Let me know if the details below this don't make sense to you.

A83-xds-gcp-authn-filter.md
```diff
@@ -283,9 +283,88 @@ a cache size change will wind up affecting the old filter instance,
...
+field on a `GcpAuthenticationFilter` object. No in-filter logic will be needed
+to separate caches per filter instance name, as Java's implementation
+will produce distinct `GcpAuthenticationFilter` instances, and therefore,
+different caches.
+
+To achieve this, we need to make several key changes to the class design.
+
+**Phase 1. Introduce Filter.Provider**
+
+Implemented in: https://github.com/grpc/grpc-java/pull/11883.
```
markdroth, line 300: Implementation links don't belong in this section; please move them down to the "Implementation" section below (see the list of Java PRs starting on line 466).

A83-xds-gcp-authn-filter.md
```diff
@@ -283,9 +283,88 @@ a cache size change will wind up affecting the old filter instance,
...
+
+The GCP Authentication filter will store the call credentials cache as a regular
+field on a `GcpAuthenticationFilter` object. No in-filter logic will be needed
+to separate caches per filter instance name, as Java's implementation
+will produce distinct `GcpAuthenticationFilter` instances, and therefore,
+different caches.
+
+To achieve this, we need to make several key changes to the class design.
+
+**Phase 1. Introduce Filter.Provider**
```
markdroth, line 298: No need for these sub-headers.  I think this is simple enough that it can be expressed in a single section.

A83-xds-gcp-authn-filter.md
markdroth: It seems like this whole section boils down to the following:

"""
Each HttpConnectionManager instance has its own filter chain.  Each filter chain will have its own list of filter instances.  Within a filter chain, filter instances are uniquely identified by the filter's name and type_url.

Filter instances are shut down when either they are removed from their filter chain or when their filter chain is removed (which may happen at client or server shutdown).
"""

I think the above communicates the intended structure unambiguously but without too many distracting details.

sergiitk: Thank you for summarizing this. I'm not sure the first paragraph is entirely correct: AFAIK the relation between HCMs and `FilterChain`s is inverted.

On the client side there's no filter chains: `HttpConnectionManager -> filter instances (list)`.
However, on the the server side it's `FilterChain (list) -> HttpConnectionManager -> filter instances (list)`.

I will rewrite your summary to reflect that, but please correct me if I'm wrong.

P.S. One of the challenges for the server side impl was keeping track of individual filter chains list items between LDS updates. We ended up using filter chain names for that -- they're required to be unique. However, they're also optional - so we generated the names using the list odred: `chain_{idx}`. I should probably cover this too.

sergiitk: Done, please take a look.

markdroth: I think we may be suffering from some terminology confusion between the L4 filter chain (the one in socket listeners, used on the server side) and the L7 filter chain (the HTTP filters inside of the HCM, used on both client and server).

On the client side, with an API listener, there is no L4 filter chain at all.  We essentially just have a single HCM instance for the channel, and the API listener contains the config for that instance.

On the server side, with a socket listener, there can be multiple L4 filter chains, each of which contains a separate HCM isntance with its own config.  As each connection comes in, we choose which L4 filter chain we're going to use, which effectively chooses which HCM instance to use for that connection.

Each HCM instance, regardless of whether it's on the client or server side, has its own HTTP filter chain.  I think the property that we want here is that the list of HTTP filters is unique only within its HCM instance, regardless of whether that instance is on client side or server side.  It's not legal to have two HTTP filters with the same name in the same HCM instance.  However, it's totally fine to have two different HCM instances that both have a filter with the same name, and those two filters should not share state at all.

I think the problem you're pointing out here is that on the server side, when we get an LDS update, we need a way to match up the L4 filter chains in the old LDS resource with those in the new LDS resource, so that we know which existing HCM state should be retained by each new HCM instance.  That's a good point, and I don't remember discussing it before -- did we?  I think this is something we need to align on cross-language.

The proposed solution seems to be to depend on the order of the L4 filter chains within the LDS resource.  However, I'm not sure that's quite ideal, since that order really isn't semantically meaningful -- the algorithm used to match L4 filter chains for incoming connections doesn't care about the order of the filter chains in the LDS resource, so swapping the order shouldn't actually change the resulting behavior.  I think swapping the state of the two HCM instances would be very surprising behavior in that situation.

An alternative would be to depend on the L4 filter chain names only, such that we retain state only if the L4 filter chain has the same name in both the old and new LDS resource.  If an L4 filter chain does not have a name, then we won't retain the HTTP filter state at all.  So we'd have to make sure that the control plane sets a name for the L4 filter chains if we want to retain this state, but this way there's no ambiguity about how to match up the HCM instances.

I agree that it's a good thing to handle the default filter chain separately, though, since that one I think we can do accurately even without the L4 filter chain having a name.  And that will be the common case anyway -- I would not expect most gRPC servers to have multiple L4 filter chains in general.

@ejona86 @dfawley Thoughts...?

sergiitk: Quite a few things to break down. I'll post them as a separate comment per thread.

> I think the problem you're pointing out here is that on the server side, when we get an LDS update, we need a way to match up the L4 filter chains in the old LDS resource with those in the new LDS resource, so that we know which existing HCM state should be retained by each new HCM instance. That's a good point, and I don't remember discussing it before -- did we? I think this is something we need to align on cross-language.

Found the discussion in our internal chat:

> Sergii Tkachenko, Feb 10, 4:07 PM
> Yes, thanks for the note. Aware that it's L4, but this is why I'm asking. We have all this logic for handling multiple MatcherChains (essentially HCMs), and that's what makes it so different from the client side. In java, this affects how we handle the filter cache on the server side. Initially I thought we'll track active filters per filter name. But it looks like we'd have to do have another dimension: per chain -> per name -> fiter

> Mark Roth, Feb 10, 4:17 PM
> Yes, I think the cache should be per filter chain -- each filter chain should be independent of any other.

Regarding cross-language alignment, the plan was for Java to establish the approach, capture it in this doc, and then for grpc-go to mirror it.

sergiitk: Second, on the per-L4-FilterChain approach:

> I think the problem you're pointing out here is that on the server side, when we get an LDS update, we need a way to match up the L4 filter chains in the old LDS resource with those in the new LDS resource, so that we know which existing HCM state should be retained by each new HCM instance. That's a good point, and I don't remember discussing it before -- did we? I think this is something we need to align on cross-language.

Found the discussion in our internal chat:

> Sergii Tkachenko, Feb 10, 4:07 PM
> Yes, thanks for the note. Aware that it's L4, but this is why I'm asking. We have all this logic for handling multiple MatcherChains (essentially HCMs), and that's what makes it so different from the client side. In java, this affects how we handle the filter cache on the server side. Initially I thought we'll track active filters per filter name. But it looks like we'd have to do have another dimension: per chain -> per name -> fiter

> Mark Roth, Feb 10, 4:17 PM
> Yes, I think the cache should be per filter chain -- each filter chain should be independent of any other.

Regarding cross-language alignment, the plan was for Java to establish the approach, capture it in this doc, and then for grpc-go to mirror it.

markdroth: > In all cases when I say Filter Chain I refer to the L4 filter chain. The reason is that this is actual field name and the proto message `config.listener.v3.FilterChain`.

In xDS, we use the term "filter chain" both in the context of L4 filters and L7 filters.  People will not assume that the term automatically means L4 filters -- you need to disambiguate.

> Found the discussion in our internal chat:

I don't see anything in that quoted discussion that talks about how we match up L4 filter chains between two LDS updates.  That's the part that I don't remember discussing.

> Regarding cross-language alignment, the plan was for Java to establish the approach, capture it in this doc, and then for grpc-go to mirror it.

The thing that we're talking about here is the behavior of the xDS API, which needs to be consistent across all languages, not just the two implementations that happen to be most similar to each other.  If Java and Go both have the same behavior in terms of matching up L4 filter chains and C-core does something different, that's not what we want.

The implementations may be different under the hood, but the behavior of the API needs to be consistent across all implementations.

sergiitk: Next, depending on the order of the L4 filter chains (by generating filter names using the list index).

> However, I'm not sure that's quite ideal, since that order really isn't semantically meaningful -- the algorithm used to match L4 filter chains for incoming connections doesn't care about the order of the filter chains in the LDS resource,

My main idea was to identify Filter Chain using a unique field. While Filter Chain name is unique, but the protocol allows to omit it altogether. My initial approach was to use the `config.listener.v3.FilterChainMatch FilterChain.filter_chain_match` field instead. It's guaranteed to be unique by the FilterChainMatch normalization procedure in Java, and  our internal FilterChainMatch object is comparable. The downside is this approach requires grpc-go to implement FilterChainMatch comparison too.

During the PR review, @ejona86 suggested that we can simplify the approach by using the list index directly, or even autogenerating Filter Chain names from it. This way, all Filter Chains with explicit names are guaranteed to match, and those without will fall back on the list order. And for stability guarantees, the control plane should just be filling in actual names.

> so swapping the order shouldn't actually change the resulting behavior. I think swapping the state of the two HCM instances would be very surprising behavior in that situation.

This won't be the case in RLQS specifically (because it relies on the filter config itself), but I how in a generic case it's possible with two Filter Chains without names to a filter instance with the same name but different configs, and then swap places.

I just talked to @ejona86 about it, and he has a good argument why it's still reasonable to use list indexes. I'll defer filling in the details to him.

sergiitk: > The thing that we're talking about here is the behavior of the xDS API, which needs to be consistent across all languages, not just the two implementations that happen to be most similar to each other. If Java and Go both have the same behavior in terms of matching up L4 filter chains and C-core does something different, that's not what we want.

This makes sense. Let's discuss it. I'll set up a meeting.

A83-xds-gcp-authn-filter.md
```diff
@@ -283,9 +283,56 @@ a cache size change will wind up affecting the old filter instance,
...
+`XdsNameResolver` and `XdsServerWrapper` respectively, due to the structural
+differences in configuration: client-side has a single `HttpConnectionManager`
+directly containing filters, while server-side can have multiple `FilterChain`
+and an optional `default_filter_chain`, each containing an
+`HttpConnectionManager` and its filters.
+
+To keep track of track of individual filter chains across LDS updates, we'll be
+using their unique names. Since filter chain names are optional but must be
+unique if provided, the implementation will generates names based on the list
+order (e.g., `chain_0`, `chain_1`). If present, `default_filter_chain` is
```
markdroth: I believe we decided that we will simply not retain any state for non-named filter chains, rather than assigning them our own names.

A83-xds-gcp-authn-filter.md
```diff
@@ -283,9 +283,56 @@ a cache size change will wind up affecting the old filter instance,
...
+tracked independently.
+
+Filter instances are shut down when either they are removed from their
+`HttpConnectionManager` or when their `HttpConnectionManager` is removed (which
+may happen at client or server shutdown).
+##### Go
```
markdroth, line 352: Nit: Please add a blank line before this one.

A83-xds-gcp-authn-filter.md
sergiitk: Done.

markdroth (approved):
Looks great, thanks!

PR #492, opened 2025-04-29
A97: xDS JWT Call Credentials

markdroth:


kyessenov (commented):
I'll let @howardjohn confirm, but I think this covers Istio functionality. For a sidecar, the service account token is mounted as "istio-token" JWT with a different audience.

A97-xds-jwt-call-creds.md
```diff
@@ -0,0 +1,210 @@
...
+To add the token to a data plane RPC, the call credential will add
+a header named `authorization`. The header value will be the string
+`Bearer ` (note trailing space) followed by the token value.
+
+### Configuring Call Credentials in xDS Bootstrap File
+
+We will add a new field to the xDS bootstrap representation of an xDS
+server:
+
+```json
```
sergiitk, line 137: Suggestion: you can use [`json5`](https://json5.org/) format to avoid syntax highlight errors produced by standard json validation.

```suggestion
```json5
```

markdroth:
@dimpavloff Thanks for the ping!  There's nothing blocking this other than finding the time to push it forward. :)

Do you have a particular gRPC language in which you're waiting for this feature?

dimpavloff:
> @dimpavloff Thanks for the ping! There's nothing blocking this other than finding the time to push it forward. :)

Ah, I know the feeling :)

> Do you have a particular gRPC language in which you're waiting for this feature?

Yes. At my company we use Go and are keen to adopt gRPC proxyless but we hit https://github.com/istio/istio/issues/53532 which is what led me here.

A97-xds-jwt-call-creds.md
```diff
@@ -0,0 +1,208 @@
...
+  // NEW FIELD!
+  // List of call creds.  Optional.  Client will apply all call creds
+  // types that it supports but will ignore any types that it does not
+  // support.
+  "call_creds": [
+    {
+      "type": <string containing call cred type>,
+      // The "config" field is optional; it may be missing if the
+      // credential type does not require config parameters.
+      "config": <JSON object containing config for the type>
```
ejona86, line 162: If config is invalid, then we fail bootstrap loading?

A97-xds-jwt-call-creds.md
```diff
@@ -0,0 +1,208 @@
...
+
+We will add a new type of call credentials called
+JwtTokenFileCallCredentials that loads a JWT token from a local file,
+with the ability to periodically reload the token.  We will also provide
+hooks to configure the use of this new call credentials type in the xDS
+bootstrap file, which was originally described in [A27].
+
+### JwtTokenFileCallCredentials
+
+Note: This section is intended for gRPC implementations that need to
```
ejona86, line 60: Java implementation note: We can use [JsonWebSignature](https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java).parse().getPayload().getExpirationTimeSeconds() to create an AccessToken with appropriate expiration. We could extend OAuth2Credentials and implement refreshAccessToken(). It takes care of stale vs expiration checking. Using JsonWebSignature is essentially re-implementing [IdToken](https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/IdToken.java), which would probably work today, but technically this is not necessarily an IdToken so we should probably avoid the existing class.

A97-xds-jwt-call-creds.md
```diff
@@ -0,0 +1,208 @@
...
+plane RPCs, which may then continue.
+
+If reading the file does not result in the cache being updated (i.e.,
+if reading the file fails or if it returns an invalid JWT token),
+[backoff](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md)
+must be applied before the next attempt may be started.  If a data plane
+RPC is started when there is no cached token available and while in
+backoff delay, it will be failed with the status from the last attempt
+to read the file.  When the backoff delay expires, the next data plane
+RPC will trigger a new attempt.  Note that no attempt should be started
```
ejona86, line 121: I'll note that this language is a bit weird for a local file access. Who cares if we continue trying to read it when the channel is idle? I know it was copied from GCP auth filter, though.

markdroth: I agree that this isn't necessarily the most obvious way of looking at this.  And you're right that this is based on the GCP auth filter, both in terms of the description here and in terms of the implementation I have in mind in C-core -- we have a general-purpose call creds type that calls an arbitrary plugin to re-fetch a token based on its expiration time, so I plan to write a plugin to read from a file rather than fetching a token via an HTTP request.

That having been said, we have seen cases in the past where local file I/O has caused significant bottlenecks on busy systems, so I think there are some benefits to treating a file-read the same as we would an HTTP request, even if I wouldn't have necessarily gone down this path if we hadn't previously built this structure for the GCP auth filter.

A97-xds-jwt-call-creds.md
markdroth: Yes.  I've added a note about that.

A97-xds-jwt-call-creds.md
ejona86: The implementation note for Java I added would have us doing something very similar.

I understand avoiding the I/O in the datapath. And I'm not really bothered by defining the behavior as you did. My comment is about the _reasoning_ expressed. This part is talking about how we really don't want to refresh when the channel is idle, "since we do not want to unnecessarily retry if the channel is idle." That doesn't seem to matter. We also already refresh on idle channels for the file-watching channel credentials.

markdroth: Given that we have seen significant I/O bottlenecks reading from local disk in the past, I do think it's reasonable to prefer not to do this reload if we don't actually need to do so.

ejona86: I have no clue what you are referencing and strongly doubt it is similar, but whatever. This is a non-normative section so we can disregard it anyway.

dfawley (approved):

PR #493, opened 2025-05-06
A98: Channelz v2

ctiller:


A98-debug-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+It also allows re-using the book-keeping already in existence for channelz in implementations, without adding another kind of book-keeping to the mix.
+
+Why not just extend channelz?
+
+There are some fundamental data model issues in that protocol that I'd like to address going forward:
+
+* chaotic-good (a transport currently implemented and in production in C++) has multiple TCP sockets per transport. This might be representable as a subchannel with multiple sockets on the client side (though we've already represented the HTTP2 transport as a socket there), but server side makes no allowance for properly describing the object hierarchy.
+* Further, the set of node types ("kinds" here-in) and their relationships are pre-baked into the protocol, and as we evolve and improve gRPC new node types are needed and new relationships will be added or removed. We should not need to update channelz each time this happens.
+
+Specific metrics have been not been carried forward from channelz as required parts of the protocol.
```
nimf, line 218:
```suggestion
Specific metrics have not been carried forward from channelz as required parts of the protocol.
```

A98-debug-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+For example, a channel has many subchannel children.
+Channelz listed specific kinds of children in its various node types - this tracking (and the need to produce it when sending an object) has caused contention issues in implementations in the past.
+Should the list of children (optionally of a particular kind) for an entity be desired - and that's common - a separate paginated service call can be made.
+Instead, this protocol lists only parents (as that set is far more stable).
+Multiple parents are allowed - to handle at least the case of C++ subchannels being owned by multiple channels.
+
+**data**: This is a list of protobuf Any objects describing the current state of this entity.
+Implementations may define their own protobufs to be exposed here, and common sets will be standardized separately.
+
+**trace**: Finally, an entity may supply a small summary of its history as a trace.
```
markdroth, line 73: Should we note here that traces can be either proactively recorded, in which case they are returned here, or recorded on-demand via the QueryTrace RPC method?

Should we impose any cap on the number of proactively recorded traces?

A98-debug-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+## Rationale
+
+The new interface is heavily inspired by channelz, but removes rigid node associations and predefined data sets.
+Much of the infrastructure required for channelz can be shared with the debug interface, and indeed one can be implemented atop the other with good fidelity (in either direction).
+
+Node ids are consistent between the protocols so that one can take a node id from the debug interface and query channelz data directly (and vice versa).
+This will allow consumers of the protocols to gradually transition from one to the other.
+It also allows re-using the book-keeping already in existence for channelz in implementations, without adding another kind of book-keeping to the mix.
+
+Why not just extend channelz?
```
markdroth, line 211: We could potentially add the new types and methods to the existing channelz interface if we wanted.

One thing worth noting is that channelz was designed for the use case of having a common UI that is able to show data for all languages with a common shape of tree.  If we could find a way to support both use-cases in a common API, that seems ideal, but we'd have to brainstorm whether there are viable ways to do that.

A98-debug-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+Entities have state and configuration, and tend be be relatively long lived - such that querying them makes sense.
+
+**id**: An entity is identified by an id.
+These ids are allocated sequentially per the rationale in A14, and implementations should use the same id space for debug entities and channelz objects.
+
+**kind**: An entity has a kind. This is a string descriptor of the general category of object that this entity describes.
+We use strings here rather than enums so that implementations are free to extend the kind space with their own objects.
+Common entity kinds will see some level of standardization across stacks, but we expect many kinds to be specific per gRPC implementation.
+Initially implementations are expected to match kinds with channelz object types:
+kind `"channel"` -> `channelz.v1.Channel`, `"subchannel"` -> `channelz.v1.Subchannel`, `"server"` -> `channelz.v1.Server`, `"socket"` -> `channelz.v1.Socket`.
```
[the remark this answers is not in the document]
ctiller: Yeah... this spec or a sister spec?

A98-debug-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+Finally, **QueryTrace** allows rich queries of live trace data to be pulled from an instance.
+
+```
+message QueryTraceRequest {
+    // The identifier of the entity to query.
+    int64 id = 1;
+    // The name of the trace to query.
+    string name = 2;
+    // The amount of time to query. Implementations may arbitrarily cap this.
+    google.protobuf.Duration duration = 3;
```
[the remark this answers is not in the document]
ctiller: let's make it 0

A98-debug-protocol.md
```diff
@@ -0,0 +1,224 @@
...
+
+Finally, **QueryTrace** allows rich queries of live trace data to be pulled from an instance.
+
+```
+message QueryTraceRequest {
+    // The identifier of the entity to query.
+    int64 id = 1;
+    // The name of the trace to query.
+    string name = 2;
+    // The amount of time to query. Implementations may arbitrarily cap this.
```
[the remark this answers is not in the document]
ctiller: We don't need to with the switch to streaming!

A98-debug-protocol.md
ctiller: Added some commentary

A98-debug-protocol.md
ctiller: ack. I think we've discussed offline directions.

PR #498, opened 2025-05-31
A100: Client-side WRR slow start configuration

altaiezior:


ejona86:
CC @dfawley

anuragagarwal561994:
@dfawley did we get a chance to check this

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,238 @@
...
+  // The value of aggression parameter should be greater than 0.0.
+  // By tuning the parameter, is possible to achieve polynomial or exponential shape of ramp-up curve.
+  //
+  // During slow start window, effective weight of an endpoint would be scaled with time factor and aggression:
+  // ``new_weight = weight * max(min_weight_percent, time_factor ^ (1 / aggression))``,
+  // where ``time_factor=(time_since_start_seconds / slow_start_time_seconds)``.
+  //
+  // As time progresses, more and more traffic would be sent to endpoint, which is in slow start window.
+  // Once host exits slow start, time_factor and aggression no longer affect its weight.
+  core.v3.RuntimeDouble aggression = 2;
```
atollena: `core.v3.*` and `type.v3.*` are xds proto. I don't think we want to have the LB policy depend on those types. You can simply use a float for the service config. Same for `min_weight_percent`. Based on other gRFC I think we usually use `google.protobuf.UInt32Value` in the service config (which uses of json when used directly).

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,238 @@
...
+}
+
+message SlowStartConfig {
+  google.protobuf.Duration slow_start_window = 1;
+  core.v3.RuntimeDouble aggression = 2;
+  type.v3.Percent min_weight_percent = 3;
+}
+```
+
+## Rationale
```
atollena: Do you have a sense of whether it would be feasable to implement this for other load balancers, such as round robin and least request, in the way that Envoy does? (I guess the problem is that the mechanism relies on weights, and weights are not implemented in round robin and least request in gRPC).

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,238 @@
...
+message ClientSideWeightedRoundRobin {
+  // ... existing fields ...
+  cluster.v3.Cluster.SlowStartConfig slow_start_config = 8;
+}
+
+message SlowStartConfig {
+  google.protobuf.Duration slow_start_window = 1;
+  core.v3.RuntimeDouble aggression = 2;
+  type.v3.Percent min_weight_percent = 3;
+}
```
atollena: (this message is already defined [in cluster.proto](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#config-cluster-v3-cluster-slowstartconfig), so we don't think need to redefine it. It's nice to repeat it here for reference though)

atollena (commented):
This would be a nice addition (it's something we've had requests for internally at Datadog). Generally the approach sounds good to me. Ideally we would be able to support this mechanism for other load balancers supported by gRPC and Envoy (lr & rr). But given that lr & rr do not implement endpoint weights, it seems difficult to fit the feature in those balancers.

client-side-weighted-round-robin-slow-start.md
altaiezior: Yes I think we can implement for these algorithms even if we don't have weights.

For round robin we when it is in the warmup period we can choose the endpoint based  on probability defined by ranom number. So can simple do something like

```
selected_endpoint = null
while (selected_endpoint == null) {
    selected_endpoint = picker.getNext()
    if (isInSlowStart(selected_endpoint) {
        if (Math.random() >= scaling_factor) {
            selected_endpoint = null
        }
    }
}
return selected_endpoint
```

For least request, we already track the number of in-flight request we can scale it by `1/scaling_factor` to achieve slow_start.

So the idea is adding a scaling_factor and pretending that this endpoint is having high stakes already bringing it closer to its actual value over the period of time.

We will need to track in both of these algorithms the time when the endpoint was first added to the picker list.

Do you think I should increase the scope of this proposal and implement slow_start in all three eligible algorithms? If the above approaches sound correct?

client-side-weighted-round-robin-slow-start.md
altaiezior: Yes we can use the same and keep it in proposal for reference, the idea was to reuse it though

client-side-weighted-round-robin-slow-start.md
altaiezior: Since as we have mentioned that in the later comment that it is already defined in cluster.proto, do I need to make any changes in the proposal for the same or will this work?

anuragagarwal561994:
@atollena I have addressed your comments regarding the proposal, let me know if I need to make any changes in it

client-side-weighted-round-robin-slow-start.md
altaiezior: @ejona86 what would be your suggestion

ejona86: There may be some possibilities here, but the configuration is inherently per-LB and the change here seems smaller to WRR than for some of the other LBs. So I feel like we can cover this LB-by-LB, and limit this gRFC to client-side WRR. The important part for this design is that `SlowStartConfig` could be used by other LBs in the future.

I'll note that Envoy uses a _completely different_ algorithm for least_request when weights are involved. And RR in Envoy is actually WRR with server-provided weights (I don't recall if it also swaps algorithm when weights vary).

altaiezior: Yes changes either way would be small, but I would also prefer that the proposal stick to WRR

How do we proceed from here @ejona86

atollena: OK, that all makes sense. I think it would be useful to provide a sentence on why this is only done for WRR is the gRFC Rationale section.

client-side-weighted-round-robin-slow-start.md
atollena: The LB policy will still need to be able to represent the slow start config without directly depending on xDS. See https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md#lb-policy-config for an example.

client-side-weighted-round-robin-slow-start.md
altaiezior: @atollena @ejona86 I have added a scopes and limitation section

client-side-weighted-round-robin-slow-start.md
altaiezior: @atollena addressed

anuragagarwal561994:
@ejona86 can I start with the implementation of the feature?

anuragagarwal561994:
@ejona86 @atollena will it be possible to review the proposal this week so that I can begin with the implementation of the feature

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
+Client-side weighted round robin slow start configuration
```
atollena: You can assign a number now, you'll get the priviledge of having A100!

```suggestion
A100: Client-side weighted round robin slow start configuration
```

Please also update the diagram name.

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+* Approver: a11r
+* Status: Draft
+* Last updated: 2025-05-31
+* Discussion at: TBD
+
+## Abstract
+
+This proposal introduces an enhancement to the existing client-side weighted_round_robin (WRR) load balancing policy in gRPC by incorporating a configurable `slow_start_config` mechanism. The intent of this feature is to gradually increase traffic to backend endpoints that are newly introduced or have recently rejoined the cluster, allowing them time to warm up and reach their optimal performance level before handling their full share of traffic. This change increases system stability and resilience in environments with dynamic scaling and volatile workloads.
+
+The design borrows from production-ready practices in other load balancers such as Envoy, where gradual traffic ramp-up (slow start) is a well-established technique for avoiding performance degradation and request failures during backend startup or recovery. The slow start feature gradually increases the traffic sent to newly added endpoints during a warmup period, allowing them to warm up their caches and establish connections before receiving full traffic load.
```
atollena: You could throw in a link to the references here:

```suggestion
The design borrows from production-ready practices in other load balancers such as Envoy, where gradual traffic ramp-up (slow start) is a [well-established technique](Envoy Slow Start Documentation) for avoiding performance degradation and request failures during backend startup or recovery. The slow start feature gradually increases the traffic sent to newly added endpoints during a warmup period, allowing them to warm up their caches and establish connections before receiving full traffic load.
```

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+This metric will help operators monitor the number of endpoints currently in slow start mode across different localities and backend services.
+
+## Implementation
+
+This will be implemented in all languages C++, Java, and Go.
+
+## References
+
+1. [Envoy Slow Start Documentation](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start)
+2. [gRFC A58][A58] - Client-side weighted round robin LB policy
```
atollena: The proper way to add references is like this:

```suggestion
[Envoy Slow Start Documentation](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start)
[gRFC A58][A58] - Client-side weighted round robin LB policy 
```
```suggestion
[Envoy Slow Start Documentation]: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/slow_start
[A58]: A58-client-side-weighted-round-robin-lb-policy.md
```

`Then you can refer to the envoy documentation for slow start [like this][Envoy Slow Start Documentation].`

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+- Description: Number of endpoints currently in slow start period
+- Labels:
+  - `grpc.lb.locality`: The locality of the endpoints
+  - `grpc.lb.backend_service`: The backend service name
+
+This metric will help operators monitor the number of endpoints currently in slow start mode across different localities and backend services.
+
+## Implementation
+
+This will be implemented in all languages C++, Java, and Go.
```
atollena: This is a good place to add the actual implementation of your choice when you have it ready (it can be a draft or closed PR that you can re-open once the gRFC has been approved).

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+
+In contrast, many modern systems adopt slow start strategies in load balancing to address these issues. These strategies allow endpoints to ramp up traffic gradually over a defined window, smoothing transitions and mitigating the risks of traffic spikes. Similar functionality exists in Envoy's load balancing policies, where slow start is implemented for round robin and least request policies.
+
+Introducing a `slow_start_config` configuration in gRPC WRR will offer these benefits within the native client policy, reducing reliance on external traffic-shaping mechanisms or manual intervention.
+
+### Related Proposals:
+* [gRFC A58][A58] - Client-side weighted round robin LB policy
```
atollena:
```suggestion
* [gRFC A58: weighted_round_robin LB policy][A58]
```

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+  // Represents the size of slow start window.
+  // If set, the newly created host remains in slow start mode starting from its creation time
+  // for the duration of slow start window.
+  google.protobuf.Duration slow_start_window = 1;
+
+  // This parameter controls the speed of traffic increase over the slow start window. Defaults to 1.0,
+  // so that endpoint would get linearly increasing amount of traffic.
+  // When increasing the value for this parameter, the speed of traffic ramp-up increases non-linearly.
+  // The value of aggression parameter should be greater than 0.0.
+  // By tuning the parameter, is possible to achieve polynomial or exponential shape of ramp-up curve.
```
atollena:
```suggestion
  // By tuning the parameter, it is possible to achieve polynomial or exponential shape of ramp-up curve.
```

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+  // Configures the minimum percentage of origin weight that avoids too small new weight,
+  // which may cause endpoints in slow start mode receive no traffic in slow start window.
+  // If not specified, the default is 10%.
+  google.protobuf.UInt32Value min_weight_percent = 3;
+}
+```
+
+### Weight Scaling During Warmup
+
+When an endpoint is first added or becomes ready after being in a non-ready state, it enters the warmup period. During this period, its weight will be scaled by a factor that increases non-linearly from `min_weight_percent` to 100% over the duration of `slow_start_window`.
```
atollena: Endpoints never start in `ready`, so you can omit this case.

```suggestion
When an endpoint becomes ready after being in a non-ready state, it enters the warmup period. During this period, its weight will be scaled by a factor that increases non-linearly from `min_weight_percent` to 100% over the duration of `slow_start_window`.
```

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+The graph illustrates how the scaling factor (effective weight) ramps up from zero to the backend weight for various aggression values, with time shown in milliseconds.
+
+The final weight used for load balancing will be:
+```
+final_weight = computed_weight * scale
+```
+
+### Subchannel Weights
+
+The existing `non_empty_since` timestamp, which is already used to check for `blackout_period`, can be repurposed to track the start of the warmup period for each endpoint. This timestamp tracks when the first non-zero load report was received, marking the beginning of its warmup period. This approach eliminates the need for additional timestamp tracking while maintaining the existing functionality.
```
atollena: So we would effectively use the mean weight (as in normal WRR) until we receive the first load report, right? I imagine that if the server has high latency, that could make slow start ineffective in that we could send many requests before receiving the first load report. I would rather use a separate time variable for this that is set to the either:
1. the time where the child entered ready, or
2. the time where the we sent the first request to the child.
Presumably the slow_start_window must be set as a function of the expected QPS and latency for slow start to be effective, so I think either of those will work fine and equally simple to implement.

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+
+The blackout period, which defaults to 10 seconds, begins when an endpoint receives its first non-zero load report. During this period, the endpoint continues to receive traffic, but instead of using the weights reported by the backend servers, the load balancer uses the mean of all backend-reported weights. This period helps prevent churn in the load balancing decisions when the set of endpoint addresses changes, ensuring that the weights used are based on stable, continuous load reporting.
+
+The slow start period also begins when the endpoint receives its first non-zero load report and applies a gradual scaling factor to the weights over a configurable duration (default 30 seconds). This scaling is applied to whatever weight is being used (either the mean weight during blackout period or the actual backend-reported weight after blackout period). The slow start period operates independently of the blackout period, meaning it will continue to scale the weights regardless of whether the blackout period is still active or has ended.
```
atollena: Where does the 30 seconds default come from? I don't see it as a default for Envoy, and it looks like the default is 0, effectively disabling slow start. We should probably use that instead.

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+
+It is recommended to keep the blackout period shorter than the slow start period. This is because when the blackout period ends, the endpoint's weight will suddenly change from the mean weight to its actual backend-reported weight. If this weight is significantly higher than the mean (e.g., 2x the mean weight), it could cause a sudden traffic spike that defeats the purpose of gradual traffic increase. By having a longer slow start period, the scaling factor will continue to gradually increase the weight even after the blackout period ends, ensuring a smooth transition to the full backend-reported weight.
```
atollena: That's a nice property, it composes well with the blackout period. The use of mean weights during the blackout period can currently cause a sudden burst of traffic, and inversely the endpoint to be overloaded during the blackout period.

client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,248 @@
...
+
+When endpoint weights become stale after the `weight_expiration_period`, the load balancer will continue to use the mean weight for load balancing. This is different from the blackout period as it's a response to weight staleness rather than initial endpoint setup. In these cases, since the endpoint weights were previously active, the slow start period is typically not triggered. However, when new weights arrive after expiration, the endpoint will enter the blackout period again, and if slow start is configured, the weights will be scaled up gradually during this period.
```
atollena: Is that an intentional behavior, or a consequence of using the same time tracking variable as the blackout period? To me this behavior makes little sense: why should slow start apply if we have a ready connection that already went through slow start, even if the load reports are stale? maybe I'm missing something?

In Envoy, when does the slow start window period start for existing load balancers that supports it? We might want to preserve that behavior.

client-side-weighted-round-robin-slow-start.md
atollena: You will need to open a PR against https://github.com/envoyproxy/envoy/blob/4c1e53f7942dc457912ad4f738e131dc191db641/api/envoy/extensions/load_balancing_policies/client_side_weighted_round_robin/v3/client_side_weighted_round_robin.proto to suggest this change to the proto.

atollena (commented):
It would be useful to see an implementation of this in one of the 3 languages. I wouldn't wait for the gRFC to be approved before starting, especially since the result can be used as a custom, private LB policy before it is merged to upstream-supported gRPC implementations.

client-side-weighted-round-robin-slow-start.md
altaiezior: Yes, I agree — this edge case was missed in my original proposal. The assumption was that since the endpoint becomes ready before receiving traffic, we would receive its load report in time to make routing decisions. However, I overlooked the case where the first load report itself might have a value of zero — for example, if the server isn’t configured to expose metrics. In such cases, the endpoint would incorrectly receive the average traffic, which defeats the purpose.

This strengthens the case that slow start should operate independently, even when no load report is available. I’ll update the proposal accordingly and explore ways to handle this scenario.

Regarding the second point — ideally, yes, slow start should be a function of expected QPS and latency. But real-world production environments introduce several nuances:
	•	We often have heterogeneous fleets behind a single endpoint (e.g., n2, n2d, c2), with varied configurations like 16 or 32 vCPUs.
	•	At application startup, caches may not be warmed, or JIT compilation (in Java) may still be in progress. Conversely, long-running pods may degrade due to accumulated errors or GC behavior — all of which affect latency and throughput.
	•	Even with identical instance types, performance may vary due to noisy neighbor effects or other runtime discrepancies.

While these are fundamentally application-level concerns, in practice, the load balancing system shouldn’t attempt to auto-tune slow start based on metrics — especially not without full application context.

So, having a static slow start configuration defined by the application owner is a more robust approach. If needed, operators can still use the above metrics to determine their own appropriate slow_start_window values

client-side-weighted-round-robin-slow-start.md
altaiezior: Yes this is the consequence of using the same time tracking variable, I will update the proposal by seeing the envoy's implementation as well. But I agree with the point that keeping a separate variable will make the feature more loosely coupled which should be better.

client-side-weighted-round-robin-slow-start.md
atollena: Thanks. I wasn't really suggesting an adaptive `slow_start_window`, it was just an observation. I think it's best to stick with what envoy currently supports right now.

Just to add to this conversation: when not using OOB (which I assume is a fairly common setup), the first load reports arrives at the end of the first *response*, which might be long after we put the endpoint in the pool, especially if lack of slow start or other protective measure causes overload.

altaiezior: @atollena so I checked the envoy implementation, what it does is that it keeps the last time the endpoint became healthy and the checks for current_time - last_hc_pass_time < slow_start_window then the endpoint is in slow start else it is working normally.

So as you suggested we can keep a time when the endpoint last enetered the ready state and from that we should determine the slow start to keep the implementation consistent with envoy.

client-side-weighted-round-robin-slow-start.md
altaiezior: There is no 30 seconds default, I have addressed this

client-side-weighted-round-robin-slow-start.md
altaiezior: I have updated the proposal to include a `ready_since` timestamp, which will be used to determine the slow_start_period.

client-side-weighted-round-robin-slow-start.md
altaiezior: I have added the implementation for Java, would love if you could review the changes :)

https://github.com/grpc/grpc-java/pull/12200

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+$$time\\_factor = \dfrac{max(time\\_since\\_ready\\_seconds, 1)}{slow\\_start\\_window\\_seconds}$$
+
+The scaling formula ensures that new endpoints receive a gradually increasing share of traffic while maintaining a minimum threshold to prevent starvation. The `aggression` parameter allows fine-tuning of the ramp-up curve, enabling either more aggressive initial scaling (values > 1.0) or more conservative approaches (values < 1.0).
+
+When an endpoint is not in the warmup period, the scale factor is set to 1.0, meaning the original weight is used without modification. This ensures that the slow start mechanism only affects endpoints during their initial warmup phase, after which they participate in normal load balancing based on their actual performance metrics.
+
+### Blackout Period vs Slow Start
+
+The WRR load balancing policy will offers two independent mechanisms for handling new endpoints: the blackout period and slow start. These mechanisms can be used independently or in combination, allowing operators to choose the approach that best fits their needs.
```
atollena:
```suggestion
The WRR load balancing policy will offer two independent mechanisms for handling new endpoints: the blackout period and slow start. These mechanisms can be used independently or in combination, allowing operators to choose the approach that best fits their needs.
```

atollena (approved):
👍

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+  google.protobuf.Duration slow_start_window = 1;
+
+  // This parameter controls the speed of traffic increase over the slow start window. Defaults to 1.0,
+  // so that endpoint would get linearly increasing amount of traffic.
+  // When increasing the value for this parameter, the speed of traffic ramp-up increases non-linearly.
+  // The value of aggression parameter should be greater than 0.0.
+  // By tuning the parameter, it is possible to achieve polynomial or exponential shape of ramp-up curve.
+  //
+  // During slow start window, effective weight of an endpoint would be scaled with time factor and aggression:
+  // ``new_weight = weight * max(min_weight_percent, time_factor ^ (1 / aggression))``,
```
dfawley: Nit: `min_weight_percent/100`

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+  // where ``time_factor=(time_since_start_seconds / slow_start_time_seconds)``.
+  //
+  // As time progresses, more and more traffic would be sent to endpoint, which is in slow start window.
+  // Once host exits slow start, time_factor and aggression no longer affect its weight.
+  google.protobuf.FloatValue aggression = 2;
+
+  // Configures the minimum percentage of origin weight that avoids too small new weight,
+  // which may cause endpoints in slow start mode receive no traffic in slow start window.
+  // If not specified, the default is 10%.
```
dfawley: Nit: some tweaks:

```suggestion
  // Configures the minimum percentage of the original weight that will be used for an endpoint
  // in slow start. This helps to avoid a scenario in which endpoints receive no traffic during the
  // slow start window. Must be between 0 and 100. If not specified, the default is 10%.
```

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Weight Scaling During Warmup
+
+When an endpoint is ready after being in a non-ready state, it enters the warmup period. During this period, its weight will be scaled by a factor that increases non-linearly from `min_weight_percent` to 100% over the duration of `slow_start_window`.
+
+The scale factor is calculated as follows:
+
+```
+time_factor = max(time_since_ready_in_seconds, 1) / slow_start_window_seconds
+scale = max(min_weight_percent, time_factor ^ (1/aggression))
```
dfawley:
```suggestion
scale = max(min_weight_percent/100, time_factor ^ (1/aggression))
```

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+When endpoint weights become stale after the `weight_expiration_period`, the load balancer will continue to use the mean weight for load balancing. This is different from the blackout period as it's a response to weight staleness rather than initial endpoint setup. In these cases, since the endpoint weights were previously active, the slow start period is typically not triggered. When new weights arrive after expiration, the endpoint will enter the blackout period again but not the slow start since there was no transition of sub-channel state and no gradullay traffic increase should ideally be expected here.
```
dfawley:
```suggestion
When endpoint weights become stale after the `weight_expiration_period`, the load balancer will continue to use the mean weight for load balancing. This is different from the blackout period as it's a response to weight staleness rather than initial endpoint setup. In these cases, since the endpoint weights were previously active, the slow start period is typically not triggered. When new weights arrive after expiration, the endpoint will enter the blackout period again but not the slow start since there was no transition of sub-channel state and no gradual traffic increase should be expected.
```

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+function get_final_weight(endpoint_weight: float, scaling_factor: float) -> float:
+    // endpoint_weight is either:
+    // - mean_weight for stale endpoints or endpoints in blackout period
+    // - actual backend-reported weight after blackout period
+    return endpoint_weight * scaling_factor
+```
+
+### xDS Integration
+
+The slow start configuration will be added to the xDS proto for the weighted round robin policy:
```
dfawley: Should this happen first? Is there an RFC for it with Envoy?  What is the status?

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+1. These algorithms don't use weights to determine endpoint selection, making the implementation of slow start more complex
+2. Additional considerations would be needed for how to gradually increase traffic to new endpoints in these algorithms
+3. The implementation details would likely differ significantly from the weighted round robin approach
+
+These other load balancing algorithms can be considered for slow start implementation in future proposals, with their own specific design considerations and requirements.
+
+## Metrics
+
+The following metric will be exposed to help monitor the slow start behavior:
```
dfawley: How important is this metric to include?  Should we consider not adding it until it's needed?

dfawley (commented):
This mostly seems fine to me, I just have some nits.  Also, since I don't see a slow start config in the envoy WRR policy proto, I'm curious if you've (or someone else has?) sent a proposal to add it there already.

anuragagarwal561994:
@dfawley https://github.com/envoyproxy/envoy/pull/40090 this is the MR I have created at envoy's end to include the slow_start_config in proto.

However it requires this grpc proposal to be approved first before it is merged to the master.

Then we will need to sync across repo and then we are good to implement this feature. I have added sample MR with the java implementation as well, please do check it as well.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Yes I have created the MR there and it will be approved once this proposal is approved.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Ideally in most of the cases the slow_start_window will be applicable only for a short duration like 1-2 mins and if everything works well, it will not be called again.

I just thought that we can implement it in disabled mode, if someone wants to keeps a track of it.

We can decide this and I will make the respective changes in the proposal

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+* Approver: a11r
+* Status: Draft
+* Last updated: 2025-05-31
+* Discussion at: TBD
+
+## Abstract
+
+This proposal introduces an enhancement to the existing client-side weighted_round_robin (WRR) load balancing policy in gRPC by incorporating a configurable `slow_start_config` mechanism. The intent of this feature is to gradually increase traffic to backend endpoints that are newly introduced or have recently rejoined the cluster, allowing them time to warm up and reach their optimal performance level before handling their full share of traffic. This change increases system stability and resilience in environments with dynamic scaling and volatile workloads.
+
+The design borrows from production-ready practices in other load balancers such as Envoy, where gradual traffic ramp-up (slow start) is a [well-established technique][Envoy Slow Start Documentation] for avoiding performance degradation and request failures during backend startup or recovery. The slow start feature gradually increases the traffic sent to newly added endpoints during a warmup period, allowing them to warm up their caches and establish connections before receiving full traffic load.
```
markdroth: s/load balancers/data planes/

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Background
+
+gRPC's WRR load balancing policy allows clients to route requests to backend endpoints in proportion to assigned weights. These weights are usually derived from backend metrics, such as CPU usage, QPS, and error rates published by the backend servers. This allows gRPC clients to adapt traffic distribution dynamically based on backend capacity.
+
+However, the current WRR implementation lacks support for any ramp-up or warm-up phase. When a new endpoint appears—either due to autoscaling, replacement, or recovery from a failure—the client immediately starts routing requests based on the full weight reported for that endpoint. This can result in overloading the endpoint before it is fully initialized, negatively impacting response times and service reliability. It is especially problematic for systems with cold caches, JIT compilation warm-up delays, or dependency initialization steps.
```
markdroth: It's not strictly true that WRR lacks any mechanism for warming up new endpoints: the blackout period is designed to do some of this.  We should mention that here and explain why slow-start is an improvement over that existing functionality.

Edit: I see you mentioned the blackout period below.  However, the wording here still implies that there is no existing mechanism for this at all, which is misleading.  I would be fine with saying something like "The current WRR implementation has a blackout period mechanism that does something similar, but it's not sufficient for all cases, for reasons described below."

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+In contrast, many modern systems adopt slow start strategies in load balancing to address these issues. These strategies allow endpoints to ramp up traffic gradually over a defined window, smoothing transitions and mitigating the risks of traffic spikes. Similar functionality exists in Envoy's load balancing policies, where slow start is implemented for round robin and least request policies.
+
+Introducing a `slow_start_config` configuration in gRPC WRR will offer these benefits within the native client policy, reducing reliance on external traffic-shaping mechanisms or manual intervention.
+
+### Related Proposals:
```
markdroth, line 44: Please also reference A79, A78, and A89 here, since they're relevant to the metric definition below.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+  // The value of aggression parameter should be greater than 0.0.
+  // By tuning the parameter, it is possible to achieve polynomial or exponential shape of ramp-up curve.
+  //
+  // During slow start window, effective weight of an endpoint would be scaled with time factor and aggression:
+  // ``new_weight = weight * max(min_weight_percent / 100, time_factor ^ (1 / aggression))``,
+  // where ``time_factor=(time_since_start_seconds / slow_start_time_seconds)``.
+  //
+  // As time progresses, more and more traffic would be sent to endpoint, which is in slow start window.
+  // Once host exits slow start, time_factor and aggression no longer affect its weight.
+  google.protobuf.FloatValue aggression = 2;
```
markdroth: Assuming 0 is not a valid value here, I think we can use `float` instead of `google.protobuf.FloatValue`.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+  // Configures the minimum percentage of the original weight that will be used for an endpoint
+  // in slow start. This helps to avoid a scenario in which endpoints receive no traffic during the
+  // slow start window. Must be between 0 and 100. If not specified, the default is 10%.
+  google.protobuf.UInt32Value min_weight_percent = 3;
```
markdroth: If 0 is not a valid value here, I think we can use `uint32` instead of `google.protobuf.UInt32Value`.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+  // ``new_weight = weight * max(min_weight_percent / 100, time_factor ^ (1 / aggression))``,
+  // where ``time_factor=(time_since_start_seconds / slow_start_time_seconds)``.
+  //
+  // As time progresses, more and more traffic would be sent to endpoint, which is in slow start window.
+  // Once host exits slow start, time_factor and aggression no longer affect its weight.
+  google.protobuf.FloatValue aggression = 2;
+
+  // Configures the minimum percentage of the original weight that will be used for an endpoint
+  // in slow start. This helps to avoid a scenario in which endpoints receive no traffic during the
+  // slow start window. Must be between 0 and 100. If not specified, the default is 10%.
```
markdroth: Please clarify the valid range: are 0 and 100 valid values, or is the valid range only between those two values?

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+  // Configuration for slow start feature
+  SlowStartConfig slow_start_config = 8;
+}
+
+message SlowStartConfig {
+  // Represents the size of slow start window.
+  // If set, the newly created host remains in slow start mode starting from its creation time
+  // for the duration of slow start window.
+  google.protobuf.Duration slow_start_window = 1;
```
markdroth: What is the default if this is unset?

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+  // This parameter controls the speed of traffic increase over the slow start window. Defaults to 1.0,
+  // so that endpoint would get linearly increasing amount of traffic.
+  // When increasing the value for this parameter, the speed of traffic ramp-up increases non-linearly.
+  // The value of aggression parameter should be greater than 0.0.
```
markdroth: I assume we're enforcing this.  If so, please change "should" to "must".

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+  // By tuning the parameter, it is possible to achieve polynomial or exponential shape of ramp-up curve.
+  //
+  // During slow start window, effective weight of an endpoint would be scaled with time factor and aggression:
+  // ``new_weight = weight * max(min_weight_percent / 100, time_factor ^ (1 / aggression))``,
+  // where ``time_factor=(time_since_start_seconds / slow_start_time_seconds)``.
```
markdroth: What is `slow_start_time_seconds`?  I don't see that defined anywhere.  Should this be `slow_start_window`?

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+  //
+  // As time progresses, more and more traffic would be sent to endpoint, which is in slow start window.
+  // Once host exits slow start, time_factor and aggression no longer affect its weight.
```
markdroth: s/host/endpoint/

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+```
+
+### Weight Scaling During Warmup
+
+When an endpoint is ready after being in a non-ready state, it enters the warmup period. During this period, its weight will be scaled by a factor that increases non-linearly from `min_weight_percent` to 100% over the duration of `slow_start_window`.
+
+The scale factor is calculated as follows:
+
+```
+time_factor = max(time_since_ready_in_seconds, 1) / slow_start_window_seconds
```
markdroth: This forumla doesn't match the comment in the proto above.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+scale = max(min_weight_percent/100, time_factor ^ (1/aggression))
+```
+
+The following image shows how different aggression values affect the scaling factor over time during the slow start window (in milliseconds):
+
+![Effect of aggression parameter on slow start scaling](A123_graphics/aggression_scaling.png)
```
markdroth: The link is broken -- I think we need to change "A123" to "A100".

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+In contrast, many modern systems adopt slow start strategies in load balancing to address these issues. These strategies allow endpoints to ramp up traffic gradually over a defined window, smoothing transitions and mitigating the risks of traffic spikes. Similar functionality exists in Envoy's load balancing policies, where slow start is implemented for round robin and least request policies.
+
+Introducing a `slow_start_config` configuration in gRPC WRR will offer these benefits within the native client policy, reducing reliance on external traffic-shaping mechanisms or manual intervention.
+
+### Related Proposals:
+* [gRFC A58: weighted_round_robin LB policy][A58]
```
markdroth: Please add an "Updated by:" header in A58 that points to this gRFC.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+final_weight = computed_weight * scale
+```
+
+### Subchannel Weights
+
+To maintain independence between the blackout period and slow start period, a new timestamp `ready_since` will be introduced to track when an endpoint transitioned to ready state. This timestamp is separate from the existing `non_empty_since` timestamp used for the blackout period.
+
+The `ready_since` timestamp is set when an endpoint transitions from a non-ready state (e.g., CONNECTING, TRANSIENT_FAILURE) to a ready state (READY). This marks the beginning of the slow start period for that endpoint. The existing `non_empty_since` timestamp continues to be used exclusively for tracking the blackout period, which begins when the first non-zero load report is received.
+
+Weight calculation in the WRR policy follows a two-step process. First, the base weight for each endpoint is computed using the formula from [gRFC A58][A58]:
```
markdroth: This part seems to basically repeat what is already covered in the previous section.  I don't think we need to spell out the formula twice.  Maybe just remove the previous section?

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+This base weight represents the endpoint's capacity based on its current performance metrics. During the warmup period, this weight is then scaled using the slow start formula:
+
+$$scaled\\_weight = weight * max(min\\_weight\\_percent, time\\_factor ^ {1/aggression})$$
+
+where,
+
+$$time\\_factor = \dfrac{max(time\\_since\\_ready\\_seconds, 1)}{slow\\_start\\_window\\_seconds}$$
+
+The scaling formula ensures that new endpoints receive a gradually increasing share of traffic while maintaining a minimum threshold to prevent starvation. The `aggression` parameter allows fine-tuning of the ramp-up curve, enabling either more aggressive initial scaling (values > 1.0) or more conservative approaches (values < 1.0).
```
markdroth: This conflicts with the proto comment above that says that aggression should be greater than 0.

A100-client-side-weighted-round-robin-slow-start.md
markdroth: Please add a link to the xDS PR.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+- All endpoints are relatively new (e.g., new deployment in Kubernetes)
+- The system has low traffic volume
+- There are many endpoints in the cluster
+
+In these cases, the slow start feature may lead to:
+- Endpoint starvation (low probability of receiving requests)
+- Non-gradual traffic increases when requests do arrive
+- Reduced effectiveness of the load balancing algorithm
+
+## Scope and Limitations
```
markdroth: This should have one more `#`, so that it's a subsection of the `Rationale` section.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+This proposal specifically focuses on implementing slow start for the weighted round robin load balancing policy. While similar slow start functionality could potentially be implemented for other load balancing algorithms like Round Robin and Least Request, these are not included in this proposal for the following reasons:
+
+1. These algorithms don't use weights to determine endpoint selection, making the implementation of slow start more complex
+2. Additional considerations would be needed for how to gradually increase traffic to new endpoints in these algorithms
+3. The implementation details would likely differ significantly from the weighted round robin approach
+
+These other load balancing algorithms can be considered for slow start implementation in future proposals, with their own specific design considerations and requirements.
+
+## Metrics
```
markdroth: This should be a subsection of the `Proposal` section.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+
+The following metric will be exposed to help monitor the slow start behavior:
+
+`grpc.lb.wrr.endpoints_in_slow_start`
+- Type: Gauge
```
markdroth: Using a gauge is going to cause conflicts whenever the application creates two channels to the same target.  I suggest making this a counter instead, which is what we do for the existing WRR metrics.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+2. Additional considerations would be needed for how to gradually increase traffic to new endpoints in these algorithms
+3. The implementation details would likely differ significantly from the weighted round robin approach
+
+These other load balancing algorithms can be considered for slow start implementation in future proposals, with their own specific design considerations and requirements.
+
+## Metrics
+
+The following metric will be exposed to help monitor the slow start behavior:
+
+`grpc.lb.wrr.endpoints_in_slow_start`
```
markdroth: Please note that this metric will start as experimental and off by default, and we will consider stabilizing it in the future.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
+A100: Client-side weighted round robin slow start configuration
+----
+* Author(s): [Anurag Agarwal](https://github.com/anuragagarwal561994)
+* Approver: a11r
```
markdroth: Please list @markdroth, @ejona86, and @dfawley as approvers.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+These other load balancing algorithms can be considered for slow start implementation in future proposals, with their own specific design considerations and requirements.
+
+## Metrics
+
+The following metric will be exposed to help monitor the slow start behavior:
+
+`grpc.lb.wrr.endpoints_in_slow_start`
+- Type: Gauge
+- Description: Number of endpoints currently in slow start period
+- Labels:
```
markdroth: This should also have the `grpc.target` label, as defined in A66.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+  - `grpc.lb.locality`: The locality of the endpoints
```
markdroth: Please reference A78 for this label.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,254 @@
...
+  - `grpc.lb.backend_service`: The backend service name
```
markdroth: Please reference A89 for this label.

markdroth (commented):
Thanks for writing this up!

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: @markdroth I am actually confused here on how to integrate this. Actually envoy already has this proto for `SlowStartConfig` which it uses for RR and LR load balancing.

Hence, there was no need of any changes in `SlowStartConfig` proto anywhere, the only change required was to inlcude it in the `ClientSideWeightedRoundRobin` proto at the envoy side.

So I think in an earlier suggestion on this proposal, I was asked to change it to a google type while envoy types were being used https://github.com/grpc/proposal/pull/498/commits/60cc61679a6b9eadbbca4e170f7043b014c3c3da

I just wanted to reference this proto for others to relate to the equations, but if we are going to use the envoy proto only then there we should make minimal changes for not confusing others. What would you suggest.

https://github.com/envoyproxy/envoy/blob/cdf3dd3b6e1f0a1274f097bd7f43350927b93f93/api/envoy/extensions/load_balancing_policies/common/v3/common.proto#L89-L113

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Same concern as stated above

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: There is no default value as such, if slow_start_window is present, only then the config is applicable. It is more of a required parameter. Other parameters are optional.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Yes this is slow_start_window, just that the slow_start_window is in Duration, here we are just calculating things at the seconds granularity. I will still change the parameter name to `slow_start_window_seconds` if that gives more clarity.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Actually there are proto validate rules at the envoy side, so I am assuming that we will not get values > 100 or < 0, however if provided as json via custom load balancing policy, I believe one might have to throw an exception.

The value is optional, if not given this will be 10%.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: It doesn't, aggression must be > 0, but it can be < 1 it is a floating value.

anuragagarwal561994:
@markdroth  I have made the respective changes and added a few comments for where I was confused, please help me resolve the same.

A100-client-side-weighted-round-robin-slow-start.md
atollena: While Mark's comment make sense if that was a new message, here I think we should stick with the existing Envoy definition of `SlowStartConfig`.

markdroth: I agree that the xDS API should use the existing xDS `SlowStartConfig` proto.  But this section of the gRFC is talking about the gRPC service config format, which may be used without xDS.  The gRPC service config format is defined in https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto, which does not and should not depend on xDS protos, and we'll need to update that file to support the new fields being added in this gRFC.

Note that while we define the gRPC service config format in protobuf form to provide a well-defined schema for it, we don't actually use that proto in our implementations; internally, gRPC takes the service config in JSON form.  This means that when using xDS, there is some translation needed to convert the xDS proto into our internal service config format.  And given that we're doing a translation anyway, there's no real benefit in having the field types exactly match between the two.

Our approach for this kind of thing has always been to define the gRPC service config fields in a way that makes sense for gRPC, independent of the xDS representation.  Just as one example, there are several key differences between the outlier detection LB policy config in gRPC and the corresponding xDS representation; gRFC A50 defines both the [gRPC service config format](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md#lb-policy-config) and the [mapping from xDS](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md#xds-integration).

A100-client-side-weighted-round-robin-slow-start.md
markdroth: If this field is required, that needs to be explicitly stated, so that all implementations know what validation rules to use here.  The current wording says "If set", which implies the exact opposite.

A100-client-side-weighted-round-robin-slow-start.md
markdroth: Same response as above.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -194,55 +240,73 @@ message SlowStartConfig {
...
+xDS PR: https://github.com/envoyproxy/envoy/pull/40090
+
+### Metrics
+
+The following metric will be exposed to help monitor the slow start behavior:
+
+`grpc.lb.wrr.endpoints_in_slow_start`
+
+- Type: Counter
+- Description: Number of endpoints currently in slow start period
```
markdroth: Please define exactly when the counter will be incremented.  I assume this will happen every time we construct a new scheduler?

A100-client-side-weighted-round-robin-slow-start.md
markdroth: The validation rules need to be stated explicitly as part of this spec, so that gRPC implementations will all have the same behavior.  If 0 and 100 are not valid values, then we need to explciitly say that.

When an implementation sees an invalid value in any config field, it must reject the config.  It may be that in the Java implementation, throwing an exception is the right way to do that, but that's certainly not true in other languages, and this gRFC should be written in a language-agnostic way.

markdroth (commented):
Sorry for the delay; I was on vacation for a few weeks, and I got Covid while I was out, and I'm still recovering...

This looks good overall, but the config content still needs work.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: I now understand the concern, I didn't know that we are maintaining a separate proto file in another repo. I will update the proposal accordingly.

altaiezior: However just one concern @markdroth, yes 0 is not a valid value here but the value can still be optional as we have assumed a default value as specified in the doc, should we keep it `optional float` or `FloatValue`. Because from the other reference that you have shared, I see that we are using `FloatValue` for when we are specifying defaults

markdroth: In proto3, any non-repeated numeric field will have the value 0 if it is not set, and explicitly setting the field to 0 is equivalent to leaving it unset.  So if a client sees the value 0, it can assume the default.  (The same is true for string fields; the default value there is the empty string.)

The only case where you need to use a wrapper type is if you need to distinguish between 0 and unset -- i.e., if 0 is actually a valid value that means something different than unset.  But that's not the case here, so there's no benefit in using the wrapper type.

altaiezior: Understood, but how about using optional here isn't that more representative?

And the implementer can also generate has functions to determine if the value is set / unset

markdroth: I don't think `optional` actually helps here, because it doesn't change the fact unset and 0 are indistinguishable from each other.

altaiezior: Sure then I will try to follow the standardards already used in the current proto file

anuragagarwal561994:
@markdroth I have made the respective changes, let me know if there are more changes required

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,343 @@
...
+  // ``new_weight = weight * max(min_weight_percent / 100, time_factor ^ (1 / aggression))``,
+  // where ``time_factor = max(time_since_start_seconds, 1) / slow_start_window_seconds``.
+  //
+  // As time progresses, more and more traffic would be sent to endpoint, which is in slow start window.
+  // Once endpoint exits slow start, time_factor and aggression no longer affect its weight.
+  float aggression = 2;
+
+  // Configures the minimum percentage of the original weight that will be used for an endpoint
+  // in slow start. This helps to avoid a scenario in which endpoints receive no traffic during the
+  // slow start window. Valid range is [0.0, 100.0]. If the value is not specified, the default is 10%.
```
markdroth, line 99: A valid range of "[0.0, 100.0]" using square brackets implies that 0 and 100 are included in the range.  If you intended 0 and 100 to be excluded from the range, you should write "(0.0, 100.0)".

If 0 is included in the valid range but we want it to default to 10% if unset, then we need to use `google.protobuf.FloatValue`.  Otherwise, we can't differentiate between it being explicitly set to 0 and it being unset.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,343 @@
...
+allow endpoints to ramp up traffic gradually over a defined window, smoothing transitions and mitigating the risks of
+traffic spikes. Similar functionality exists in Envoy's load balancing policies, where a slow start is implemented for
+round-robin and the least request policies.
+
+Introducing a `slow_start_config` configuration in gRPC WRR will offer these benefits within the native client policy,
+reducing reliance on external traffic-shaping mechanisms or manual intervention.
+
+### Related Proposals:
+
+* [gRFC A24][A24]
```
markdroth: Please have the display text for each of these bullets list the title of each gRFC -- e.g., change "gRFC A24" to "A24: Load Balancing Policy Configuration".

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,343 @@
...
+  // The value of aggression parameter must be greater than 0.0.
+  // By tuning the parameter, it is possible to achieve polynomial or exponential shape of ramp-up curve.
+  //
+  // During slow start window, effective weight of an endpoint would be scaled with time factor and aggression:
+  // ``new_weight = weight * max(min_weight_percent / 100, time_factor ^ (1 / aggression))``,
+  // where ``time_factor = max(time_since_start_seconds, 1) / slow_start_window_seconds``.
+  //
+  // As time progresses, more and more traffic would be sent to endpoint, which is in slow start window.
+  // Once endpoint exits slow start, time_factor and aggression no longer affect its weight.
+  float aggression = 2;
```
markdroth: In the xDS config, this is a double.  Should it be a double here too, so that we don't lose precision?

A24-lb-policy-config.md
```diff
@@ -6,6 +6,7 @@ Load Balancing Policy Configuration
 * Implemented in: C-core
 * Last updated: 2018-12-05
 * Discussion at: https://groups.google.com/d/topic/grpc-io/K03NV5H8HoE/discussion
+* Updated By: [A100-client-side-weighted-round-robin-slow-start.md](A100-client-side-weighted-round-robin-slow-start.md)
```
markdroth: This isn't necessary.  A100 does not actually modify A24 at all.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,343 @@
...
+
+Introducing a `slow_start_config` configuration in gRPC WRR will offer these benefits within the native client policy,
+reducing reliance on external traffic-shaping mechanisms or manual intervention.
+
+### Related Proposals:
+
+* [gRFC A24][A24]
+* [gRFC A58][A58]
+* [gRFC A66][A66]
+* [gRFC A78][A78]
```
markdroth: Please add an "Updated-by:" header in A78 pointing to this gRFC.

markdroth (commented):
This is looking really good!  Only a few comments remaining.

For future reference, please don't force-push to a PR after reviews have started, because that makes it very difficult for a reviewer to tell what's changed since their last review pass.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Got it, so it's not like 0 and 100 are not valid values, they doesn't make much sense but technically they are valid values and won't make much difference in the execution of the slow start logic.

So I was referring to the envoy proto and there `type.v3.Percent` is used. Since in the later section I have also covered steps to convert from xDS configuration, I thought it would make much more sense to keep their ranges similar. In case tomorrow we have Istio implementing it, this would be much more convinient and prevent confusions.

What would be your opinion about the same, I think inclusive ranges are fine and instead we can just use `FloatValue` here?

A24-lb-policy-config.md
altaiezior: It modifies LB policy config so I thought it needs to be updated, I will revert

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Only this is remaining and rest of the changes have been addressed

markdroth: The xDS type `type.v3.Percent` does indeed allow both 0 and 100 as explicit values.  So I think we should use `google.protobuf.FloatValue` here, so that we can differentiate between 0 and unset.

altaiezior: This is done

markdroth (approved):
Looks great!

atollena (approved):

anuragagarwal561994:
@ejona86 @dfawley can you help review and merge the proposal?

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,342 @@
...
+}
+
+message SlowStartConfig {
+  google.protobuf.Duration slow_start_window = 1;
+  core.v3.RuntimeDouble aggression = 2;
+  type.v3.Percent min_weight_percent = 3;
+}
+```
+
+xDS PR: https://github.com/envoyproxy/envoy/pull/40090
```
dfawley, line 245: It seems this is closed now.  Will it be possible to reopen it and get it merged?

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,342 @@
...
+
+### Metrics
+
+The following metric will be exposed to help monitor the slow start behavior:
+
+`grpc.lb.wrr.endpoints_in_slow_start`
+
+- Type: Counter
+- Description: Number of endpoints currently in the slow start period. This is incremented when a new scheduler is
+  created.
```
dfawley, line 265: Schedulers are recreated pretty frequently, aren't they?

Should this be incremented at the start of the connection instead?

dfawley (commented):
One concern (and a minor change will be needed if we can't keep the same envoy PR number):

A100-client-side-weighted-round-robin-slow-start.md
markdroth: Most of the existing similar WRR metrics get updated when we recreate the scheduler, so I think this is consistent with the existing pattern.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Yes we can re-open this, so this was auto-closed because the envoy team is waiting on this proposal to get merged first, then we can merge there

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Yes so whenever the scheduler is re-created, we are collecting the endpoints latest weight and at that time we are also updating the respective metrics.

In java it is happening in every 1 second, so there is not additional processing it is just done at the time of computing weights.

dfawley: Is it usable that way (and how)?

altaiezior: So it is helpful like other metrics too which tells us that there are currently endpoints which are in slow start period, may not be very helpful in case the slow start duration is small. But if it is high, then it can be of use to inspect why may be some pods are not getting the traffic.

Moreover instead I feel that the current metrics we are exposing should actually be gauges instead of counters anyways, because I don't think rate of these metrics makes much sense. I have just added to make things consistent with the current workflow.

dfawley: I guess the way you consume these metrics is by computing the _rate_, and then the rate is essentially your gauge?  I'm not sure how I feel about this as a general strategy.  Is that a recommendation from OTel?

If our other metrics are this way, then we can approve this for this PR regardless.

altaiezior: Yes I mostly reused the template implementation of other metrics from related Proposals

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Yes I have reinitiated the conversation there let me try getting it reopened

A100-client-side-weighted-round-robin-slow-start.md
markdroth: Yes, Doug, that's exactly the way these metrics are intended to be used.  That's the expected pattern for basically any counter metric.  There's essentially no other useful way to do it, given the need to aggregate the data across tasks.

dfawley: > That's the expected pattern for basically any counter metric

Not really.  Most counters are accumulating the number of occurrences of a thing.  I.e. the number of RPC failures.  You _can_ use the rate for a metric like that to see how often it's happening, but the raw number still means something interesting in isolation.

Here, we are accumulating events at a rate where the only useful way of looking at it is via its rate of change relative to the rate at which we rebuild the scheduler (which is configurable), and the raw number is mostly meaningless on its own.

It might be better if there was also a counter for the total number of scheduler rebuilds so you could combine the two metrics to find out what you probably really want to know.  I.e. "number of endpoints in slow start in every scheduler rebuild" / "number of scheduler rebuilds" = "number of endpoints in slow start".

markdroth: In practice, no one ever looks at raw counter values.  It's simply not useful, since the data is always aggregates across tasks and possibly other dimensions, and the only way to do that aggregation is to take the rate.

You're right that this is a case where the counter will be incremented multiple times on each scheduler rebuild, which is a little different than the more typical case of a counter that tracks individual discrete events.  But I think the way it winds up being used is basically the same in practice.  Keep in mind that the scheduler rebuilds happen on a pretty regular schedule, so it will be easy to see spikes on the graph without a separate counter for rebuilds.

In any case, this concern is really not unique to the new metric being added in this gRFC, because the existing WRR metrics work the same way.  Let's stick with the existing pattern, and if someone has a problem with these metrics, we can address that later.

dfawley (approved):

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: @dfawley MR is now open and I have requested review

anuragagarwal561994:
@ejona86 a quick follow up on the proposal, as the envoy side MR is also merged. I will be starting the implementations in the upcoming month now

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,342 @@
+A100: Client-side weighted round-robin slow start configuration
+----
+
+* Author(s): [Anurag Agarwal](https://github.com/anuragagarwal561994)
+* Approver: Mark Roth (@markdroth), Eric Anderson (@ejona86), Doug Fawley (@dfawley)
+* Status: Draft
+* Last updated: 2025-05-31
+* Discussion at: TBD
```
ejona86, line 8: It looks like this still needs a thread started on the mailing list https://groups.google.com/g/grpc-io

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,342 @@
...
+}
+```
+
+### Subchannel Weights and Scaling During Warmup
+
+To maintain independence between the blackout period and slow start period, a new timestamp `ready_since` will be
+introduced to track when an endpoint transitioned to ready state. This timestamp is separate from the existing
+`non_empty_since` timestamp used for the blackout period.
+
+The `ready_since` timestamp is set when an endpoint transitions from a non-ready state (e.g., CONNECTING,
```
ejona86, line 110: Re-starting slow start each time an endpoint becomes READY seems much more aggressive than Envoy, and doesn't seem to serve much purpose. I'd expect options more like:
1. When an endpoint is added (since UNHEALTHY endpoints are filtered, this will re-trigger slow start when it becomes HEALTHY; dunno if there's interactions with stateful session affinity plumbing). This seems very close to Envoy ("If no active healthcheck is configured per cluster, immediately upon joining the cluster.")
2. The first time an endpoint is READY, and the first time it goes READY after having been TRANSIENT_FAILURE. This seems pretty normal for gRPC

It would seem bad for MAX_CONNECTION_AGE to keep triggering slow-start, and MAX_CONNECTION_AGE will trigger CONNECTING.

A100-client-side-weighted-round-robin-slow-start.md
```diff
@@ -0,0 +1,342 @@
...
+    - `grpc.lb.locality`: The locality of the endpoints [gRFC A78][A78]
+    - `grpc.lb.backend_service`: The backend service name [gRFC A89][A89]
+    - `grpc.target`: gRPC channel target [gRFC A66][A66]
+
+This metric will help operators monitor the number of endpoints currently in slow start mode across different localities
+and backend services.
+
+Please note that this metric will start as experimental and off by default, and we will consider stabilizing it in the
+future.
+
```
ejona86, line 276: There's no "Temporary environment variable protection" section. How will the implementation process go here? Seems the changes should be behind a variable until at least someone reports it is working.

ejona86 (commented):
Nicely written. My READY comment will probably cause some discussion. I see this as very close to ready.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: I have added the thread now, waiting for it be approved

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Yes that makes sense, so should I add the second one? That slow start should trigger when an endpoint becomes ready for the first time or transitioning from TRANSIENT_FAILURE.

Will also have to make changes with respect to the variable name, something like `added_since`?

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: I was referring to the current metrics implementations, they are turned off by default and if someone enables them via lets say opentelemetry they get enabled. There is no environment vairable, although this is not very well documented yet for how a user can enable these metrics. I also struggled and got to know about them when I was going through the code base in java

A100-client-side-weighted-round-robin-slow-start.md
ejona86: This will need input from @markdroth and @dfawley. We discussed it (on the 7th?) and realized it was in weak conflict with the blackout period behavior. I'm fine with only "fixing" this one, or changing both this and blackout period behavior. I also have a sneaky suspicion that Envoy implemented blackout period differently than we did. But overall there's several different views on the subject and we need to align.

A100-client-side-weighted-round-robin-slow-start.md
ejona86: I'm not concerned about the metrics, but the feature in general. The missing section would normally have been here.

A100-client-side-weighted-round-robin-slow-start.md
altaiezior: Is there a regular meeting or Slack channel where I can reach the project contributors directly? PR follow-ups can be slow, and I’d like a quicker way to clarify open questions

anuragagarwal561994:
@ejona86 can you help me propose the changes or help me connect with you so we can close this proposal

altaiezior:
@ejona86 can you help me close this proposal?

davinci26:
@markdroth @ejona86 @atollena @dfawley @altaiezior I was wondering if there is anything we can do to push the proposal across the finish line and I think we would most likely be able to sponsor the grpc-go implementation as well!

While it is possible at least in `grpc-go` to implement this as a custom load balancer we think it would be a great feature for project, so we wanted to see if there is a plan to get this merged

PR #499, opened 2025-06-04
L123: Add Python get_target() API

rajatnaik0405:
This design document proposes a public get_target() API for gRPC Python, to expose the target address string


A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+* Approver: TBD
+* Status: Draft
+* Updated: 2025-05-29
+* Implementation Owner: Rajat Naik
+* Reviewers: TBD
+* Language: Python
+
+## 1. Abstract
+
+Introduce a new public API `get_target()` in gRPC Python that returns the channel's target string. This API provides a stable way for clients and libraries (such as Google Pub/Sub) to retrieve the underlying channel endpoint for observability, debugging, and routing.
```
sreenithi: remove the reference to Google Pub/Sub, I don't think it's necessary in this gRFC

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+
+## 2. Background
```
sreenithi: This whole Background section is wrong as well. You can just mention that there is no existing direct way to get the channel target from the Channel object, because of which users have to resort to indirect methods to find it and link the Github issue. You don't need to be discussing the Github issue or how Google Pub/Sub is using it here.

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+def get_target(self) -> Optional[str]:
+    """Returns the verbatim channel target if the channel is active, otherwise None."""
+```
+
+### 5.2 Implementation
+
+1. Use the existing `cygrpc.Channel.check_connectivity_state(try_to_connect=False)` to retrieve the current raw connectivity state.
+2. Map the returned `cygrpc` enum to `grpc.ChannelConnectivity` via `_common.CYGRPC_CONNECTIVITY_STATE_TO_CHANNEL_CONNECTIVITY`.
+3. If the mapped state is not one of `{IDLE, CONNECTING, READY}`, return `None` immediately.
+4. Otherwise, call `self._channel.target().decode("utf-8")`. Catch `UnicodeDecodeError` and return `None` if decoding fails.
```
sreenithi: this will change corresponding to [my review comment](https://github.com/grpc/grpc/pull/39536/files#r2128057086) in the implementation PR:

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+
+```python
+from typing import Optional
+import grpc
+from grpc._cython import cygrpc
+from grpc.framework.common import _common
+
+class ChannelWithTarget(grpc.Channel):
```
sreenithi: do not change the name of the class, it should match the exact existing class where you are adding this function to

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+1. Use the existing `cygrpc.Channel.check_connectivity_state(try_to_connect=False)` to retrieve the current raw connectivity state.
+2. Map the returned `cygrpc` enum to `grpc.ChannelConnectivity` via `_common.CYGRPC_CONNECTIVITY_STATE_TO_CHANNEL_CONNECTIVITY`.
+3. If the mapped state is not one of `{IDLE, CONNECTING, READY}`, return `None` immediately.
+4. Otherwise, call `self._channel.target().decode("utf-8")`. Catch `UnicodeDecodeError` and return `None` if decoding fails.
+
+```python
+from typing import Optional
+import grpc
+from grpc._cython import cygrpc
+from grpc.framework.common import _common
```
sreenithi: the class definition is enough for this section, these import lines are redundant, can be removed.

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+# After shutting down the channel
+tchannel.close()
+print(channel.get_target())  # Expected: None
+```
+
+## 7. Alternatives Considered
+
+1. **Maintain Private Internals:** Continue using private attributes (`_channel` or `_thunk`) to retrieve the target.
+
+   * Drawback: Breaks whenever internals change; no stability guarantee.
```
sreenithi: this doesn't look relevant. should be removed

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+
+   * Drawback: Error-prone, duplicative, and cumbersome to maintain.
+
+By adding a public method on `grpc.Channel`, users get a first-class, stable API that works regardless of future refactoring.
+
+## 8. Rationale
+
+* **Simplicity:** A single method call (`get_target()`) is clearer than complex workarounds.
+* **Stability:** As part of the public surface, this API will receive compatibility guarantees.
+* **Usability:** Libraries like Google Pub/Sub or other monitoring tools can rely on this method without custom logic.
```
sreenithi: again, I don't think we should mention Google Pub/Sub here

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+
+## 9. Implementation Plan
+
+1. **Modify `channel.py` in gRPC Python core:** Add the `get_target()` definition.
+2. **Update `_common.py`:** Ensure `CYGRPC_CONNECTIVITY_STATE_TO_CHANNEL_CONNECTIVITY` mapping includes all relevant states.
```
sreenithi: I don't see any such update to `_common.py` in the [PR](https://github.com/grpc/grpc/pull/39536) as of today, and I think it will work with the existing code in `_common.py` without any changes.

This document is to talk about any changes to the existing design or code of the gRPC library, this line should be removed if there are no changes made there.

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+3. **Add unit tests:** Cover:
+
+   * Active channel returns the correct target.
+   * Closed or failed channel returns `None`.
+4. **Update documentation:** Add a note about `get_target()` in the public API docs and README.
+5. **Release:** Include in the next minor version (e.g., 1.50.0).
```
sreenithi: The release version is wrong. The release when this change can most likely go in is in 1.74.0

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+## 10. Backwards Compatibility
+
+* Existing code that does not call `get_target()` remains unaffected.
+* Introducing a new method does not break any existing public APIs.
+
+## 11. References
+
+* [gRPC Python Repository](https://github.com/grpc/grpc/tree/master/src/python)
+* [A6: Client Retries Design](https://github.com/grpc/proposal/blob/master/A6-client-retries.md)
+* Internal Pub/Sub issue: `broken due to interceptor architecture` (Private link)
```
sreenithi: This section needs to be updated. 
1) I don't think the Python repository reference is needed here.
2) The `Client retries design` seems irrelevant to me, can you explain how the reference was used in this comment?
3) The Pub/Sub issue was an external Github issue, not sure why it says Internal link, and the link is not updated too

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
+# A100: Expose Channel Target via Public API
```
sreenithi: Only common design changes across all languages supported by gRPC should be an `Ax` change. This is a language specific change, so should be `Lx`, and the name of the file must include `python` 
For example: https://github.com/grpc/proposal/blob/master/L13-python-interceptors.md

Also the number used in the name of the file and in this first line must be the same.

A91-get-target-api.md
```diff
@@ -0,0 +1,155 @@
...
+   * Closed or failed channel returns `None`.
+4. **Update documentation:** Add a note about `get_target()` in the public API docs and README.
+5. **Release:** Include in the next minor version (e.g., 1.50.0).
+
+## 10. Backwards Compatibility
+
+* Existing code that does not call `get_target()` remains unaffected.
+* Introducing a new method does not break any existing public APIs.
+
+## 11. References
```
sreenithi: Please also add a section called `Implementation` and link your PR there, like in https://github.com/grpc/proposal/blob/master/L13-python-interceptors.md

sreenithi (changes requested):

PR #501, opened 2025-06-18
A99: xDS Aggregate Cluster Resources

markdroth:
CC @fuqianggao


dfawley (approved):

ejona86 (commented):
Is this still up-in-the-air? I remember discussing it some, and it sounded like other options probably weren't available. But I'm also not sure if any movement forward started up again.

markdroth:
Yeah, this is still up in the air, and we never got back to it.  I'll mark this PR as a draft for now.

PR #502, opened 2025-06-22
Nogil mode  support for gRPC Python

aditya-1112:


aditya-1112:
The above proposal only addresses the failing aio tests (segfaults) issue (https://github.com/grpc/grpc/pull/39993), other errors are yet to be fixed.

PR #494, opened 2025-05-12
Proposal for max_outstanding_streams in ResourceQuota

siddharthnohria:

PR #504, opened 2025-06-30
L126: C-core: Support ALTS Credentials in Google Default Credentials

anniefrchz:


L123-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,83 @@
+# Support ALTS Credentials in Google Default Credentials
+
+--------------------------------------------------------------------------------
+
+*   Author(s): anniefrchz
+*   Approver: a11r
+*   Status: Draft
+*   Implemented in: C++
+*   Last updated: 2025/06/30
+*   Discussion at: <google group thread> (filled after thread exists)
```
rockspore: Please send an email (for broadcasting purposes) there and update this with the link.

L123-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,83 @@
+# Support ALTS Credentials in Google Default Credentials
```
rockspore: Let's replace ALTS credentials with ALTS Hard Bound Call Credentials. ALTS Credentials would be confusing since there is already the ALTS channel credentials.

L123-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,83 @@
...
+*   Status: Draft
+*   Implemented in: C++
+*   Last updated: 2025/06/30
+*   Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+This proposal outlines a change to the gRPC Core C-API to support
+alts-credentials configurations within `grpc_google_default_credentials`. This
+enhancement specifically allows a secondary set of channel credentials for ALTS
```
rockspore: Call credentials not channel credentials.

L123-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,83 @@
...
+to be provided alongside the default credentials.
+
+### Background
+
+The existing `grpc_google_default_credentials_create` function only allowed for
```
rockspore: Let's not frame it as a limitation of the existing API. Instead, we could say that we want to support hard bound call credentials specific to ALTS and therefore have the need to modify the existing C-core API.

L123-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,83 @@
...
+GRPCAPI grpc_channel_credentials* grpc_google_default_credentials_create(
+    grpc_call_credentials* tls_credentials,
+    grpc_call_credentials* alts_credentials);
+```
+
+This new function accepts two arguments:
+
+1.  `tls_credentials`: The primary call credentials, consistent with the
+    existing API. Usually. default back to the TLS connection.
+2.  `alts_credentials`: A secondary set of channel credentials to be used
```
rockspore: call credentials.

L123-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,83 @@
...
+    specifically for ALTS connections.
+
+When a channel is created with this function, the gRPC runtime will select the
+appropriate credentials based on the transport security type required for a
```
rockspore: The transport security type is populated after the secure connection is established and indicates the channel type (ie, TLS or ALTS). It's not a requirement. Please reword this paragraph.

L123-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,83 @@
...
+
+This approach was decided upon after initial feedback suggested modifying the
+existing API.
+
+## Rationale
+
+The primary motivation for this change is to enable seamless support for hybrid
+security environments on a single gRPC channel.
+
+The advantages of this approach are:
```
rockspore: Another important advantage is that since wrapper languages like Python already call this API, it will be straightforward to support ALTS bound credentials there by passing in the right ALTS call creds there.

rockspore (commented):
Thanks for writing up the proposal! Made some comments assuming we are fine with scoping this as a C-Core only gRFC.

I'll defer the gRPC TLs to decide whether an across-language one is needed (and I can help with that if so).

L123-core-add-alts-google-call-credentials.md
anniefrchz: Prending approval on the grpc.io group.

ejona86:
FYI: https://github.com/grpc/proposal/pull/496 is also L123

anniefrchz:
> FYI: #496 is also L123

Modify the numnber.

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
+# Support ALTS Hard Bound Call Credentials in Google Default Credentials
```
markdroth: Please change this to "L124: C-core: Support ALTS Hard Bound Call Credentials in Google Default Credentials"

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+
+--------------------------------------------------------------------------------
+
+*   Author(s): anniefrchz
+*   Approver: a11r
```
markdroth: Please change this to "markdroth".

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+*   Status: Draft
+*   Implemented in: C++
+*   Last updated: 2025/06/30
```
markdroth: Please use ISO date format (YYYY-MM-DD).

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
+# Support ALTS Hard Bound Call Credentials in Google Default Credentials
+
+--------------------------------------------------------------------------------
```
markdroth: This line isn't necessary.

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+
+The existing `grpc_google_default_credentials_create` function allows the
+configuration of a single set of call credentials. In some scenarios, a client
+would want to communicate with services that support hard bound credentials over
+ALTS. This proposal addresses the possibility to support this use case on a
+channel initialized with Google's default credentials.
+
+### Proposal
+
+To maintain backward compatibility with the existing C-API, this proposal
```
markdroth: It's odd to say "To maintain backward compatibility with the existing C-API", since we're making a breaking API change here.  I would remove that phrase.

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+*   Discussion at: https://groups.google.com/g/grpc-io/c/7rRWghiS95E/m/nENwU3BtCgAJ
+
+## Abstract
+
+This proposal outlines a change to the gRPC Core C-API to support
+alts-credentials configurations within `grpc_google_default_credentials`. This
+enhancement specifically allows a secondary set of call credentials for ALTS to
+be provided alongside the default credentials.
+
+### Background
```
markdroth: This should have only two `#` characters.

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+
+The existing `grpc_google_default_credentials_create` function allows the
+configuration of a single set of call credentials. In some scenarios, a client
+would want to communicate with services that support hard bound credentials over
+ALTS. This proposal addresses the possibility to support this use case on a
+channel initialized with Google's default credentials.
+
+### Proposal
```
markdroth: This should have only two `#` characters.

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+    supporting ALTS bound credentials will be straightforward by passing them as
+    an argument to the existing API.
+
+## Implementation
+
+The implementation for this proposal has been completed and merged into the main
+gRPC repository.
+
+*   Pull Request: grpc/grpc#39770
+*   Key Commit: The changes were integrated via commit ca2e8c9.
```
markdroth: This isn't necessary, since the commit can be found from the PR.

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+```c
+GRPCAPI grpc_channel_credentials* grpc_google_default_credentials_create(
+    grpc_call_credentials* tls_credentials,
+    grpc_call_credentials* alts_credentials);
+```
+
+This new function accepts two arguments:
+
+1.  `tls_credentials`: The primary call credentials, consistent with the
+    existing API. Usually. default back to the TLS connection.
```
markdroth: I don't understand this part:

"Usually. default back to the TLS connection."

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+
+After a secure connection is established, the gRPC runtime identifies the
+channel's transport security type, which indicates whether the underlying
+channel is using a protocol like ALTS or TLS. The runtime then selects the
+appropriate call credentials for that connection. If the determined transport
+security type is ALTS, the provided alts_credentials will be used. For all other
+transport types, the primary call_credentials are used, maintaining the default
+behavior.
+
+This approach was decided upon after initial feedback suggested modifying the
```
markdroth: I don't think we need this paragraph.  This is already covered in the "Rationale" section below.

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+    pull request favored this path to avoid an unnecessary new API.
+*   Improved Discoverability: Developers only need to be aware of a single
+    function for creating Google default credentials. The optional nature of the
+    second parameter would make the basic use case simple while allowing for the
+    more advanced dual-credential scenario when needed.
+*   Logical Cohesion: Since the new functionality is an extension of the
+    existing credential creation process, incorporating it into the original
+    function maintains logical cohesion. The function's responsibility is
+    expanded rather than duplicated across multiple functions.
+*   Cross-language support: Since wrapper languages like Python use this API,
```
markdroth: Do we intend to also make API changes for C++, Python, or other wrapped languages to expose this in their APIs?  If so, we'll need other gRFCs for those -- and for those APIs, we can't make breaking changes.

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+
+This new function accepts two arguments:
+
+1.  `tls_credentials`: The primary call credentials, consistent with the
+    existing API. Usually. default back to the TLS connection.
+2.  `alts_credentials`: A secondary set of call credentials to be used
+    specifically for ALTS connections.
+
+After a secure connection is established, the gRPC runtime identifies the
+channel's transport security type, which indicates whether the underlying
```
markdroth: s/channel/connection/

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,85 @@
...
+channel is using a protocol like ALTS or TLS. The runtime then selects the
```
markdroth: s/channel/connection/

L124-core-add-alts-google-call-credentials.md
anniefrchz: I removed the empty lines but i am not sure if you meant that ones or the "---------------------"

L124-core-add-alts-google-call-credentials.md
anniefrchz: I do not think we intend to make API changes on the other wrapped APIs. So far this is only scoped for this specific API.

L124-core-add-alts-google-call-credentials.md
anniefrchz: I meant that this set of credentials usually is used for TLS connections. I rephrased this a little, but I can also remove it as the name already implies it.

L124-core-add-alts-google-call-credentials.md
markdroth: I meant the one with the "---------------------".  If you look at the formatted page, this adds a second line that isn't necessary:

https://github.com/grpc/proposal/blob/1d45a7880133c088649f48d4112996163db75dc9/L124-core-add-alts-google-call-credentials.md

L124-core-add-alts-google-call-credentials.md
```diff
@@ -39,21 +36,18 @@ GRPCAPI grpc_channel_credentials* grpc_google_default_credentials_create(
 This new function accepts two arguments:
 
 1.  `tls_credentials`: The primary call credentials, consistent with the
-    existing API. Usually. default back to the TLS connection.
+    existing API. Usually, for TLS connections.
```
markdroth: This is a fragment, not a complete sentence: "Usually, for TLS connections."  Please change that to say "This is usually used for TLS connections."

L124-core-add-alts-google-call-credentials.md
markdroth: How are we actually going to use this if we don't expose this option in the APIs that applications actually use?

L124-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,78 @@
+# L124: C-core: Support ALTS Hard Bound Call Credentials in Google Default Credentials
```
markdroth: Looks like L124 and L125 are already taken (https://github.com/grpc/proposal/pull/494 and https://github.com/grpc/proposal/pull/505).  Please change this to L126.

Also, please rename the file accordingly.

anniefrchz: Done.

L124-core-add-alts-google-call-credentials.md
anniefrchz: Done

L124-core-add-alts-google-call-credentials.md
anniefrchz: Done

L126-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,92 @@
...
+
+After a secure connection is established, the gRPC runtime identifies the
+connection's transport security type, which indicates whether the underlying
+connection is using a protocol like ALTS or TLS. The runtime then selects the
+appropriate call credentials for that connection. If the determined transport
+security type is ALTS, the provided alts_credentials will be used. For all other
+transport types, the primary call_credentials are used, maintaining the default
+behavior.
+
+Additionally, external customers will de able to create GoogleDefaultCredentials
```
markdroth: I assume this is intended to talk about a C++ API change, not a C-core API change?  If so, we need to state that explicitly -- that intent is not at all clear from this text.

Also, what about other wrapped languages, like python?

L126-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,92 @@
...
+by setting a GoogleDefaultCredentialsOptions value into their standard call.
+For this addition, the proposed struct `GoogleDefaultCredentialsOptions` will hold
+a boolean that will be default to false. Callers of the GoogleDefaultCredentials()
+API will be able to set use_alts to false value, if required to indicate the
+request for the underlying bound token call credentials.
+
+```c
+struct GoogleDefaultCredentialsOptions {
```
markdroth, line 80: How is this struct actually used?  We need to show the complete API changes we're proposing here.

L126-core-add-alts-google-call-credentials.md
anniefrchz: Other wrapped languages are not a concern AFAIK, since the priority is giving C++ libraries access.
Python is in scope but we don't plan to have public API changes on the library for that so far.

Could we maybe just state the C++ change for now and note that expansion on other languages' public APIs  may be discussed further ?

markdroth: Yes, if we're not doing any wrapped languages right now, that seems fine.

L126-core-add-alts-google-call-credentials.md
anniefrchz: Done.

L126-core-add-alts-google-call-credentials.md
```diff
@@ -48,19 +49,48 @@ security type is ALTS, the provided alts_credentials will be used. For all other
...
+credentials.
 
 ```c
+typedef struct {
+  struct {
+    const char* param;
+    const char* value;
+  } QueryParam;
+
+  const QueryParam* query_params;
```
markdroth: How will the function know how many elements are in this array?

L126-core-add-alts-google-call-credentials.md
```diff
@@ -48,19 +49,48 @@ security type is ALTS, the provided alts_credentials will be used. For all other
...
 behavior.
 
-Additionally, external customers will de able to create GoogleDefaultCredentials
-by setting a GoogleDefaultCredentialsOptions value into their standard call.
-For this addition, the proposed struct `GoogleDefaultCredentialsOptions` will hold
-a boolean that will be default to false. Callers of the GoogleDefaultCredentials()
-API will be able to set use_alts to false value, if required to indicate the
-request for the underlying bound token call credentials.
+The hard-bound call credentials will be created through
+`grpc_google_compute_engine_credentials_create`. This function has a reserved
```
markdroth: The existing reserved argument isn't really worth mentioning, since we're allowed to make breaking changes in the C-core API.  I think it's sufficient to just say that we're changing the API of this function.

L126-core-add-alts-google-call-credentials.md
```diff
@@ -48,19 +49,48 @@ security type is ALTS, the provided alts_credentials will be used. For all other
...
+GRPCAPI grpc_call_credentials* grpc_google_compute_engine_credentials_create(
+    grpc_google_compute_engine_credentials_options* options);
+```
+
+Additionally, external customers for the public C++ library will de able to
+create GoogleDefaultCredentials by setting a GoogleDefaultCredentialsOptions
+value into their standard call. For this addition, the proposed struct
+`GoogleDefaultCredentialsOptions` will hold a boolean that will be default to
+false. Callers of the GoogleDefaultCredentials() API will be able to set
+use_alts to false value, if required to indicate the request for the underlying
```
markdroth: s/use_alts/use_alts_call_credentials/

L126-core-add-alts-google-call-credentials.md
```diff
@@ -48,19 +49,48 @@ security type is ALTS, the provided alts_credentials will be used. For all other
...
+transport protocol in the form of query parameters pairs, the caller will be
+able to obtain ALTS hard-bound credentials instead of the standard default call
+credentials.
 
 ```c
+typedef struct {
+  struct {
+    const char* param;
+    const char* value;
+  } QueryParam;
```
markdroth: Do we actually need the full generality of passing in arbitrary query params here?  Why not just pass in a single bool that says if it's for ALTS, just like we're doing for the C++ API?

Keep in mind that if there's a lot of translation needed between the public C++ API and the C-core API, we're going to have to reimplement that translation logic in every wrapped language when we eventually need to support this in those languages.

L126-core-add-alts-google-call-credentials.md
```diff
@@ -48,19 +49,48 @@ security type is ALTS, the provided alts_credentials will be used. For all other
...
+
+  const QueryParam* query_params;
+} grpc_google_compute_engine_credentials_options;
+
+GRPCAPI grpc_call_credentials* grpc_google_compute_engine_credentials_create(
+    grpc_google_compute_engine_credentials_options* options);
+```
+
+Additionally, external customers for the public C++ library will de able to
```
markdroth, line 71: Please create separate subsections for the C-core changes and the C++ changes, just to make it clear that there are two different layers here.

L126-core-add-alts-google-call-credentials.md
anniefrchz: We attempted to make the C-Core API agnostic to ALTS. SInce only C++ uses it there might not be a lot of trouble there. Wrapper languages create grpc_call_credentials in their own stack. Also opens up the possibility of expanding other configuration parameters here.

If this does not make sense either way, we could roll this back to use a single boolean as well (which I guess will be the easier implementation.

WDYT? @rockspore

markdroth: Do we have any other use-cases today where we will need other query params here?  If not, this fully general interface seems like overkill.

I think we can just use an options struct with a bool field, just like you're proposing for the C++ layer.  In the future, if we have additional options we want to add, we can just add them to the options struct, and it won't even be a breaking API change.

L126-core-add-alts-google-call-credentials.md
```diff
@@ -76,8 +78,10 @@ create GoogleDefaultCredentials by setting a GoogleDefaultCredentialsOptions
 value into their standard call. For this addition, the proposed struct
 `GoogleDefaultCredentialsOptions` will hold a boolean that will be default to
 false. Callers of the GoogleDefaultCredentials() API will be able to set
-use_alts to false value, if required to indicate the request for the underlying
-bound token call credentials.
+use_alts_call_credentials to false value, if required to indicate the request
+for the underlying bound token call credentials.
+
+### C++ Changes
```
markdroth: This heading needs to move up above the previous paragraph, which talks about the C++ API.

L126-core-add-alts-google-call-credentials.md
anniefrchz: Done!
Thanks :)

L126-core-add-alts-google-call-credentials.md
```diff
@@ -0,0 +1,122 @@
...
+The hard-bound call credentials will be created through
+`grpc_google_compute_engine_credentials_create`. This function will have a
+`grpc_google_compute_engine_credentials_options` parameter. The caller will be
+able to obtain ALTS hard-bound credentials instead of the standard default call
+credentials by toggling the corresponding flag use_alts_call_credentials in the
+structure.
+
+```c
+typedef struct {
+  bool use_alts_call_credentials = false;
```
markdroth: The name `use_alts_call_credentials` makes sense in the context of the C++ `GoogleDefaultCredentials()` API, where we're creating a channel credentials object and telling it what type of call credentials to use to attach to the channel credentials object.  However, in the case of `grpc_google_compute_engine_credentials_create()`, the parameter is telling us what kind of call credentials to create, not what kind of call credentials to use, so I suggest calling this parameter something like `alts_hard_bound`.

markdroth (approved):
Just one minor nit remaining.  Feel free to merge after fixing.

PR #505, opened 2025-07-04
L125: Python Abseil Logging Support

sreenithi:
Changes to add Abseil logging support in gRPC Python hence resolving grpc/grpc#38703.

Implementation PR:
- grpc/grpc#39779

### DIFF
--- a/L125: python-absl-logging-support.md
+++ b/L125: python-absl-logging-support.md
@@ -63,11 +63,6 @@ However, we considered that this might cause too much overhead to users to add a
 Hence, the current approach to allow users to disable using an enironment variable was chosen.
 
 
-Secondly, we also tried adding `absl/log/initialize.cc` directly as a Core dependency instead of a specific Cython level dependency. But as we have multiple languages wrapped on the core layer, this implied that every wrapped language got an additional dependency on `absl/log/initialize.cc`. This is not too much of an overhead given that this is a single additional file and Core already depends on the abseil package. 
-
-However this file is only needed for Python, and hence it seemed to be a better option to use this file as Python specific dependency without affecting Core and other wrapped languages.
-
-
 ## Implementation
 
 The automatic call to `absl::InitializeLog()` will be included in the `_initialize()` Cython function of `src/python/grpcio/grpc/_cython/cygrpc.pyx` file as follows:
--- a/L125: python-absl-logging-support.md
+++ b/L125: python-absl-logging-support.md
@@ -1,0 +1,96 @@
+Title
+----
+* Author(s): sreenithi
+* Approver: sergiitk, asheshvidyut
+* Status: Draft
+* Implemented in: Python
+* Last updated: 04-07-2025
+* Discussion at: <google group thread> (filled after thread exists)
+xxxxxxx
+## Abstract
+
+This proposal describes the cahnges necessary to add absl logging support in gRPC Python hence solving the user warning:
+
+```sh
+WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
+``` 
+
+
+## Background
+
+gRPC Core changed its logging system to use absl logging as per [L117](https://github.com/grpc/proposal/blob/master/L117-core-replace-gpr-logging-with-abseil-logging.md).
+
+After this change, gRPC python users started seeing the warning log - 
+`WARNING: All log messages before absl::InitializeLog() is called are written to STDERR`.
+
+For gRPC C++, the users are expected to invoke `absl::InitializeLog()`, but gRPC Python users cannot invoke a C++ API call. Hence this proposal explains the changes necessary to gRPC Python to implicitly invoke `absl::InitializeLog()` in the Cython layer at the time of initialization.
+
+> Note that if abseil fixes https://github.com/abseil/abseil-cpp/issues/1656, gRPC Core itself will be able to invoke absl::InitializeLog(), but until then this is a workaround.
+
+### Related Proposals: 
+* https://github.com/grpc/proposal/blob/master/L117-core-replace-gpr-logging-with-abseil-logging.md
+
+## Proposal
+
+To resolve the warning and include the call to `absl::InitializeLog()` within gRPC Python, the following changes must be made:
+
+1. Add `abseil-cpp/absl/log/initialize.cc` as a dependency in the Cython layer
+2. Call `absl::InitializeLog()` in `src/python/grpcio/grpc/_cython/cygrpc.pyx` automatically at the time of initialization.
+
+However, abseil-cpp currently [doesn't allow `absl::InitializeLog()` to be called more than once](https://github.com/abseil/abseil-cpp/issues/1656), and will result in an error like:
+
+```
+[globals.cc : 105] RAW: absl::log_internal::SetTimeZone() has already been called
+```
+
+The above changes may hence cause problems to a small percentage of users who may be calling `absl::InitializeLog()` themselves or may be dependent on other libraries which already invokes `absl::InitializeLog()`.
+
+To solve this problem, users will be provided with the option of setting an environment variable `DISABLE_ABSL_INIT_LOG` to opt-out of this automatic abseil log initialization.
+
+
+## Rationale
+
+
+An alternate approach that was considered to provide users with the option of disabling automatic absl log initialization is by using an additional import statement in the code that will disable the call to `absl::InitializeLog()` along with the usual import statement like:
+
+```py
+import grpc.no_absl_log_init
+import grpc
+```
+
+However, we considered that this might cause too much overhead to users to add an additional import statement everywhere in the codebase that calls `import grpc`. And it may also lead to situations where some part of users' codebase may include the additional import statement while others don't, hence causing conflicting behaviour.
+
+Hence, the current approach to allow users to disable using an enironment variable was chosen.
+
+
+Secondly, we also tried adding `absl/log/initialize.cc` directly as a Core dependency instead of a specific Cython level dependency. But as we have multiple languages wrapped on the core layer, this implied that every wrapped language got an additional dependency on `absl/log/initialize.cc`. This is not too much of an overhead given that this is a single additional file and Core already depends on the abseil package. 
+
+However this file is only needed for Python, and hence it seemed to be a better option to use this file as Python specific dependency without affecting Core and other wrapped languages.
+
+
+## Implementation
+
+The automatic call to `absl::InitializeLog()` will be included in the `_initialize()` Cython function of `src/python/grpcio/grpc/_cython/cygrpc.pyx` file as follows:
+```py
+## src/python/grpcio/grpc/_cython/cygrpc.pyx
+
+# absl::InitializeLog() imported in grpc.pxi
+# cdef extern from "absl/log/initialize.h" namespace "absl":
+#  void InitializeLog() nogil
+
+disable_absl_init_log = os.getenv("DISABLE_ABSL_INIT_LOG")
+
+#
+# initialize gRPC
+#
+cdef _initialize():
+  if not disable_absl_init_log:
+    InitializeLog()
+
+  # rest of the function
+
+_initialize()
+```
+
+### Implemented in PR:
+https://github.com/grpc/grpc/pull/39779/


L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+* Author(s): sreenithi
+* Approver: sergiitk, asheshvidyut
+* Status: Draft
+* Implemented in: Python
+* Last updated: 04-07-2025
+* Discussion at: <google group thread> (filled after thread exists)
+xxxxxxx
+## Abstract
+
+This proposal describes the cahnges necessary to add absl logging support in gRPC Python hence solving the user warning:
```
asheshvidyut:
```suggestion
This proposal describes the changes necessary to add absl logging support in gRPC Python hence solving the user warning:
```

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+#
+# initialize gRPC
+#
+cdef _initialize():
+  if not disable_absl_init_log:
+    InitializeLog()
+
+  # rest of the function
+
+_initialize()
```
asheshvidyut: Do we need to define this function? or we can directly do this?
```
import os

disable_absl_init_log = os.getenv("DISABLE_ABSL_INIT_LOG")

if not disable_absl_init_log:
    InitializeLog()
```

sreenithi: The `_initialize()` function was already there and has some more parts to it too. I'm just adding the call to InitializeLog() here.

Also side note, I think it would be better to keep this implementation discussion on the implementation PR instead of this proposal.

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
+Title
```
asheshvidyut: Is the file name matching the convention?

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+The above changes may hence cause problems to a small percentage of users who may be calling `absl::InitializeLog()` themselves or may be dependent on other libraries which already invokes `absl::InitializeLog()`.
+
+To solve this problem, users will be provided with the option of setting an environment variable `DISABLE_ABSL_INIT_LOG` to opt-out of this automatic abseil log initialization.
+
+
+## Rationale
+
+
```
sergiitk: TODO: add a header to make it clear this describes an alternative approach

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+Hence, the current approach to allow users to disable using an enironment variable was chosen.
+
+
+## Implementation
+
+The automatic call to `absl::InitializeLog()` will be included in the `_initialize()` Cython function of `src/python/grpcio/grpc/_cython/cygrpc.pyx` file as follows:
+```py
+## src/python/grpcio/grpc/_cython/cygrpc.pyx
+
+# absl::InitializeLog() imported in grpc.pxi
```
sergiitk: TODO: break up to a separate snippet

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+import grpc
+```
+
+However, we considered that this might cause too much overhead to users to add an additional import statement everywhere in the codebase that calls `import grpc`. And it may also lead to situations where some part of users' codebase may include the additional import statement while others don't, hence causing conflicting behaviour.
+
+Hence, the current approach to allow users to disable using an enironment variable was chosen.
+
+
+## Implementation
+
```
sergiitk: Add a section describing whether multiple `import grpc` may cause the `_initialize` to be called multiple times. Consider different conditions where this may happen (manual module reload, threads, etc). If we conclude this is possible, add a variable to track the state of whether we called initialized log.

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+The automatic call to `absl::InitializeLog()` will be included in the `_initialize()` Cython function of `src/python/grpcio/grpc/_cython/cygrpc.pyx` file as follows:
+```py
+## src/python/grpcio/grpc/_cython/cygrpc.pyx
+
+# absl::InitializeLog() imported in grpc.pxi
+# cdef extern from "absl/log/initialize.h" namespace "absl":
+#  void InitializeLog() nogil
+
+disable_absl_init_log = os.getenv("DISABLE_ABSL_INIT_LOG")
```
sergiitk: Let's use `os.environ.get()` instead. I've shared my reasoning on preferring it over `os.getenv()` [here](https://github.com/astral-sh/ruff/issues/3608#issuecomment-3165769725).

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+
+However, we considered that this might cause too much overhead to users to add an additional import statement everywhere in the codebase that calls `import grpc`. And it may also lead to situations where some part of users' codebase may include the additional import statement while others don't, hence causing conflicting behaviour.
+
+Hence, the current approach to allow users to disable using an enironment variable was chosen.
+
+
+## Implementation
+
+The automatic call to `absl::InitializeLog()` will be included in the `_initialize()` Cython function of `src/python/grpcio/grpc/_cython/cygrpc.pyx` file as follows:
+```py
```
sergiitk: I'm pretty sure we can use `pyx` here

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+
+After this change, gRPC python users started seeing the warning log - 
+`WARNING: All log messages before absl::InitializeLog() is called are written to STDERR`.
+
+For gRPC C++, the users are expected to invoke `absl::InitializeLog()`, but gRPC Python users cannot invoke a C++ API call. Hence this proposal explains the changes necessary to gRPC Python to implicitly invoke `absl::InitializeLog()` in the Cython layer at the time of initialization.
+
+> Note that if abseil fixes https://github.com/abseil/abseil-cpp/issues/1656, gRPC Core itself will be able to invoke absl::InitializeLog(), but until then this is a workaround.
+
+### Related Proposals: 
+* https://github.com/grpc/proposal/blob/master/L117-core-replace-gpr-logging-with-abseil-logging.md
```
sergiitk: nit: we tend to use [reference links](https://github.github.com/gfm/#reference-link) for this. Example: https://github.com/grpc/proposal/blob/343b4d8a21f4724d75ced970cc1fc00c997ef4d5/A83-xds-gcp-authn-filter.md?plain=1#L26-L35

L125: python-absl-logging-support.md
```diff
@@ -0,0 +1,91 @@
...
+* Author(s): sreenithi
+* Approver: sergiitk, asheshvidyut
+* Status: Draft
+* Implemented in: Python
+* Last updated: 04-07-2025
+* Discussion at: <google group thread> (filled after thread exists)
+xxxxxxx
+## Abstract
+
+This proposal describes the changes necessary to add absl logging support in gRPC Python hence solving the user warning:
```
sergiitk: While I don't see a clear guideline for these, but other gRFCs tend to wrap lines at 80 characters. Example: https://github.com/grpc/proposal/blob/343b4d8a21f4724d75ced970cc1fc00c997ef4d5/A83-xds-gcp-authn-filter.md?plain=1#L10-L18

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+`WARNING: All log messages before absl::InitializeLog() is called are written to STDERR`.
+
+For gRPC C++, the users are expected to invoke `absl::InitializeLog()`, but gRPC
+Python users cannot invoke a C++ API call. Hence this proposal explains the
+changes necessary to gRPC Python to implicitly invoke `absl::InitializeLog()`
+in the Cython layer at the time of initialization.
+
+> Note that if abseil fixes https://github.com/abseil/abseil-cpp/issues/1656,
+> gRPC Core itself will be able to invoke absl::InitializeLog(), but until then
+> this is a workaround.
```
sergiitk: Looking at the ticket - I don't think its point is to allow to "Core itself" ..  "to invoke absl::InitializeLog()", but to "Make absl::InitializeLog safe to be called multiple times", making the call a no-op if already been called. We'll still have to call it from Cython.

I think we should remove this block.

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+[doesn't allow `absl::InitializeLog()` to be called more than once](https://github.com/abseil/abseil-cpp/issues/1656),
+and will result in an error like:
+
+```
+[globals.cc : 104] RAW: absl::log_internal::SetTimeZone() has already been called
+```
+
+The above changes may hence cause problems to a small percentage of users who
+may be calling `absl::InitializeLog()` themselves or may be dependent on other
+libraries which already invokes `absl::InitializeLog()`.
```
sergiitk: Users with a custom build that's linking directly against gRPC Cython code, and then call `absl::InitializeLog()` (directly or transitively).

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+
+## Background
+
+gRPC Core changed its logging system to use absl logging as per
+[L117][L117].
+
+After this change, gRPC python users started seeing the warning log -
+`WARNING: All log messages before absl::InitializeLog() is called are written to STDERR`.
```
sergiitk, line 26: How do we reproduce this? I was only able to get it printed with `export GRPC_VERBOSITY=INFO` (which also prints out `Log level INFO is not suitable for production. Prefer WARNING or ERROR. However if you see this message in a debug environment or test environment it is safe to ignore this message.`, or with `export GRPC_TRACE=connectivity_state` (or any other module).

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+```
+[globals.cc : 104] RAW: absl::log_internal::SetTimeZone() has already been called
+```
+
+The above changes may hence cause problems to a small percentage of users who
+may be calling `absl::InitializeLog()` themselves or may be dependent on other
+libraries which already invokes `absl::InitializeLog()`.
+
+To solve this problem, users will be provided with the option of setting an
+environment variable `DISABLE_ABSL_INIT_LOG` to opt-out of this automatic abseil
```
sergiitk: Should this be `GRPC_PYTHON_DISABLE_ABSL_INIT_LOG`?

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+cdef _initialize():
+  if not disable_absl_init_log:
+    InitializeLog()
+
+  # rest of the function
+
+_initialize()
+```
+
+### Behaviour with multiple imports of gRPC
```
sergiitk:
```suggestion
### Behavior with multiple imports of gRPC
```

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+
+Given that `absl::InitializeLog()` must be called exactly once, we had to
+consider if this current implementation can cause any issues with multiple calls
+to `import grpc` in scenarios like:
+
+1. multiple imports for type checking
```
sergiitk: Formatting: start with a capital letter, finish with a period.

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+build time), Python's import system guarantees that it will run only once when
+a process first imports grpc even on manual reload.
+
+All other tested scenarios like threads or child processes — either reuse this
+single, first-time setup or (if they are a new, clean process) safely run their
+own single setup. In all cases, the function is never run twice in a way that
+would cause an error.
+
+(For a more detailed explanation and examples of each scenario tested,
+see [PR #40724](https://github.com/grpc/grpc/pull/40724/).)
```
sergiitk: No need for parens.
```suggestion
For a more detailed explanation and examples of each scenario tested,
see [PR #40724](https://github.com/grpc/grpc/pull/40724).
```

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+`absl::InitializeLog()` will be imported in `_cython/_cygrpc/grpc.pxi`, and the
+automatic call to it will then be included in the `_initialize()` function of
+`_cython/cygrpc.pyx` file as follows:
+
+```pxi
+## src/python/grpcio/grpc/_cython/_cygrpc/grpc.pxi
+
+# absl::InitializeLog() imported in grpc.pxi
+# cdef extern from "absl/log/initialize.h" namespace "absl":
+#  void InitializeLog() nogil
```
sergiitk: We've moved it to a separate file, no need to comment it out anymore

sergiitk (approved):
LGTM, but let's ask Mark to give this a look once the remaining feedback is addressed.

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+
+```sh
+WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
+```
+
+
+## Background
+
+gRPC Core changed its logging system to use absl logging as per
+[L117][L117].
```
sergiitk:
```suggestion
[L117].
```

L125-python-absl-logging-support.md
sergiitk: Discussed IRL, to be rephrased to make it clear the call may be moved to gRPC Core depending on how https://github.com/abseil/abseil-cpp/issues/1656 is resoved.

L125-python-absl-logging-support.md
sreenithi: I was able to reproduce by setting `GRPC_VERBOSITY=debug`

L125-python-absl-logging-support.md
sreenithi: This is yet another Americal vs British English differences I suppose. I was using the British English spelling, however changed as suggested now :)

L125-python-absl-logging-support.md
sreenithi: right, good catch

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
+Title
```
markdroth: This should say "L125: Python Abseil Logging Support".

L125-python-absl-logging-support.md
```diff
@@ -0,0 +1,153 @@
...
+----
+* Author(s): sreenithi
+* Approver: sergiitk, asheshvidyut
+* Status: Draft
+* Implemented in: Python
+* Last updated: 17-09-2025
```
markdroth: Nit: Please use ISO date format (YYYY-MM-DD).

PR #506, opened 2025-07-15
L127: C++: SPIFFE Bundle Map support in Providers

gtcooke94:


L123-cpp-spiffe-root-provider.md
```diff
@@ -0,0 +1,48 @@
+L123: Spiffe Bundle Map support in Root Providers
```
markdroth: L123 is already taken (https://github.com/grpc/proposal/pull/496).  Please change this to L127.

Also, please rename the file accordingly.

markdroth: Please add "C++:" after the gRFC number and before the title.

L123-cpp-spiffe-root-provider.md
```diff
@@ -0,0 +1,48 @@
...
+* Author(s): gtcooke94
+* Approver: markdroth
+* Status: Draft
+* Implemented in: core, cpp
+* Last updated: 2025-07-15
+* Discussion at: TODO
+
+## Abstract
+
+The purpose of this proposal is to add public API support for SPIFFE bundle maps in root certificate file watcher providers. [gRFC A87](https://github.com/grpc/proposal/blob/master/A87-mtls-spiffe-support.md) details the broader internals for this support.
```
markdroth: You can use a relative link instead of an absolute one, since the file you're linking to is in the same directory.

Also, instead of repeating the link multiple times, just define it once on its own line like this:

```
[A87]: A87-mtls-spiffe-support.md
```

Then you can refer to it in text by just saying "[A87]" (or "[gRFC A87][A87]" if you want to change the link text).

L123-cpp-spiffe-root-provider.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Background
+
+gRPC supports SPIFFE bundle maps as root certificate material per [gRFC A87](https://github.com/grpc/proposal/blob/master/A87-mtls-spiffe-support.md). Public APIs to configure these roots is needed.
```
markdroth: s/is needed/are needed/

L123-cpp-spiffe-root-provider.md
```diff
@@ -0,0 +1,48 @@
...
+### Related Proposals:
+* [gRFC A87](https://github.com/grpc/proposal/blob/master/A87-mtls-spiffe-support.md)
+
+## Proposal
+
+This document proposes to extend the C-Core and C++ APIs as follows:
+
+
+### C-Core
+C-Core APIs are always subject to change - we will simply add an argument to the existing constructor in https://github.com/grpc/grpc/blob/79769b35d04535259592ac1b0a98d65f63203f06/include/grpc/credentials.h#L663-L666.
```
markdroth: No need for the code pointer.  But let's be more specific about what we're changing:

"""
In the C-core API, we will add a new `spiffe_bundle_map_path` parameter to the `grpc_tls_certificate_provider_file_watcher_create()` function, which will now look like this:
"""

markdroth: We need to be explicit about what the behavior of the new parameter is.  In particular, what happens if both `root_cert_path` and `spiffe_bundle_map_path` are set?

L123-cpp-spiffe-root-provider.md
```diff
@@ -0,0 +1,48 @@
...
+```
+GRPCAPI grpc_tls_certificate_provider*
+grpc_tls_certificate_provider_file_watcher_create(
+    const char* private_key_path, const char* identity_certificate_path,
+    const char* root_cert_path, const char* spiffe_bundle_map_path, unsigned int refresh_interval_sec);
+```
+
+### C++
+While the existing C++ API is marked experimental, we don't _want_ to break existing users. Thus, in https://github.com/grpc/grpc/blob/79769b35d04535259592ac1b0a98d65f63203f06/include/grpcpp/security/tls_certificate_provider.h#L109-L112, we will add a constructor with the `spiffe_bundle_map_path` argument.
```
markdroth: Code pointer not necessary; we're already describing the API change.

L123-cpp-spiffe-root-provider.md
```diff
@@ -0,0 +1,48 @@
...
+
+FileWatcherCertificateProvider(const std::string& private_key_path,
+                                const std::string& identity_certificate_path,
+                                const std::string& root_cert_path,
+                                const std::string& spiffe_bundle_map_path,
+                                unsigned int refresh_interval_sec);
+```
+
+### Other Providers
+This proposal _only_ aims to support file-based SPIFFE Bundle Maps via the file watcher providers. The `StaticDataCertificateProvider` structure is left as future work - we do not want to expose our internal SPIFFE utilities.
```
markdroth: The last part of this is not clear: "we do not want to expose our internal SPIFFE utilities".  I suggest changing that to say something like "This will involve broadening the API surface to expose a type for the SPIFFE bundle map."

L123-cpp-spiffe-root-provider.md
gtcooke94: Done

gtcooke94: Done

L123-cpp-spiffe-root-provider.md
gtcooke94: Done4

L123-cpp-spiffe-root-provider.md
gtcooke94: Done

L123-cpp-spiffe-root-provider.md
gtcooke94: Done

gtcooke94: Done

L123-cpp-spiffe-root-provider.md
gtcooke94: Removed

L123-cpp-spiffe-root-provider.md
gtcooke94: Replaced

L127-cpp-spiffe-root-provider.md
```diff
@@ -0,0 +1,53 @@
+L127: C++: SPIFFE Bundle Map support in Root Providers
+----
+* Author(s): gtcooke94
+* Approver: markdroth
+* Status: Draft
+* Implemented in: core, cpp
+* Last updated: 2025-07-15
+* Discussion at: TODO
```
markdroth: Please create a mailing list thread and add the link here, as per the gRFC process.  Note that that starts the 10-day clock before we can merge this.

L127-cpp-spiffe-root-provider.md
```diff
@@ -0,0 +1,53 @@
...
+
+FileWatcherCertificateProvider(const std::string& private_key_path,
+                                const std::string& identity_certificate_path,
+                                const std::string& root_cert_path,
+                                const std::string& spiffe_bundle_map_path,
+                                unsigned int refresh_interval_sec);
+```
+
+### Other Providers
+This proposal _only_ aims to support file-based SPIFFE Bundle Maps via the file watcher providers. The `StaticDataCertificateProvider` structure is left as future work. This will involve broadening the API surface to expose a type for the SPIFFE bundle map.
```
markdroth: Please add an "Implementation" section that has a list of implementation PRs.

markdroth (approved):

L127-cpp-spiffe-root-provider.md
gtcooke94: Added a header

L127-cpp-spiffe-root-provider.md
gtcooke94: Done

L127-cpp-spiffe-root-provider.md
```diff
@@ -50,4 +50,7 @@ FileWatcherCertificateProvider(const std::string& private_key_path,
 ```
 
 ### Other Providers
-This proposal _only_ aims to support file-based SPIFFE Bundle Maps via the file watcher providers. The `StaticDataCertificateProvider` structure is left as future work. This will involve broadening the API surface to expose a type for the SPIFFE bundle map.
\ No newline at end of file
+This proposal _only_ aims to support file-based SPIFFE Bundle Maps via the file watcher providers. The `StaticDataCertificateProvider` structure is left as future work. This will involve broadening the API surface to expose a type for the SPIFFE bundle map.
+
+## Implementation
+PR will be linked when created.
```
markdroth: I think you can list all of the PRs you've been working on that are directly related to this effort:
- https://github.com/grpc/grpc/pull/39148
- https://github.com/grpc/grpc/pull/39445
- https://github.com/grpc/grpc/pull/39708

You can add to the list as you finish the remaining work.

markdroth (approved):

PR #507, opened 2025-07-25
A101: xDS-Based setting SNI and server certificate SAN validation

kannanjgithub:


A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+
+During Tls handshake, the server presents its certificate to the client for authentication. For servers
+serving multiple domains, the client needs to indicate which domain it is requesting, so that the server
+can present the certificate is has for that domain. The client does this at the time of Tls handshaking
+via Server Name Indication (SNI). When using `XdsChannelCredentials` for a channel, the gRPC client needs
+to be configured by the xDS server with what value to send for SNI and the gRPC client should use it for
+the Tls handshake.
+
+In [A29][A29] for TLS security in xDS-managed connections, it
+proposed that the `SNI` field from [UpstreamTlsContext.SNI][UTC_SNI]
```
ejona86: Lower-case the two SNI references here, as the field is lower-case.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+When using `XdsChannelCredentials` for the transport, hostname validation
+is turned off and instead SAN matching is performed against [UpstreamTlsContext.match_subject_alt_names][match_subject_alt_names].
+This proposal adds checking of SAN against the SNI provided by the client as well.
+
+For an overview of securing connections in the envoy proxy using SNI 
+and SAN validation, see [envoy-SNI].
+
+[UTC_SNI]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L42
+[A29]: A29-xds-tls-security.md
+[envoy-SNI]: https://www.envoyproxy.io/docs/envoy/latest/_sources/start/quick-start/securing.rst.txt
```
ejona86: I think you want to link to https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/securing instead

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+[match_subject_alt_names]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L407
+
+## Proposal
+This proposal has two parts:
+* Setting SNI
```
ejona86: This does not render well. If you click the three dots (...) at the top-right of this file in github and "View field" then you can see how it renders.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+Tls connections using the fields from [UpstreamTlsContext][UTC]
+in the CDS update.
+
+1. If [UpstreamTlsContext][UTC] specifies the SNI to use, then
+it will be used.
+
+2. If [UpstreamTlsContext][UTC] specifies `auto_sni_host`, then
+SNI will be set to the hostname, which is either the logical
+DNS name for DNS clusters or the endponit hostname for EDS
+clusters, as in the case of the hostname useed for authority
```
ejona86: s/useed/used/

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+rewriting [Ar 81-hostname][A81-hostname].
```
ejona86: This link is pretty incomprehensible in the sentence.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+[match_subject_alt_names]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L407
+
+## Proposal
+This proposal has two parts:
+* Setting SNI
+xDS-managed gRPC clients will set SNI for the Tls handshake for 
+Tls connections using the fields from [UpstreamTlsContext][UTC]
+in the CDS update.
+
+1. If [UpstreamTlsContext][UTC] specifies the SNI to use, then
```
ejona86: "specifies the SNI to use" is very weak and can mean many things. Say specifically the sni field, like "If UpstreamTlsContext specifies `sni`" (to mirror the next item). Although it'd be even more clear "If UpstreamTlsContext.sni is not empty".

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+
+[UTC_SNI]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L42
+[A29]: A29-xds-tls-security.md
+[envoy-SNI]: https://www.envoyproxy.io/docs/envoy/latest/_sources/start/quick-start/securing.rst.txt
+[match_subject_alt_names]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L407
+
+## Proposal
+This proposal has two parts:
+* Setting SNI
+xDS-managed gRPC clients will set SNI for the Tls handshake for
```
ejona86: This is only done when using XdsChannelCredentials.

A98-SNI-setting-and-SNI-SAN-validation.md
ejona86: I think we want to change the tense. Remove "it proposed that" (and replace s/would be/was/). The words don't add much and make it less clear.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+During Tls handshake, the server presents its certificate to the client for authentication. For servers
+serving multiple domains, the client needs to indicate which domain it is requesting, so that the server
+can present the certificate is has for that domain. The client does this at the time of Tls handshaking
+via Server Name Indication (SNI). When using `XdsChannelCredentials` for a channel, the gRPC client needs
+to be configured by the xDS server with what value to send for SNI and the gRPC client should use it for
+the Tls handshake.
+
+In [A29][A29] for TLS security in xDS-managed connections, it
+proposed that the `SNI` field from [UpstreamTlsContext.SNI][UTC_SNI]
+would be ignored. In proposal seeks to start using this and other fields
```
ejona86: s/In proposal/This proposal/ ? Although, really, that sentence could be deleted because this would be part of the proposal.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+for setting the SNI by the gRPC client.
+
+When using `XdsChannelCredentials` for the transport, hostname validation
```
ejona86: hostname matching _is_ a form of SAN matching. How about "When using `XdsChannelCredentials` for the channel, SAN matching is performed against UpstreamTlsContext.match_subject_alt_names instead a typical hostname"? ("typical hostname" is a bit weak, but I think gets the point across. Actually defining how it is different I think gets wordy quickly.)

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+is turned off and instead SAN matching is performed against [UpstreamTlsContext.match_subject_alt_names][match_subject_alt_names].
+This proposal adds checking of SAN against the SNI provided by the client as well.
```
ejona86: "provided by the client" can mean the client chose the SNI name. What we mean here is "This proposal adds SAN matching for the same name as the client used for SNI." (or the like)

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+DNS name for DNS clusters or the endponit hostname for EDS
+clusters, as in the case of the hostname useed for authority
+rewriting [Ar 81-hostname][A81-hostname].
+
+[UTC]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L29
+[A81-hostname]: https://github.com/grpc/proposal/blob/4f833c5774e71e94534f72b94ee1b9763ec58516/A81-xds-authority-rewriting.md?plain=1#L85
+
+* Server SAN validation against SNI used
+
+If `auto_sni_san_validation` is set in the [UpstreamTlsContext][UTC]
```
ejona86: s/set/true/

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+clusters, as in the case of the hostname useed for authority
+rewriting [Ar 81-hostname][A81-hostname].
+
+[UTC]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L29
+[A81-hostname]: https://github.com/grpc/proposal/blob/4f833c5774e71e94534f72b94ee1b9763ec58516/A81-xds-authority-rewriting.md?plain=1#L85
+
+* Server SAN validation against SNI used
+
+If `auto_sni_san_validation` is set in the [UpstreamTlsContext][UTC] 
+gRPC client will perform validation for a DNS SAN matching the SNI value
```
ejona86: We should emphasize this is a DNS SAN. The normal matching allows other SAN types, but only the DNS type will be checked here.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+`UpstreamTlsContext` is either passed down to child policies via
+channel arguments or is put in sub-channel attribute wrapped in a
+`SslContextProviderSupplier`, depending on the language. So far the same
+information has been used for creating the Ssl connection for all
+the subchannels in the cluster so it sufficed to have a single instance of
+the provider of this Tls/Ssl context, instantiated by the LB policy
+and set in the ClusterImpl LB policy helper, which then set this
+provider object as an address attribute of all the subchannels created
+by this helper. For setting the SNI to the hostname, a separate instance
+of the provider needs to be created for each subchannel, augmented with
```
ejona86: I don't think we need a separate instance. Instead, we just add the hostname as an argument when calling it.

The main reason to have the supplier is to have a lifetime for the certificate providers. Since the same certificate providers are used, it should still work out reasonably well. We could create different instances in the Helper, but we'd then need another level to share the certificate providers across those instances.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,138 @@
...
+the hostname to use for the subchannel. So the creation of this provider
+will have to move from the LB policy's accepting addresses to the LB policy
+helper creating subchannel when invoed by the child LB policy. The `UpstreamTlsContext.SNI`
+would already be available to this provider from the parsed Cluster resource. 
+
+##### Caching
+If several or all of the endpoints in a cluster have the same hostname, then
```
ejona86: I'm not sure this is actually a problem, as long as we are sharing certificate provider instances. We might need to confirm that is true cross-language, though.

ejona86 (commented):
Sending what I have

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Agree, I didn't look closely how the SslContextProviderSupplier set into the subchannel attributes by the ClusterImpl LB policy is accessed and used. In SecurityProtocolNegotiators it retreives it from the EAG attributes, and in a similar way it can also retrieve the hostname from the subchannel attributes and float it on to sslContextProviderSupplier.updateSslContext.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Updated.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,121 @@
...
+rewriting][A81-hostname].
+
+[UTC]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L29
+[A81-hostname]: https://github.com/grpc/proposal/blob/4f833c5774e71e94534f72b94ee1b9763ec58516/A81-xds-authority-rewriting.md?plain=1#L85
+
+* Server SAN validation against SNI used
+
+If `auto_sni_san_validation` is true in the [UpstreamTlsContext][UTC] 
+gRPC client will perform validation for a DNS SAN matching the SNI value 
+sent. The normal matching when using `TlsCredentials' for the channel
```
ejona86: This is the XdsChannelCredentials matching

A98-SNI-setting-and-SNI-SAN-validation.md
ejona86: The "two parts" are still unintelligible. Like, the file structure is unclear.

A98-SNI-setting-and-SNI-SAN-validation.md
ejona86: The text is unchanged...

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,121 @@
...
+### Setting SNI
+#### Tls handshake time changes
+As mentioned in [A29 implementation details][A29_impl-details] the
+`UpstreamTlsContext` is either passed down to child policies via
+channel arguments or is put in sub-channel attribute wrapped in a
+`SslContextProviderSupplier`, depending on the language. The `UpstreamTlsContext.SNI`
+would already be available to this provider supplier  from the parsed Cluster resource.
+At the time of Tls protocol negotiation, when this provider supplier is 
+invoked to set the SslContext, the hostname from the channel attributes
+also will be passed, to determine the SNI to be set for the Tls handshake.
```
ejona86: The "also will be passed" is unclear whether that's already the case or being changed. I expect this part is language-specific today, on whether it is available already or not. So we should probably say that this gRFC is adding it, because it wasn't discussed before.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: But looking at [XdsX509ExtendedTrustManager](https://github.com/grpc/grpc-java/blob/7040417eeecb69aa4f012e6ee01da249abbca6d4/xds/src/main/java/io/grpc/xds/internal/security/trust/XdsX509TrustManager.java#L193) it is only matching DNS names against the SANs so I thought we are doing the same but just matching  them against the SNI passed by the client.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Changed to "also will be passed now" . Also removed some details but added some other to make it more  accurate and relevant.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: > The "two parts" are still unintelligible. Like, the file structure is unclear.
 
Used numbering and made the text continue after the respective headings.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+Tls connections using the fields from [UpstreamTlsContext][UTC] in the CDS update.
+
+    i. If `UpstreamTlsContext.sni` specifies the SNI to use, then
+it will be used.
+
+    ii. If [UpstreamTlsContext][UTC] specifies `auto_sni_host`, then
+SNI will be set to the hostname, which is either the logical
+DNS name for DNS clusters or the endpoint hostname for EDS
+clusters, as in the case of the hostname used for [authority
+rewriting][A81-hostname].
```
atollena: Shouldn't it be in the opposite order?

Quoting https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/securing#connect-to-an-endpoint-with-sni:

> When multiple options are configured, auto_sni has the highest priority, followed by auto_host_sni, followed by the fixed sni.

The behavior for gRPC needs to be consistent with Envoy.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Proposal
+This proposal has two parts:
+1. Setting SNI: When using `XdsChannelCredentials` for the channel, gRPC clients will set SNI for the Tls handshake for 
+Tls connections using the fields from [UpstreamTlsContext][UTC] in the CDS update.
+
+    i. If `UpstreamTlsContext.sni` specifies the SNI to use, then
+it will be used.
+
+    ii. If [UpstreamTlsContext][UTC] specifies `auto_sni_host`, then
```
atollena: s/auto_sni_host/auto_host_sni/

https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto#extensions-transport-sockets-tls-v3-upstreamtlscontext

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+* Last updated: 2025-07-28
+
+## Abstract
+
+gRPC will add support for setting Server Name Indication (SNI) and validation of server certificate's
+Subject Alternative Names (SANs) aginst the SNI that was used.
+
+### Background
+
+During Tls handshake, the server presents its certificate to the client for authentication. For servers
```
atollena: FYI, this is actually required by the http2 standard ([paragraph 9.2 of the http2 spec](https://datatracker.ietf.org/doc/html/rfc7540#section-9.2)):

> HTTP/2 clients MUST indicate the target domain name when negotiating TLS.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+need to be enhanced to be <UpstreamTlsContext, String> to hold the SNI as well, and the client
+`SslContext` provider for a particular key will create a `TrustManager` instance that takes the 
+SNI to validate the SANs against and set it in the `SslContext` it provides.
+
+[A29_SAN-matching]: https://github.com/grpc/proposal/blob/master/A29-xds-tls-security.md#server-authorization-aka-subject-alt-name-checks
+[match_subject_alt_names]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L407
+[UTC]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L29
+
+#### Behavior when SNI is not indicated in UpstreamTlsContext
+When `UpstreamTlsContext` has neither of `SNI` and `auto_sni_host` values set, the current behavior will continue, i.e. SNI will be set to the xds hostname from `GrpcRoute`.
```
atollena:
```suggestion
When `UpstreamTlsContext` has neither of `SNI` nor `auto_sni_host` values set, the current behavior will continue, i.e. SNI will be set to the xds hostname from `GrpcRoute`.
```

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
+A98: xDS-Based setting SNI and server certificate SAN validation
```
markdroth: The gRFC numbers are already taken for A98 (https://github.com/grpc/proposal/pull/493), A99 (https://github.com/grpc/proposal/pull/501), and A100 (https://github.com/grpc/proposal/pull/498).  This should be A101.

Please update the text here and rename the file accordingly.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+DNS name for DNS clusters or the endpoint hostname for EDS
+clusters, as in the case of the hostname used for [authority
+rewriting][A81-hostname].
+
+[UTC]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L29
+[A81-hostname]: https://github.com/grpc/proposal/blob/4f833c5774e71e94534f72b94ee1b9763ec58516/A81-xds-authority-rewriting.md?plain=1#L85
+
+2. Server SAN validation against SNI used: If `auto_sni_san_validation` is true in the [UpstreamTlsContext][UTC] 
+gRPC client will perform validation for a DNS SAN matching the SNI value 
+sent. The normal matching when using `TlsCredentials' for the channel
```
markdroth: Need to change the apstrophe to a backquote after `TlsCredentials`.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+it will be used.
+
+    ii. If [UpstreamTlsContext][UTC] specifies `auto_sni_host`, then
+SNI will be set to the hostname, which is either the logical
+DNS name for DNS clusters or the endpoint hostname for EDS
+clusters, as in the case of the hostname used for [authority
+rewriting][A81-hostname].
+
+[UTC]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L29
+[A81-hostname]: https://github.com/grpc/proposal/blob/4f833c5774e71e94534f72b94ee1b9763ec58516/A81-xds-authority-rewriting.md?plain=1#L85
```
markdroth: This link should point to "A81-xds-authority-rewriting.md#xds-resource-validation".

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Setting SNI
+#### Tls handshake time changes
+As mentioned in [A29 implementation details][A29_impl-details] the
+`UpstreamTlsContext` is either passed down to child policies via
+channel arguments or is put in sub-channel attribute wrapped in a
+`SslContextProviderSupplier`, depending on the language. The `UpstreamTlsContext.SNI`
+would already be available to this provider supplier  from the parsed Cluster resource.
+At the time of Tls protocol negotiation, when this provider supplier is 
+used to invoked to set the SslContext, the hostname from the channel attributes
```
markdroth: This doesn't parse grammatically: "is used to invoked to set the SslContext"

A98-SNI-setting-and-SNI-SAN-validation.md
markdroth: I assume `GrpcRoute` is meant to refer to either the k8s gateway API or the GCP AppNet API?  If so, it's confusing to reference that here, since gRPC doesn't directly deal with those APIs.

I think this should say that it will use the authority derived from the target URI used to create the gRPC channel?

What would Envoy's behavior be if neither `sni` nor `auto_sni_host` are set?  Is it going to be the equivalent of our behavior, or are we going to have a different default than Envoy?  If the latter, we need to call that out and say why we're doing that.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Related Proposals:
+* [gRFC A29: xDS-Based Security for gRPC Clients and Servers][A29]
+* [gRFC A81: xDS Authority Rewriting][A81]
+
+[A29]: A29-xds-tls-security.md
+[A81]: A81-xds-authority-rewriting.md
+
+### Setting SNI
+#### Tls handshake time changes
```
markdroth: Do we need a sub-heading here that encompasses everything in the header on the line above?

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+[A29]: A29-xds-tls-security.md
+[envoy-SNI]: https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/securing
+[match_subject_alt_names]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L407
+
+## Proposal
+This proposal has two parts:
+1. Setting SNI: When using `XdsChannelCredentials` for the channel, gRPC clients will set SNI for the Tls handshake for 
+Tls connections using the fields from [UpstreamTlsContext][UTC] in the CDS update.
+
+    i. If `UpstreamTlsContext.sni` specifies the SNI to use, then
```
markdroth: Are there any validation rules for the content of these fields?

It looks like Envoy has PGV rules that would reject the resource if the `sni` field is longer than 255 characters.  Do we want that same restriction?  If so, we need to explicitly state it.

A98-SNI-setting-and-SNI-SAN-validation.md
markdroth: What's the behavior if `auto_sni_host` is true but the EDS endpoint doesn't have a hostname set?

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+and SAN validation, see [envoy-SNI].
+
+[UTC_SNI]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L42
+[A29]: A29-xds-tls-security.md
+[envoy-SNI]: https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/securing
+[match_subject_alt_names]: https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/common.proto#L407
+
+## Proposal
+This proposal has two parts:
+1. Setting SNI: When using `XdsChannelCredentials` for the channel, gRPC clients will set SNI for the Tls handshake for
```
markdroth: What happens if both `sni` and `auto_host_sni` are set?  Which one takes prededence?

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+## Proposal
+This proposal has two parts:
+1. Setting SNI: When using `XdsChannelCredentials` for the channel, gRPC clients will set SNI for the Tls handshake for 
+Tls connections using the fields from [UpstreamTlsContext][UTC] in the CDS update.
+
+    i. If `UpstreamTlsContext.sni` specifies the SNI to use, then
+it will be used.
+
+    ii. If [UpstreamTlsContext][UTC] specifies `auto_sni_host`, then
+SNI will be set to the hostname, which is either the logical
```
markdroth: s/logical DNS name for DNS clusters/DNS name for logical DNS clusters/

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+[A29]: A29-xds-tls-security.md
+[A81]: A81-xds-authority-rewriting.md
+
+### Setting SNI
+#### Tls handshake time changes
+As mentioned in [A29 implementation details][A29_impl-details] the
+`UpstreamTlsContext` is either passed down to child policies via
+channel arguments or is put in sub-channel attribute wrapped in a
+`SslContextProviderSupplier`, depending on the language. The `UpstreamTlsContext.SNI`
+would already be available to this provider supplier  from the parsed Cluster resource.
```
markdroth: Most of the description of the plumbing here is very Java-specific.  As an example, there is nothing called a "provider supplier" in C-core.

Can we back up a bit and describe how this is going to work more generically, so that it's clear how this will work in all languages?  If we have to put sub-sections to provide language-specific details, that's fine, but the overall structure should be described in a language-agnostic way.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+[`match_subject_alt_names`][match_subject_alt_names] in `CertificateValidationContext`.
+If `auto_sni_san_validation` is set in the [UpstreamTlsContext][UTC], matching will be 
+performed against the SNI that was used by the client, and this validation will replace
+the [`match_subject_alt_names`][match_subject_alt_names] if set. This verification occurs
+in the TrustManager of the SslContext which is created using the cert store indicated by 
+`CertificateValidationContext` in `UpstreamTlsContext` which is either a managed cert store
+or the system root cert store. 
+
+#### Caching for the SslContext 
+The `SslContextProviderSupplier` (named so because it supplies both client and server
```
markdroth: This is another area where the description is very Java-specific.  Please describe this in a language-agnostic way.

I don't know if C-core even has anything equivalent to the cache you're talking about here.  But it may be that I am just not sufficiently familiar with the Java code to map this to the equivalent thing in C-core.

A98-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,117 @@
...
+SslContext providers) creates a provider for the client `SslContext` and today 
+maintains a cache of `UpstreamTlsContext` to the client `SslContext` provider instances. 
+For the SNI requirement, the `TrustManager` in the `SslContext` needs to 
+be aware of the SNI to validate the SAN against, so a different `TrustManager` instance needs 
+to be created for each SNI to use for the same `UpstreamTlsContext`, so this cache's key will 
+need to be enhanced to be <UpstreamTlsContext, String> to hold the SNI as well, and the client
+`SslContext` provider for a particular key will create a `TrustManager` instance that takes the 
+SNI to validate the SANs against and set it in the `SslContext` it provides.
+
+[A29_SAN-matching]: https://github.com/grpc/proposal/blob/master/A29-xds-tls-security.md#server-authorization-aka-subject-alt-name-checks
```
markdroth: Don't need the "https://github.com/grpc/proposal/blob/master/" part.  This should be a relative link.

markdroth (commented):
Thanks for doing this, Kannan!

atollena (commented):
Thanks for this work, SNI has been is a big missing piece for our own xds deployment.

A98-SNI-setting-and-SNI-SAN-validation.md
ejona86: It's fair to point out, but actually not relevant. It's really be "required" well before HTTP/2. But SNI and the HTTP/2 spec only considers WWW.

When you have SPIFFE certs, or just doing xds, the "target domain name" may not be a valid domain name. SNI prohibits using invalid domain names, so you may have to drop SNI. That's what happened in Java; C++ instead was choosing to send invalid SNI.

So while we could mention HTTP/2 requires it, I don't think it actually adds any weight.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Envoy [docs](https://www.envoyproxy.io/docs/envoy/latest/start/quick-start/securing) say 
```
When multiple options are configured, auto_sni has the highest priority, followed by auto_host_sni, followed by the fixed sni.
```

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: In that case envoy tries to use the UpstreamTlsContext.sni as per the code [here](https://github.com/envoyproxy/envoy/blob/64cb65c7d1a1b8d1d779ac2424ac530addac5762/source/common/tls/client_context_impl.cc#L114). If the final value it decides to use is still empty it doesn't set SNI at all as the lines immediately following show.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Yes, changed it now, thanks.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Updated this section as per comment.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Fixed.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Just made it "### Setting SNI during Tls handshake".

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Added this towards the end.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: However note that we are not implementing auto_sni in this proposal.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Envoy [doesn't set SNI](https://github.com/envoyproxy/envoy/blob/64cb65c7d1a1b8d1d779ac2424ac530addac5762/source/common/tls/client_context_impl.cc#L120) if it is empty after assigning it using the precedence rules. I have added this as a 3rd rule in the "Proposal" section now instead of having it down here.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: To rephrase this section I need to know more details of how other languages are handling the dynamic creation of the Client SslContext. In Java it is the `CertProviderClientSslContextProvider` that dynamically provides the SslContext. This provider instance is what is cached in a map against `UpstreamTlsContext` and the proposal in this section is to make the key include  the SNI and the provider set the SNI in the TrustStore of the SslContext it provides.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -38,23 +38,21 @@ and SAN validation, see [envoy-SNI].
...
-DNS name for DNS clusters or the endpoint hostname for EDS
-clusters, as in the case of the hostname used for [authority
-rewriting][A81-hostname].
+   ii. If `UpstreamTlsContext.sni` specifies the SNI to use, then it will be used.
+
+   iii. Otherwise no SNI will be set for the Tls handshake.
 
 [UTC]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L29
-[A81-hostname]: https://github.com/grpc/proposal/blob/4f833c5774e71e94534f72b94ee1b9763ec58516/A81-xds-authority-rewriting.md?plain=1#L85
+[A81-hostname]: https://github.com/grpc/proposal/blob/4f833c5774e71e94534f72b94ee1b9763ec58516/A81-xds-authority-rewriting.md#xds-resource-validation
```
markdroth: Please remove the "https://github.com/grpc/proposal/blob/4f833c5774e71e94534f72b94ee1b9763ec58516/" part.  This should be a relative link.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -38,23 +38,21 @@ and SAN validation, see [envoy-SNI].
...
+logical DNS clusters or the endpoint hostname for EDS clusters, as in the case of the hostname used for [authority rewriting][A81-hostname].
 
-    ii. If [UpstreamTlsContext][UTC] specifies `auto_sni_host`, then
-SNI will be set to the hostname, which is either the logical
-DNS name for DNS clusters or the endpoint hostname for EDS
-clusters, as in the case of the hostname used for [authority
-rewriting][A81-hostname].
+   ii. If `UpstreamTlsContext.sni` specifies the SNI to use, then it will be used.
+
+   iii. Otherwise no SNI will be set for the Tls handshake.
```
markdroth: Isn't this going to be a breaking change for gRPC?  Currently, we unconditionally set SNI based on the channel's default authority, which is generally derived from the target URI, which means that that's what happens if SNI is not configured.  But with this proposal, we will change the behavior in that case to stop setting SNI.  This will affect any control plane that doesn't set SNI.

I don't know enough about the TLS handshake behavior to know how big of a problem this is likely to be in practice.  @ejona86, thoughts?

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -64,21 +62,25 @@ allows other SAN types, but only the DNS type will be checked here.
...
+
+##### Language specific example
+As an example, in Java, the ClusterImpl LB policy creates the `SslContextProviderSuppler` wrapping the
+`UpstreamTlsContext` and puts it in the subchannel wrapper when its child policy creates a subchannel. At the time of Tls protocol negotiation
+for the subchannel, the hostname from the channel attributes also should be passed to this provider supplier to determine the SNI to be set for 
+the Tls handshake. The hostname will be set in the callback object that is given to the `SslContextProviderSupplier`, to be invoked with the 
+`SslContext` when the client Ssl Provider instantiated by this supplier has the `SslContext` available. This value along with the 
+`UpstreamTlsContext` available in the `SslContextProviderSupplier` will be used to decide the SNI to be used for the handshake.
 
 [A29_impl-details]: https://github.com/grpc/proposal/blob/master/A29-xds-tls-security.md#implementation-details
```
markdroth: Please remove the "https://github.com/grpc/proposal/blob/master/" part.  This should be a relative link.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -64,21 +62,25 @@ allows other SAN types, but only the DNS type will be checked here.
 [A29]: A29-xds-tls-security.md
```
markdroth, line 62: Please add an "Updated by:" header in A29 linking to this gRFC.

A98-SNI-setting-and-SNI-SAN-validation.md
markdroth: Still need to address this part of my original comment:

> I assume `GrpcRoute` is meant to refer to either the k8s gateway API or the GCP AppNet API? If so, it's confusing to reference that here, since gRPC doesn't directly deal with those APIs.
> 
> I think this should say that it will use the authority derived from the target URI used to create the gRPC channel?

A98-SNI-setting-and-SNI-SAN-validation.md
markdroth: The text of this gRFC needs to make the exact precedence behavior clear.  The current text does not make it clear that these rules are intended to be "first match wins".  It also doesn't make it clear that the first rule is considered non-matching if `auto_sni_host` is set but the endpoint does not have a hostname associated with it.

My rubric for a good gRFC is that we should be able to hand it to an L4 on our team who is familiar with the gRPC code base but was not involved in writing this gRFC, and they should be able to implement the design correctly without having to figure out details like these.  The Envoy code base is a good reference when writing a gRFC, but it's not by itself a spec, and implementors of a gRFC should not need to look at it to figure out how things are supposed to work.

A98-SNI-setting-and-SNI-SAN-validation.md
markdroth: The updated description is still not generic enough.  C-core doesn't even have an `SslContext` object the way you're describing it here.

Here's a quick summary of the current structure in C-core:
- We use an [`XdsCertificateProvider`](https://github.com/grpc/grpc/blob/73c0f8ac9c7e448f439a084455f37f9d9ed3a6bb/src/core/xds/grpc/xds_certificate_provider.h#L43), which is essentially a wrapper around the cert providers for the root and identity certs.  Our TlsCreds code supports a single cert provider instance for both root and identity certs, but xDS can configure a different cert provider instance for each one, so we need a layer of indirection to combine them.  The `XdsCertificateProvider` also contains the SAN matching rules to be used in the handshake.
- The xds_cluster_impl LB policy [creates an `XdsCertificateProvider` based on the CDS resource](https://github.com/grpc/grpc/blob/73c0f8ac9c7e448f439a084455f37f9d9ed3a6bb/src/core/load_balancing/xds/xds_cluster_impl.cc#L595), and then [adds it to channel args](https://github.com/grpc/grpc/blob/73c0f8ac9c7e448f439a084455f37f9d9ed3a6bb/src/core/load_balancing/xds/xds_cluster_impl.cc#L603) to be passed down to child policies.
- When a subchannel is created, we [use the channel credentials to create a security connector](https://github.com/grpc/grpc/blob/73c0f8ac9c7e448f439a084455f37f9d9ed3a6bb/src/core/transport/endpoint_transport_client_channel_factory.cc#L51), which is what actually does the handshake.  In `XdsCredentials`, if the `XdsCertificateProvider` was passed down in channel args, then we [create a TLS security connector using the `XdsCertificateProvider`](https://github.com/grpc/grpc/blob/73c0f8ac9c7e448f439a084455f37f9d9ed3a6bb/src/core/credentials/transport/xds/xds_credentials.cc#L139).
- If the EDS hostname is set, we [pass it down in a per-endpoint channel arg](https://github.com/grpc/grpc/blob/73c0f8ac9c7e448f439a084455f37f9d9ed3a6bb/src/core/xds/grpc/xds_endpoint_parser.cc#L197).

To support SNI, I think we'll need the following changes:
- Change the `XdsCertificateProvider` to include the SNI settings from CDS.
- Change `XdsCredentials` to use the SNI settings from the `XdsCertificateProvider`, along with the channel arg containing the EDS hostname, to determine the SNI to use.  It will then pass that name when creating the TLS security connector.

A98-SNI-setting-and-SNI-SAN-validation.md
markdroth: I don't think C-core has anything like this cache at all.  See my description above for details.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,122 @@
...
+[A81]: A81-xds-authority-rewriting.md
+
+### Setting SNI during Tls handshake
+As mentioned in [A29 implementation details][A29_impl-details] the `UpstreamTlsContext` is either 
+passed down to child policies via channel arguments or a similar mechanism, depending on the language, 
+and the SslContext is instantiated using the truststore location indicated by the `UpstreamTlsContext`. 
+This SslContext is then used to initiate the Tls handshake for the transport and this is when the SNI is sent
+for the `ClientHello` frame of the handshake. To determine the SNI, we need both the `UpstreamTlsContext` and 
+the hostname for the endpoint. The hostname attribute is already stored in the subchannel wrapper by the 
+xds_cluster_impl policy when its child policy creates a subchannel. Once the `SslContext` is available during the
```
markdroth: Is the fact that this is already used in the xds_cluster_impl policy actually relevant?  It's true that we have that in C-core, but in our implementation, that is actually a different code path than the attribute that gets passed into the internal subchannel for use in the TLS handshake.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Yes, it will be a change of existing behavior. As the authority derived from the Xds channel target URI is anyway meaningless for choosing a certificate by the server, setting the SNI will be breaking only those cases where the server required some some value to be set even if invalid.

So while this gRFC might break some clients, those would be cases vulnerable to breakage anyway if the server started refusing the connection.
Java and Go are not setting SNI (Java sets it but the Ssl library seems to drop it if it is not in a domain name format (haven't tested with Xds hostname _in_ domain name format).

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: I thought it would make the reading easier without having to remember or cross-reference where the attribute value is coming from. My intention was to specify either which xds resource the hostname attribute originates from, which is the endpoint resource or the LB policy that sets it into the subchannel attributes, and I ended up choosing the latter. In c-core even if the TLS handshake gets the hostname via a different code path than the subchannel attribute, it would still have been copied from it at some point, right?

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

kannanjgithub (commented):
.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Yes, it would be better to say authority derived from the channel's target URI rather than the hostname from GrpcRoute definition in the config APIs, however after checking what envoy does when neither `UpstreamTlsContext.sni` nor `auto_host_sni` is set, we are now not setting any SNI in this case.
I think that is same as the breaking change we are discussing in point iii under the _Proposal_ section, for not setting any SNI in this case similar to envoy.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: I have updated this section again to remove references to `SslContext` and describe in a language agnostic way.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -64,15 +64,20 @@ allows other SAN types, but only the DNS type will be checked here.
...
-Tls handshake phase of the transport creation (the creation of which depends on the choice of the certificate provider 
-infra to use as indicated by the `UpstreamTlsContext`), the fields from `UpstreamTlsContext` and the hostname 
-from the channel attributes will be used to determine the SNI to set for the handshake.
+passed down to child policies via channel arguments or a similar mechanism, depending on the language.
+[A29 implementation details][A29_impl-details] also talks about a `CertificateProvider` object that represents 
+a plugin that provides the required certificates and keys to the gRPC application. When Tls handshake is
+initiated for a channel that is using `XdsCredentials`, this `CertificateProvider` object is used to
+provide the certs and trust roots for establishing the secure connection. During this handshake we need 
+to set the SNI to use for the `ClientHello` frame of the handshake. To determine the SNI, we need both the 
+`UpstreamTlsContext` and the hostname for the endpoint.
```
markdroth: We don't need the entire `UpstreamTlsContext`; what we need is just the `auto_host_sni` and `sni` fields.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -92,11 +97,11 @@ matches the Subject Alternative Names specified in the server certificate agains
 If `auto_sni_san_validation` is set in the [UpstreamTlsContext][UTC], matching will be 
 performed against the SNI that was used by the client, and this validation will replace
 the [`match_subject_alt_names`][match_subject_alt_names] if set. This verification occurs
-in the TrustManager of the SslContext which is created using the cert store indicated by 
+in the TrustManager created by the `CertificateProvider` using the cert store indicated by
```
markdroth: I think this last sentence belongs in the Java-specific section.  C-core does not have anything called `TrustManager`, nor do we have a cert store.

In C-core, `XdsCredentials` [constructs a cert verifier](https://github.com/grpc/grpc/blob/73c0f8ac9c7e448f439a084455f37f9d9ed3a6bb/src/core/credentials/transport/xds/xds_credentials.cc#L159) that does the cert verification based on the settings in the `XdsCertificateProvider`.  That will need to be changed such that the `XdsCertificateProvider` also encodes the value of the `auto_sni_san_validation` field and the SNI used by the client.

A98-SNI-setting-and-SNI-SAN-validation.md
markdroth: Regardless of what we decide the actual behavior should be, this gRFC should not mention `GrpcRoute` at all, because that object does not exist in the xDS API.  There will be cases in OSS where people don't use the k8s gateway API or the GCP AppNet API to generate the xDS resources and therefore don't know what `GrpcRoute` is.  This gRFC is defining the client behavior, and the client sees only the xDS API, so we need to define the behavior in terms of that API.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,127 @@
...
+* [gRFC A81: xDS Authority Rewriting][A81]
+
+[A29]: A29-xds-tls-security.md
+[A81]: A81-xds-authority-rewriting.md
+
+### Setting SNI during Tls handshake
+As mentioned in [A29 implementation details][A29_impl-details] the `UpstreamTlsContext` is either 
+passed down to child policies via channel arguments or a similar mechanism, depending on the language.
+[A29 implementation details][A29_impl-details] also talks about a `CertificateProvider` object that represents 
+a plugin that provides the required certificates and keys to the gRPC application. When Tls handshake is
```
markdroth: s/application/implementation/

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -0,0 +1,127 @@
...
+
+[A29_impl-details]: A29-xds-tls-security.md#implementation-details
+[UTC_SNI]: https://github.com/envoyproxy/envoy/blob/ee2bab9e40e7d7649cc88c5e1098c74e0c79501d/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L42
+
+### SAN SNI validation
+The server certificate validation described in [A29 SAN matching][A29_SAN-matching]
+matches the Subject Alternative Names specified in the server certificate against 
+[`match_subject_alt_names`][match_subject_alt_names] in `CertificateValidationContext`.
+If `auto_sni_san_validation` is set in the [UpstreamTlsContext][UTC], matching will be 
+performed against the SNI that was used by the client, and this validation will replace
```
markdroth, line 101: Which fields in the server's cert will be matched against?  I assume this will be done as DNS SAN matching?

Also, I assume that this uses the SNI used by the client, regardless of how that SNI was determined?  So, for example, if `auto_host_sni` was set but there is no EDS hostname for the endpoint, but `sni` is set, then it would use the value of the `sni` field?

Please explicitly state what the matching behavior is here.

A101-SNI-setting-and-SNI-SAN-validation.md
markdroth: If the only reason why Java wasn't sending SNI is that the xDS hostname is not in DNS format, then anyone who is using a DNS-format name will see a behavior change even in Java.

I'd still like to hear thoughts from @ejona86 here.  But I think that at minimum, we should consider adding some sort of knob at least temporarily to get back the old behavior, so that users have a way to migrate.

A101-SNI-setting-and-SNI-SAN-validation.md
markdroth: Quoting from [the "xDS Resource Validation" section in A81](https://github.com/grpc/proposal/blob/master/A81-xds-authority-rewriting.md#xds-resource-validation):

> When validating an EDS resource, we will look at the [Endpoint.hostname field](https://github.com/envoyproxy/envoy/blob/b65de1f56850326e1c6b74aa72cb1c9777441065/api/envoy/config/endpoint/v3/endpoint_components.proto#L89). The string value of this field will be included in a per-endpoint resolver hostname attribute in the parsed resource struct that is passed to the XdsClient watcher. Note that the resolver attribute used here should be a general-purpose one, not something specific to EDS; for example, in the future, we may want to use this same attribute to expose per-endpoint hostnames in channelz (see [gRFC A14](https://github.com/grpc/proposal/blob/master/A14-channelz.md)).

The structure here should be:
1. The EDS validation logic sets the hostname attribute.
2. The code in xds_cluster_impl then uses the attribute from (1) to implement authority rewriting.
3. We are now saying that the `XdsCredentials` code will also use the same attribute from (1) to set SNI.

In other words, (2) and (3) should depend on (1), but (3) should not depend on (2).  As a result, there should be no need to mention the xds_cluster_impl policy in this gRFC.

kannanjgithub: But (1) is into the address attributes that is passed down the child LB policies (for example [here](https://github.com/grpc/grpc-java/blob/c643e68aa1faf7f1aabf8e789030c44323fc4f77/xds/src/main/java/io/grpc/xds/ClusterResolverLoadBalancer.java#L415) in Java).
(2) is [here](https://github.com/grpc/grpc-java/blob/c643e68aa1faf7f1aabf8e789030c44323fc4f77/xds/src/main/java/io/grpc/xds/ClusterImplLoadBalancer.java#L246) where the address attribute is set as the subchannel attribute by the ClusterImpl LB Helper, that is then used by the picker for authority rewriting. 

It is the latter that actually makes it available in the transport handling code, right?

markdroth: It sounds like Java's implementation of A81 doesn't match what I intended when I wrote it.  Maybe the wording didn't make the intent sufficiently clear?

My intention was that (1) should be sufficient to make the attribute available to the transport.  I think it basically has to be that way if we're going to use it in channelz even in non-xDS cases, where there is no xds_cluster_impl LB policy.

(2) should be needed only to use that attribute in the xds_cluster_impl policy's picker to do authority rewriting.  It should not affect whether the attribute is available in the transport.

kannanjgithub: (2) is set into the subchannel wrapper's args for use by the picker, that is inline with what A81 says. I had mistaken that the wrapper's args end up in the real subchannel as well eventually and that's how they become available for the transport. I must have missed the code flow from (1) which is what makes them available to the transport, or there is some divergence from A81 that I need to check for.  From the existing Java code I thought (1) only propagates the hostname from EdsUpdate to the child LB policy via the address attributes.

kannanjgithub: No, there is no divergence. (1) is what makes (2) possible also by sending the attribute down the LB tree, then then it comes back up via createSubChannel from the child policy, I see it is already available to the transport in this way and (2) is only derived from this.

markdroth: Super.  Then let's remove the reference to xds_cluster_impl in this gRFC.

kannanjgithub: All references? xds_cluster_impl is what makes available the `CertificateProvider` for the transport via subchannel args, so we can retain that part?

markdroth: The fact that the xds_cluster_impl policy creates the XdsCertificateProvider isn't a new thing we're introducing in this gRFC, so we don't need to say that that.  But we should say that we're changing the xds_cluster_impl policy to pass down the new information from CDS so that it's available in the transport.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Yes, the SNI for matching will be the same SNI that was sent. I have mentioned this here now. About the precedence rules, it seems the more appropriate place is the section describing sending SNI, so I have added it towards the end of the `Setting SNI during Tls handshake` section above.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Added a separate environment variable for reverting back to sending the xds authority instead of setting no SNI.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: My bad, I thought I removed this section when I added "iii. Else, no SNI will be set for the Tls handshake." in the Proposal section but seems like I either forgot to do so or cross edited from multiple windows and reintroduced this text. Removed it now.

A98-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: I misunderstood that doing matching for only 3 types was more restrictive than what `TlsCredentials` does. Modified it now.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -66,23 +66,27 @@ allows other SAN types, but only the DNS type will be checked here.
...
-information from `UpstreamTlsContext.sni` and `UpstreamTlsContext.auto_host_sni` will also be set into the
-`CertificateProvider`. When the Tls handling code uses the certs and trust roots from the `CertificateProvider`
+SNI related fields from the parsed `UpstreamTlsContext` and the hostname for the endpoint. 
+To determine the SNI `UpstreamTlsContext.sni` and `UpstreamTlsContext.auto_host_sni` from the parsed
+cluster resource will also be set into the `CertificateProvider` by the xds_cluster_impl policy. 
+When the Tls handling code uses the certs and trust roots from the `CertificateProvider`
 to establish the connection, it will also now determine the SNI to set based on the parsed sni related fields
-available in the `CertificateProvider` and also the hostname available in the subchannel attributes.
+available in the `CertificateProvider` and the hostname in the endpoint attributes.
+The precedence order mentioned in the `Proposal1` section will be used to determine the SNI to use. For example,
```
markdroth: Suggest changing ```in the `Proposal1` section``` to `at the top of the [Proposal](#proposal)` section.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -66,23 +66,27 @@ allows other SAN types, but only the DNS type will be checked here.
...
+if `UpstreamTlsContext.auto_host_sni` was set but there is no EDS hostname for the endpoint, but 
+`UpstreamTlsContext.sni` is set, then it would use the value of the `UpstreamTlsContext.sni` if set. 
+If no SNI value is determined, then it will no set SNI for the Tls handshake.
```
markdroth: s/no set/not set/

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -66,23 +66,27 @@ allows other SAN types, but only the DNS type will be checked here.
...
+
+[A81_xds_resource_validation]: A81-xds-authority-rewriting.md#xds-resource-validation
```
markdroth: This doesn't appear to be used anywhere.  And it duplicates the "A81-hostname" link above.

A101-SNI-setting-and-SNI-SAN-validation.md
```diff
@@ -66,23 +66,27 @@ allows other SAN types, but only the DNS type will be checked here.
...
 
 ##### Language specific example
 As an example, in Java, the ClusterImpl LB policy creates the `SslContextProviderSuppler` wrapping the
 `UpstreamTlsContext` and puts it in the subchannel wrapper when its child policy creates a subchannel. At the time of Tls protocol negotiation
-for the subchannel, the hostname from the channel attributes also should be passed to this provider supplier to determine the SNI to be set for 
+for the subchannel, the hostname from the endpoint address attributes also should be passed to this provider supplier to determine the SNI to be set for
```
markdroth: s/endpoint address/endpoint/

markdroth (commented):
Thanks for all those changes, Kannan!  This is looking really good.

I'd still like Eric's input on the question about the breaking behavior change, to make sure we're in agreement about how to handle that.  But I think that's the only really substantive issue remaining open.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Done.

A101-SNI-setting-and-SNI-SAN-validation.md
kannanjgithub: Removed.

atollena (approved):

A101-SNI-setting-and-SNI-SAN-validation.md
ejona86: Yeah, if at all possible we should have a flag to allow sending SNI even when not configured for xDS, because it was the behavior that was used for a while, and there was even evidence some users were trying to use it (https://github.com/grpc/grpc-java/issues/11750).

kannanjgithub: Added mention of separate flag GRPC_USE_CHANNEL_AUTHORITY_IF_NO_SNI_APPLICABLE for sending xds channel authority if no SNI was determined.

markdroth (approved):
Looks great!

PR #511, opened 2025-08-28
A103: xDS Composite Filter

markdroth:


A103-xds-composite-filter.md
```diff
@@ -0,0 +1,171 @@
...
+        Within this field:
+        - [default_value](https://github.com/envoyproxy/envoy/blob/cdd19052348f7f6d85910605d957ba4fe0538aec/api/envoy/config/core/v3/base.proto#L648):
+          This field must be present. The configured value will be capped at
+          100%.
+        - runtime_key: This field will be ignored, since gRPC does not
+          have a runtime system.
+- matcher: gRPC will not support this deprecated field.
+
+We will also support per-route overrides via the
+[`envoy.extensions.common.matching.v3.ExtensionWithMatcherPerRoute`
```
ejona86, line 99: So this means we can have filters defined wholly in the RouteConfiguration? The Listener would no longer enumerate all the filters being used.

A103-xds-composite-filter.md
```diff
@@ -0,0 +1,171 @@
...
+will use only those entries in the parsed metadata map that correspond
+to `google.protobuf.Struct` type.
+
+We will also add support for the following attributes on the gRPC server
+side only (these attributes are not relevant on the client side):
+- `source.ip`
+- `source.port`
+- `connection.requested_server_name`
+- `connection.tls_version`
+- `connection.sha256_peer_certificate_digest`
```
ejona86: How are these represented in a `Struct`? What is the format for these?

A103-xds-composite-filter.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Filter Behavior
+
+Because all of the CEL attributes that we are exposing are available
```
ejona86: I need to think about this more in Java. This is harder for us than just processing it in the filter itself. I'll need to think about how we'd want to implement the approach mentioned here.

A103-xds-composite-filter.md
markdroth: Yes, that's correct.

In the future, we'll also add ECDS support, which would be another way for the Listener to not enumerate all of the filters being used, even if there is no per-route override.

A103-xds-composite-filter.md
markdroth: These aren't encoded in a `Struct`, they're just string or integer values, as per the Envoy docs:

https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes.html#upstream-attributes

The text above that mentions `google.protobuf.Struct` is talking about the values of the `xds.route_metadata.filter_metadata` attribute, not about these attributes.

(This question made me notice an internal problem that might cause changes for `xds.route_metadata.filter_metadata`, but I'll follow up on that separately.)

ejona86 (approved):

A103-xds-composite-filter.md
```diff
@@ -0,0 +1,171 @@
...
+* [A39: xDS HTTP Filter Support][A39]
+* [A36: xDS-Enabled Servers][A36]
+* [A77: xDS Server-Side Rate Limiting][A77] (pending)
+* [A83: xDS GCP Authentication Filter][A83]
+
+[A27]: A27-xds-global-load-balancing.md
+[A36]: A36-xds-for-servers.md
+[A39]: A39-xds-http-filters.md
+[A77]: https://github.com/grpc/proposal/pull/414
+[A83]: A83-xds-gcp-authn-filter.md
```
AgraVator: The links seem to be broken...

A103-xds-composite-filter.md
AgraVator: A39 says that 

1. `There must be at least one filter in the list, or the Listener resource will be NACKed.` Is this no longer true ?
2. `Note that gRPC will not fail validation if the map key specifies a filter instance name that does not exist in the HttpConnectionManager filter list. This is because during an update, the xDS client code cannot know which HttpConnectionManager config is currently being used.` Do we ignore such filters currently ? and will we continue doing so ?

markdroth: This gRFC does not modify A39 in any way.  Both of those statements from A39 are still true.

A103-xds-composite-filter.md
```diff
@@ -98,22 +97,51 @@ Within it, gRPC will look at the following fields:
 
 We will also support per-route overrides via the
 [`envoy.extensions.common.matching.v3.ExtensionWithMatcherPerRoute`
-proto](https://github.com/envoyproxy/envoy/blob/0685d7bf568485eb112df2a9c73248cb8bfc1c37/api/envoy/extensions/common/matching/v3/extension_matcher.proto#L39C9-L39C37).
-In this proto, the
+proto](https://github.com/envoyproxy/envoy/blob/0685d7bf568485eb112df2a9c73248cb8bfc1c37/api/envoy/extensions/common/matching/v3/extension_matcher.proto#L39C9-L39C37)
+proto.  In this proto, the
```
ejona86: "ExtensionWithMatcherPerRoute proto proto" is maybe emphasizing it is a proto a bit too much :). Looking at the other changes, I guess you want to remove "proto" from the link itself? Although it also seems more precise to say "message" here like you did elsewhere.

A103-xds-composite-filter.md
```diff
@@ -98,22 +97,51 @@ Within it, gRPC will look at the following fields:
...
+in LDS, we know whether the resource is an API listener (client side)
+or socket listener (server side), so we reject filters that are not
+supported on the side they are being configured on.  However, because
+the composite filter's per-route override config may be delivered via
+RDS instead of LDS, it will in this case not be possible for gRPC to
+detect that a filter is being configured on an unsupported side when
+validating the RDS resource.  Instead, the composite filter will need
+to handle this on a per-RPC basis: if a nested filter chain includes a
+filter that is not supported on the side that it is running on, it will
+fail the RPC with status UNAVAILABLE.  Note that if the problematic
```
ejona86, line 140: The previous sentence said the "nested filter chain" will fail the RPC if a bad filter is present. This sentence is spelling that out further, but it says "may", even though that seemed required by the previous sentence. Is it okay for an implementation to fail after the RPC has been seen by earlier filters in the chain?

ejona86 (approved):

A103-xds-composite-filter.md
markdroth: Thanks, fixed.

A103-xds-composite-filter.md
markdroth: I intended that the implementation can choose whether to fail it before or after the earlier filters.  I've attempted to clarify the wording.

PR #519, opened 2025-10-21
A80: Export TCP Telemetry from gRPC

AananthV:


ctiller (approved):

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,96 @@
+# A80: Export TCP Telemetry from gRPC
+
+**Author(s)**: Aananth V (@aananthv), Nana Pang (@nanahpang), Yash Tibrewal (@yashykt), Yousuk Seung  (@yousukseung)
```
markdroth: The formatting is a little messed up here -- e.g., the "Approvers" list is on the end of the "Authors" line instead of being on its own line.

Please format this as a bulletted list, as per [the template](https://github.com/grpc/proposal/blob/master/GRFC-TEMPLATE.md?plain=1).

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,96 @@
...
+**Approver(s)**: Craig Tiller (@ctiller), Mark Roth (@markdroth)  
+**Status**: In Review  
+**Implemented in**: C-Core  
+**Last updated**: Oct 21, 2025
```
markdroth: Nit: Please put the date in ISO format (YYYY-MM-DD).

markdroth: I don't think we need a new mailing list thread, since https://groups.google.com/g/grpc-io/c/AyT0LVgoqFs was already created for #428.  But please add the link to that thread here.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,96 @@
...
+### *Suggested Metric Collection Algorithms*
+
+#### Per-Connection Metrics
+
+* Set TCP\_CONNECTION\_METRICS\_RECORD\_INTERVAL to a default of 5 minutes.  
+  * Implementations can choose to make it configurable.  
+* For each new connected TCP socket, set an initial alarm of 10% to 110% (randomly selected) of TCP\_CONNECTION\_METRICS\_RECORD\_INTERVAL.  
+* When the alarm fires \-  
+  * Use `getsockopt(TCP_INFO)` or equivalent method to retrieve and record connection metrics.  
+  * Re-arm the alarm with TCP\_CONNECTION\_METRICS\_RECORD\_INTERVAL and repeat.
```
markdroth: Should this use the same 10% to 110% jitter as the initial alarm?

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,96 @@
...
+| grpc.tcp.bytes\_sent | Counter (integer) | By | None | network.local.address network.local.port network.peer.address network.peer.port | Total bytes sent by TCP including retransmissions and spurious retransmissions.  Corresponds to `tcpi_bytes_sent` from `struct tcp_info`. |
+| grpc.tcp.bytes\_retransmitted | Counter (integer) | By | None | network.local.address network.local.port network.peer.address network.peer.port | Total bytes sent by TCP except those sent for the first time. A byte sequence may be retransmitted multiple times and will be counted multiple times as retransmitted. Retransmission counts include spurious retransmissions.  Corresponds to `tcpi_bytes_retrans` from `struct tcp_info`. |
+| **Per-Connection Op Metrics** |  |  |  |  |  |
+| grpc.tcp.connection\_count | Gauge (integer) | {connection} | None | network.local.address network.local.port network.peer.address network.peer.port | Number of active TCP connections. |
```
markdroth: Should this be `UpDownCounter` instead of `Gauge`?

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,96 @@
...
+
+\*   \[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\]: gRPC Non-Per-Call Metrics Framework
+
+## Proposal
+
+This document proposes exporting the following TCP metrics from gRPC to improve the network debugging capabilities for gRPC users.
+
+| Name | Type | Unit | Labels | Optional Labels | Description |
+| ----- | :---- | :---- | :---- | :---- | :---- |
+| **Per-Connection Metrics** |  |  |  |  |  |
```
markdroth: This would be a lot easier to read if we split it up into three separate tables, rather than having 3 subsections in a single large table.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,96 @@
...
+| grpc.tcp.min\_rtt | Histogram (floating-point) | {s} | None | network.local.address network.local.port network.peer.address network.peer.port | TCP's current estimate of minimum round trip time (RTT). It can be used as an indication of the network health between two endpoints. Corresponds to `tcpi_min_rtt` from `struct tcp_info`. |
```
markdroth: This table is really wide, which makes it hard to read.

It looks like every single metric has the same set of optional labels.  I suggest removing that column from the table and instead just having text above that says "All metrics have the following optional labels:" followed by a bulletted list of these four optional labels.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,96 @@
...
+* For each new connected TCP socket,   
+  * Set `writes_since_last_latency_measurement_` to a random integer in \[0, TCP\_LATENCY\_RECORD\_FREQUENCY). Increment this value for every write.  
+  * Perform any prerequisites needed for the socket to support timestamping.   
+    * For Linux TCP, this involves making a setsockopt(SO\_TIMESTAMPING) call with the flag value set to SOF\_TIMESTAMPING\_SOFTWARE | SOF\_TIMESTAMPING\_OPT\_ID | SOF\_TIMESTAMPING\_OPT\_TSONLY | SOF\_TIMESTAMPING\_OPT\_ID\_TCP | SOF\_TIMESTAMPING\_OPT\_STATS.  
+* If `writes_since_last_latency_measurement_` % TCP\_LATENCY\_RECORD\_FREQUENCY \= 0  
+  * Enable latency measurement for the write.   
+    * For Linux TCP, this involves splitting the write into two chunks based on the buckets listed above and adding a SO\_TIMESTAMPING cmsg header to the sendmsg call with the flags SOF\_TIMESTAMPING\_TX\_SCHED | SOF\_TIMESTAMPING\_TX\_SOFTWARE on the sampled chunk and SOF\_TIMESTAMPING\_TX\_ACK on the other chunk.   
+    * There is only one chunk if the write is less than 1024 bytes, in that case all three flags are set on this chunk.  
+  * Set `writes_since_last_latency_measurement_` to 1 and repeat.
+
```
markdroth, line 111: Please add a "Metrics Stability" section, as per the original PR for this gRFC:

https://github.com/nanahpang/proposal/blob/d95b572f30b1494ac0a7240812b2948902c25ec1/A80-grpc-metrics-for-tcp-connection.md#metric-stability

markdroth, line 111: Please add a "Temporary environment variable protection" section, as per the original PR for this gRFC:

https://github.com/nanahpang/proposal/blob/d95b572f30b1494ac0a7240812b2948902c25ec1/A80-grpc-metrics-for-tcp-connection.md#temporary-environment-variable-protection

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,96 @@
...
+\[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\] provides a framework for adding non-per-call metrics in gRPC. This document uses that framework to expose the proposed TCP Latency metrics.
+
+### *Related Proposals:*
+
+\*   \[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\]: gRPC Non-Per-Call Metrics Framework
+
+## Proposal
+
+This document proposes exporting the following TCP metrics from gRPC to improve the network debugging capabilities for gRPC users.
```
markdroth: Just to avoid future questions, let's explicitly state that these metrics will be implemented inside the gRPC endpoint layer.

markdroth (commented):
This looks good!  My comments are mostly minor cosmetic things, but there are two small substantive comments.

A80-tcp-telemetry.md
AananthV: The initial jitter exists so that sockets created in bulk (eg: binary start) do not collect and export at the same time (causing usage spikes). I don't think we need a jitter for re-arming.

A80-tcp-telemetry.md
AananthV: I actually sent out a separate email here: https://groups.google.com/g/grpc-io/c/MoLrWPsFB3s

Added that instead.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+* network.local.address
+* network.local.port
+* network.peer.address
+* network.peer.port
+
+### Per-Connection Metrics
+
+| Name | Type | Unit | Description |
+| ----- | :---- | :---- | :---- |
+| grpc.tcp.min\_rtt | Histogram (floating-point) | {s} | TCP's current estimate of minimum round trip time (RTT). It can be used as an indication of the network health between two endpoints. Corresponds to `tcpi_min_rtt` from `struct tcp_info`. |
```
ejona86: Looks like all of these `{s}` should just be `s`, to mean seconds. (The next line has By/s, which has it correct.)

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. **TCP socket state** can be retrieved using the `getsockopt()` system call with `level` set to `IPPROTO_TCP` and `optname` set to `TCP_INFO`. The state is returned in a `struct tcp_info` which gives details about the TCP connection. At present, the machinery to collect such information is available on Linux 2.6 or later kernels.  
+2. **Per-message transmission timestamps** can be collected from TCP sockets using the [SO\_TIMESTAMPING](https://docs.kernel.org/networking/timestamping.html) interface. At present, this is available on Linux 2.6 or later kernels. These timestamps can be very valuable for diagnosing network level issues and can be used to break down the time spent in the "network".
+
+\[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\] provides a framework for adding non-per-call metrics in gRPC. This document uses that framework to expose the proposed TCP Latency metrics.
+
+### *Related Proposals:*
+
+\*   \[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\]: gRPC Non-Per-Call Metrics Framework
```
ejona86: The bullet here is busted.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+This document proposes collecting and exposing new TCP Endpoint Level Telemetry to gRPC for improved network analysis and debugging.
+
+## Background
+
+The Linux Kernel exposes two telemetry hooks that can be used to collect TCP-level metrics.
+
+1. **TCP socket state** can be retrieved using the `getsockopt()` system call with `level` set to `IPPROTO_TCP` and `optname` set to `TCP_INFO`. The state is returned in a `struct tcp_info` which gives details about the TCP connection. At present, the machinery to collect such information is available on Linux 2.6 or later kernels.  
+2. **Per-message transmission timestamps** can be collected from TCP sockets using the [SO\_TIMESTAMPING](https://docs.kernel.org/networking/timestamping.html) interface. At present, this is available on Linux 2.6 or later kernels. These timestamps can be very valuable for diagnosing network level issues and can be used to break down the time spent in the "network".
+
+\[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\] provides a framework for adding non-per-call metrics in gRPC. This document uses that framework to expose the proposed TCP Latency metrics.
```
ejona86: Are you really meaning to put A79 in square brackets here and elsewhere? Personally, I think it should have "gRFC A79" when introduced, and not just presume the reader will assume "a letter followed by a number is a gRFC."

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+    * For Linux TCP, this involves making a setsockopt(SO\_TIMESTAMPING) call with the flag value set to SOF\_TIMESTAMPING\_SOFTWARE | SOF\_TIMESTAMPING\_OPT\_ID | SOF\_TIMESTAMPING\_OPT\_TSONLY | SOF\_TIMESTAMPING\_OPT\_ID\_TCP | SOF\_TIMESTAMPING\_OPT\_STATS.  
+* If `writes_since_last_latency_measurement_` % TCP\_LATENCY\_RECORD\_FREQUENCY \= 0  
+  * Enable latency measurement for the write.   
+    * For Linux TCP, this involves splitting the write into two chunks based on the buckets listed above and adding a SO\_TIMESTAMPING cmsg header to the sendmsg call with the flags SOF\_TIMESTAMPING\_TX\_SCHED | SOF\_TIMESTAMPING\_TX\_SOFTWARE on the sampled chunk and SOF\_TIMESTAMPING\_TX\_ACK on the other chunk.   
+    * There is only one chunk if the write is less than 1024 bytes, in that case all three flags are set on this chunk.  
+  * Set `writes_since_last_latency_measurement_` to 1 and repeat.
+
+### Metric Stability
+All metrics added in this proposal will start as experimental. The long term goal will be to de-experimentalize them and have them be on by default, but the exact criteria for that change are TBD.
```
ejona86: Is it our goal to have these on by default? The per-connection metrics are sorta hard to use without the labels, but those have too high a cardinality to be generally useful.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+
+\[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\] provides a framework for adding non-per-call metrics in gRPC. This document uses that framework to expose the proposed TCP Latency metrics.
+
+### *Related Proposals:*
+
+\*   \[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\]: gRPC Non-Per-Call Metrics Framework
+
+## Proposal
+
+This document proposes exporting the following TCP metrics from gRPC to improve the network debugging capabilities for gRPC users. These metrics will be implemented inside the gRPC endpoint layer. All metrics have the following optional labels:
```
ejona86, line 31: What having some combination of `grpc.target`/`grpc.lb.locality`/`grpc.lb.backend_service` as an optional label discussed? The only reason I'm asking is because there's the part where we're wanting these to be on by default, and the per-connection metrics look useless for most people today.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+* When the alarm fires \-  
+  * Use `getsockopt(TCP_INFO)` or equivalent method to retrieve and record connection metrics.  
+  * Re-arm the alarm with TCP\_CONNECTION\_METRICS\_RECORD\_INTERVAL and repeat.  
+  * Before the socket is closed, cancel the alarm set above, and retrieve and record connection metrics, providing observability for short-lived connections as well. This will also allow collection of `grpc.tcp.recurring_retransmits`
+
+### Per-Connection Op Metrics
+
+| Name | Type | Unit | Description |
+| ----- | :---- | :---- | :---- |
+| grpc.tcp.connection\_count | UpDownCounter (integer) | {connection} | Number of active TCP connections. |
```
ejona86, line 65: Do we care about the connection rate? If so, this should be two counters (opened+closed).

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+* The `grpc.tcp.syscall_writes` and `grpc.tcp.write_size` metrics will be updated whenever we write to the socket.  
+* The `grpc.tcp.syscall_reads` and `grpc.tcp.read_size` metrics will be updated whenever we read from the socket.
+
+### Per-Write Metrics
+
+| Name | Type | Unit | Labels | Description |
+| ----- | :---- | :---- | :---- | :---- |
+| grpc.tcp.sender\_latency | Histogram (floating-point) | {s} | None | Time taken by the TCP socket to write the first byte of a write onto the NIC. This includes the latency incurred by traffic shaping, qdisc, throttling, and pacing at the sender. Corresponds to the time taken between the final `SCHED` timestamp and the `SENT` timestamp. Sampled periodically. |
+| grpc.tcp.transfer\_latency | Histogram (floating-point) | {s} | size (Bytes)<sup>1</sup> | Time taken to transmit the first size bytes of a write. Transfer latency is measured from when the first byte is handed to the NIC until TCP receives the acknowledgement for the last byte. Corresponds to the time taken between the `SENT` timestamp and the `ACKED` timestamp. Sampled periodically. |
```
ejona86: The size label has no namespacing, and how it is written it is confusing what the label string is ("size" vs "size (Bytes)"; especially because the table has it the full string in backticks; surely that isn't actually the label name?).

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+
+<sup>1</sup> Since transfer latency is strongly affected by the write size, it is broken down into different buckets based on the size of the write. Further, we will only measure the latencies of certain benchmark sample sizes to get measurements that are unaffected by the write sizes. The proposed buckets are:
+
+| Buffered Size | Benchmark Size | `Size (Bytes)` |
+| :---- | :---- | :---- |
+| \[0, 1KiB) | Whole buffer | 1024 |
```
ejona86, line 89: Should we have a note about why we're combining sub-1024 and 1024? I assume that's because it can go out as one packet so ¹ doesn't influence it as much.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+| \[1KiB, 8KiB) | First 1KiB | 1024 |
+| \[8KiB, 64KiB) | First 8KiB | 8196 |
+| \[64KiB, 256KiB) | First 64KiB | 65536 |
+| \[256KiB, 2MiB) | First 256KiB | 262144 |
+| \[2MiB, \+inf) | First 2MiB | 2097152 |
+
+#### Suggested Metric Collection Algorithm
+
+* Set TCP\_LATENCY\_RECORD\_FREQUENCY to a default of 1 in 1000 writes.
```
ejona86: When is this done? Once per process? 1 seems pretty low, especially if you're saying this will be on by default.

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Metric Stability
+All metrics added in this proposal will start as experimental. The long term goal will be to de-experimentalize them and have them be on by default, but the exact criteria for that change are TBD.
+
+### Temporary environment variable protection
+This proposal does not include any features enabled via external I/O, so it does not need environment variable protection.
+
+## Implementation
+
+Will be implemented in C-Core to start with.
```
ejona86: We need to say *somewhere* that it will be normal for other languages to implement only a subset of the metrics.

A80-tcp-telemetry.md
AananthV: Made the label name "size" and changed the table to "`size` (Bytes)"

A80-tcp-telemetry.md
AananthV: Done. Added a line.

A80-tcp-telemetry.md
AananthV: This is 1 out of every 1000 writes in each connection. It is also the default sampling rate stated in the Fathom paper: https://dl.acm.org/doi/pdf/10.1145/3603269.3604815

A80-tcp-telemetry.md
AananthV: Moving this to the other comment.

A80-tcp-telemetry.md
AananthV: I am open to this if @ctiller @markdroth feel this is useful.

A80-tcp-telemetry.md
AananthV: We discussed this earlier and came to the conclusion that the # of active connections is a more important metric for now. We can explore connection rate in the future if needed.

A80-tcp-telemetry.md
```diff
@@ -17,51 +18,34 @@ The Linux Kernel exposes two telemetry hooks that can be used to collect TCP-lev
 1. **TCP socket state** can be retrieved using the `getsockopt()` system call with `level` set to `IPPROTO_TCP` and `optname` set to `TCP_INFO`. The state is returned in a `struct tcp_info` which gives details about the TCP connection. At present, the machinery to collect such information is available on Linux 2.6 or later kernels.  
 2. **Per-message transmission timestamps** can be collected from TCP sockets using the [SO\_TIMESTAMPING](https://docs.kernel.org/networking/timestamping.html) interface. At present, this is available on Linux 2.6 or later kernels. These timestamps can be very valuable for diagnosing network level issues and can be used to break down the time spent in the "network".
 
-\[[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\] provides a framework for adding non-per-call metrics in gRPC. This document uses that framework to expose the proposed TCP Latency metrics.
+\[[gRFC A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)\] provides a framework for adding non-per-call metrics in gRPC. This document uses that framework to expose the proposed TCP Latency metrics.
```
markdroth: This can be a relative link rather than an absolute one.  Also, you can provide the link just once and refer to it in multiple places in the document, like this:

```
[gRFC A79][A79] provides a framework for adding non-per-call metrics in gRPC.
This document uses that framework to expose the proposed TCP Latency metrics.

### Related Proposals:

* [gRFC A79: Non-Per-Call Metrics Framework][A79]

[A79]: A79-non-per-call-metrics-architecture.md
```

A80-tcp-telemetry.md
markdroth: I think there could still be use cases where the various connections wind up converging on similar times over a long period of time.  For example, consider what happens if there's a brief period of thread pool starvation that delays some timer callbacks from running, and then they all run at once when the starvation ends.  Applying the jitter for each successive timer seems like it would protect against that sort of thing, and I don't see much of a down-side to doing so.

Note that our [backoff spec](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md) does apply jitter on each connection attempt, so that seems like a good precedent to follow.

A80-tcp-telemetry.md
markdroth: Are these metrics actually useful without these optional labels?  I'm not sure if the aggregation would work for the histograms.  The counters would probably be fine, though.

If the metrics are not useful without these optional labels, then we should make the labels non-optional.  However, in that case, the cardinality will be way too high to ever enable these metrics by default, so we should say that we never plan to enable them by default.

If the metrics *are* useful without these optional labels, then we can consider enabling them by default at some point in the future.  In that case, we can also consider adding some other useful labels.  However, we'd need a bit more discussion as to which labels to add.  Note that these metrics would be used for both client-side connections and server-side connections, but the labels that Eric suggested would be available on client side only -- and even then, they would apply only for endpoint connections, not for side-channel connections like auth token fetches.

A80-tcp-telemetry.md
AananthV: Done

A80-tcp-telemetry.md
AananthV: Done.

A80-tcp-telemetry.md
AananthV: Currently, the metrics (without labels) are good to gauge the overall networking performance of a server (vs an individual connection). I think they can be enabled by default.

I do think that we can add more labels to be more useful generally, but maybe we can add them as an amendment to this gRFC after discussion? WDYT.

A80-tcp-telemetry.md
ejona86: Oh, I read this as "1 to 1000." It is an integer right, based on writes_since_last_latency_measurement_ below, so it would actually be "1000"? Although the name is strange for the value it holds.

A80-tcp-telemetry.md
ejona86: To repeat: "Are you really meaning to put A79 in square brackets here and elsewhere?"

A80-tcp-telemetry.md
ejona86: Repeating: "The size label has no namespacing". How do we feel about that @markdroth @dfawley

A80-tcp-telemetry.md
ejona86: Since I just realized these metrics apply to the server as well, connection rate is important. As a service owner, a flood of connections would be important.

A80-tcp-telemetry.md
ejona86: For "Per-Connection Metrics":
The histograms are sorta useless without buckets based on destination, as the expected latency/speed varies wildly. The 99% when a long link is actually broken could be useful, except then you don't know what network link is broken. Retransmits are a bit more useful, but, again, you wouldn't know which network link is broken, nor even whether this was an inbound connection or outbound connection.

That leaves bytes_sent and packets_sent. bytes_sent will be pretty close to grpc.client.attempt.sent_total_compressed_message_size, so that by itself is probably not too important in isolation. That just leaves packets_sent.

What is the expectation on using this if this is on by default? Someone would see there is a problem, then run tcpdump for all traffic on a server having trouble?

For "Per-Connection Op Metrics": I can't say much about what labels would generally be useful here. I don't know why this would be on by default. Except for `grpc.tcp.connection_count`; that would generally be useful, although really you want to know inbound or outbound connection.

For "Per-Write Metrics": Same problem as "Per-Connection Metrics" histograms.

ejona86 (commented):
So you understand why I'm looking at this so much more closely now vs before: Earlier this was "some metrics y'all wanted for yourselves". And I knew why you'd want them, and why they'd be useful to others, but there wasn't an idea we'd move quickly to implement this cross-language for wide availability. With the idea these are going to be on by default, that means these are seen as being more important/widely used metrics. They are now intended for the masses. Thus I started reviewing them as such.

Since I'm going to be unavailable next week: I'd said the comments I need to say. Once the threads are actually resolved (not just marked resolved), it seems it would be good to go.

A80-tcp-telemetry.md
AananthV: What does this mean?

Do you want it to be something like `grpc.transfer_size`?

A80-tcp-telemetry.md
AananthV: Renamed it to TCP_LATENCY_RECORD_PERIOD

A80-tcp-telemetry.md
AananthV: Added `grpc.tcp.connections_created`.

A80-tcp-telemetry.md
AananthV: (resolving based on offline discussion)

Updated the Metric Stability section to: 

> All metrics added in this proposal will start as experimental. The long term goal will be to de-experimentalize them and potentially have some metrics be on by default, but the exact criteria for that change are TBD. We may also add new labels (eg: `grpc.lb.locality`, `grpc.lb.backend_service`) in the future. This gRFC will be amended when this happens.

A80-tcp-telemetry.md
AananthV: Removed the square brackets.

ejona86 (approved):
I had thought `grpc.tcp.connections_created` would be paired with a `closed` counter, as that's what we've done in the past instead of UpDown counters. But I can live with this.

Thank you!

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,120 @@
...
+| grpc.tcp.bytes\_retransmitted | Counter (integer) | By | Total bytes sent by TCP except those sent for the first time. A byte sequence may be retransmitted multiple times and will be counted multiple times as retransmitted. Retransmission counts include spurious retransmissions.  Corresponds to `tcpi_bytes_retrans` from `struct tcp_info`. |
+
+#### Suggested Metric Collection Algorithm
+
+* Set TCP\_CONNECTION\_METRICS\_RECORD\_INTERVAL to a default of 5 minutes.  
+  * Implementations can choose to make it configurable.  
+* For each new connected TCP socket, set an initial alarm of 10% to 110% (randomly selected) of TCP\_CONNECTION\_METRICS\_RECORD\_INTERVAL.  
+* When the alarm fires \-  
+  * Use `getsockopt(TCP_INFO)` or equivalent method to retrieve and record connection metrics.  
+  * Re-arm the alarm with 10% to 110% (randomly selected) of TCP\_CONNECTION\_METRICS\_RECORD\_INTERVAL and repeat.
```
ejona86: I can understand why the initial alarm might be set to 10%. Why would we let the re-arming be as low as 10%?

A80-tcp-telemetry.md
```diff
@@ -0,0 +1,120 @@
...
+| grpc.tcp.min\_rtt | Histogram (floating-point) | s | TCP's current estimate of minimum round trip time (RTT). It can be used as an indication of the network health between two endpoints. Corresponds to `tcpi_min_rtt` from `struct tcp_info`. |
+| grpc.tcp.delivery\_rate | Histogram (floating-point) | By/s | TCP’s most recent measure of the connection’s "non-app-limited" throughput. The term non-app-limited means that the link is saturated by the application. The delivery rate is only reported when it is non-app-limited.  Corresponds to `tcpi_delivery_rate` from `tcp_info` when `tcpi_delivery_rate_app_limited` is `false`. |
+| grpc.tcp.packets\_sent  | Counter (integer) | {packet}  | Total packets sent by TCP including retransmissions and spurious retransmissions.  Corresponds to `tcpi_data_segs_out` from `struct tcp_info`. |
+| grpc.tcp.packets\_retransmitted | Counter (integer) | {packet} | Total packets sent by TCP except those sent for the first time. A packet may be retransmitted multiple times and will be counted multiple times as retransmitted. Retransmission counts include spurious retransmissions.  Corresponds to `tcpi_total_retrans` from `struct tcp_info`. |
```
AgraVator, line 44: Since tcpi_total_retrans tracks cumulative retransmissions over a specific connection's lifetime, it should be recorded as a Gauge. Using a Counter seems incorrect because a reset to zero in the kernel would break the monotonicity required for Counters, whereas a Gauge accurately captures the connection's current state and handles resets gracefully.

AananthV: Under what conditions is the `tcpi_total_retrans` reset in the kernel?

A80-tcp-telemetry.md
AananthV: Changed to "90% to 110%"

A80-tcp-telemetry.md
AgraVator: You're right—it doesn't reset mid-connection, so a monotonic Counter is appropriate here.

My concern is that struct tcp_info provides absolute cumulative totals, while the OpenTelemetry Counter API expects increments (deltas). Since implementations are forced to calculate these deltas manually between polling intervals, shouldn't we explicitly state that deltas must be reported? Without it, implementations might mistakenly report the ever-growing absolute kernel values directly.

PR #520, opened 2025-11-04
A106: xDS Unified Matcher and CEL Integration

sergiitk:
xDS Unified Matcher and CEL Integration

Rendered version: https://github.com/sergiitk/grfc-proposal/blob/a106-xds-unified-matcher-and-cel/A106-xds-unified-matcher-and-cel.md

### DIFF
--- a/A106-xds-unified-matcher-and-cel.md
+++ b/A106-xds-unified-matcher-and-cel.md
@@ -1,0 +1,413 @@
+A106: xDS Unified Matcher and CEL Integration
+======
+
+* Author(s): Sergii Tkachenko (@sergiitk)
+* Approver: Mark Roth (@markdroth)
+* Status: In Review
+* Implemented in:
+* Last updated: 2025-11-03
+* Discussion at:
+  - [ ] TODO(sergiitk): insert google group thread
+
+## Abstract
+
+TODO(sergiitk): new abstract
+
+## Background
+
+TODO(sergiitk): new background
+
+1.  [Unified Matcher API].
+2.  One of the matching mechanisms will be [CEL](https://cel.dev/) (Common
+    Expression Language).
+
+### Related Proposals
+
+* [gRFC A41: xDS RBAC Support][A41]
+* [gRFC A77: xDS Server-Side Rate Limiting][A77] (WIP)
+* [gRFC A92: xDS ExtAuthz Support][A92] (WIP)
+* [gRFC A93: xDS ExtProc Support][A93] (WIP)
+
+[A41]: A41-xds-rbac.md
+[A77]: https://github.com/grpc/proposal/pull/414
+[A92]: https://github.com/grpc/proposal/pull/481
+[A93]: https://github.com/grpc/proposal/pull/484
+
+[Unified Matcher API]: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/matching/matching_api.html
+[Unified Matcher API Support]: #unified-matcher-api-support
+[Unified Matcher: Filter Integration]: #unified-matcher-filter-integration
+[Unified Matcher: Input Extensions]: #unified-matcher-input-extensions
+[Unified Matcher: Matching Extensions]: #unified-matcher-matching-extensions
+[Unified Matcher: `Matcher`]: #unified-matcher-matcher
+[Unified Matcher: `OnMatch`]: #unified-matcher-onmatch
+[Unified Matcher: `MatcherList`]: #unified-matcher-matcherlist
+[Unified Matcher: `MatcherTree`]: #unified-matcher-matchertree
+[Unified Matcher: `HttpRequestHeaderMatchInput`]: #unified-matcher-httprequestheadermatchinput
+[Unified Matcher: `HttpAttributesCelMatchInput`]: #unified-matcher-httpattributescelmatchinput
+[Unified Matcher: `StringMatcher`]: #unified-matcher-stringmatcher
+[Unified Matcher: `CelMatcher`]: #unified-matcher-celmatcher
+
+[`cel.expr.CheckedExpr`]: https://github.com/google/cel-spec/blob/master/proto/cel/expr/checked.proto
+[CEL Integration]: #cel-integration
+[CEL Runtime Restrictions]: #cel-runtime-restrictions
+[Supported CEL Variables]: #supported-cel-variables
+
+## Proposal
+
+### Unified Matcher API Support
+
+[Unified Matcher API] is an adaptable framework that can be used in any xDS
+component that needs matching features.
+
+Envoy provides two syntactically equivalent Unified Matcher definitions:
+[`envoy.config.common.matcher.v3.Matcher`](https://github.com/envoyproxy/envoy/blob/e3da7ebb16ad01c2ac7662758a75dba5cdc024ce/api/envoy/config/common/matcher/v3/matcher.proto)
+and
+[`xds.type.matcher.v3.Matcher`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto),
+which is the preferred version for all new APIs using Unified Matcher. If
+`envoy.config.common.matcher.v3.Matcher` is provided, we will interpret it as is
+`xds.type.matcher.v3.Matcher`.
+
+In this iteration the following Unified Mather extensions will be supported:
+
+1.  Inputs:
+    1.  [Unified Matcher: `HttpRequestHeaderMatchInput`]
+    2.  [Unified Matcher: `HttpAttributesCelMatchInput`]
+2.  Matchers:
+    1.  [Unified Matcher: `StringMatcher`] (standard matcher)
+    1.  [Unified Matcher: `CelMatcher`]
+
+#### Unified Matcher: Filter Integration
+
+When implementing Unified Matcher API, a filter must define the following:
+
+-   Supported protocol-specific actions (see [Unified Matcher: `OnMatch`]).
+-   Supported [Unified Matcher: Input Extensions].
+-   Supported [Unified Matcher: Matching Extensions], including any additional
+    limitations on their inputs.
+-   Filter-specific default no-match behavior (f.e. xDS resource NACK).
+
+##### Unified Matcher: `Matcher`
+
+While the Unified Matcher API allows for matcher trees of arbitrary depth, gRPC
+will reject any matcher definition with a tree depth greater than 100, NACKing
+the xDS resource.
+
+We will support the following fields in the
+[`xds.type.matcher.v3.Matcher`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L22)
+message:
+
+-   `matcher_type`: One of the following must be present and valid:
+    -   [`matcher_list`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L126):
+        A valid [Unified Matcher: `MatcherList`] message.
+    -   [`matcher_tree`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L129):
+        A valid [Unified Matcher: `MatcherTree`] message.
+-   [`on_no_match`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L135):
+    Specifies the action executed if no match is found in the `matcher_list` or
+    `matcher_tree`. If set, must be a valid [Unified Matcher: `OnMatch`].
+    If not set, refer to filter's default no-match behavior.
+
+##### Unified Matcher: `OnMatch`
+
+We will support the following fields in the
+[`xds.type.matcher.v3.Matcher.OnMatch`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L24)
+message:
+
+-   `on_match`: One of the following must be present and valid:
+    -   [`matcher`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L33):
+        A nested [Unified Matcher: `Matcher`] for more complex, tree-like
+        matching logic.
+    -   [`action`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L36):
+        A [`TypedExtensionConfig`] containing a protocol-specific action to
+        take.
+
+The following fields will be ignored by gRPC:
+
+-   `keep_matching`: Not supported in the initial implementation, may be added
+    later.
+
+##### Unified Matcher: `MatcherList`
+
+We will support the following fields in the
+[`xds.type.matcher.v3.Matcher.MatcherList`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L43)
+message:
+
+-   [`matchers`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L96):
+    A list of
+    [`Matcher.MatcherList.FieldMatcher`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L87)
+    messages. Must contain at least 1 item.
+    -   [`predicate`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L89):
+        Must be set and contain a valid
+        [`Matcher.MatcherList.Predicate`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L45)
+        message.
+        -   `match_type`: One of the following must be present and valid:
+            -   [`single_predicate`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L73):
+                A
+                [`Matcher.MatcherList.Predicate.SinglePredicate`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L47)
+                message. The return type of the input must match the input type
+                of the matcher.
+                -   [`input`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L50):
+                    A valid [`TypedExtensionConfig`]. Must be present and
+                    contain one of the input extensions
+                    [supported by the filter][Unified Matcher: Filter Integration].
+                    Must have return type compatible with the `matcher`.
+                -   `matcher`: One of the following must be present and valid:
+                    -   [`value_match`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L56):
+                        A valid [Unified Matcher: `StringMatcher`]. Only
+                        compatible with `input` that returns a string.
+                    -   [`custom_match`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L60):
+                        A valid [`TypedExtensionConfig`] containing one of the
+                        matching extensions
+                        [supported by the filter][Unified Matcher: Filter Integration].
+                        Must have input type compatible with the `input`. Must
+                        return a boolean indicating the status of the match.
+            -   [`or_matcher`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L76):
+                A
+                [`Matcher.MatcherList.Predicate.PredicateList`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L65)
+                message.
+                -   [`predicate`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L66):
+                    A list of `Matcher.MatcherList.Predicate` messages. Must
+                    contain at least 2 items. Returns true if any of them are
+                    true.
+            -   [`and_matcher`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L79):
+                A
+                [`Matcher.MatcherList.Predicate.PredicateList`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L65)
+                message.
+                -   [`predicate`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L66):
+                    A list of `Matcher.MatcherList.Predicate` messages. Must
+                    contain at least 2 items. Returns true if all of them are
+                    true.
+            -   [`not_matcher`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L82):
+                A nested `Matcher.MatcherList.Predicate` message. Returns the
+                inverted result of predicate evaluation.
+    -   [`on_match`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L92):
+        Must be set and contain a valid [Unified Matcher: `OnMatch`] message.
+
+##### Unified Matcher: `MatcherTree`
+
+We will support the following fields in the
+[`xds.type.matcher.v3.Matcher.MatcherTree`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L99)
+message:
+
+-   [`input`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L106):
+    A valid [`TypedExtensionConfig`]. Must be present and contain one of the
+    input extensions
+    [supported by the filter][Unified Matcher: Filter Integration]. Must have
+    return type compatible with the `matcher`.
+-   `tree_type`: One of the following must be present and valid:
+    -   [`exact_match_map`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L114):
+        A
+        [`Matcher.MatcherTree.MatchMap`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L101)
+        message. Only compatible with `input` that returns a string.
+        -   [`map`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L102):
+            A map from a string to a valid [Unified Matcher: `OnMatch`] message.
+            Must contain at least 1 pair.
+    -   [`prefix_match_map`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L117):
+        A
+        [`Matcher.MatcherTree.MatchMap`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L101)
+        message. Only compatible with `input` that returns a string.
+        -   [`map`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L102):
+            A map from a string to a valid [Unified Matcher: `OnMatch`] message.
+            Must contain at least 1 pair.
+    -   [`custom_match`](https://github.com/cncf/xds/blob/b4127c9b8d78b77423fd25169f05b7476b6ea932/xds/type/matcher/v3/matcher.proto#L120):
+        A valid [`TypedExtensionConfig`] containing one of the matching
+        extensions
+        [supported by the filter][Unified Matcher: Filter Integration]. Must
+        have input type compatible with the `input`. Must return a boolean
+        indicating the status of the match.
+
+##### Unified Matcher: Input Extensions
+
+###### Unified Matcher: `HttpRequestHeaderMatchInput`
+
+Returns a `string` containing the value of the header with name specified in
+`header_name`.
+
+We will support the following fields in the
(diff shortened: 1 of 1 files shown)


sergiitk:
Split from #414

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+-   `on_match`: One of the following must be present:
+    -   [`matcher`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L33)
+        ([Unified Matcher: `Matcher`]): A nested matcher that allows for
+        building more complex, tree-like matching logic.
+    -   [`action`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L36)
+        ([`TypedExtensionConfig`]): If set, must contain one of the
+        protocol-specific actions
+        [supported by the filter][Unified Matcher: Filter Integration].
+-   [`keep_matching`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L45)
+    (bool): If this field is set in a context in which it's not supported, the
```
ejona86, line 178: It is not at all clear how the validator knows if the field is supported. The proto description doesn't really say anything either.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+        the tree evaluation is terminated, and considered an
+        [unsuccessful match][Unified Matcher: Filter Integration].
+-   `keep_matching` determines whether a successful match within an `OnMatch`
+    should be considered "terminal" for the current `Matcher` being evaluated.
+    It essentially answers the question: "After processing this `OnMatch`,
+    should the current matcher stop looking for more matches, or continue?"
+    -   If `keep_matching` is `false` (the usual case), finding this `OnMatch`
+       is terminal. The `Action` is added (or the nested `Matcher` is
+       evaluated), and the current matcher stops searching.
+    -   If `keep_matching` is `true`, the `Action` is added (or nested
```
ejona86, line 421: How does the caller know that the Action is keep_matching=true and it should "behave as if the Matcher did not match"?

I could understand "only consider the last Action as being the one to take," but if there is no `on_no_match`, then the last action could be for this keep_matching=true case. the `keep_matching` protobuf doesn't say what happens either if there is no terminal matcher. Does the entire match fail and the action here discarded?

ejona86: There should not be any "added" idea here. Actions for keep_matching shouldn't mix with the actual selected action. [Envoy uses a callback](http://google3/third_party/envoy/src/source/extensions/filters/http/rate_limit_quota/filter.cc;l=260;rcl=836283804) to learn about keep_matching=true actions. We should say that it is communicated separately in the gRFC.

A106-xds-unified-matcher-and-cel.md
markdroth: This should be handled the same way as the set of allowed actions: it's basically determined by the calling filter when it calls the code to parse the matcher tree from the xDS proto.  For example, RLQS might decide that it supports `keep_matching`, but the composite filter might decide that it doesn't.  So if we get a matcher with `keep_matching` in the RLQS filter config, that's fine, but if we get a matcher with `keep_matching` in the composite filter config, that would generate a validation error.

The gRFC should use the same language to describe how this works for the set of allowed actions and for `keep_matching`, since they are both handled the same way.

A106-xds-unified-matcher-and-cel.md
markdroth: The way we implemented this in C-core, the matcher returns a list of matching actions and a bool indicating whether the match succeeded.  When we get a matching action, we append it to the list of matching actions to return.  If `keep_matching` is false (the default), then we stop searching at this point and return true to indicate that we found a match.  However, if `keep_matching` is true, then we keep going as if we didn't find a match.  If we get to the end without finding any match (not even an `on_no_match`), then we return false.

If the matcher returns true, then we know that the last entry in the list of returned actions is the one we should actually take, and any earlier entries in the list are the ones that matched but had `keep_matching` set to true.  If the match returns false, then *all* actions in the list are the ones that matched but had `keep_matching` set to true.

If a given implementation wants to use a callback API instead, that's fine too.  But it would need to pass a bool to the callback indicating whether it's the final match or whether it's a `keep_matching` match.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+the matcher tree evaluation.
+
+##### Behavior for Unsuccessful Matches
+
+The match is considered unsuccessful:
+
+1.  If no match found after evaluating the [Unified Matcher: `Matcher`] AND
+2.  `on_no_match` field is unset OR its evaluation is unsuccessful.
+
+The filter may define any behavior for an unsuccessful match, f.e. NACK the xDS
```
markdroth, line 135: It cannot NACK the xDS resource on an unsuccessful match, because the actual matching happens only after the xDS resource has already been validated.

More generally, though, I'm not sure we actually need this statement at all, since this doesn't come into play until the matcher returns.  At that point, the logic is not part of the matcher; it's part of the component that is calling the matcher.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+    [unsuccessful match behavior][Unified Matcher: Filter Integration].
+
+#### Unified Matcher: `OnMatch`
+
+We will support the following fields in the
+[`xds.type.matcher.v3.Matcher.OnMatch`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L24)
+message:
+
+-   `on_match`: One of the following must be present:
+    -   [`matcher`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L33)
```
markdroth, line 170: There has historically been some confusion about the exact matching semantics here, so I think it's worth describing the intended behavior here, even if it duplicates the description in the proto file.  For this field, I would say something like this:

If the matcher does not return a successful match, then the context in which this `OnMatch` is evaluated is considered not to have matched.  Examples:
- In a matcher list, if a predicate matches the input but has an `OnMatch` containing a nested matcher that does not return a successful match, the matcher list will treat this as if the predicate did not match the input (i.e., it will continue on to the next predicate).
- In a prefix match map, if the longest matching prefix that matches the input has an `OnMatch` containing a nested matcher that does not return a successful match, the prefix match map will treat this as if the prefix did not match the input (i.e., it will look for a shorter prefix that matches the input).
- In an exact match map, if the prefix that matches the input has an `OnMatch` containing a nested matcher that does not return a successful match, the exact match map will treat the match as unsuccessful.  (In this case, there is no need to continue searching, because there cannot possibly be another match.)

Note that if the context in which the `OnMatch` is evaluated is considered not to have matched, then `on_no_match` will be used if present.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+   combination of thereof, additional parameters to mathers that support it,
+   etc.
+
+#### Unified Matcher: Filter Integration
+
+When implementing Unified Matcher API, a filter must define the following for
+each [Unified Matcher: `Matcher`] field in its config:
+
+1.  Supported protocol-specific actions.
+2.  Supported [Unified Matcher: Input Extensions].
```
markdroth, line 109: The input extensions should not need to be provided by the individual caller.  There should be a central registry of all supported input extensions, and any of them should be usable in any matcher that provides a compatible context.

Instead, the caller needs to provide the matcher context type to the parser, so that it can reject the resource if the matcher specifies an input extension that is not compatible with the context that the caller will use for actually evaluating the matcher.  For example, an input extension that gets a header value from an RPC will only be able to get that value from an RPC matcher context; if the caller is instead using (e.g.) a match context for TCP connections, then it's invalid to use that input extension.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+3.  Filter-specific behavior for unsuccessful matches.
```
markdroth, line 110: I assume this is talking about the `keep_matching` behavior?  If so, the wording isn't very clear -- those are not unsuccessful matches; they're matches that aren't going to cause the search to immediately terminate successfully.

If this is not talking about `keep_matching`, then I'm not sure what it's referring to.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+2.  **Matcher Action**: If a matcher evaluates to true, an associated action is
+   taken. This action could be anything from selecting a route to applying a
+   filter.
+3.  **Matcher Input**: This extracts the data from the Matcher Context and
+   provides it to the matchers evaluate. For example, it may get a specific
+   header from the request provided in the matcher context.
+4.  **Matcher Context**: This holds the input data and any other contextual
+   information needed during the matching process. It may include information
+   about the request being processed, the response, connection info, a
+   combination of thereof, additional parameters to mathers that support it,
```
markdroth, line 100: s/mathers/matchers/

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+
+## Proposal
+
+### Unified Matcher
+
+#### Unified Matcher: Core Concepts
+
+The Unified Matcher API revolves around a few key concepts:
+
+1.  **Matcher**: A matcher is a rule that evaluates to true or false based on
```
markdroth, line 87: It doesn't just evaluate to true or false; it also returns action(s) associated with the matched condition(s).

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+
+[`StringValue`]: https://protobuf.dev/reference/protobuf/google.protobuf/#string-value
+[`TypedExtensionConfig`]: https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/core/v3/extension.proto#L14
+[RE2_wiki]: https://github.com/google/re2/wiki/Syntax
+
+## Proposal
+
+### Unified Matcher
+
+#### Unified Matcher: Core Concepts
```
markdroth, line 83: The "Unified Matcher:" prefix on each of these subsection headers seems redundant, since they're subsections of a section with exactly that name.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+
+The Unified Matcher API revolves around a few key concepts:
+
+1.  **Matcher**: A matcher is a rule that evaluates to true or false based on
+   some properties of the input data. Matchers can be simple (e.g., checking if
+   a header has a specific value) or complex (e.g., a boolean combination of
+   other matchers, nested matchers, etc).
+2.  **Matcher Action**: If a matcher evaluates to true, an associated action is
```
markdroth, line 91: I suggest reversing the order of these last three elements.  I think we want to describe the match context first, since that's important information about what kinds of things we're going to be matching.  Then we can describe the input extension, which extracts individual pieces of information from the context.  And then finally, we can describe the action, which is the main result of matching against that input.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+[CEL: Supported Functions]: #cel-supported-functions
+[CEL: Supported Variables]: #cel-supported-variables
+
+[`StringValue`]: https://protobuf.dev/reference/protobuf/google.protobuf/#string-value
+[`TypedExtensionConfig`]: https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/core/v3/extension.proto#L14
+[RE2_wiki]: https://github.com/google/re2/wiki/Syntax
+
+## Proposal
+
+### Unified Matcher
```
markdroth, line 81: This section is incredibly detailed, which makes it a bit hard to see the big picture or understand how the various parts relate to each other.  I think it needs a clearer overview before we start getting into the fine details.  It would also help to split up the technical details into sections that provide a bit more context for the information.

I suggest reorganizing it a bit, ideally around the following lines:

1. High level overview (does not include specific proto fields):
   - core concepts (what matchers are used for)
   - overview of the general matcher structure
   - the three types of matchers (matcher list, exact match map, and prefix match map)
2. Parsing logic:
   - what information needs to be provided by the calling component to validate the config
   - show specific proto fields and what validation needs to be done for each
   - the result of successful parsing (a matcher object that can actually be evaluated against some input)
3. Evaluation logic:
   - what information needs to be provided by the calling component to evaluate the match against an input
   - what the result of the match is (a list of zero or more matching actions, and a bool indicating if the match was successful)

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+3.  **Matcher Input**: This extracts the data from the Matcher Context and
+   provides it to the matchers evaluate. For example, it may get a specific
+   header from the request provided in the matcher context.
+4.  **Matcher Context**: This holds the input data and any other contextual
+   information needed during the matching process. It may include information
+   about the request being processed, the response, connection info, a
+   combination of thereof, additional parameters to mathers that support it,
+   etc.
+
+#### Unified Matcher: Filter Integration
```
markdroth, line 103: Although the two cases we currently have for the unified matcher (RLQS and the composite filter) are both filters, the unified matcher API is not really specific to filters at all.  It can also be used for L4 filter chain selection, L7 routing, and many other things.  So we should try to make sure our language here is a bit more generic.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+
+-   `on_match`: One of the following must be present:
+    -   [`matcher`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L33)
+        ([Unified Matcher: `Matcher`]): A nested matcher that allows for
+        building more complex, tree-like matching logic.
+    -   [`action`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L36)
+        ([`TypedExtensionConfig`]): If set, must contain one of the
+        protocol-specific actions
+        [supported by the filter][Unified Matcher: Filter Integration].
+-   [`keep_matching`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L45)
```
markdroth, line 177: Similar to my comment above about the `matcher` field, I think it's important to spell out the matching behavior here.  I suggest something like this:

If set to true, then if this `OnMatch` produces a successful match (i.e., either by containing an action or by containing a nested matcher that produces a successful match), the match will not be considered successful.  The matching action(s) will still be returned, but the matcher will not terminate its search; instead, it will continue looking for subsequent matches.  Examples:
- In a matcher list, if a predicate matches the input but has an `OnMatch` containing an action and has `keep_matching` set to true, then the action will be returned, but the matcher list will treat this as if the predicate did not match the input (i.e., it will continue on to the next predicate).
- In a prefix match map, if the longest matching prefix that matches the input has an `OnMatch` containing an action and has `keep_matching` set to true, then the action will be returned, but the prefix match map will treat this as if the prefix did not match the input (i.e., it will look for a shorter prefix that matches the input).
- In an exact match map, if the prefix that matches the input has an `OnMatch` containing an action and has `keep_matching` set to true, the action will be returned, but the exact match map will treat the match as unsuccessful.  (In this case, there is no need to continue searching, because there cannot possibly be another match.)

If all matches found have `keep_matching` set to true, then the matcher is considered to not match, which means that `on_no_match` will be used if present.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+                    [supported by the filter][Unified Matcher: Filter Integration].
+                    Must have return type compatible with the `matcher`.
+                -   `matcher`: One of the following must be present:
+                    -   [`value_match`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L64)
+                        ([Unified Matcher: `StringMatcher`]): Only compatible
+                        with `input` that returns a string.
+                    -   [`custom_match`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L68)
+                        ([`TypedExtensionConfig`]): Must contain one of the
+                        matching extensions
+                        [supported by the filter][Unified Matcher: Filter Integration].
```
markdroth, line 214: Similar to input types, I don't think these matcher types should be provided by the caller.  They should be in a centralized registry of supported matcher types.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+            at least 1 pair.
+
+The following are not supported by gRPC in the initial implementation and will
+result in xDS resource NACK:
+
+-   `custom_match`
+
+#### Unified Matcher: Input Extensions
+
+In this iteration, the following Unified Mather extensions will be supported:
```
markdroth, line 266: s/Mather/Matcher/

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+
+1.  [Unified Matcher: `StringMatcher`](standard matcher)
+2.  [Unified Matcher: `CelMatcher`]
+
+##### Unified Matcher: `StringMatcher`
+
+Compatible with [Unified Matcher: Input Extensions] that return a `string`.
+
+We will support the following fields in the
```
markdroth, line 306: I don't think we need to spell out the fields in `StringMatcher`, because all languages should already have code to handle this type -- it's not new with this gRFC.  It was originally added for RBAC (see [gRFC A41](https://github.com/grpc/proposal/blob/master/A41-xds-rbac.md)), and we later added support for it in RDS (see [gRFC A63](https://github.com/grpc/proposal/blob/master/A63-xds-string-matcher-in-header-matching.md)).  We should just refer to those gRFCs here.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+
+#### Unified Matcher: Evaluation Flow
+
+**Goal:** To produce a list of matching `Action`s.\
+**Matcher Context:** The input to the Matcher evaluation tree. Provided at
+runtime by the filter, contains the input data. May contain any other contextual
+information relevant to the filter. Implementation note: when setting an input,
+consider memory footprint. For example, instead of resolving all headers in
+advance, provide them in a lazy-loading wrapper.
+**Matching Process:** Starting with a top-level `Matcher`, there will be one
```
markdroth, line 368: Looks like this isn't being formatted on its own line, as I assume was intended.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+    [CEL: Runtime Restrictions] and only contain [CEL: Supported Variables]. If
+    the conversion or the validation step fail, gRPC will NACK the xDS resource.
+-   [`description`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/cel.proto#L36):
+    An optional string. May be ignored or used for testing/debugging.
+
+#### Unified Matcher: Evaluation Flow
+
+**Goal:** To produce a list of matching `Action`s.\
+**Matcher Context:** The input to the Matcher evaluation tree. Provided at
+runtime by the filter, contains the input data. May contain any other contextual
```
markdroth, line 364: Not sure what you mean by "May contain any other contextual information relevant to the filter."  I don't think that's going to be the case.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+    -   Each rule has a **Condition** (`Predicate`) and a **Result** (`OnMatch`).
+    -   **Condition Checking:** To check a typical condition:
+        1.  Using the `input` extension, extract a specific piece of data from
+            the **Matcher Context** (e.g., the value of requests's `:host`
+            header).
+        2.  Using the `matcher` extension, compare the extracted data against
+            the matcher's criteria (e.g., "is it equal to `'example.com'`?",
+            "does it start with `'api.'`?").
+    -   Conditions can be combined using AND, OR, NOT, or nested via `OnMatch`.
+    -   **First Match Wins:** The *first* rule whose **Condition** is true has
```
markdroth, line 382: There are two cases where this will not be true, which I mentioned above: (1) the `OnMatch` can contain a nested matcher that does not return a successful match, and (2) the `OnMatch` could have `keep_matching` set.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+        its **Result** executed.
```
markdroth, line 383: Instead of saying "has its result executed", I would say something like "returns the corresponding action or evaluates the corresponding nested matcher".

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+
+-   **[Prefix Map Matcher][Unified Matcher: `MatcherTree`]:** Similar to the Map
+    Matcher, but uses prefix matching (a Trie data structure).
+    -   Using the `input` extension, it extracts a specific string value from
+        the **Matcher Context**.
+    -   It finds all entries in the map whose keys are prefixes of the input
+        string.
+    -   The **Result** is chosen based the the key with the *longest* matching
+        prefix.
+    -   Undefined behavior: If there are multiple prefixes of the same greatest
```
markdroth, line 400: I don't see how this case is possible.  There cannot be two strings of the same length that are both prefix matches for the same input string.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+        length, their **Result**s are all processed in order of Trie traversal.
+        There's no other tie-breaking rule like alphabetical order among the
+        tied keys.
+
+**[Result][Unified Matcher: `OnMatch`]:** When a match occurs, the `OnMatch`
+dictates the outcome:
+
+-   It can contain an `Action` to be added to the results.
+-   It can contain a nested `Matcher`, triggering a further round of matching.
+    -   The tree is validated to not contain matchers with the tree depth
```
markdroth, line 410: This is handled at resource validation time, so it does not need to be considered at evaluation time.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+         greater than `16`. If this three depth is reached at runtime,
+        the tree evaluation is terminated, and considered an
+        [unsuccessful match][Unified Matcher: Filter Integration].
+-   `keep_matching` determines whether a successful match within an `OnMatch`
+    should be considered "terminal" for the current `Matcher` being evaluated.
+    It essentially answers the question: "After processing this `OnMatch`,
+    should the current matcher stop looking for more matches, or continue?"
+    -   If `keep_matching` is `false` (the usual case), finding this `OnMatch`
+       is terminal. The `Action` is added (or the nested `Matcher` is
+       evaluated), and the current matcher stops searching.
```
markdroth, line 420: As noted above, if the nested matcher does not return a successful match, then the current matcher will continue searching.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+    -   If `keep_matching` is `true`, the `Action` is added (or nested
+       `Matcher` evaluated), but the current matcher *continues* to look for
+       more matches. The overall process is not considered complete until an
+       `OnMatch` with `keep_matching` set to `false` is encountered.
+
+**Default/No Match:** Any `Matcher` can have a default `OnMatch` to use if none
+of its primary conditions or map lookups succeed. See details in
+[Unified Matcher: `Matcher`] and [Unified Matcher: Filter Integration].\
+**The Output:** A list of `Action`s accumulated from all triggered `OnMatch`
```
markdroth, line 429: As mentioned above, it needs to also return a bool indicating whether the match was successful.  Otherwise, the caller would not be able to tell the difference between a successful match and a match that found only actions with `keep_matching` set to true.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+produce the same form for either one.
+
+We will support the following `Matcher` fields:
+
+-   `matcher_type`: One of the following must be present:
+    -   [`matcher_list`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L134)
+        ([Unified Matcher: `MatcherList`]).
+    -   [`matcher_tree`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L137)
+        ([Unified Matcher: `MatcherTree`]).
+-   [`on_no_match`](https://github.com/cncf/xds/blob/2ac532fd44436293585084f8d94c6bdb17835af0/xds/type/matcher/v3/matcher.proto#L143):
```
markdroth, line 158: This is another field for which we need to clearly spell out its effect on matcher evaluation.  Specifically:

A matcher will use its `on_no_match` under either of the following conditions:
- The matcher found no matches in its list or tree.
- The matcher found matches in its list or tree, but all of those matches had either (a) nested matchers that did not return a successful match or (b) `keep_matching` set to true.

If `on_no_match` is unset, then the matcher will immediately return false.  Otherwise, `on_no_match` will be evaluated.  In this case, if the `on_no_match` contains a nested matcher, then the result of the nested matcher determines the result of the parent matcher.  And regardless of whether `on_no_match` contains a nested matcher or an action, if it has `keep_matching` set to true, the parent matcher will not be considered successful.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+**Evaluation (detailed):**
+
+1.  The `matcher_list` evaluates its matchers in order.
+2.  The first matcher is evaluated.
+    -   The `input` executes `HttpRequestHeaderMatchInput` extension:
+        -   The extension logic extracts the value of the `x-user-segment`
+            header from the Matcher Context.
+        -   The `input` returns `standard-user-1`.
+    -   The `StringMatcher` is evaluated (standard matcher):
+        -   The input is a string `standard-user-1`, which is the correct input
```
markdroth, line 499: The fact that it's the correct input type for this matcher doesn't matter during evaluation.  That should be checked at resource validation time.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+       `OnMatch` with `keep_matching` set to `false` is encountered.
+
+**Default/No Match:** Any `Matcher` can have a default `OnMatch` to use if none
+of its primary conditions or map lookups succeed. See details in
+[Unified Matcher: `Matcher`] and [Unified Matcher: Filter Integration].\
+**The Output:** A list of `Action`s accumulated from all triggered `OnMatch`
+results. Generally, only a single `Action` will be returned, unless
+`keep_matching` is enabled and multiple matches found.
+
+#### Unified Matcher: Evaluation Examples
```
markdroth, line 433: These examples are really useful.  I think we should also include some examples that show the edge cases I desribed above (e.g., a nested matcher returning unsuccessful, how that and `keep_matching` affect use of `on_no_match`, and how `on_no_match` influences the final result of the matcher).

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+### CEL Integration
+
+We will support request metadata matching via CEL expressions.
+
+CEL evaluation environment is a set of available variables and extension
+functions in a CEL program. We will match
+[Envoy CEL environment](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes)
+and CEL interpreter configuration.
+
+#### CEL: `CelExpression`
```
markdroth, line 681: As above, the "CEL:" prefix on all the sub-headings seems redundant.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+
+---
+
+### CEL Integration
+
+We will support request metadata matching via CEL expressions.
+
+CEL evaluation environment is a set of available variables and extension
+functions in a CEL program. We will match
+[Envoy CEL environment](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes)
```
markdroth, line 678: This statement reads like we will support everything documented for Envoy.  The subsections below clarify this, but to avoid confusion, I suggest changing the wording here to make it clear that we will support only certain subsets of what Envoy provides.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+-   **Performance Overhead**: Evaluating CEL expressions for every request
+    introduces computational overhead. While designed to be fast, it will be
+    slower than simple, hard-coded logic or basic string comparisons.
+
+## Implementation
+
+Will be implemented in C-core, Java, Go, and Node as part of either RLQS
+([A77]) or Composite Filter ([A103]), whichever happens to be implemented first
+in any given language. Role Based Access Control (RBAC, [A41]) currently does
+not support the Unified Matcher API in gRPC, though it is supported by Envoy,
```
markdroth, line 901: I think this last statement doesn't really belong in the "Implementation" section.  Instead, let's move it up to the "Rationale" section under a sub-heading called "Using Unified Matchers in Other Parts of xDS".

Note that RBAC is only one of many places where the unified matcher API can be used.  As I mentioned above, it can also be used for L4 filter chain selection in the gRPC server, and it can be used for route selection.  We can say that we are not going to proactively add support for unified matchers in any of those places, but we will structure the code in a reusable way so that we can easily add it in more places in the future if needed.

markdroth (commented):
Thanks for writing this up, Sergii!  Sorry it took me so long to do a review pass.

The content is very good.  Most of my comments are about readability and clarify of edge cases.

A106-xds-unified-matcher-and-cel.md
shivaspeaks, line 100: s/a combination ~of~ thereof/a combination thereof

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+    An optional string. May be ignored or used for testing/debugging.
+
+#### Unified Matcher: Evaluation Flow
+
+**Goal:** To produce a list of matching `Action`s.\
+**Matcher Context:** The input to the Matcher evaluation tree. Provided at
+runtime by the filter, contains the input data. May contain any other contextual
+information relevant to the filter. Implementation note: when setting an input,
+consider memory footprint. For example, instead of resolving all headers in
+advance, provide them in a lazy-loading wrapper.
```
shivaspeaks, line 367: Do you mean lazy-loading wrapper _with caching_ or _without caching_? IMO there are pros and cons for both, but without caching seems fine since the access is "usually" once because xDS matching logic tends toward simplicity.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+In the initial implementation only the `request` variable is supported in CEL
+expressions. We will adapt
+[Envoy's Request Attributes](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes#request-attributes)
+for gRPC.
+
+| Attribute           | Type                  | gRPC source                  | Envoy Description                                           |
+|---------------------|-----------------------|------------------------------|-------------------------------------------------------------|
+| `request.path`      | `string`              | Full method name             | The path portion of the URL.                                |
+| `request.url_path`  | `string`              | Same as `request.path`       | The path portion of the URL without the query string.       |
+| `request.host`      | `string`              | Authority                    | The host portion of the URL.                                |
```
ejona86, line 785: What authority are we using on client-side? We were mostly looking at RBAC when looking at the mapping, but server-side has an easily-accessible authority. We don't know the actual authority on client-side until we choose the transport... which is after client filters.

We can use the same authority as we use for xds routing, which ignores per-RPC overrides and any of the recent (SNI-related) authority override goo.

kannanjgithub: What about user-agent? This header is added by the gRPC transport that the filter doesn't know the value.

A106-xds-unified-matcher-and-cel.md
```diff
@@ -0,0 +1,902 @@
...
+| `timestamp(x)`, `timestamp.get*(x)`, `duration`    | Date/time functions.                                                                          |
+| `in`, `[]`                                         | Map/list indexing.                                                                            |
+| `has(m.x)`                                         | (macro) Returns true if the map `m` has the string `"x"` as a key.                            |
+| `int`, `uint`, `double`, `string`, `bytes`, `bool` | Conversions and identities.                                                                   |
+| `==`, `!=`, `>`, `<`, `<=`, `>=`                   | Comparisons.                                                                                  |
+| `or`, `&&`, `+`, `-`, `/`, `*`, `%`, `!`           | Basic functions.                                                                              |
+
+#### CEL: Supported Variables
+
+In the initial implementation only the `request` variable is supported in CEL
```
kannanjgithub, line 776: We need the response variable as well, for A93 ext-proc filter.

kannanjgithub: A93 has been updated to mention that using CEL evaluator is not necessary for the ext-proc filter.

PR #524, opened 2025-12-01
A107: TLS Private Key Offloading

gtcooke94:
gRFC for TLS Private Key Offloading

### DIFF
--- a/A107-tls-private-key-offloading.md
+++ b/A107-tls-private-key-offloading.md
@@ -1,0 +1,502 @@
+A107 - TLS Private Key Offloading
+----
+* Author: @gtcooke94
+* Approver: ejona86
+* Status: In Review
+* Implemented in: C++, Go
+* Last updated: 2025-12-01
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+This document outlines gRPC's plan to support TLS private key offloading, allowing a separate module (e.g., hardware) to handle private key signing during TLS handshakes for enhanced security and flexibility. 
+
+## Background
+
+A server possesses an identity certificate and a private key to identify itself during a handshake for authentication. During the TLS handshake, the server uses its private key to sign the transcript of the handshake, thereby validating ownership of said private key to the client. Currently, gRPC exclusively supports directly accessible, in-memory private keys; however, specialized private key storage solutions exist. TLS Private Key Offloading is a process by which an application performing a TLS handshake can use a private key to sign data *without* having direct access to the private key itself. This document will detail gRPC's prospective support for private key offloading.
+
+This feature will have the following requirements/assumptions:
+
+* TLS1.3 or TLS1.2 with modern ciphersuites that use ECDHE (Elliptic Curve Diffie-Hellman Ephemeral) for key exchange.  
+  * HTTP/2 mandates TLS\_ECDHE\_{RSA/ECDSA}\_WITH\_AES\_128\_GCM\_SHA256 as a baseline ciphersuite, explicitly prohibiting older, non-ephemeral Diffie-Hellman methods.  
+  * In non-ephemeral Diffe-Hellman key exchange, the private key could be used for other cryptographic operations (e.g. decryption with an RSA key). These will not be supported.
+
+
+### Related Proposals: 
+* TODO
+
+## Proposal
+
+The crypto libraries that each gRPC implementation uses have support for TLS Private Key Offloading. BoringSSL, for example, has the following interface:
+
+```c
+struct ssl_private_key_method_st {
+  enum ssl_private_key_result_t (*sign)(SSL *ssl, uint8_t *out, size_t *out_len,
+                                        size_t max_out,
+                                        uint16_t signature_algorithm,
+                                        const uint8_t *in, size_t in_len);
+
+  enum ssl_private_key_result_t (*decrypt)(SSL *ssl, uint8_t *out,
+                                           size_t *out_len, size_t max_out,
+                                           const uint8_t *in, size_t in_len);
+
+  enum ssl_private_key_result_t (*complete)(SSL *ssl, uint8_t *out,
+                                            size_t *out_len, size_t max_out);
+};
+
+```
+
+For TLS\>=1.2 with ECDHE, the `decrypt` method is not needed. We will provide an interface through which users can provide a signing function, with the base functionality being a function that takes in unsigned bytes and a signature algorithm and returns signed bytes. The `complete` function is an implementation detail for gRPC to handle.
+
+```
+string signed_bytes sign(string algorithm, string unsigned_bytes)
+```
+
+ 
+
+
+### Temporary environment variable protection
+
+This feature will be explicitly configured by users, thus no environment variable protection is needed. If a user does not configure TLS Private Key Offloading, it will not happen.
+
+## Rationale
+
+
+Private key offloading is designed to support signing outside of the existing process, for example in a hardware module or via an RPC \- thus this API should support asynchronous operations in languages where that is possible (Golang's crypto/tls does **not** support asynchronous private key signing). 
+
+We are largely restricted by the underlying security libraries in each language. In the following sections, each language's API will be discussed as they are dependent upon the SSL library interfaces. Further, each language has different expectations for the sign functions on whether raw bytes or a digest is expected.
+
+
+## Implementation
+
+### C-Core
+
+#### *APIs*
+
+BoringSSL provides an asynchronous API for private key signing, so we will provide an asynchronous, functional API using callbacks. The C-Core APIs used to configure this feature will rely on a new certificate provider implementation.
+
+We will create an `InMemoryCertificateProvider` that takes a set of root certs and a list of `PemKeyCertPair`. Further, these values can be manually updated by the user.  
+We will also decouple the root provider and the identity provider in [the tls\_credentials\_options](http://google3/third_party/grpc/src/core/credentials/transport/tls/grpc_tls_credentials_options.h;l=77;rcl=731487339). Users should be able to specify, for example, an `InMemoryCertificateProvider` that implements private key offloading and a `FileWatcherCertificateProvider` for the root certificates. We will mark the [coupled API](http://google3/third_party/grpc/include/grpcpp/security/tls_credentials_options.h;l=55-56;rcl=682352913) as deprecated.
+
+```c
+using PrivateKey = std::variant<absl::string_view, CustomPrivateKeySign>
+
+class PemKeyCertPair {
+ public:
+  PemKeyCertPair(PrivateKey private_key, absl::string_view cert_chain);
+}
+
+// Implements a provider that uses in-memory data that can be modified in a thread-safe manner.
+class InMemoryCertificateProvider : grpc_tls_certificate_provider {
+ public:
+  InMemoryCertificateProvider(std::string root_certificates, PemKeyCertPairList pem_key_cert_pairs);
+  
+  // thread safe updates
+  UpdateRoot(std::string root_certificates);
+  UpdateIdentity(PemKeyCertPairList pem_key_cert_pairs); 
+}
+
+struct grpc_tls_credentials_options {
+  // Deprecated. Use `set_root_certificate_provider` and 
+  // `set_identity_certificate_provider` instead. 
+  void set_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider>   certificate_provider);
+}
+
+// Sets the `grpc_tls_certificate_provider` to provide identity data.
+void set_identity_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider> certificate_provider);
+
+// Sets the `grpc_tls_certificate_provider` to provide root data.
+void set_root_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider> certificate_provider);
+```
+
+#### *Internals*
+
+1. User configures gRPC credentials with a PemKeyCertPair providing the CustomPrivateKeySign function.
+
+   [`PemKeyCertPair` already exists](http://google3/third_party/grpc/src/core/credentials/transport/tls/ssl_utils.h;l=153-167;rcl=786762336) but only accepts a `string` private key.  We will modify this to optionally accept a private key signing function.
+
+```c
+
+// A user's implementation MUST invoke `done_callback` with the signed bytes. This will let gRPC take control when the async operation is complete.
+// MUST not block
+// MUST support concurrent calls
+using CustomPrivateKeySign = absl::AnyInvocable<void(
+    absl::string_view data_to_sign,
+    SignatureAlgorithm signature_algorithm,
+    absl::AnyInvocable<void(absl::StatusOr<std::string> signed_data)> done_callback
+)>;
+
+
+struct PemKeyCertPair {
+  std::string private_key;  // PEM-encoded private key.
+  std::string cert_chain;   // PEM-encoded certificate chain.
+  // Asynchronous private key signing function. Mutually exclusive with private_key.
+  CustomPrivateKeySign private_key_sign;
+  // Associated ctors, etc
+}
+
+
+
+// Enum class representing TLS signature algorithm identifiers from BoringSSL.
+// The values correspond to the SSL_SIGN_* macros in <openssl/ssl.h>.
+enum class SignatureAlgorithm : uint16_t {
+  kRsaPkcs1Sha256 = 0x0401,           // SSL_SIGN_RSA_PKCS1_SHA256
+  kRsaPkcs1Sha384 = 0x0501,           // SSL_SIGN_RSA_PKCS1_SHA384
+  kRsaPkcs1Sha512 = 0x0601,           // SSL_SIGN_RSA_PKCS1_SHA512
+  kEcdsaSecp256r1Sha256 = 0x0403,    // SSL_SIGN_ECDSA_SECP256R1_SHA256
+  kEcdsaSecp384r1Sha384 = 0x0503,    // SSL_SIGN_ECDSA_SECP384R1_SHA384
+  kEcdsaSecp521r1Sha512 = 0x0603,    // SSL_SIGN_ECDSA_SECP521R1_SHA512
+  kRsaPssRsaeSha256 = 0x0804,         // SSL_SIGN_RSA_PSS_RSAE_SHA256
+  kRsaPssRsaeSha384 = 0x0805,         // SSL_SIGN_RSA_PSS_RSAE_SHA384
+  kRsaPssRsaeSha512 = 0x0806,         // SSL_SIGN_RSA_PSS_RSAE_SHA512
+};
+
+
+
+```
+
+2. gRPC's TLS stack configures the SSL\_CTX with the custom SSL\_PRIVATE\_KEY\_METHOD.
+
+   We will add a type, `TlsPrivateKeyOffloadContext` to manage the state of this operation and store it as `ex_data` on the SSL object.
+
+```c
+
+// State associated with an SSL object for async private key operations.
+struct TlsPrivateKeyOffloadContext {
+  CustomPrivateKeySign private_key_sign;
+  absl::StatusOr<std::string> signed_bytes;
+
+  // TSI handshake state needed to resume.
+  tsi_handshaker* handshaker;
+  tsi_handshaker_on_next_done_cb notify_cb;
+}
+
+
+
+// Callback function to be invoked when the user's async sign operation is complete.
+// This function is curried with 'ctx' using absl::bind_front.
+static void TlsOffloadSignDoneCallback(TlsPrivateKeyOffloadContext* ctx,
+                                         absl::StatusOr<std::string> signed_data) {
+
+  if (signed_data.ok()) {
+    ctx->signed_data = std::move(signed_data);
+  }
+  //... handle other cases
+
+  // Notify the TSI layer to re-enter the handshake.
+  // This call is thread-safe as per TSI requirements for the callback.
+  if (ctx->notify_cb) {
+    ctx->notify_cb(ctx->handshaker, ctx->notify_user_data, TSI_OK);
+  }
+}
+
+```
+
+ 
+
+We then will implement [BoringSSL's ssl\_private\_key\_method\_st](http://google3/third_party/openssl/boringssl/src/include/openssl/ssl.h;l=1352-1356;rcl=814357088) with the user's signing function.
+
+```c
+
+static enum ssl_private_key_result_t TlsPrivateKeySignWrapper(
+    SSL* ssl, uint8_t* out, size_t* out_len, size_t max_out,
+    uint16_t signature_algorithm, const uint8_t* in, size_t in_len) {
+  // Get and fill the TlsPrivateKeyOffloadContext.
+  TlsPrivateKeyOffloadContext* ctx = static_cast<TlsPrivateKeyOffloadContext*>(
+      SSL_get_ex_data(...)
+  // Fill important info
+  // ... 
+  // Create the completion callback by binding the current context.
+  auto done_callback =
+      absl::bind_front(TlsOffloadSignDoneCallback, ctx);
+
+  // Call the user's async sign function
+  // The contract with the user is that they MUST invoke the callback when complete in their implementation, and their impl MUST not block. 
+  ctx->private_key_sign(data_to_sign, signature_algorithm,
+                       std::move(done_callback));
+
+  return ssl_private_key_retry;
+}
+
+static enum ssl_private_key_result_t TlsPrivateKeyOffloadComplete(
+    SSL* ssl, uint8_t* out, size_t* out_len, size_t max_out) {
+  // Get TlsPrivateKeyOffloadContext.
+  TlsPrivateKeyOffloadContext* ctx = static_cast<TlsPrivateKeyOffloadContext*>(
+      SSL_get_ex_data(...);
+  // Various checks
+  // ...
+  // Important bit is moving the signed data where it needs to go 
+  memcpy(out, ctx->signed_data.data(), ctx->signed_data.length());
+  *out_len = ctx->signed_data.length();
+  // Tell BoringSSL we're done
+  return ssl_private_key_success;
+}
+
+static const SSL_PRIVATE_KEY_METHOD TlsOffloadPrivateKeyMethod = {
+    TlsPrivateKeySignWrapper,
+    nullptr,  // decrypt not implemented for this use case
+    TlsPrivateKeyOffloadComplete};
+
+```
+
+3. When a handshake starts, tsi\_handshaker\_next is called.  
+4. The TSI layer creates and configures an SSL object, attaching the TlsPrivateKeyOffloadContext (including the user's function and TSI callback info) as ex\_data.  
+5. tsi\_handshaker\_next calls SslDoHandshake (in ssl\_util.cc), which calls SSL\_do\_handshake.  
+6. BoringSSL eventually requires the signature, calls our SSL\_PRIVATE\_KEY\_METHOD sign impl.  
+7. TlsPrivateKeySignWrapper invokes the user's private\_key\_sign, providing the `TlsOffloadSignDoneCallback`. It returns ssl\_private\_key\_retry.  
+8. SSL\_do\_handshake returns, SSL\_get\_error is SSL\_ERROR\_WANT\_PRIVATE\_KEY\_OPERATION.  
+9. Seeing SSL\_ERROR\_WANT\_PRIVATE\_KEY\_OPERATION, SslDoHandshake returns TSI\_ASYNC . tsi\_handshaker\_next returns TSI\_ASYNC. [gRPC core yields control.](http://google3/third_party/grpc/src/core/handshaker/security/security_handshaker.cc;l=426-431;rcl=807452496)  
+10. The user's async operation completes, invoking the `TlsOffloadSignDoneCallback`.  
+11. The callback stores the result in TlsPrivateKeyOffloadContext and calls the notify\_cb (which is tsi\_handshaker\_on\_next\_done\_cb).
+
+    This callback is part of TSI and is [provided by the caller](http://google3/third_party/grpc/src/core/handshaker/security/security_handshaker.cc;l=422-425;rcl=807452496) of `tsi_handshaker_next`. In our case in gRPC, this is [the `OnHandshakeNextDoneGrpcWrapper` function](http://google3/third_party/grpc/src/core/handshaker/security/security_handshaker.cc;l=402;rcl=807452496) .
+
+    This returns control to gRPC Core, resuming the handshake.
+
+12. gRPC core calls tsi\_handshaker\_next again.  
+13. This leads to SSL\_do\_handshake being called again.  
+14. BoringSSL, seeing the pending operation, now calls TlsPrivateKeyOffloadComplete.  
+15. TlsPrivateKeyOffloadComplete retrieves the signed\_data from TlsPrivateKeyOffloadContext and provides it to BoringSSL.  
+16. The handshake continues with the signed data.
+
+### C++
+
+For the C++ APIs, we will try to alias as much of the C-Core API as we can. C++ has its own `IdentityKeyCertPair` that converts to `PemKeyCertPair` from C-Core, so we will add similar code to that struct to support a signing function.  
+The `TlsCredentialsOptions` will be similarly updated to differentiate between an identity provider and a root provider.  
+We will also create a C++ wrapper around the `InMemoryCertificateProvider` in the same way the other provider implementations are wrapped.
+
+```c
+
+using grpc_core::SignatureAlgorithm;
+using grpc_core::CustomPrivateKeySignCallback;
+using grpc_core::PrivateKey;
+
+// This is the C++ API for PemKeyCertPair
+struct IdentityKeyCertPair {
+  PrivateKey private_key;
+  std::string certificate_chain;
+  // Associated ctors, etc
+}
+
+class TlsCredentialsOptions {
+  // Deprecated. Use `set_identity_certificate_provider` and 
+  // `set_root_certificate_provider` instead.
+  void set_certificate_provider(
+      std::shared_ptr<CertificateProviderInterface> certificate_provider);
+
+  void set_identity_certificate_provider(
+      std::shared_ptr<CertificateProviderInterface> certificate_provider);
+
+  void set_root_certificate_provider(
+      std::shared_ptr<CertificateProviderInterface> certificate_provider);
+}
+
+// Basic wrapping of C-Core
+class InMemoryCertificateProvider : public CertificateProviderInterface {
+
+  static std::shared_ptr<InMemoryCertificateProvider> Create(std::string root_cert, std::vector<IdentityKeyCertPair> key_cert_pairs);
+
+  // thread safe updates
+  UpdateRoot(std::string root_certificates);
+  UpdateIdentity(absl::Span<IdentityKeyCertPair> key_cert_pairs); 
+}
+
+// To use, set the provider on the TlsCredentialsOptions
+```
+
+### Python
+
+Python wraps the C-Core implementation. Currently, Python's security configuration wraps the `SslCredentials` instead of the `TlsCredentials`. We will update gRPC-Python internally to use `TlsCredentials` (PR [\#40878](https://github.com/grpc/grpc/pull/40878)). Then, the task of supporting this feature in Python is similar to C++. We will wrap the new types and split the certificate provider on `tls_credentials_options` to support different root and identity providers.
+
+The current Python API is very simple:
+
+```py
+PrivateKeySignDoneCallback = Callable[[Optional[bytes], bool], None]
+# Note: SignatureAlgorithm corresponds to C-core's enum class SignatureAlgorithm.
+CustomPrivateKeySign = Callable[[bytes, SignatureAlgorithm, PrivateKeySignDoneCallback], None]
+
+
+def ssl_channel_credentials_with_custom_signer(
+    *,
+    private_key_sign_fn: CustomPrivateKeySign,
+    root_certificates: Optional[bytes] = None,
+    certificate_chain: Optional[bytes] = None,
+) -> ChannelCredentials:
+    """Creates a ChannelCredentials for use with an SSL-enabled Channel.
+
+    Args:
+
+      private_key_sign_fn: A function implementing
+        `CustomPrivateKeySign`.
+        `CustomPrivateKeySign` MUST NOT block, MUST be thread-safe, and MUST call
+        `PrivateKeySignDoneCallback` on completion.
+      root_certificates: The PEM-encoded root certificates as a byte string,
+        or None to retrieve them from a default location chosen by gRPC
+        runtime.
+      certificate_chain: The PEM-encoded certificate chain as a byte string
+        to use or None if no certificate chain should be used.
+
+    Returns:
+      A ChannelCredentials for use with an SSL-enabled Channel.
+    """
+
+
+```
+
+We won't significantly refactor the Python API surface \- instead we will allow the `private_key` input to be a signing function.
+
+The most complex piece of this is the implementation \- C-Core/Cython/Python must handle the calling of the user provided Python signing function from C which must invoke a callback that is passed to it. This will involve creating bridge types between the Python user sign function and the expected `absl::AnyInvocable` as well as bridging the callback that is passed to the user sign function while managing the GIL and asynchronous nature of the signing. This is technically feasible with Cython. A proof-of-concept of this structure [is written here.](https://source.corp.google.com/piper///depot/google3/experimental/users/gregorycooke/python_cpp_wrapping/) 
+
+```py
+# Example Usage
+
+# example_signer is a basic implementation CustomPrivateKeySign
+def example_signer(
+    unsigned_data: bytes,
+    algorithm: SignatureAlgorithm,
+    on_done: PrivateKeySignDoneCallback,
+) -> None:
+    # Manually sign the bytes.
+    signed_bytes = ...
+    if signed_bytes:
+        on_done(signed_bytes, True)
+    else:
+        on_done(None, False)
+
+
+
+
+# Now the user is in their application configuring gRPC 
+# Create creds with the custom signer
+creds = ssl_channel_credentials_with_custom_signer(<some_root>, example_signer, <some_chain>)
+
+```
+
+### Go
+
+Golang's `crypto/tls` package does not directly support asynchronous operations for the `Signer` interface, but this is not a problem due to the structure of goroutines. The API will simply look like implementing the `crypto/tls` `Signer` API. In this library, the [`Certificate` type's](http://google3/third_party/go/gc/src/crypto/tls/common.go;l=1559;rcl=815505302)  `PrivateKey` is a [`crypto.PrivateKey`](https://pkg.go.dev/crypto#PrivateKey), which is `Any`, but it must implement the [`Signer` interface](https://pkg.go.dev/crypto#Signer). Notably, in Golang, the user's function should expect a *digest* of the transcript as input.
+
+```go
+// From crypto.tls
+type Signer interface {
+	// Public returns the public key corresponding to the opaque,
+	// private key.
+	Public() PublicKey
+
+	// Sign signs digest with the private key, possibly using entropy from
+	// rand. For an RSA key, the resulting signature should be either a
+	// PKCS #1 v1.5 or PSS signature (as indicated by opts). For an (EC)DSA
+	// key, it should be a DER-serialised, ASN.1 signature structure.
+	//
+	// Hash implements the SignerOpts interface and, in most cases, one can
+	// simply pass in the hash function used as opts. Sign may also attempt
+	// to type assert opts to other types in order to obtain algorithm
+	// specific values. See the documentation in each package for details.
+	//
+	// Note that when a signature of a hash of a larger message is needed,
+	// the caller is responsible for hashing the larger message and passing
+	// the hash (as digest) and the hash function (as opts) to Sign.
+	Sign(rand io.Reader, digest []byte, opts SignerOpts) (signature []byte, err error)
+}
+
+cert = tls.Certificate{
+Certificate: derChain, //[][]byte
+PrivateKey:  PrivateKeySigner, //crypto.PrivateKey
+}
+```
+
+This is the interface that a user will implement and pass to gRPC-Go via a `tls.Certificate`.  
+We will create a similar `InMemoryCertProvider` that takes and provides a `tls.Certificate`. This will live in the advancedTLS package.
+
+```go
+
+// InMemoryCertProvider is a certprovider.Provider implementation that holds and serves a
+// single, predefined tls.Certificate instance.
+type InMemoryCertProvider struct {
+    tlsCert *tls.Certificate
+}
+
+func NewInMemoryCertProvider(cert *tls.Certificate) (*InMemoryCertProvider, error) {
+    // Checks 
+    return &InMemoryCertProvider{tlsCert: *cert}, nil
+}
+
+// Implements the `KeyMaterial` interface
+// Thread safe access
+func (i *InMemoryCertProvider) KeyMaterial(_ context.Context) (*KeyMaterial, error) {...}
+
+// Allow the user to update with new data
+// Thread safe updates
+func (provider *InMemoryCertProvider) UpdateCertificate(tlsCert *tls.Certificate) {...}
+
+```
+
+This will be piped around by the existing infrastructure to configure the `crypto/tls` library with no additional changes needed.
+
+```go
+// Example Usage
+
+// ExampleSigner is a basic implementation of crypto.Signer
+type ExampleSigner struct {}
+
+// Public returns the public key corresponding to the opaque, private key.
+func (s *ExampleSigner) Public() crypto.PublicKey {
+	return <some key>
+}
+
+// Sign signs the digest with the private key.
+func (s *ExampleSigner) Sign(r io.Reader, digest []byte, opts crypto.SignerOpts) ([]byte, error) {
+       signed_bytes := <do stuff with digest>
+	return signed_bytes, nil
+}
+
+// Now the user is in their application configuring gRPC 
+// Create the custom signer  
+customSigner, err := NewExampleSigner()
+
+// Create the tls.Certificate
+tlsCert := &tls.Certificate{
+    Certificate: <pem-certificate>
+    Signer: customSigner
+}
+
+// Create the InMemoryCertProvider and configure gRPC with it
+provider, err := NewInMemoryCertProvider(tlsCert)
+serverOpts := &advancedtls.Options{
+	IdentityOptions: advancedtls.IdentityCertificateOptions{
+		IdentityProvider: provider,
+	},
+}
+
+// Down the line update
+provider.UpdateCertificate(&tls.Certificate(<some updated thing>))
+
+```
+
+### Java
+
+After initial investigation, we will not pursue implementing this feature in Java right now. Due to the different designs of the options of underlying security library APIs, there is no cohesive convenience API that we could add to gRPC-Java. A user can still implement this themselves and use the `AdvancedTlsKeyProvider` APIs to do whatever they want with the `PrivateKey` interface. Particularly, a user could create their own signature provider and globally register it to be used. gRPC-Java, as a library, will not do global registration.
+
+# **Observability**
+
+The following non-per-call metric and labels will be added. These will allow a user insight into the offloaded operations and will be an aid in debugging failures.
+
+| Name | Disposition | Description |
+| :---- | :---- | :---- |
+| `grpc.target` | required | Indicates the target of the gRPC channel for which this handshake is occurring |
+| `grpc.security.handshaker.offloaded_private_key_sign.algorithm` | optional | The signature algorithm used to sign |
+| `grpc.security.handshaker.offloaded_private_key_sign.status` | optional | The status code return for the private key sign function |
+
+| Metric | Type | Unit | Labels | Description |
+| :---- | :---- | :---- | :---- | :---- |
+| `grpc.security.handshaker.offloaded_private_key_sign_duration` | Histogram | float64/double s | `grpc.target, grpc.security.handshaker.offloaded_private_key_sign.algorithm, grpc.security.handshaker.offloaded_private_key_sign.status` | How long the offloaded private key signing took |
+
+For the latency metric, we will use the buckets as defined in [gRFC A66](https://github.com/grpc/proposal/blob/fcabdfdbd50b3c088f5a5c2bf925755781ec076e/A66-otel-stats.md?plain=1#L360-L369).
+
+The `grpc.security.handshaker.offloaded_private_key_sign.algorithm` label will contain both a name and key length.
+
+
+## Open issues (if applicable)
+
+* The C APIs may be built around a class rather than straight passing `absl::AnyInvocable` or `std::function`.
+* The exact additions vs. breakages to the C-Core APIs and C++ APIs will be decided during PR reviews. Some of the APIs we are modifying are experimental-but-old, so we want to be careful when making breaking changes.
--- a/A107-tls-private-key-offloading.md
+++ b/A107-tls-private-key-offloading.md
@@ -70,7 +70,15 @@ signature algorithm and returns signed bytes.
 string signed_bytes sign(string algorithm, string unsigned_bytes)
 ```
 
-### C-Core
+### C-Core / C++
+
+There are three distinct changes:
+1. Creating an InMemory Certificate Provider
+     * Must be done to cleanly accept the PrivateKeySigner instead of just a static PEM key or a file path.
+2. Splitting Certificate Providers into Root and Identity Providers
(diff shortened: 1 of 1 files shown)


A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,502 @@
...
+	//
+	// Hash implements the SignerOpts interface and, in most cases, one can
+	// simply pass in the hash function used as opts. Sign may also attempt
+	// to type assert opts to other types in order to obtain algorithm
+	// specific values. See the documentation in each package for details.
+	//
+	// Note that when a signature of a hash of a larger message is needed,
+	// the caller is responsible for hashing the larger message and passing
+	// the hash (as digest) and the hash function (as opts) to Sign.
+	Sign(rand io.Reader, digest []byte, opts SignerOpts) (signature []byte, err error)
```
salrashid123, line 420: there's an obscure usecase for supporting `MessageSigner` here for restricted TPM keys

* https://github.com/golang/go/issues/75656

gtcooke94: Thanks for the heads up - I took a look at the change and I think someone implementing `crypto.MessageSigner`, and I think it should "just work" since crypto/tls supports a `crypto.MessageSigner` as a `PrivateKey` in the `tls.Certificate`.

So I think gRPC will pretty blindly pass this through while configuring `tls.Config`. If the user implements a `crypto.MessageSigner` with a `SignMessage` it'll use that

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,502 @@
...
+
+
+```
+
+We won't significantly refactor the Python API surface \- instead we will allow the `private_key` input to be a signing function.
+
+The most complex piece of this is the implementation \- C-Core/Cython/Python must handle the calling of the user provided Python signing function from C which must invoke a callback that is passed to it. This will involve creating bridge types between the Python user sign function and the expected `absl::AnyInvocable` as well as bridging the callback that is passed to the user sign function while managing the GIL and asynchronous nature of the signing. This is technically feasible with Cython. A proof-of-concept of this structure [is written here.](https://source.corp.google.com/piper///depot/google3/experimental/users/gregorycooke/python_cpp_wrapping/) 
+
+```py
+# Example Usage
```
sergiitk: I'd like to restart the conversation we had elsewhere.

Would it make sense to abstract this call away? Make the signer just return the bytes, and the callback will be handled by our custom function? With this approach, we'll avoid the problem where users forgot to call the callback (f.e. by not wrapping their method into `try ... finally`).

The main point is to abstract away internal implementation details as much as possible.

```py
# Internal wrapper
def _grpc_invoke_custom_private_key_sign(
    private_key_sign_fn: CustomPrivateKeySign,
    unsigned_data: bytes,
    algorithm: SignatureAlgorithm,
    on_done: PrivateKeySignDoneCallback,
) -> None:
    try:
        signed_bytes = private_key_sign_fn(unsigned_data, algorithm)
        if signed_bytes:
            on_done(signed_bytes, True)
        else:
            on_done(None, False)
    except Exception as e:
        logging.warning("Exception during custom private key sign: %s", e)
        on_done(None, False)


# User's implementation example
def example_signer(
    unsigned_data: bytes, algorithm: SignatureAlgorithm
) -> Optional[bytes]:
    # Manually sign the bytes.
    signed_bytes = ...
    return signed_bytes
```

sergiitk: Spoke with @matthewstevenson88 - he explained some of the async patterns our users might apply when `CustomPrivateKeySign`. One of them is creating a thread that will do the computation, passing `on_done` callback to the thread, while immediately returning from the `CustomPrivateKeySign`. This is a good argument for not introducing an internal wrapper.

I need think about this some more, and consider a few nuances: 

- Whether a thread will actually be useful for offloading CPU-bound tasks - because if GIL is stuck on this thread waiting for the signing to complete, there won't be any parallelization/concurrency (unless it  calls into a Python C extension that disables GIL)
- Whether we still can use the wrapper approach, but with a different interface, like using the [Future](https://docs.python.org/3/library/concurrent.futures.html#future-objects) interface. `Future` should make it easier for the user to work with various `PoolExecutor`s.
- Whether we should also allow an `async` method signature (make `private_key_sign_fn` a `Union[CustomPrivateKeySign, CustomPrivateKeySignAsync]`) - which may be required if it's common for the HW offloading libraries the do the signing to use async/await pattern ([asyncio](https://docs.python.org/3/library/asyncio.html) python).

sergiitk: For the reference, 

> GIL and performance considerations
>
> Unlike the [multiprocessing](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing) module, which uses separate processes to bypass the [global interpreter lock](https://docs.python.org/3/glossary.html#term-global-interpreter-lock) (GIL), the threading module operates within a single process, meaning that all threads share the same memory space. However, the GIL limits the performance gains of threading when it comes to CPU-bound tasks, as only one thread can execute Python bytecode at a time. Despite this, threads remain a useful tool for achieving concurrency in many scenarios.
>
> — https://docs.python.org/3/library/threading.html#gil-and-performance-considerations

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,502 @@
...
+
+  // thread safe updates
+  UpdateRoot(std::string root_certificates);
+  UpdateIdentity(absl::Span<IdentityKeyCertPair> key_cert_pairs); 
+}
+
+// To use, set the provider on the TlsCredentialsOptions
+```
+
+### Python
```
sergiitk, line 245: We should mention that `gevent` is not supported, both in the gRFC and the `ssl_channel_credentials_with_custom_signer` docstring.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,502 @@
...
+        to use or None if no certificate chain should be used.
+
+    Returns:
+      A ChannelCredentials for use with an SSL-enabled Channel.
+    """
+
+
+```
+
+We won't significantly refactor the Python API surface \- instead we will allow the `private_key` input to be a signing function.
```
sergiitk: I don't think this in relevant in the last iteration - we provided a new method instead of modifying the existing one to support `private_key`

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,502 @@
...
+        on_done(signed_bytes, True)
+    else:
+        on_done(None, False)
+
+
+
+
+# Now the user is in their application configuring gRPC 
+# Create creds with the custom signer
+creds = ssl_channel_credentials_with_custom_signer(<some_root>, example_signer, <some_chain>)
```
sergiitk: note - we require arguments to be passed as keyword arguments now. 
```suggestion
creds = ssl_channel_credentials_with_custom_signer(
    private_key_sign_fn=example_signer,
    root_certificates=b"<some_root>",
    certificate_chain=b"<some_chain>",
)
```

A107-tls-private-key-offloading.md
salrashid123: yeah, if anything implements `messagesigner`, it'd get automatically pickedup and used.

as mentioned internally, that the [advancedtls](https://pkg.go.dev/google.golang.org/grpc/security/advancedtls) pakage which AFAIK, is only in go, you can offload TLS now

* https://github.com/salrashid123/grpc-cert-provider

A107-tls-private-key-offloading.md
gtcooke94: Have you had the chance to consider these options? We need to move relatively quickly on this, and that prototype that I made earlier in the year seemed viable? Can we keep this and disable the GIL?

Can we allow for `Future` or `asyncio` with the cython integration?

I'll lean on you for these decisions since you are the Python expert

gtcooke94: I've been toying around with how we wrap this more

The C++ layers knows it is wrapping Python. What if we just make the C++ wrapper layer responsible for the async. 

i.e.
```
... implementing the interface ...
class PrivateKeySignerPyWrapper : PrivateKeySigner {

... details, assume self_->sign() is calling a cython method that's holding a python callable and is going to call into Python, here's where we _actually_ call it  ...
  
  thread::Detach({}, [s = std::string(input), self = shared_from_this(),
                       context] {
    PyGILState_STATE gstate;
    gstate = PyGILState_Ensure();
    // This is calling into Python
    self->sign_(s, CompletionCallbackForPy, context, self->user_data_);
    PyGILState_Release(gstate);
  });

}
```

We could go even further in this case, the python sign function doesn't need to know about a callback, calling the callback, or handling async stuff - it just simply returns signed data sync.

```
  thread::Detach({}, [s = std::string(input), self = shared_from_this(),
                       context] {
    PyGILState_STATE gstate;
    gstate = PyGILState_Ensure();
    // This is calling into Python
    std::string signed_data = self->sign_(s, context, self->user_data_);
    PyGILState_Release(gstate);
    // C++ wrapping layer is responsible for the async and calling the callback
    CompletionCallback(signed_data);
  });
```

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
+A107 - TLS Private Key Offloading
+----
+* Author: @gtcooke94
+* Approver: ejona86
+* Status: C++ implemented
+* Implemented in: C++, Go
+* Last updated: 2026-03-18
+* Discussion at: https://groups.google.com/g/grpc-io/c/N02jVxPd_4Y/m/n34PWOyKBgAJ?e=48417069## Abstract
```
markdroth: Need a couple of newlines before `## Abstract`.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+We are largely restricted by the underlying security libraries in each language.
+In the following sections, each language's API will be discussed as they are
+dependent upon the SSL library interfaces. Further, each language has different
+expectations for the sign functions on whether raw bytes or a digest is
+expected.
+
+
+## Implementation
+
+### C-Core
```
markdroth: All of these language-specific details belong in the "Proposal" section, not the "Implementation" section.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+}
+
+// Implements a provider that uses in-memory data that can be modified in a thread-safe manner.
+class InMemoryCertificateProvider : grpc_tls_certificate_provider {
+ public:
+  InMemoryCertificateProvider(std::string root_certificates, PemKeyCertPairList pem_key_cert_pairs);
+  
+  // thread safe updates
+  UpdateRoot(std::string root_certificates);
+  UpdateIdentity(PemKeyCertPairList pem_key_cert_pairs);
```
markdroth: Need to list the return type for these methods.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+  // Deprecated. Use `set_root_certificate_provider` and 
+  // `set_identity_certificate_provider` instead. 
+  void set_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider>   certificate_provider);
+}
+
+// Sets the `grpc_tls_certificate_provider` to provide identity data.
+void set_identity_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider> certificate_provider);
+
+// Sets the `grpc_tls_certificate_provider` to provide root data.
+void set_root_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider> certificate_provider);
```
markdroth: These methods should be inside of the `grpc_tls_credentials_options` struct.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+### C-Core
+
+#### *APIs*
+
+BoringSSL provides an asynchronous API for private key signing, so we will
+provide an asynchronous, cancellable API using callbacks. The C-Core APIs used
+to configure this feature will rely on a new certificate provider
+implementation.
+
+We will create an `InMemoryCertificateProvider` that takes a set of root certs
```
markdroth, line 82: There are really 3 distinct API changes here, and I think it would be helpful to discuss each of them separately:

1. Adding `InMemoryCertificateProvider`, and deprecating `StaticDataCertificateProvider`.
2. Allowing a separate cert provider for root and identity certs.
3. Adding the new `PrivateKeySigner` API.

I think we should first list the changes, then talk about each one in detail in its own subsection.  For each one, we should describe why we're making the change and then document the API we're introducing.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+}
+
+// Sets the `grpc_tls_certificate_provider` to provide identity data.
+void set_identity_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider> certificate_provider);
+
+// Sets the `grpc_tls_certificate_provider` to provide root data.
+void set_root_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider> certificate_provider);
+```
+
+#### *Internals*
```
markdroth: I don't think we need most of these implementation details here.  I think it's sufficient to say that the PrivateKeySigner will be integrated into the TLS handshake inside of the TLS TSI implementation.

The only thing from this section that I think we want to retain is the `PrivateKeySigner` API.  And we should include discussion of how it will work for both sync and async flows.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+will detail gRPC's prospective support for private key offloading.
+
+This feature will have the following requirements/assumptions:
+
+* TLS1.3 or TLS1.2 with modern ciphersuites that use ECDHE (Elliptic Curve Diffie-Hellman Ephemeral) for key exchange.  
+  * HTTP/2 mandates TLS\_ECDHE\_{RSA/ECDSA}\_WITH\_AES\_128\_GCM\_SHA256 as a baseline ciphersuite, explicitly prohibiting older, non-ephemeral Diffie-Hellman methods.  
+  * In non-ephemeral Diffe-Hellman key exchange, the private key could be used for other cryptographic operations (e.g. decryption with an RSA key). These will not be supported.
+
+
+### Related Proposals:
```
markdroth: This should reference gRFCs A66 and A79.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+globally register it to be used. gRPC-Java, as a library, will not do global
+registration.
+
+# **Observability**
+
+The following non-per-call metric and labels will be added. These will allow a
+user insight into the offloaded operations and will be an aid in debugging
+failures.
+
+| Name | Disposition | Description |
```
markdroth: It's not clear from the text that this table lists labels.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+security library APIs, there is no cohesive convenience API that we could add to
+gRPC-Java. A user can still implement this themselves and use the
+`AdvancedTlsKeyProvider` APIs to do whatever they want with the `PrivateKey`
+interface. Particularly, a user could create their own signature provider and
+globally register it to be used. gRPC-Java, as a library, will not do global
+registration.
+
+# **Observability**
+
+The following non-per-call metric and labels will be added. These will allow a
```
markdroth: Please reference gRFC A79 here.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+
+| Name | Disposition | Description |
+| :---- | :---- | :---- |
+| `grpc.target` | required | Indicates the target of the gRPC channel for which this handshake is occurring |
+| `grpc.security.handshaker.offloaded_private_key_sign.algorithm` | optional | The signature algorithm used to sign |
+| `grpc.security.handshaker.offloaded_private_key_sign.status` | optional | The status code return for the private key sign function |
+
+| Metric | Type | Unit | Labels | Description |
+| :---- | :---- | :---- | :---- | :---- |
+| `grpc.security.handshaker.offloaded_private_key_sign_duration` | Histogram | float64/double s | `grpc.target, grpc.security.handshaker.offloaded_private_key_sign.algorithm, grpc.security.handshaker.offloaded_private_key_sign.status` | How long the offloaded private key signing took |
```
markdroth: I don't think we actually implemented this metric in C-core.  Do we still need to do that?

Implementing this in C-core raises a layering question.  The TLS TSI handshaker is the thing that has this data to export, but the non-per-call metrics APIs are gRPC-specific APIs, which means that the TSI implementation would need to use gRPC-specific APIs.  Won't that make it harder for this TSI implementation to be used outside of gRPC?  Are we okay with that?

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+
+For the latency metric, we will use the buckets as defined in [gRFC
+A66](https://github.com/grpc/proposal/blob/fcabdfdbd50b3c088f5a5c2bf925755781ec076e/A66-otel-stats.md?plain=1#L360-L369).
+
+The `grpc.security.handshaker.offloaded_private_key_sign.algorithm` label will contain both a name and key length.
```
markdroth: This comment should go in the table above, where we describe that label.

Also, please provide examples of supported values.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,590 @@
...
+
+The following non-per-call metric and labels will be added. These will allow a
+user insight into the offloaded operations and will be an aid in debugging
+failures.
+
+| Name | Disposition | Description |
+| :---- | :---- | :---- |
+| `grpc.target` | required | Indicates the target of the gRPC channel for which this handshake is occurring |
+| `grpc.security.handshaker.offloaded_private_key_sign.algorithm` | optional | The signature algorithm used to sign |
+| `grpc.security.handshaker.offloaded_private_key_sign.status` | optional | The status code return for the private key sign function |
```
markdroth: IIRC, there are limitations in OT that make it desirable to keep label names short, and these are both extremely long.  I suggest something like `grpc.tls.private_key_sign_algorithm` for the first one.  The second one can probably just reuse the `grpc.status` label from A66.

markdroth (commented):
Thanks for writing this up!

A107-tls-private-key-offloading.md
gtcooke94: Ah my bad

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done, and added the PRs to the implementation section

A107-tls-private-key-offloading.md
gtcooke94: Done, PTAL

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Swapped fully to being the C++ API class and it's semantics.

A107-tls-private-key-offloading.md
gtcooke94: Done, PTAL

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Changed top left to "Label Name"

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: We have no security metrics yet (that's happening ~now, we are working on the concept review) - this will be included in that work.

The layering question is one of the prime questions for our concept and design.

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
```diff
@@ -31,6 +33,11 @@ This feature will have the following requirements/assumptions:
 
 
 ### Related Proposals: 
+* [A66]
+* [A79]
```
markdroth: Please change these lines to:

```
* [A66: OpenTelemetry Metrics][A66]
* [A79: Non-per-call Metrics Architecture][A79]
```

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
-using PrivateKey = std::variant<absl::string_view, CustomPrivateKeySign>
-
-class PemKeyCertPair {
- public:
-  PemKeyCertPair(PrivateKey private_key, absl::string_view cert_chain);
-}
+The updates will return an `absl::Status` to the caller, and further a
+`ValidateCredentials` API is provided.
 
+```c
```
markdroth: This should say `c++` instead of just `c`.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
-tls\_credentials\_options](http://google3/third_party/grpc/src/core/credentials/transport/tls/grpc_tls_credentials_options.h;l=77;rcl=731487339).
-Users should be able to specify, for example, an `InMemoryCertificateProvider`
-that implements private key offloading and a `FileWatcherCertificateProvider`
-for the root certificates. We will mark the [coupled
-API](http://google3/third_party/grpc/include/grpcpp/security/tls_credentials_options.h;l=55-56;rcl=682352913)
-as deprecated.
+the user. This provides a strict super-set of functionality of the current
+`StaticDataCertificateProvider` - calling `UpdateRoot` and `UpdateIdentity` on
+an `InMemoryCertificateProvider` at setup is the same as a
+`StaticDataCertificateProvider`.
```
markdroth, line 88: This should explicitly state that we are going to deprecate `StaticDataCertificateProvider`.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
+  //   PEM block.
+  absl::Status ValidateCredentials() const;
 }
+```
 
-// Sets the `grpc_tls_certificate_provider` to provide identity data.
-void set_identity_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider> certificate_provider);
+#### Splitting Certificate Providers into Identity and Root Providers
+We will also decouple the root provider and the identity provider in [the
+tls\_credentials\_options](http://google3/third_party/grpc/src/core/credentials/transport/tls/grpc_tls_credentials_options.h;l=77;rcl=731487339)
```
markdroth: Links to codesearch don't work in OSS.  Please use github links instead.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
+signer must be done for the identity, but is unrelated to how the root is
+provided. Currently, identity (private key and certificate chain) presentation
+to gRPC is coupled with the roots presentation to gRPC.
 
-// Sets the `grpc_tls_certificate_provider` to provide root data.
-void set_root_certificate_provider(grpc_core::RefCountedPtr<grpc_tls_certificate_provider> certificate_provider);
+```c++
+class TlsCredentialsOptions {
+ public:
+  [[deprecated(
```
markdroth, line 130: Please put a comment here saying that this is the existing method, which is now marked as deprecated.

Also, please add a comment before the last two methods indicating that those are the new methods being added in this design.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
-We are largely restricted by the underlying security libraries in each language.
-In the following sections, each language's API will be discussed as they are
-dependent upon the SSL library interfaces. Further, each language has different
-expectations for the sign functions on whether raw bytes or a digest is
-expected.
-
-
-## Implementation
-
 ### C-Core
```
markdroth: This should say "C-core / C++", since it covers both.

A107-tls-private-key-offloading.md
markdroth: Before the subsections describing each of the three changes in detail, please add a bulletted list of the three changes, with just a few words about each to say why we're doing it.  That way, the reader has a bit of context about why we're making these changes before we dive into the details.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
+      std::shared_ptr<CertificateProviderInterface> certificate_provider);
+  void set_identity_certificate_provider(
+      std::shared_ptr<CertificateProviderInterface> certificate_provider);
 ```
 
-#### *Internals*
+#### Private Key Signing
 
-We detail the async flow below. In the sync case, a signature is simply returned when it is asked for.
+BoringSSL provides an asynchronous API for private key signing, so we will
```
markdroth, line 145: We should call out the fact that this is supported only if gRPC is built with BoringSSL.  We should also discuss what the behavior will be if it's not.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
+`Cancel` will be called by gRPC in the case that an in-flight operation must be
+shutdown, and the implementer should use this function to ensure any resources
+being used for an asynchronous signing operation are properly shut down and
+released.
 
-   [`PemKeyCertPair` already exists](http://google3/third_party/grpc/src/core/credentials/transport/tls/ssl_utils.h;l=153-167;rcl=786762336) but only accepts a `string` private key.  We will modify this to optionally accept a private key signer.
 
-```c
+```c++
+using PrivateKey = std::variant<absl::string_view, CustomPrivateKeySign>
```
markdroth: This type is not actually part of the public API.  I don't think we need to include it here.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
+  std::string private_key;
+  std::string certificate_chain;
+};
+
+// A struct that stores the credential data presented to the peer in handshake
+// to show local identity. The private_key and certificate_chain should always
+// match. The private_key can be either a PEM string or a PrivateKeySigner.
+// The PrivateKeySigner will only work with gRPC binaries compiled with
+// BoringSSL.
+struct GRPCXX_DLL IdentityKeyOrSignerCertPair {
```
markdroth: We also need to point out that we're adding a new overload of the `InMemoryCertificateProvider::UpdateIdentityKeyCertPair()` method that takes this new type.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
+  void set_certificate_provider(
+      std::shared_ptr<CertificateProviderInterface> certificate_provider);
+  void set_root_certificate_provider(
+      std::shared_ptr<CertificateProviderInterface> certificate_provider);
+  void set_identity_certificate_provider(
+      std::shared_ptr<CertificateProviderInterface> certificate_provider);
 ```
 
-#### *Internals*
+#### Private Key Signing
```
markdroth, line 143: I suggest reorganizing this section such that it first describes the `PrivateKeySigner` API, and then discusses how the key signer can be configured via `InMemoryCertificateProvider`.

A107-tls-private-key-offloading.md
```diff
@@ -54,102 +61,114 @@ struct ssl_private_key_method_st {
...
+The implementation of `PrivateKeySigner::Sign` can be sync or async. In the sync
+case, the implementer must return a signature or a non-ok status.  In the async
+case, the implementer must kick off their async operations then return a
+`PrivateKeySigner::AsyncSigningHandle`. The async operation must call the
+provided `on_sign_complete` function with the signing result when the operation
+is complete. The implementer must also implement `PrivateKeySigner::Cancel`.
+`Cancel` will be called by gRPC in the case that an in-flight operation must be
+shutdown, and the implementer should use this function to ensure any resources
+being used for an asynchronous signing operation are properly shut down and
+released.
```
markdroth: The details like type named and method names are already covered in the comments in the API definition, so it's a bit redundant to cover them here.  Instead, I think this should just focus on a high-level description of the capabilities of the API -- e.g., we should say that the implementation can choose whether to return synchronously or asynchronously via a callback.

A107-tls-private-key-offloading.md
```diff
@@ -568,23 +478,58 @@ interface. Particularly, a user could create their own signature provider and
...
+expected.
+
+
+## Implementation
+
+* https://github.com/grpc/grpc/pull/40878 - Migrate Python to TlsCredentials under the hood
+* https://github.com/grpc/grpc/pull/41490 - Separate cert provider into a root and identity provider
+* https://github.com/grpc/grpc/pull/41484 - Create InMemoryCertificateProvider
+* https://github.com/grpc/grpc/pull/41606 - Implement PrivateKeySigner in C-Core and C++
+* https://github.com/grpc/grpc/pull/41701 - Implement in Python and Cython
```
markdroth: Suggest splitting this into separate lists by language -- i.e., the C-core/C++ changes in one list, and the python changes in a separate list.

A107-tls-private-key-offloading.md
```diff
@@ -568,23 +478,58 @@ interface. Particularly, a user could create their own signature provider and
...
 | :---- | :---- | :---- |
 | `grpc.target` | required | Indicates the target of the gRPC channel for which this handshake is occurring |
-| `grpc.security.handshaker.offloaded_private_key_sign.algorithm` | optional | The signature algorithm used to sign |
-| `grpc.security.handshaker.offloaded_private_key_sign.status` | optional | The status code return for the private key sign function |
+| `grpc.tls.private_key_sign_algorithm` | optional | The signature algorithm used to sign |
+| `grpc.status` | optional | The status code return for the private key sign function. From [A66] |
+
+The `grpc.tls.private_key_sign_algorithm` label will contain both a name and key
+length. For example, `RsaPkcs1Sha256`. For more, see the `SignatureAlgorithm`
+enum in the C section above.
```
markdroth: This text should go in the "Description" field of the table for the row for the `grpc.tls.private_key_sign_algorithm` label, rather than in a separate paragraph.

A107-tls-private-key-offloading.md
```diff
@@ -568,23 +478,58 @@ interface. Particularly, a user could create their own signature provider and
...
+* https://github.com/grpc/grpc/pull/41606 - Implement PrivateKeySigner in C-Core and C++
+* https://github.com/grpc/grpc/pull/41701 - Implement in Python and Cython
+
+
 # **Observability**
 
-The following non-per-call metric and labels will be added. These will allow a
-user insight into the offloaded operations and will be an aid in debugging
-failures.
+The following non-per-call metrics [A79] and labels will be added. These will
```
markdroth: Suggest changing this to:

"""
We will add a new metric using the non-per-call metric framework described in [A79].  This will ...
"""

A107-tls-private-key-offloading.md
```diff
@@ -568,23 +478,58 @@ interface. Particularly, a user could create their own signature provider and
...
+allow a user insight into the offloaded operations and will be an aid in
+debugging failures.
 
-| Name | Disposition | Description |
+| Label Name | Disposition | Description |
 | :---- | :---- | :---- |
 | `grpc.target` | required | Indicates the target of the gRPC channel for which this handshake is occurring |
```
markdroth: The description for this label should note that it's the existing label defined in A66.

A107-tls-private-key-offloading.md
```diff
@@ -568,23 +478,58 @@ interface. Particularly, a user could create their own signature provider and
...
 
-The following non-per-call metric and labels will be added. These will allow a
-user insight into the offloaded operations and will be an aid in debugging
-failures.
+The following non-per-call metrics [A79] and labels will be added. These will
+allow a user insight into the offloaded operations and will be an aid in
+debugging failures.
 
-| Name | Disposition | Description |
+| Label Name | Disposition | Description |
```
markdroth: Suggest adding a sentence in its own paragraph before this table:

"""
The new metric will use the following labels:
"""

A107-tls-private-key-offloading.md
```diff
@@ -568,23 +478,58 @@ interface. Particularly, a user could create their own signature provider and
...
-| `grpc.security.handshaker.offloaded_private_key_sign.algorithm` | optional | The signature algorithm used to sign |
-| `grpc.security.handshaker.offloaded_private_key_sign.status` | optional | The status code return for the private key sign function |
+| `grpc.tls.private_key_sign_algorithm` | optional | The signature algorithm used to sign |
+| `grpc.status` | optional | The status code return for the private key sign function. From [A66] |
+
+The `grpc.tls.private_key_sign_algorithm` label will contain both a name and key
+length. For example, `RsaPkcs1Sha256`. For more, see the `SignatureAlgorithm`
+enum in the C section above.
 
 | Metric | Type | Unit | Labels | Description |
```
markdroth: Suggest adding a sentence before this table:

"""
Here is the new metric we will add:
"""

A107-tls-private-key-offloading.md
```diff
@@ -568,23 +478,58 @@ interface. Particularly, a user could create their own signature provider and
...
+| Label Name | Disposition | Description |
 | :---- | :---- | :---- |
 | `grpc.target` | required | Indicates the target of the gRPC channel for which this handshake is occurring |
-| `grpc.security.handshaker.offloaded_private_key_sign.algorithm` | optional | The signature algorithm used to sign |
-| `grpc.security.handshaker.offloaded_private_key_sign.status` | optional | The status code return for the private key sign function |
+| `grpc.tls.private_key_sign_algorithm` | optional | The signature algorithm used to sign |
+| `grpc.status` | optional | The status code return for the private key sign function. From [A66] |
+
+The `grpc.tls.private_key_sign_algorithm` label will contain both a name and key
+length. For example, `RsaPkcs1Sha256`. For more, see the `SignatureAlgorithm`
```
markdroth: We will want the values for this label to be the same in all languages.  Is that going to be a problem if we define the values via a C++ enum?

A107-tls-private-key-offloading.md
markdroth: If we haven't implemented this as part of this feature, how about we leave the metric out of this gRFC and include it in the forthcoming gRFC for all of the security-related metrics?

markdroth (commented):

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Oops, I keep thinking I've gotten them all

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Took out most of the detail, left the bit about the user being responsible for cleaning up resources during Cancellation

A107-tls-private-key-offloading.md
gtcooke94: Removed

A107-tls-private-key-offloading.md
gtcooke94: Added

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: I think there are relatively standard names - we'll want to convert to a string and not use the enum directly. Called that out

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
```diff
@@ -119,55 +128,36 @@ to gRPC is coupled with the roots presentation to gRPC.
...
 
 #### Private Key Signing
 
 BoringSSL provides an asynchronous API for private key signing, so we will
-provide an asynchronous, cancellable API using callbacks. The C-Core APIs used
-to configure this feature will rely on a new certificate provider
-implementation.
+provide an asynchronous, cancellable API using callbacks. Thus, this is only supported in gRPC builds with BoringSSL. If a
+`PrivateKeySigner` is used in a non-BoringSSL build, the user should expect
+failure.
```
markdroth: Let's be more specific here.  What exactly will fail?  I assume the failure won't be seen until handshake time, right?

A107-tls-private-key-offloading.md
```diff
@@ -224,6 +214,34 @@ class PrivateKeySigner {
   // from a previous call to Sign().
   virtual void Cancel(std::shared_ptr<AsyncSigningHandle> handle) = 0;
 };
+```
+
+The C-Core APIs used
+to configure this feature will rely on a new certificate provider
+implementation.
```
markdroth: We've already described the `InMemoryCertificateProvider` above, so we don't need to introduce the fact that we're adding a new cert provider impl here.  Instead, this should just say something like "The private key signer can be configured via the `InMemoryCertificateProvider` using a new `IdentityKeyOrSignerCertPair` struct, as shown below."

A107-tls-private-key-offloading.md
```diff
@@ -504,28 +524,32 @@ expected.
 
 ## Implementation
 
-* https://github.com/grpc/grpc/pull/40878 - Migrate Python to TlsCredentials under the hood
-* https://github.com/grpc/grpc/pull/41490 - Separate cert provider into a root and identity provider
-* https://github.com/grpc/grpc/pull/41484 - Create InMemoryCertificateProvider
-* https://github.com/grpc/grpc/pull/41606 - Implement PrivateKeySigner in C-Core and C++
-* https://github.com/grpc/grpc/pull/41701 - Implement in Python and Cython
+* C-Core/C++
+    * https://github.com/grpc/grpc/pull/40878 - Migrate Python to TlsCredentials under the hood
```
markdroth: This one belongs in the python list, not the C-core/C++ list.

markdroth (commented):
Just a few comments left.

Note that there are also a couple of unresolved comments from my previous review pass that still need to be addressed.

Thanks!

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Update - it should fail when we try to configure it (the update fn should return a bad status) - https://github.com/grpc/grpc/blob/fb1b22f38a5232744439e3c290913f50da3f927c/src/core/credentials/transport/tls/grpc_tls_certificate_distributor.cc#L342-L345

This flow happens when the user calls `UpdateIdentityKeyCertPair` on their `InMemoryCertificateProvider`.

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
```diff
@@ -216,9 +217,8 @@ class PrivateKeySigner {
 };
 ```
 
-The C-Core APIs used
-to configure this feature will rely on a new certificate provider
-implementation. 
+The private key signer can be configured via the `InMemoryCertificateProvider`
+using a new `IdentityKeyOrSignerCertPair` struct, as shown below."
```
markdroth: Please remove trailing `"`.

A107-tls-private-key-offloading.md
markdroth: Still need to resolve this question.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+    private_key_bytes = client_private_key()
+    signature = sign_private_key(
+        data_to_sign, private_key_bytes, signature_algorithm
+    )
+    return signature
+
+def async_signer_worker(data_to_sign, signature_algorithm, on_complete):
+    """
+    Meant to be used as an async function for a thread, for example
+    """
```
sergiitk: Async function has a special meaning in python: https://docs.python.org/3/library/asyncio.html. I recommend changing the term to "concurrent":
```suggestion
def concurrent_signer_worker(data_to_sign, signature_algorithm, on_complete):
    """
    Meant to be used as an concurrent function for a thread, for example
    """
```

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+    private_key_bytes = client_private_key()
+    signature = sign_private_key(
+        data_to_sign, private_key_bytes, signature_algorithm
+    )
+    on_complete(signature)
+
+
+def no_op_cancel():
+    pass
```
sergiitk: Let's add a comment explaining why we pass:

> Threads cannot be destroyed, stopped, suspended, resumed, or interrupted
> https://docs.python.org/3/library/threading.html#threading.TIMEOUT_MAX

We should clarify that in other cases (like sending external RPC, or running a separate process) the logic should be to cancel the underlying concurrent call.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+        args=(data_to_sign, signature_algorithm, on_complete),
+    ).start()
+    return no_op_cancel
+
+
+
+
+# Now the user is in their application configuring gRPC 
+# Create creds with the custom signer
+creds = ssl_channel_credentials_with_custom_signer(<some_root>, <sync or async example_signer, <some_chain>)
```
sergiitk: Note that we we require kwargs now.

```suggestion
creds = grpc.experimental.ssl_channel_credentials_with_custom_signer(
    private_key_sign_fn=your_signer_fn
    certificate_chain=your_cert_chain,
    root_certificates=your_root_certs,
)
```

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+):
+    """
+    Takes in data_to_sign and signs it using the test private key with a sync return
+    """
+    private_key_bytes = client_private_key()
+    signature = sign_private_key(
+        data_to_sign, private_key_bytes, signature_algorithm
+    )
+    return signature
+
```
sergiitk, line 351: Let's separate these two examples, make it

```md
(break triple tick)

Concurrent example:

(new triple tick)
```

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+        data_to_sign, private_key_bytes, signature_algorithm
+    )
+    on_complete(signature)
+
+
+def no_op_cancel():
+    pass
+
+
+def async_client_private_key_signer(
```
sergiitk: Async function has a special meaning in python: https://docs.python.org/3/library/asyncio.html. I recommend changing the term to "concurrent":

```suggestion
def concurrent_client_private_key_signer(
```

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+    """
+    Takes in data_to_sign and signs it using the test private key with a sync return
+    """
+    private_key_bytes = client_private_key()
+    signature = sign_private_key(
+        data_to_sign, private_key_bytes, signature_algorithm
+    )
+    return signature
+
+def async_signer_worker(data_to_sign, signature_algorithm, on_complete):
```
sergiitk:
```suggestion
def _thread_signer_worker(data_to_sign, signature_algorithm, on_complete):
```

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+    Meant to be used as an async function for a thread, for example
+    """
+    private_key_bytes = client_private_key()
+    signature = sign_private_key(
+        data_to_sign, private_key_bytes, signature_algorithm
+    )
+    on_complete(signature)
+
+
+def no_op_cancel():
```
sergiitk:
```suggestion
def _no_op_cancel():
```

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+    """
+    private_key_bytes = client_private_key()
+    signature = sign_private_key(
+        data_to_sign, private_key_bytes, signature_algorithm
+    )
+    return signature
+
+def async_signer_worker(data_to_sign, signature_algorithm, on_complete):
+    """
+    Meant to be used as an async function for a thread, for example
```
sergiitk: Let's make it clear it's used in `concurrent_client_private_key_signer`

A107-tls-private-key-offloading.md
sergiitk: Also, let's make it clear it's used in `concurrent_client_private_key_signer`

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+        target=async_signer_worker,
+        args=(data_to_sign, signature_algorithm, on_complete),
+    ).start()
+    return no_op_cancel
+
+
+
+
+# Now the user is in their application configuring gRPC 
+# Create creds with the custom signer
```
sergiitk: reword?

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+def ssl_channel_credentials_with_custom_signer(
+    *,
+    private_key_sign_fn: "CustomPrivateKeySign",
+    root_certificates: Optional[bytes] = None,
+    certificate_chain: bytes,
+) -> grpc.ChannelCredentials:
+    """Creates a ChannelCredentials for use with an SSL-enabled Channel with a custom signer.
+
+    THIS IS AN EXPERIMENTAL API.
+    This API will be removed in a future version and combined with `grpc.ssl_channel_credentials`.
```
sergiitk, line 291: Let's add the note on this not working with gevent here too.

sergiitk (commented):
Python: left notes on improving the usage example

A107-tls-private-key-offloading.md
markdroth: This comment made me realize that this is actually a very bad example to use here, because we really don't want users writing key signing implementations that don't support cancellation.

Please replace this example with one that can actually cancel the operation.  For example, consider a case where the key signer might issue an RPC to do the key signing; when the signing request gets cancelled, the underlying RPC can also be cancelled.

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Swapped to a more pseduo-code example

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,559 @@
...
+    private_key_bytes = client_private_key()
+    signature = sign_private_key(
+        data_to_sign, private_key_bytes, signature_algorithm
+    )
+    return signature
+```
+
+Concurrent example:
+
+```
```
sergiitk: nit: add `py` syntax

A107-tls-private-key-offloading.md
sergiitk: I agree with Mark - having an actual cancellable example is much better than using `def on_cancel: pass` antipattern. 

However, I've looked at the new pseudo-code example, and not sure if it's useful anymore. At least the thread demonstrated the flow better. Now it's not clear where "on_complete" is called and why `concurrent_client_private_key_signer` needs `_do_signing_concurrent` in the first place.

I think we should replace it with more realistic cancellable example, f.e. using something like `multiprocessing.Process` or RPC cancellation https://github.com/grpc/grpc/tree/master/examples/python/cancellation#cancellation-on-the-client-side.

@gtcooke94 — Let @asheshvidyut know if you need help drafting that.

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Done, PTAL

markdroth (approved):

markdroth:
This looks great!  I'd like approval from @easwars and @sergiitk before merging.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,536 @@
...
+	//
+	// Note that when a signature of a hash of a larger message is needed,
+	// the caller is responsible for hashing the larger message and passing
+	// the hash (as digest) and the hash function (as opts) to Sign.
+	Sign(rand io.Reader, digest []byte, opts SignerOpts) (signature []byte, err error)
+}
+
+cert = tls.Certificate{
+Certificate: derChain, //[][]byte
+PrivateKey:  PrivateKeySigner, //crypto.PrivateKey
```
easwars: Nit: Can we add a tab to these fields in this literal struct initialization syntax to make it more readable. Thanks.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,536 @@
...
+provider, err := NewInMemoryCertProvider(tlsCert)
+serverOpts := &advancedtls.Options{
+	IdentityOptions: advancedtls.IdentityCertificateOptions{
+		IdentityProvider: provider,
+	},
+}
+
+// Down the line update
+provider.UpdateCertificate(&tls.Certificate(<some updated thing>))
+
```
easwars, line 495: For completeness sake, does it also make sense to show how this options struct will be used to create a credentials and how it will be passed to create the grpc channel?

easwars (approved):

A107-tls-private-key-offloading.md
gtcooke94: Done

A107-tls-private-key-offloading.md
gtcooke94: Yes - I actually removed this line about UpdateCertificate because I think it's just confusing for this example.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,535 @@
...
+    data_to_sign, signature_algorithm, on_complete
+):
+    """Takes in data_to_sign and signs it concurrently using multiprocessing."""
+    p = multiprocessing.Process(
+        target=_sign_worker,
+        args=(data_to_sign, signature_algorithm, on_complete),
+    )
+    p.start()
+
+    # Per the Python API, return a callable matching PrivateKeySignCancel
```
sergiitk: Instead, let's explain the flow? Something like "This will be called when ...`

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,535 @@
...
+    def cancel():
+        p.terminate()
+        p.join()
+
+    return cancel
+
+
+# In the Python code configuring gRPC, create creds with the custom signer
```
sergiitk: nit: now this is a part of "Concurrent example", which may be a bit confusing. Let's move it to the top, so the flow is as follows:

---

Usage example:

```py
creds = grpc.experimental.ssl_channel_credentials_with_custom_signer(
    private_key_sign_fn=your_signer_fn
    certificate_chain=your_cert_chain,
    root_certificates=your_root_certs,
)
```

Synchronous (blocking) signer example:

```py
def sync_client_private_key_signer(
    data_to_sign,
    signature_algorithm,
    on_complete,
): ...
```

Concurrent signer example:

```py
import multiprocessing
...
```

A107-tls-private-key-offloading.md
sergiitk: Looks so much better, thank you!

sergiitk (approved):
Python looks good! Just left a few very minor notes / suggestions

A107-tls-private-key-offloading.md
gtcooke94: PTAL

A107-tls-private-key-offloading.md
gtcooke94: Good call, moved around

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,537 @@
...
+}
+
+// Create the InMemoryCertProvider and configure gRPC with it
+provider, err := NewInMemoryCertProvider(tlsCert)
+clientOpts := &advancedtls.Options{
+	IdentityOptions: advancedtls.IdentityCertificateOptions{
+		IdentityProvider: provider,
+	},
+}
+clientTLSCreds, err := advancedtls.NewClientCreds(options)
```
ejona86: s/options/clientOpts/ ?

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,537 @@
...
+## Rationale
+
+
+Private key offloading is designed to support signing outside of the existing
+process, for example in a hardware module or via an RPC \- thus this API should
+support asynchronous operations in languages where that is possible (Golang's
+crypto/tls does **not** support asynchronous private key signing). 
+
+We are largely restricted by the underlying security libraries in each language.
+In the following sections, each language's API will be discussed as they are
```
ejona86: "following sections" accidentally left over after this text moved?

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,537 @@
+A107 - TLS Private Key Offloading
+----
+* Author: @gtcooke94
+* Approver: ejona86
+* Status: C++ implemented
```
ejona86: This was the status of the gRFC, with a fixed enumeration of values. So it would have just been "Implemented". The next line would give specifics as to the languages that are complete.

Although with #275 just merged, you could delete this line instead. We removed it because it was essentially always wrong and multiple of the states only applied while it was a PR which didn't provide value.

A107-tls-private-key-offloading.md
```diff
@@ -0,0 +1,537 @@
...
+* Implemented in: C++, Go
```
ejona86: This is for languages that already have it implemented. Is Go already implemented? I don't see `InMemoryCertProvider`; was it renamed?

ejona86 (commented):
Looks like this has reached consensus. I did notice a few minor things while checking if people's comments were addressed.

ejona86 (approved):
That should have been approved, since there is consensus. Do fix up the few minor comments, though.

dfawley (approved):

A107-tls-private-key-offloading.md
gtcooke94: Deleted

A107-tls-private-key-offloading.md
gtcooke94: Just C++, removed Go

A107-tls-private-key-offloading.md
gtcooke94: Moved it above each language section

A107-tls-private-key-offloading.md
gtcooke94: Fixed

PR #529, opened 2025-12-23
A110: Child Channel Options

AgraVator:


A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
+# **A110: Child Channel Configuration Plugins**
+
+* **Author(s)**: [Abhishek Agrawal](mailto:agrawalabhi@google.com)
```
markdroth: There are a couple of fields missing here, in particular the "approver" and "mailing list link" fields.

Please make sure you are following the gRFC template: https://github.com/grpc/proposal/blob/master/GRFC-TEMPLATE.md

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+* **Status**: Draft
+* **To be implemented in**: Core, Java, Go
+* **Last updated**: 2025-12-23
+
+## **Abstract**
+
+This proposal introduces a mechanism to configure "child channels"—channels created internally by gRPC components (such as xDS control plane channel). Currently, these internal channels cannot easily inherit configuration (like metric sinks and interceptors) from the user application without relying on global state. This design proposes a language-specific approach to injecting configuration: using `functional interfaces` in Java, `DialOptions` in Go, and `ChannelArgs` in Core.
```
markdroth: I don't think we want to specify a different approach in each language.  I think what we want here is to describe an approach in a language-agnostic way, which we will then use in all languages.

To be clear, I'm fine with specifying details of how things will work in individual languages if that is helpful to clarify the design.  But the main goal of an "A"-series gRFC is to specify a feature that will work essentially the same way in all languages, and we need to approach it from that perspective.

You may want to look at several of the recent "A"-series gRFCs for examples.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+
+## **Background**
+
+Complex gRPC ecosystems often require the creation of auxiliary channels that are not directly instantiated by the user application. Two primary examples are:
+
+1. **xDS (Extensible Discovery Service)**: When a user creates a channel with an xDS target, the gRPC library internally creates a separate channel to communicate with the xDS control plane. Currently, users have limited ability to configure this internal control plane channel.
+2. **Advanced Load Balancing (RLS, GrpcLB):** Policies like RLS (Route Lookup Service) and GrpcLB, as well as other high-level libraries built on top of gRPC, frequently create internal channels to communicate with look-aside load balancers or backends.
```
markdroth: RLS isn't a public API and shouldn't be mentioned here.

I think better examples to cite here are ext_authz (see WIP gRFC A92 in https://github.com/grpc/proposal/pull/481) and ext_proc (see WIP gRFC A93 in https://github.com/grpc/proposal/pull/484).  Note that those cases are not LB policies, so we should consider whether the plumbing needs to be different to support those cases.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+
+### **The Problem**
+
+There is currently no standardized way to configure behavior for these child channels.
+
+* **Metrics**: Users need to configure metric sinks so that telemetry from internal channels can be read and exported.
+* **Interceptors**: Users may need to apply specific interceptors (e.g., for authentication, logging, or tracing) to internal traffic.
+* **No Global State**: These configurations cannot be set globally (e.g., using static singletons) because different parts of an application may require different configurations, such as different metric backends or security credentials.
```
markdroth: This seems like it should be a separate paragraph rather than a bullet in this list.  It's not an example of what state needs to be propagated to child channels; it's a comment about why global state is not a good way to propagate the state to child channels.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
+# **A110: Child Channel Configuration Plugins**
```
markdroth: I think this should be called "Child Channel Options".  The mechanism we're describing here isn't really a plugin thing.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+Complex gRPC ecosystems often require the creation of auxiliary channels that are not directly instantiated by the user application. Two primary examples are:
+
+1. **xDS (Extensible Discovery Service)**: When a user creates a channel with an xDS target, the gRPC library internally creates a separate channel to communicate with the xDS control plane. Currently, users have limited ability to configure this internal control plane channel.
+2. **Advanced Load Balancing (RLS, GrpcLB):** Policies like RLS (Route Lookup Service) and GrpcLB, as well as other high-level libraries built on top of gRPC, frequently create internal channels to communicate with look-aside load balancers or backends.
+
+### **The Problem**
+
+There is currently no standardized way to configure behavior for these child channels.
+
+* **Metrics**: Users need to configure metric sinks so that telemetry from internal channels can be read and exported.
```
markdroth: It's not clear to me what this list is trying to communicate, because each of the three items here is something completely different:
- Metrics are one of the main features supported by stats plugins, but if that's what you're talking about, then I would expect you to also mention tracing.
- I don't think we need to talk about interceptors specifically here, since the mechanism we're describing isn't really specifically about interceptors; it's about setting options for child channels in general.
- The "no global state" thing isn't even talking about state that may be propagated to child channels; it's talking about why we can't use global state as a mechanism for how to do this propagation.  I think that should just be a separate paragraph.

I think we should instead start this section by talking about the problem that actually motivated this design, which is the need to set StatsPlugins on a per-channel basis and also propagate them to child channels.  We can then discuss why it's not sufficient to just use the global registry.

This discussion should reference gRFCs A66 and A72, which describe StatsPlugins.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+* **Interceptors**: Users may need to apply specific interceptors (e.g., for authentication, logging, or tracing) to internal traffic.
+* **No Global State**: These configurations cannot be set globally (e.g., using static singletons) because different parts of an application may require different configurations, such as different metric backends or security credentials.
+
+## **Proposal**
+
+The proposal creates a "plugin" or configuration injection style for internal channels. The implementation varies by language to match existing idioms, but the goal remains consistent: allow the user to pass a configuration object or function that the internal channel factory applies during creation.
```
markdroth: As I mentioned above, I think we need to describe the overall approach here in a language-agnostic way, so that the intent of the design is clear.  I'm fine with going into language-specific details if that turns out to be useful, but the vast majority of the design should be described in a language-agnostic way.  The language-specific parts, if any, should be added only for the purpose of clarifying how the general approach will be translated into reality in each language.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+
+### **Java**
+
+In Java, the configuration will be achieved by accepting functions (callbacks). The API allows users to pass a `Consumer<ManagedChannelBuilder<?>>` (or a similar functional interface). When an internal library (e.g., xDS, RLS, gRPCLB) creates a child channel, it applies this user-provided function to the builder before building the channel.
+
+
+* #### 1\. Configuration Interface
```
markdroth: The numbers on these subsections don't add any value.  Please remove them.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+    builder.maxInboundMessageSize(1024 * 1024);
+  };
+
+  // 2. Apply it to the parent channel
+  ManagedChannel channel = ManagedChannelBuilder.forTarget("xds:///my-service")
+    .childChannelConfigurer(myInternalConfig) // <--- Configuration injected here
+    .build();
+  ```
+
+* #### 5\. Out-of-Band (OOB) Channels
```
markdroth: What is an out-of-band channel?  Is this some Java-specific concept?

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+
+  We do not propose applying child channel configurations to Out-of-Band (OOB) channels at this time. To maintain architectural flexibility and avoid breaking changes in the future, we will modify the implementation to use a `noOp()` MetricSink for OOB channels rather than inheriting the parent channel's sink.
+
+  Furthermore, we acknowledge that certain configurations will not function out-of-the-box for `resolvingOobChannel` due to its specific initialization requirements.
+
+### **Go**
+
+In Go, configuration for child channels is be achieved by passing a specific `DialOption` to the parent channel. This option encapsulates a slice of other a slice of *other* `DialOption`s that are applied exclusively to any internal child channels created by the parent.
```
markdroth: The words "a slice of other" are repeated twice.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+
+* #### 1\. Configuration Mechanism
+
+  We define a new channel argument key. The value associated with this key is a pointer to a `grpc_channel_args` struct, managed via a pointer vtable to ensure correct ownership and copying.
+
+  ```c
+  // A pointer argument key that internal components (like xDS) look for.
+  // The value is a pointer to a grpc_channel_args struct containing the subset 
+  // of options (e.g., specific socket mutators, user agents) for child channels.
+  #define GRPC_ARG_CHILD_CHANNEL_ARGS "grpc.internal.child_channel_args"
```
markdroth: This can't start with `grpc.internal.`, since args with that prefix are not actually allowed to be passed into C-core from the outside.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+  ```
+
+* #### 2\. Internal Implementation
+
+  Internal components that create channels (specifically `XdsClient`) are updated to look for this argument. When present, these arguments are merged with the default internal arguments required by the component.
```
markdroth: What does "default internal arguments required by the component" mean?  Can you give an example?

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+
+* #### 3\. Usage Example (User-Side Code)
+
+  In this design, a user configures the parent channel by defining a subset of arguments intended for child channels, packing them into a standard `grpc_arg`, and passing them to the parent. This allows for granular control over internal components, such as the xDS control plane connection, without affecting the parent channel stack.
+
+  ```c
+  // 1. Prepare the Child Config (The Subset)
+  // In this example, we configure a specific Socket Mutator for the child channel.
```
markdroth: This is not a good example to use here, because socket mutator is not a public API.

I suggest using a StatsPlugin as an example, since that's the thing we're actually trying to enable with this design.  In fact, I think we should use a StatsPlugin example for all languages.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+  grpc_socket_mutator* my_mutator = CreateMySocketMutator();
+  grpc_arg child_arg = grpc_channel_arg_socket_mutator_create(my_mutator);
+  grpc_channel_args child_args_struct = {1, &child_arg};
+
+  // 2. Pack the Subset into the Parent's Arguments
+  // The child_args_struct is wrapped as a pointer argument.
+  // We use a VTable to ensure the nested struct is safely copied or
+  // destroyed during the parent channel's lifetime.
+  grpc_arg parent_arg = grpc_channel_arg_pointer_create(
```
markdroth: This is using a C-core API, which is not what we'd want applications to use.

Instead, we would probably just add a method to `grpc::ChannelArguments` that looks something like this:

```
void SetChildChannelArgs(grpc::ChannelArguments child_args);
```

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+  );
+
+  // 3. Create the Parent Channel
+  // The parent channel receives the nested argument but does not apply
+  // the socket mutator to itself.
+  grpc_channel_args parent_args = {1, &parent_arg};
+  auto channel = grpc::CreateCustomChannel(
+      "xds:///my-service", 
+      grpc::InsecureChannelCredentials(), 
+      grpc::ChannelArguments::FromC(parent_args)
```
markdroth: There is no `FromC()` method on `grpc::ChannelArguments`, although that method does exist on some of our internal channel args types.  This looks an awful lot like an AI hallucination. ;P

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+
+## **Rationale**
+
+### **Why not Global Configuration?**
+
+We reject global configuration (static variables) because it prevents multi-tenant applications from isolating configurations. For example, one client may need to export metrics to Prometheus, while another in the same process exports to Cloud Monitoring.
+
+### **Why Language-Specific Approaches?**
+
+While the concept is cross-language, the mechanisms for channel creation differs significantly:
```
markdroth: These differences are really just implementation details.  I think the overall approach is the same across languages, and we should describe it as such.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+* **Interceptors**: Users may need to apply specific interceptors (e.g., for authentication, logging, or tracing) to internal traffic.
+* **No Global State**: These configurations cannot be set globally (e.g., using static singletons) because different parts of an application may require different configurations, such as different metric backends or security credentials.
+
+## **Proposal**
+
+The proposal creates a "plugin" or configuration injection style for internal channels. The implementation varies by language to match existing idioms, but the goal remains consistent: allow the user to pass a configuration object or function that the internal channel factory applies during creation.
+
+### **Java**
+
+In Java, the configuration will be achieved by accepting functions (callbacks). The API allows users to pass a `Consumer<ManagedChannelBuilder<?>>` (or a similar functional interface). When an internal library (e.g., xDS, RLS, gRPCLB) creates a child channel, it applies this user-provided function to the builder before building the channel.
```
markdroth: The API proposed for Java and Go seems to be only for channels, not for servers.  However, the xDS, ext_authz, and ext_proc cases will also happen on the gRPC server side.  Do we need to solve this problem also on the server side?

(The C-core approach will work on both client and server sides.)

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,215 @@
...
+
+## **Abstract**
+
+This proposal introduces a mechanism to configure "child channels"—channels created internally by gRPC components (such as xDS control plane channel). Currently, these internal channels cannot easily inherit configuration (like metric sinks and interceptors) from the user application without relying on global state. This design proposes a language-specific approach to injecting configuration: using `functional interfaces` in Java, `DialOptions` in Go, and `ChannelArgs` in Core.
+
+## **Background**
+
+Complex gRPC ecosystems often require the creation of auxiliary channels that are not directly instantiated by the user application. Two primary examples are:
+
+1. **xDS (Extensible Discovery Service)**: When a user creates a channel with an xDS target, the gRPC library internally creates a separate channel to communicate with the xDS control plane. Currently, users have limited ability to configure this internal control plane channel.
```
markdroth: In the xDS case, the same xDS channel will be used in multiple channels for the same target.  We need to document what the behavior will be if the application creates multiple channels to the same target with different child channel args.  (It will just use whatever child channel args are passed to the first channel to be created.)

markdroth (commented):
The overall design looks good, but I think the doc needs work.

My main concern is that we need a much better description of the overall design in a language-agnostic way, rather than just saying that we have a different design for each language.  It's true that each language has its own APIs for how per-channel options are set, but the overall goal here is still to have a way to pass a set of options to be used in child channels.  The semantics for that should be the same in all languages.

A110-child-channel-plugins.md
AgraVator: A ManagedChannel backed used for LB's own RPCs needs. Something similar should be available in other languages as well.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+* [A92: xDS ExtAuthz Support](https://github.com/grpc/proposal/pull/481)
+* [A93: xDS ExtProc Support](https://github.com/grpc/proposal/pull/484)
+
+### The Problem
+
+The primary motivation for this feature is the need to configure observability on a per-child-channel basis.
+
+* StatsPlugins & Tracing: Users need to configure metric sinks (as described in gRFC [A66](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md) and [A72](https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md)) so that telemetry from internal channels is correctly tagged and exported.
+* Interceptors: Users may need to apply specific interceptors (e.g., for authentication, logging, or tracing) to internal traffic.
+
```
kannanjgithub, line 48: About interceptors for authentication - since the same configuration will be applied to all child channels, but something like auth tokens is going to be different per server like xDs or the DNS servers used by gRPC LB, i.e the same auth tokens is likely not going to work with both.

atollena: This is true for most options. I don't see any reason that the options for the xDS management server and a service like `ExtProc`/`ExtAuthz` must share the same set of interceptors, stats configuration, and other channel options (I'm also interested in setting client keepalive settings for the management server, hopefully this design address this as well -- I opened https://github.com/grpc/grpc/issues/39187 a while back, happy to give you more context if needed).

The access patterns of those auxiliary channels have reasons to be different. If we can make this design account for that, I think it would make it more useful.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+* Last updated: 2025-12-24
+* Discussion at: [Mailing List Link TBD]
+
+## Abstract
+
+This proposal introduces a mechanism to configure "child channels", channels created internally by gRPC components (such as xDS control plane channel). Currently, these channels are often opaque to the user, making it difficult to inject necessary configurations for metrics, tracing etc from the user application. This design proposes an approach for users to pass configuration options to these internal channels.
+
+## Background
+
+Complex gRPC ecosystems often require the creation of auxiliary channels that are not directly instantiated by the user application. Two primary examples are:
```
ejona86: s/Two/Three/

A110-child-channel-plugins.md
ejona86: Yeah, we should remove authentication from this list. I think we would strongly discourage injecting interceptors into child channels for authentication. We already deal with authentication in other ways anyway; that was essential for getting any child channel working. Also, I'd say it is clearly improper to assume what the child channel is doing. Logging/tracing/metrics can be added independent of what the channel is being used for; the configuration doesn't have to assume "oh, this child channel is for the xds server, let's use configuration X."

I agree that the child channels can have different configuration from their parent.

If you want to enable keepalive for child channel X but not Y, then we'd have to build in some way to identify the child channel. The design right now is targeting "same configuration for all children of this particular channel," which would probably "work" for keepalive, although technically it may be assuming a particular usage. Bootstrap does still seem a better option for the keepalive, as you really want to say "all connections to this control plane should use this setting."

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+* [A93: xDS ExtProc Support](https://github.com/grpc/proposal/pull/484)
+
+### The Problem
+
+The primary motivation for this feature is the need to configure observability on a per-child-channel basis.
+
+* StatsPlugins & Tracing: Users need to configure metric sinks (as described in gRFC [A66](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md) and [A72](https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md)) so that telemetry from internal channels is correctly tagged and exported.
+* Interceptors: Users may need to apply specific interceptors (e.g., for authentication, logging, or tracing) to internal traffic.
+
+These configurations cannot be set globally because different parts of an application may require different configurations, such as different metric backends or security credentials.
```
ejona86: Or we just mention the reason we have this problem: libraries. Libraries want to configure their channels, but can't configure globally.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+* `O_child` is opaque to `P`. `P` does not apply these options to itself.
+* `O_child` is carried in `P`'s state, available for extraction by internal components.
+
+### Propagation
+When an internal component (e.g., an xDS client factory or an auth filter) attached to `P` needs to create a Child Channel `C`:
+1.  It retrieves `O_child` from `P`.
+2.  It applies `O_child` to the configuration of `C`.
+
+### Precedence and Merging
+The Child Channel `C` typically requires some internal configuration `O_internal` (e.g., specific target URIs, bootstrap credentials, or internal User-Agent strings).
```
ejona86: There's a lot of internal configuration. In the Java design, some would be able to be changed by O_child, but other parts not. I really don't know what this specifically is trying to call out. This set of examples here doesn't help me much, as Java wouldn't let you change the target URI or credentials, and we have no case where we're setting an internal User-Agent string.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+
+If multiple Parent Channels (`P1`, `P2`) point to the same xDS target but provide *different* Child Channel Options (`O_child1`, `O_child2`):
+* Behavior: The shared client is created using the options from the first parent channel that triggers its creation (e.g., `O_child1`).
+* Subsequent Usage: When `P2` requests the client, it receives the existing shared client. `O_child2` is effectively ignored for that specific shared resource.
+
+### Language Implementations
+
+#### Java
+
+In Java, the configuration will be achieved by accepting functions (callbacks). The API allows users to pass a `Consumer<ManagedChannelBuilder<?>>` (or a similar functional interface). When an internal library (e.g., xDS, gRPCLB) creates a child channel, it applies this user-provided function to the builder before building the channel.
```
ejona86: s/before building/before further configuring/ to adhere to the defined `O_internal` behavior, right?

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+
+  Use the standard `java.util.function.Consumer` and define a new public API interface, `ChildChannelConfigurer`, to encapsulate the configuration logic for auxiliary channels.
+
+  ```java
+  import java.util.function.Consumer;
+  import io.grpc.ManagedChannelBuilder;
+
+  // Captures the intent of the plugin.
+  // Consumes a builder to modify it before the channel is built
+  public interface ChildChannelConfigurer extends Consumer<ManagedChannelBuilder<?>> {
```
ejona86: The interface should be a _channel_ configurator, not a _child channel_ configurator. It's only the ManagedChannelBuilder API that then applies that to child channels. I think it would be quite fair for someone to make one of these configurations, and call it directly to apply to the parent channel.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+#### Java
+
+In Java, the configuration will be achieved by accepting functions (callbacks). The API allows users to pass a `Consumer<ManagedChannelBuilder<?>>` (or a similar functional interface). When an internal library (e.g., xDS, gRPCLB) creates a child channel, it applies this user-provided function to the builder before building the channel.
+
+* ##### Configuration Interface
+
+  Use the standard `java.util.function.Consumer` and define a new public API interface, `ChildChannelConfigurer`, to encapsulate the configuration logic for auxiliary channels.
+
+  ```java
+  import java.util.function.Consumer;
```
ejona86: There's no point in having the interface if we're going to have it extend Consumer. We would just use Consumer directly. However, because it was added in Android API level 24, we can't actually use Consumer at all.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+  // Consumes a builder to modify it before the channel is built
+  public interface ChildChannelConfigurer extends Consumer<ManagedChannelBuilder<?>> {
+      // Inherits accept(T t) from Consumer
+  }
+  ``` 
+
+* ##### API Changes
+
+  * ManagedChannelBuilder: Add `ManagedChannelBuilder#childChannelConfigurer(ChildChannelConfigurer childChannelConfigurer)` to allow users to register this configurer.
+  * ServerBuilder: Add `ServerBuilder#childChannelConfigurer(ChildChannelConfigurer configurer)` to allow users to provide configuration for any internal channels created by the server (e.g., connections to external authorization or processing services).
```
ejona86: I think this would be only for XdsServerBuilder. Child channels are not available for regular servers.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+  ManagedChannel channel = ManagedChannelBuilder.forTarget("xds:///my-service")
+      .childChannelConfigurer(myInternalConfig) // <--- Configuration injected here
+      .build();    
+  ```
+
+* ##### Out-of-Band (OOB) Channels
+
+  We do not propose applying child channel configurations to Out-of-Band (OOB) channels at this time. To maintain architectural flexibility and avoid breaking changes in the future, we will modify the implementation to use a `noOp()` MetricSink for OOB channels rather than inheriting the parent channel's sink.
+
+  Furthermore, we acknowledge that certain configurations will not function out-of-the-box for `resolvingOobChannel` due to its specific initialization requirements.
```
ejona86: What were the problems here? We should call it out more; "certain configurations" could be very common, in which case that would be a problem.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+
+  // Apply it to the parent channel
+  ManagedChannel channel = ManagedChannelBuilder.forTarget("xds:///my-service")
+      .childChannelConfigurer(myInternalConfig) // <--- Configuration injected here
+      .build();    
+  ```
+
+* ##### Out-of-Band (OOB) Channels
+
+  We do not propose applying child channel configurations to Out-of-Band (OOB) channels at this time. To maintain architectural flexibility and avoid breaking changes in the future, we will modify the implementation to use a `noOp()` MetricSink for OOB channels rather than inheriting the parent channel's sink.
```
ejona86: I think we can remove this section, since https://github.com/grpc/grpc-java/pull/12589 is merged.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+    void SetChildChannelArgs(const ChannelArguments& args);
+  ```
+
+* ##### Usage Example (User-Side Code)
+
+  ```c
+  // TODO(AgraVator): add example for configuring child channel observability
+  ```
+
+## Rationale
```
ejona86, line 397: We need to say _somewhere_ that we are not trying to provide any identity information about which channel is being created. The configuration provided by this gRFC is required to be uniform across child channels of a particular channel.

christama (approved):

A110-child-channel-plugins.md
AgraVator: This is there in rationale.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,220 @@
...
+The user API must allow "nesting" of channel options. A user creating a Parent Channel `P` can provide a set of options `O_child`.
+* `O_child` is opaque to `P`. `P` does not apply these options to itself.
+* `O_child` is carried in `P`'s state, available for extraction by internal components.
+
+### Propagation
+When an internal component (e.g., an xDS client factory or an auth filter) attached to `P` needs to create a Child Channel `C`:
+1.  It retrieves `O_child` from `P`.
+2.  It applies `O_child` to the configuration of `C`.
+
+### Precedence and Merging
```
ejona86: We should say somewhere that the child channel options override global channel options. Although I hope in practice that doesn't really matter much.

A110-child-channel-plugins.md
dfawley: Are we sure this should be a non-goal?  E.g. we might have xds client connections needing a different set of options from RLS connections/etc.  It seems pretty plausible given that one team will own the control plane, but a different team would own the ext_proc services, so different credentials might be needed or different telemetry configuration may be desired.

This doesn't have to be hard; it could be a simple map from name->config.

A110-child-channel-plugins.md
```diff
@@ -32,7 +34,7 @@ Complex gRPC ecosystems often require the creation of auxiliary channels that ar
 The primary motivation for this feature is the need to configure observability on a per-child-channel basis.
 
 * StatsPlugins & Tracing: Users need to configure metric sinks (as described in gRFC [A66](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md) and [A72](https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md)) so that telemetry from internal channels is correctly tagged and exported.
-* Interceptors: Users may need to apply specific interceptors (e.g., for authentication, logging, or tracing) to internal traffic.
+* Interceptors: Users may need to apply specific interceptors (e.g., for logging, or tracing) to internal traffic.
 
 These configurations cannot be set globally because different parts of an application may require different configurations, such as different metric backends or security credentials.
```
ejona86: "security credentials" is still here, even though we're saying not to use this for authentication any more.

A110-child-channel-plugins.md
ejona86: This is not for credentials. Both of those channels would send telemetry to the same destination (configured here), so they'd use the same credentials. We would solve multi-ownership problems by allowing applying telemetry _multiple times_, a different time for each consumer/configuration.

I don't disagree that someone may want to identify the purpose of the child channel, but that then means _we_ need some way to identify the purpose of the child channel. That's a bigger lift and opens up lots of complexities on how you will categorize the child channels' purposes. In our TL discussion about it, nobody could come up with a strong need for the info, we so left it for future work if/when it is needed.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
-# **A110: Child Channel Configuration Plugins**
+A110: Child Channel Options
+----
+* Author(s): [Abhishek Agrawal](mailto:agrawalabhi@google.com)
+* Approver: a11r
```
markdroth: Please set this to "markdroth, ejona86, dfawley".

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
+* Status: In Review
+* Implemented in: Core, Java, Go
```
markdroth: It isn't actually implemented anywhere yet.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
-* **Status**: Draft
-* **To be implemented in**: Core, Java, Go
-* **Last updated**: 2025-12-23
+This proposal introduces a mechanism to configure "child channels", channels
+created internally by gRPC components (such as xDS control
+user, making it difficult to inject necessary configurations for
+metrics, tracing etc from the user application. This design proposes
+an approach for users to pass configuration options to these
+internal channels.
+gRPC willl support the xDS Extension Config Discovery Service (ECDS),
```
markdroth: This looks like something you copied from another gRFC, but I don't think it belongs here.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
 
-## **Abstract**
+## Background
 
-This proposal introduces a mechanism to configure "child channels"—channels created internally by gRPC components (such as xDS control plane channel). Currently, these internal channels cannot easily inherit configuration (like metric sinks and interceptors) from the user application without relying on global state. This design proposes a language-specific approach to injecting configuration: using `functional interfaces` in Java, `DialOptions` in Go, and `ChannelArgs` in Core.
+Complex gRPC ecosystems often require the creation of auxiliary channels that
+are not directly instantiated by the user application. The primary examples are:
 
-## **Background**
+1. xDS (Extensible Discovery Service): When a user creates a channel with an xDS
```
markdroth: The "x" in "xDS" does *not* stand for "Extensible".  I suggest just removing the parenthetical here.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
+
+## Abstract
 
-* **Author(s)**: [Abhishek Agrawal](mailto:agrawalabhi@google.com)
-* **Status**: Draft
-* **To be implemented in**: Core, Java, Go
-* **Last updated**: 2025-12-23
+This proposal introduces a mechanism to configure "child channels", channels
+created internally by gRPC components (such as xDS control
+user, making it difficult to inject necessary configurations for
```
markdroth: What does "xDS control user" mean?  I think maybe you meant "such as a channel created to an xDS server"?

Also, you say "This proposal introduces a mechanism to configure child channels..., making it difficult to...".  That reads like you're saying that you're introducing a mechanism that makes it difficult to do something, which I don't think is what you mean here.

I suggest rewriting this paragraph to something like this:

"""
There are several use cases where gRPC internally creates a "child channel" for communicating with a control plane.  Because these channels are created internally rather than being created by the application, it is currently difficult to inject necessary configuration for these channels, which makes it hard to configure things like metrics or tracing from the application.  This design proposes a mechanism to allow applications to pass configuration options to these child channels.
"""

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
+an approach for users to pass configuration options to these
+internal channels.
+gRPC willl support the xDS Extension Config Discovery Service (ECDS),
 
-## **Abstract**
+## Background
 
-This proposal introduces a mechanism to configure "child channels"—channels created internally by gRPC components (such as xDS control plane channel). Currently, these internal channels cannot easily inherit configuration (like metric sinks and interceptors) from the user application without relying on global state. This design proposes a language-specific approach to injecting configuration: using `functional interfaces` in Java, `DialOptions` in Go, and `ChannelArgs` in Core.
+Complex gRPC ecosystems often require the creation of auxiliary channels that
+are not directly instantiated by the user application. The primary examples are:
```
markdroth, line 23: I'd like to also have an example here of doing this from within an LB policy.  The obvious example is RLS, but since that isn't actually a public API, I guess we'll have to cite grpclb as the example instead.  However, let's note that we're just using grpclb as an example for how we will handle this plumbing in LB policies, and that this proposal does not actually mandate any behavior changes for grpclb specifically.  (In C-core, at least, there's a bunch of crufty legacy behavior for grpclb that I don't want to touch at this point, so we will not implement this proposal for grpclb.)

Are Java or Go going to have any problems plumbing the child channel options down into resolvers or LB policies?  @ejona86 @dfawley @easwars

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
-1. **xDS (Extensible Discovery Service)**: When a user creates a channel with an xDS target, the gRPC library internally creates a separate channel to communicate with the xDS control plane. Currently, users have limited ability to configure this internal control plane channel.
-2. **Advanced Load Balancing (RLS, GrpcLB):** Policies like RLS (Route Lookup Service) and GrpcLB, as well as other high-level libraries built on top of gRPC, frequently create internal channels to communicate with look-aside load balancers or backends.
+* [A27: xDS-Based Global Load Balancing](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)
+* [A66: Otel Stats](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md)
+* [A72: OpenTelemetry Tracing](https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md)
+* [A92: xDS ExtAuthz Support](https://github.com/grpc/proposal/pull/481)
+* [A93: xDS ExtProc Support](https://github.com/grpc/proposal/pull/484)
 
-### **The Problem**
+### The Problem
```
markdroth: I don't think we need a separate subsection for this.  Instead, please move the text of this section up above the "Related Proposals" subsection.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
 
-There is currently no standardized way to configure behavior for these child channels.
+The primary motivation for this feature is the need to configure observability
+on a per-child-channel basis.
 
-* **Metrics**: Users need to configure metric sinks so that telemetry from internal channels can be read and exported.
-* **Interceptors**: Users may need to apply specific interceptors (e.g., for authentication, logging, or tracing) to internal traffic.
-* **No Global State**: These configurations cannot be set globally (e.g., using static singletons) because different parts of an application may require different configurations, such as different metric backends or security credentials.
+* StatsPlugins & Tracing: Users need to configure metric sinks (as described in
```
markdroth: The wording here is strange.  A StatsPlugin configures both metrics and tracing, so I think this should be written like this:

"""
StatsPlugins: Users use these plugins to configure metrics and tracing (as described in ...)
"""

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
-The proposal creates a "plugin" or configuration injection style for internal channels. The implementation varies by language to match existing idioms, but the goal remains consistent: allow the user to pass a configuration object or function that the internal channel factory applies during creation.
+## Proposal
 
-### **Java**
+We introduce the concept of **Child Channel Options**. This is a configuration
+container attached to a parent channel that is strictly designated for use by
+its children.
 
-In Java, the configuration will be achieved by accepting functions (callbacks). The API allows users to pass a `Consumer<ManagedChannelBuilder<?>>` (or a similar functional interface). When an internal library (e.g., xDS, RLS, gRPCLB) creates a child channel, it applies this user-provided function to the builder before building the channel.
+### Encapsulation
```
markdroth: I don't think we need these first 4 sub-headings (Encapsulation, Propagation, Precedence and Merging, and Shared Resources).  All of the text of those sub-headings can simply be right under the Proposal heading.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
+see [gRFC A27](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)).
+
+If multiple Parent Channels (`P1`, `P2`) point to the same xDS target but
+provide *different* Child Channel Options (`O_child1`, `O_child2`):
+
+* Behavior: The shared client is created using the options from the first parent
+  channel that triggers its creation (e.g., `O_child1`).
+* Subsequent Usage: When `P2` requests the client, it receives the existing
+  shared client. `O_child2` is effectively ignored for that specific shared
+  resource.
```
markdroth, line 113: @ejona86 @dfawley Should we define this specific behavior, or should we just say that it's up to the implementation?  I don't feel strongly either way, but I wanted to make sure we're all on the same page.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
+* `O_child` is carried in `P`'s state, available for extraction by internal
+  components.
+* The configuration provided by `O_child` is strictly uniform across all child
+  channels of a particular parent channel.
 
-  Use the standard `java.util.function.Consumer` and define a new public API interface, `ChildChannelConfigurer`, to encapsulate the configuration logic for auxiliary channels.
+### Propagation
+
+When an internal component (e.g., an xDS client factory or an auth filter)
+attached to `P` needs to create a Child Channel `C`:
```
markdroth: What about multi-level propagation?  For example, what if we create a channel to the ext_authz server, and that channel uses xDS?  We would then need a way to pass options to xDS channel within the ext_authz channel within the top-level gRPC channel.

Given the use-cases that we're targetting, I think we should say that `O_child` applies to all child channels, no matter how deeply nested.  In other words, when we apply `O_child` to channel `C`, we should also configure channel `C` to use `O_child` for *its* children.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
+  gRFC [A66](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md)
+  and [A72](https://github.com/grpc/proposal/blob/master/A72-open-telemetry-tracing.md))
+  so that telemetry from internal channels is correctly tagged and exported.
+* Interceptors: Users may need to apply specific interceptors (e.g., for
+  logging, or tracing) to internal traffic.
 
-## **Proposal**
+These configurations cannot be set globally because different parts of an
+application may require different configurations, such as different metric
+backends.
```
markdroth: This isn't actually the reason why we can't use the global stats plugin registry.  The use-case we care about is setting a StatsPlugin for one particular channel, in which case we want that same StatsPlugin to also be used for any child of that channel.

For example, let's say that we create two channels, one to target A and one to target B, both of which create their own child channel to ext_authz server target Z.  If we create the channel to target A with a specific StatsPlugin, then we want that StatsPlugin to also be used for the child channel to target Z created by the channel to target A.  We do *not* want it to be used for the child channel to target Z created by the channel to target B, because we did not configure the StatsPlugin for the channel to target B.

We can't do that from the global registry, for a couple of reasons.  First, the global registry can only select channels based on things like the target URI, so we would have to select it based on target Z, which would attach it to the child channels for the channels to both A and B.  And second, it's hard for the application that registers the global StatsPlugin to know what target URIs will be used for child channels.

Reading further, it looks like you talk about this in the "Rationale" section below.  I suggest moving all of this discussion there, and changing the text here to say something like "Global configuration is not sufficient for these use cases for the reasons described in the 'Rationale' section below."

A110-child-channel-plugins.md
```diff
@@ -148,68 +251,57 @@ In Go, configuration for child channels is be achieved by passing a specific `Di
   }
   ```
 
-### **Core (C/C++)**
+##### Core (C/C++)
```
markdroth: This needs one fewer `#`.

A110-child-channel-plugins.md
```diff
@@ -1,141 +1,244 @@
...
+These configurations cannot be set globally because different parts of an
+application may require different configurations, such as different metric
+backends.
 
-The proposal creates a "plugin" or configuration injection style for internal channels. The implementation varies by language to match existing idioms, but the goal remains consistent: allow the user to pass a configuration object or function that the internal channel factory applies during creation.
+## Proposal
 
-### **Java**
+We introduce the concept of **Child Channel Options**. This is a configuration
+container attached to a parent channel that is strictly designated for use by
```
markdroth: It's not just parent channels, it's also servers.  Please make that clear throughout this description.

A110-child-channel-plugins.md
```diff
@@ -148,68 +251,57 @@ In Go, configuration for child channels is be achieved by passing a specific `Di
...
+  // E.g., add a custom tracing interceptor specifically for child channels
+  child_args.SetPointer(GRPC_ARG_TRACING_PROVIDER, my_tracing_provider);
 
-We reject global configuration (static variables) because it prevents multi-tenant applications from isolating configurations. For example, one client may need to export metrics to Prometheus, while another in the same process exports to Cloud Monitoring.
+  grpc::ChannelArguments parent_args;
+  // Pass the nested args up 
+  parent_args.SetChildChannelArgs(child_channel_args);
+
+  std::shared_ptr<grpc::Channel> channel =
+      grpc::CreateCustomChannel("xds:///my-service", credentials, parent_args);
```
markdroth: There is no `GRPC_ARG_TRACING_PROVIDER` channel arg, and there is no interceptor involved here.

Here's how it will actually look in C++:

```c++
// Create OTel StatsPlugin.
grpc::OpenTelemetryPluginBuilder ot_plugin_builder;
// ...set options on builder...
auto ot_plugin = builder.Build();
assert(ot_plugin.ok());
// Add the StatsPlugin to both child args and parent args.
grpc::ChannelArguments child_args;
grpc::ChannelArguments parent_args;
ot_plugin->AddToChannelArguments(&child_args);
ot_plugin->AddToChannelArguments(&parent_args);
// Add child args to parent args.
parent_args.SetChildChannelArgs(child_args);
// Create channel with parent args.
auto channel = grpc::CreateCustomChannel(
    "xds:///my-service", credentials, parent_args);
```

And here's an example of how this will work on the server side:

```c++
grpc::ServerBuilder server_builder;
// Create OTel StatsPlugin.
grpc::OpenTelemetryPluginBuilder ot_plugin_builder;
// ...set options on builder...
auto ot_plugin = builder.Build();
assert(ot_plugin.ok());
// Add the StatsPlugin to both child args and the server builder.
grpc::ChannelArguments child_args;
ot_plugin->AddToChannelArguments(&child_args);
ot_plugin->AddToServerBuilder(&server_builder);
// Add the child args to the server builder.
server_builder.SetChildChannelArgs(child_args);
// Start the server.
auto server = server_builder.BuildAndStart();
```

A110-child-channel-plugins.md
```diff
@@ -148,68 +251,57 @@ In Go, configuration for child channels is be achieved by passing a specific `Di
...
-      "xds:///my-service", 
-      grpc::InsecureChannelCredentials(), 
-      grpc::ChannelArguments::FromC(parent_args)
-  );
+  We add a helper method to the C++ `ChannelArguments` class to simplify packing
+  the nested arguments safely.
 
+  ```cpp
+    // Sets the channel arguments to be used for child channels.
+    void SetChildChannelArgs(const ChannelArguments& args);
```
markdroth, line 353: We also need to add this same method on the `ServerBuilder` object for the server side.

A110-child-channel-plugins.md
markdroth: There is no concept of an out-of-band channel in C-core.  But it looks like you've removed this anyway.

markdroth (commented):
This is looking much better!

A110-child-channel-plugins.md
```diff
@@ -92,18 +85,16 @@ configuration `O_internal` (e.g., target URIs, or internal interceptors).
...
-on the target URI (
-see [gRFC A27](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)).
+often pooled and shared across multiple parent channels or servers within a
+process based on the target URI (see
+[gRFC A27](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)).
 
-If multiple Parent Channels (`P1`, `P2`) point to the same xDS target but
-provide *different* Child Channel Options (`O_child1`, `O_child2`):
+If multiple Parent Channels/Servers (`P1`, `P2`) point to the same xDS target
+but provide *different* Child Channel Options (`O_child1`, `O_child2`):
```
markdroth: Suggest wording this to make it clear that xDS is just one example of a generally applicable pattern that may pop up eslewhere in the future:

"""
In some cases, child channels may be shared across multiple parent channels/servers.  For example, the xDS control plane channel is shared across multiple channels or servers as described in [A27].  However, it is possible for each parent channel or server to be created with different child options.  In such cases, we will do the following:
"""

A110-child-channel-plugins.md
```diff
@@ -251,15 +242,15 @@ into these internal channels from both entry points.
...
 In gRPC Core, we utilize the existing `ChannelArgs` mechanism recursively to
 pass configuration to internal channels. We define a standard argument key whose
 value is a pointer to another `grpc_channel_args` structure. This "Nested
-Arguments" pattern allows the parent channel to carry a specific subset of
-arguments intended solely for its children.
+Arguments" pattern allows the parent channel or server to carry a specific
+subset of arguments intended solely for its children.
 
-* ##### Configuration Mechanism
+* #### Configuration Mechanism
```
markdroth: This needs one more `#` character, so that it's nested underneath the "Core (C/C++)" section.

A110-child-channel-plugins.md
```diff
@@ -271,37 +262,77 @@ arguments intended solely for its children.
   #define GRPC_ARG_CHILD_CHANNEL_ARGS "grpc.child_channel.args"
   ```
 
-* **API Changes**
+* #### API Changes
```
markdroth: This needs one more `#` character.

A110-child-channel-plugins.md
```diff
@@ -271,37 +262,77 @@ arguments intended solely for its children.
...
+  We add a helper method to the C++ `ChannelArguments` and `ServerBuilder`
+  classes to simplify packing the nested arguments safely.
 
   ```cpp
     // Sets the channel arguments to be used for child channels.
     void SetChildChannelArgs(const ChannelArguments& args);
   ```
 
-* ##### Usage Example (User-Side Code)
+* #### Usage Example (User-Side Code)
```
markdroth: This needs one more `#` character.

markdroth (commented):
This is looking really good!  The only really substantive remaining comment from me is the one from my last review pass about making sure we can plumb this into channels created via an LB policy or resolver.

A110-child-channel-plugins.md
```diff
@@ -85,13 +85,16 @@ configuration `O_internal` (e.g., target URIs, or internal interceptors).
...
-Certain internal channels, specifically the **xDS Control Plane Client**, are
-often pooled and shared across multiple parent channels or servers within a
-process based on the target URI (see
-[gRFC A27](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md)).
+In some cases, child channels may be shared across multiple parent
+channels/servers. For example, the xDS control plane channel is shared across
+multiple channels or servers as described
+in [gRFC A27](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md).
+However, it is possible for each parent channel or server to be created with
+different child options. In such cases, we will do the following:
```
markdroth: If you're keeping the following paragraph, we don't need the "In such cases, we will do the following:" sentence here.

A110-child-channel-plugins.md
ejona86: In Java it seems we're settling on a function in the Helper where you pass it a ChannelBuilder and it sets whatever options it wants (which is the same API we're exposing on the ChannelBuilder, where the caller provides such a function). In Go/C that's be similar to a getter than returns options, that the caller then combines with their own.

A110-child-channel-plugins.md
ejona86: I think I'm in the same boat, of not being too particular. Up to this point we've defined it per-use-case.

A110-child-channel-plugins.md
markdroth: That approach seems fine.  @easwars, I assume Go would do something similar?

In C-core, I don't think we even need a special getter, because the child channel args will themselves just be a channel arg in the parent channel, and those channel args are already plumbed down into the LB policies.  Any concerns about that?

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,341 @@
...
+2. It applies `O_child` to the configuration of `C`.
+3. It should also configure channel `C` to use `O_child` for its children.
+
+The Child Channel `C` typically requires some internal
+configuration `O_internal` (e.g., target URIs, or internal interceptors).
+
+* Merge Rule: `O_child` and `O_internal` are merged. If the environment supports
+  global channel options, `O_child` options override global channel options.
+* Conflict Resolution: Mandatory internal settings (`O_internal`) generally take
+  precedence over user-provided child options (`O_child`) to ensure correctness.
```
easwars, line 96: Why is the user-provided options given less precedence here? There is more than one example of existing options where the user-provided option overrides values provided by things like the name resolver or the credentials. The reasoning being the user knows what they are doing, and if they are doing something stupid, we throw our hands up in the air.

A110-child-channel-plugins.md
easwars: In Go, the Channel passes a `BuildOptions` struct when creating a resolver or an LB policy. The struct is different for the two cases. But currently it contains things like credentials, dialer etc that the resolver and balancer should use if and when they are creating a Channel themselves.

We *could* add a new field in this struct to contain the child Channel options. As a long-term cleanup, we *could* deprecate some of the existing fields and move them to the newly added options field as well.

For the server, it should be simpler as the xDS-enabled server *is* the one that attempts to create the xDS client.

A110-child-channel-plugins.md
ejona86: @easwars, what are the examples you were thinking about?

I think this rule is basically the same as what we do for global options. The global sets new defaults, but individual call sites can still override. Since we're expecting to inject this child options stuff with global configuration, it seems we should use the same resolution order for the same reasons: individual call sites know better than a canned config that is the same for multiple channels.

Maybe the confusion is just "what is `O_internal` talking about?"

easwars: The example that I had in mind was how we determine the value of the `:authority` header to use. At least in Go, the name resolver can specify a value for the address, while the credentials can also override it, and the user can specify as override as well. In this case we prefer the value specified by the user.

But I get your point in this case. Here the value specified by the user is the global option while the value specified in the xDS configuration (for example) is more specific, and therefore is to be preferred.

ejona86 (approved):

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,369 @@
...
+    ```go
+    // WithChildDialOptions returns a ServerOption that specifies a list of 
+    // DialOptions to be applied to the server's internal child channels 
+    // (e.g., the xDS control plane connection).
+    func WithChildDialOptions(opts ...DialOption) ServerOption {
+        return newFuncServerOption(func(o *serverOptions) {
+            o.childDialOptions = opts
+        })
+    }
+    ```
```
easwars, line 272: Nit: Server options usually don't being with the `With` prefix. So, this can be named `ChildChannelOptions` instead.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,369 @@
...
+* [A92: xDS ExtAuthz Support](https://github.com/grpc/proposal/pull/481)
+* [A93: xDS ExtProc Support](https://github.com/grpc/proposal/pull/484)
+
+## Proposal
+
+We introduce the concept of **Child Channel Options**. This is a configuration
+container attached to a parent channel/server that is strictly designated for
+use by its children.
+
+The user API must allow "nesting" of channel options. A user creating a Parent
```
easwars: What nesting are we talking about here? It looks like the options are only propagated one level deep. Am I missing something?

AgraVator: The options are actually intended to propagate recursively to all child channels, no matter how deeply nested, to support cases like an ext_authz child channel that itself uses the xds:/// scheme to connect to its target (which in turn creates an internal control plane child channel). Have updated the proposal to clarify this.

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,363 @@
...
+pass configuration options to these child channels.
+
+## Background
+
+Complex gRPC ecosystems often require the creation of auxiliary channels that
+are not directly instantiated by the user application. The primary examples are:
+
+1. xDS: When a user creates a channel with an xDS target, the gRPC library
+   internally creates a separate channel to communicate with the xDS control
+   plane.
```
easwars: The XdsClient has the ability to create gRPC channels to multiple xDS management servers (as part of the support we added for federation). In Go (and I believe in C++), these gRPC channels are created lazily, i.e., at the time when a resource needs to be requested from management A, a gRPC channel is created to it.

Now, how does this work when we have multiple parent channels. Will the child options of the first parent channel, `P1`,  that creates the XdsClient be applied to all child channels (even though `P1` might not have anything to do with a child channel that was created when another parent channel `P2` requested a resource from a different management server).

AgraVator: I think this was addressed at https://github.com/grpc/proposal/pull/529/changes#diff-8a3140e3e8002e31a20d2188875d3f70478ae6a195a4bbb7a48ebedd604ced90R90-R105. Do let me know if u meant something else.

easwars: I had seen that. But I felt it didn't quite address the nuance involved with the XdsClient in the sense that it can create multiple child channels, and not all of them are created at the same time and not all of them as a result of resource watches created by a single parent.

easwars (approved):

A110-child-channel-plugins.md
```diff
@@ -102,70 +104,72 @@ Options (`O_child1`, `O_child2`):
...
 
+**LB Policies and Resolvers**
+
+Some LB policies and resolvers may need to create child channels. We use `grpclb` as an example for how this plumbing will be handled in LB policies. Note that this proposal does not mandate any behavior changes for `grpclb` specifically.
+
+To support this, the child channel options must be plumbed down into resolvers and LB policies. Here are examples of how a component like `grpclb` would use this plumbing:
+
+* Java: The `Helper` will provide a function that accepts a `ChannelBuilder` and applies the child channel options to it.
+* Go: A new field will be added to the `BuildOptions` struct (passed when creating a resolver or LB policy) to contain the child channel options.
+* C-core: No special plumbing is needed because the child channel args are simply passed as channel arguments, which are already available to LB policies.
```
markdroth: We should also say that the LB policy will need to propagate both the individual child channel args and the arg containing the child channel args to the child channel.

A110-child-channel-plugins.md
markdroth: I think the behavior we'd want here is to pass the child channel options to the XdsClient when the XdsClient is created, and then that XdsClient instance will use those same child channel options for any child channel it creates over the lifetime of the XdsClient.

I don't see any other way this could reasonably work, because after the XdsClient is created, there's no other place where we'd pass down channel options to use from a "parent" channel.

markdroth (approved):
Just one small remaining comment, otherwise looks great!

A110-child-channel-plugins.md
```diff
@@ -0,0 +1,383 @@
...
+  components.
+* The configuration provided by `O_child` is strictly uniform across all child
+  channels of a particular parent channel/server.
+
+When an internal component (e.g., an xDS client factory) attached to `P`
+needs to create a Child Channel `C`:
+
+1. It retrieves `O_child` from `P`.
+2. It applies `O_child` to the configuration of `C`.
+3. It should also configure channel `C` to use `O_child` for its children.
```
ejona86, line 81: In "LB Policies and Resolvers", C-core has it pretty explicit which component is responsible for this ("configure channel C to use O_child"). It isn't clear to me which internal component is responsible in Go and Java. And looking at https://github.com/grpc/grpc-java/pull/12578, I don't see this done at all.

kannanjgithub: In `GrpcXdsTransportFactory`, the constructor for `GrpcXdsTransport` after configuring the xDS control plane child channel using channelConfigurator, should pass it recursively:
```
ManagedChannelBuilder<?> channelBuilder = Grpc.newChannelBuilder(target, channelCredentials)
            .keepAliveTime(5, TimeUnit.MINUTES);
channelConfigurator.configureChannelBuilder(channelBuilder);
// Missing: channelBuilder.childChannelConfigurator(channelConfigurator);
```
In `ManagedChannelImpl`, inside `createResolvingOobChannelBuilder` helper, out-of-band channels are created and configured, but the configurator is not recursively set:
```
    ResolvingOobChannelBuilder builder = new ResolvingOobChannelBuilder();
   // Configures the channel C (the OOB channel)
   channelConfigurator.configureChannelBuilder(builder);
     
   // Missing: builder.childChannelConfigurator(channelConfigurator);
```

AgraVator: updated, was waiting to merge this before updating in https://github.com/grpc/grpc-java/pull/12578

A110-child-channel-plugins.md
```diff
@@ -119,40 +119,63 @@ Some LB policies and resolvers may need to create child channels. We use
...
-  propagate both the individual child channel args and the
-  `GRPC_ARG_CHILD_CHANNEL_ARGS` argument containing the child channel args to
-  the child channel.
+To support this, the child channel options must be plumbed down into
+resolvers and LB policies. Each internal component that creates a child
+channel `C` is explicitly responsible for applying and propagating those
+options:
+
+* Java: When an LB policy creates an out-of-band (OOB) child channel via
+  `LoadBalancer.Helper` (e.g., `createResolvingOobChannelBuilder()` or
```
ejona86, line 128: Given the current `ManagedChannelBuilder.childChannelConfigurator()` API, I suspect we'd want "whatever is calling build() should be applying the child config." Otherwise, that component can't compose its own configuration to child channels; if it calls childChannelConfigurator() it will overwrite any previous configuration.

Either we could move the responsibility to whatever component calls build(), or we could have `managedChannelBuilder.childChannelConfigurator()` _append_ the provided configurator. It currently overwrites the previous configurator. (Note that we don't have to have a list, we could just make a new configurator that calls the old configurator first, then the newly-passed configurator.)

Note that Go and C's approaches here allow a component to supplement the child configuration; only Java and only createResolvingOobChannelBuilder() has the limitation.

@kannanjgithub, thoughts?

AgraVator: Yeah, we could do something like below for composition

```java
@Override
public ManagedChannelImplBuilder childChannelConfigurator(ChannelConfigurator channelConfigurator) {
  checkNotNull(channelConfigurator, "channelConfigurator");
  ChannelConfigurator previous = this.channelConfigurator;
  this.channelConfigurator = builder -> {
    previous.configureChannelBuilder(builder);
    channelConfigurator.configureChannelBuilder(builder);
  };
  return this;
}

kannanjgithub: Agree, the chained composition approach is simpler and preserves any previously applied child channel configurator.

PR #536, opened 2026-01-30
A114: WRR Support for Custom Backend Metrics

sauravzg:
This proposal allows configuring the weighted_round_robin policy to use custom backend metrics (via ORCA) for computing endpoint weights, instead of relying solely on standard utilization metrics. This aligns gRPC with xDS custom backend metric support.

### DIFF
--- a/A114-wrr-metric-names-for-computing-utilization.md
+++ b/A114-wrr-metric-names-for-computing-utilization.md
@@ -1,0 +1,165 @@
+# A114: WRR Support for Custom Backend Metrics
+
+- Author(s): sauravzg
+- Approver: markdroth
+- Status: In review
+- Implemented in:
+- Last updated: 2026-01-30
+- Discussion at:
+
+## Abstract
+
+This proposal updates the client-side `weighted_round_robin` (WRR) load balancing policy to support customizable utilization metrics. It adds a new configuration field `metric_names_for_computing_utilization` to the WRR LB policy config. This allows users to specify which backend metrics should be used to compute endpoint weights, enabling the use of custom metrics (via ORCA named metrics) instead of relying solely on the default `application_utilization` or `cpu_utilization`.
+
+## Background
+
+The existing `weighted_round_robin` policy (defined in [gRFC A58][A58]) calculates endpoint weights based on standard metrics provided by the backend via ORCA (Open Request Cost Aggregation) load reports. Specifically, it uses `application_utilization` if available, and falls back to `cpu_utilization`.
+
+However, services may want to drive load balancing decisions based on other resources, such as memory utilization, queue depth, or custom application-defined metrics. The [xDS Custom Backend Metrics][A51] specification (ORCA) supports reporting arbitrary named metrics, and xDS has updated its WRR implementation to allow selecting these metrics for utilization calculation.
+
+To support advanced load balancing scenarios, gRPC's WRR policy needs to support this flexibility.
+
+### Related Proposals
+
+- [A58: Client-Side Weighted Round Robin LB Policy][A58]
+- [A51: Custom Backend Metrics][A51]
+
+## Proposal
+
+### Service Config Update
+
+We will add a new field `metric_names_for_computing_utilization` to the [`WeightedRoundRobinLbConfig`][WeightedRoundRobinLbConfigProto] message in the Service Config.
+
+```protobuf
+message WeightedRoundRobinLbConfig {
+  // ... existing fields ...
+
+  // A list of metric names to use for computing utilization.
+  //
+  // By default, endpoint weight is computed based on the 'application_utilization'
+  // field reported by the endpoint.
+  //
+  // If 'application_utilization' is not set, then utilization will instead be
+  // computed by taking the max of the values of the metrics specified here.
+  //
+  // For map fields in the ORCA proto, the string will be of the form
+  // "<map_field_name>.<map_key>". For example, the string "named_metrics.foo"
+  // will mean to look for the key "foo" in the ORCA "named_metrics" field.
+  //
+  // If none of the specified metrics are present in the load report, then
+  // 'cpu_utilization' is used instead.
+  repeated string metric_names_for_computing_utilization = 7;
+}
+```
+
+### Weight Calculation Logic
+
+The weight calculation logic in the WRR policy will be updated to determine the `utilization` value as follows from the [`OrcaLoadReport`][OrcaLoadReportProto]
+
+1.  **Check `application_utilization`**: If the backend reports `application_utilization` (value > 0), use it.
+2.  **Check Custom Metrics**: If `application_utilization` is not reported (or is 0), and `metric_names_for_computing_utilization` is configured:
+    - Iterate through the specified metric names.
+    - **Resolve Metric Value**:
+      - If the name is of the format `field.key` (e.g., `named_metrics.foo`), look up the map field `field` and retrieve the value for `key`.
+      - If the name is a simple field name (e.g., `cpu_utilization`, `mem_utilization`), look up the `field`.
+    - **Compute Max**: Track the maximum value among all successfully resolved, positive ( > 0), finite metrics.
+    - If a max value is found, use it as the `utilization`.
+3.  **Fallback to `cpu_utilization`**: If neither of the above yields a value, use `cpu_utilization` (utilizing the existing logic).
+
+#### Pseudocode
+
+```
+function GetUtilization(report, configured_metrics):
+  # 1. Prefer application_utilization if present
+  if report.application_utilization > 0:
+    return report.application_utilization
+
+  # 2. Check Custom Metrics
+  max_util = null
+
+  for metric_name in configured_metrics:
+    value = null
+
+    if metric_name contains ".":
+      # Map lookup (e.g. "named_metrics.foo" -> map="named_metrics", key="foo")
+      map_name, key = split_on_first_dot(metric_name)
+      if report has map field map_name:
+         value = report[map_name][key]
+    else:
+      # Root field lookup (e.g. "mem_utilization") via Reflection
+      if report has field metric_name:
+         value = report[metric_name]
+
+    # Only consider valid, non-nan, positive values
+    if value is not null and !is_nan(value) and value > 0:
+      if max_util is null or value > max_util:
+        max_util = value
+
+  if max_util is not null:
+    return max_util
+
+  # 3. Fallback
+  return report.cpu_utilization
+```
+
+#### Implementation Notes
+
+Since `OrcaLoadReport` is often exposed as a language-specific proxy object rather than a raw Protobuf message (e.g., in Java and C++), implementations are **not** expected to use Protobuf reflection to look up arbitrary fields. Instead, implementations should manually handle:
+
+- **Map Fields**: Lookups in the `named_metrics` map (e.g., `named_metrics.foo`). The string is split on the first dot: `foo.bar.baz` looks up key `bar.baz` in map `foo`.
+- **Standard Fields**: Explicit lookups for known fields (e.g., `cpu_utilization`).
+
+Support for any new standard fields added to `OrcaLoadReport` in the future will require explicit code changes in the implementation. This behavior is consistent with the [current Envoy implementation](https://github.com/envoyproxy/envoy/blob/35749578db375f5fe8ac5dd293cb7c4efb689611/source/common/orca/orca_load_metrics.cc#L47-L73).
+
+#### Validity and Edge Cases
+
+- **Nan Values**: As shown above, `NaN` values in reports are explicitly ignored to prevent undefined behavior in weight calculations. This is relevant because behavior of `max` on `NaN` in c++ is inconsistent based on order.
+- **Zero and Negative Values**: Any value <= 0.0 is treated as missing and ignored. This is consistent with [gRFC A58][A58] and the [Envoy implementation](https://github.com/envoyproxy/envoy/blob/113f2c2d1015913256a2a8c6f9f97d0622623f45/source/extensions/load_balancing_policies/client_side_weighted_round_robin/client_side_weighted_round_robin_lb.cc#L214-L218).
+- **Bound Checks**: The final selected `utilization` is subject to the standard validation logic from [gRFC A58][A58] (e.g., ensuring the value is positive) before being used to compute weight to avoid undefined behavior in weight calculations.
+- **Normalization**: The WRR policy does not normalize reported metrics; the application is responsible for this.
+
+The rest of the weight calculation formula (using QPS, EPS, and penalty) from [gRFC A58][A58] remains unchanged.
+
+### xDS Integration
+
+We will support the `metric_names_for_computing_utilization` field in the xDS [`ClientSideWeightedRoundRobin`][ClientSideWeightedRoundRobinProto] policy.
+
+When converting the xDS configuration to the gRPC Service Config `WeightedRoundRobinLbConfig`, the `metric_names_for_computing_utilization` field should be copied over directly.
+
+### Temporary environment variable protection
+
+The features described in this proposal will be guarded by the environment variable `GRPC_EXPERIMENTAL_WRR_CUSTOM_METRICS`, which defaults to `false`.
+
+## Rationale
+
+- **Consistency with Envoy**: This design mirrors the corresponding feature in Envoy, ensuring consistent behavior for xDS-controlled clients.
+- **Flexibility**: Allows users to define load balancing weights based on the actual bottleneck resource of their application (e.g., memory-bound services).
+- **Backward Compatibility**: The default behavior (using `application_utilization` or `cpu_utilization`) remains unchanged if the new field is not configured.
+
+## Implementation
+
+This will be implemented in all languages C++, Java, and Go.
+
+### C++
+
+- **xDS Integration**: Update [`ClientSideWeightedRoundRobinLbPolicyConfigFactory`](https://github.com/grpc/grpc/blob/f7f13023412c1a589af7558eb0b9f8f664a76431/src/core/xds/grpc/xds_lb_policy_registry.cc#L68) to copy over the new field from the xDS configuration.
+- **Config**: Update [`WeightedRoundRobinLbConfig`](https://github.com/grpc/grpc/blob/f7f13023412c1a589af7558eb0b9f8f664a76431/src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc#L133) to include `metric_names_for_computing_utilization`.
+- **Weight Calculation Logic**: Update callers of [`EndpointWeight::MaybeUpdateWeight`](https://github.com/grpc/grpc/blob/f7f13023412c1a589af7558eb0b9f8f664a76431/src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc#L212) to implement the new utilisation selection logic.
+
+### Java
+
+- **xDS Integration**: Update [`convertWeightedRoundRobinConfig`](https://github.com/grpc/grpc-java/blob/a9f73f4c0aa5617aa2b6ae6ac805693915899b6a/xds/src/main/java/io/grpc/xds/LoadBalancerConfigFactory.java#L286) to copy over the field from the xDS configuration.
+- **Config**: Update [`WeightedRoundRobinLoadBalancerConfig`](https://github.com/grpc/grpc-java/blob/a9f73f4c0aa5617aa2b6ae6ac805693915899b6a/xds/src/main/java/io/grpc/xds/WeightedRoundRobinLoadBalancer.java#L716) to include `metric_names_for_computing_utilization`.
+- **Weight Calculation Logic**: Update [`OrcaPerRequestUtil`](https://github.com/grpc/grpc-java/blob/a9f73f4c0aa5617aa2b6ae6ac805693915899b6a/xds/src/main/java/io/grpc/xds/WeightedRoundRobinLoadBalancer.java#L364) to implement the new utilisation selection logic.
+
+### Go
+
+- **xDS Integration**: Update the configuration struct and conversion logic in [`converter.go`](https://github.com/grpc/grpc-go/blob/c05cfb3693bf18086810e671a6f9c05f296e0183/internal/xds/xdsclient/xdslbregistry/converter/converter.go#L241) to copy over the new field from the xDS configuration.
+- **Config**: Update the configuration struct in [`config.go`](https://github.com/grpc/grpc-go/blob/7985bb44d26ecbeb8950996d028e38a0de08070b/balancer/weightedroundrobin/config.go#L26) to include `metric_names_for_computing_utilization`.
+- **Weight Calculation Logic**: Update the weight update function in [`balancer.go`](https://github.com/grpc/grpc-go/blob/7985bb44d26ecbeb8950996d028e38a0de08070b/balancer/weightedroundrobin/balancer.go#L546) to implement the new utilisation selection logic.
+
+[A58]: A58-client-side-weighted-round-robin-lb-policy.md
+[A51]: A51-custom-backend-metrics.md
+[ClientSideWeightedRoundRobinProto]: https://github.com/envoyproxy/envoy/blob/7242d5ad170523d7936849e596d261e3502c3886/api/envoy/extensions/load_balancing_policies/client_side_weighted_round_robin/v3/client_side_weighted_round_robin.proto#L86
+[WeightedRoundRobinLbConfigProto]: https://github.com/grpc/grpc-proto/blob/ec99424f3b7dae9db194f848b4cea52ecfae07af/grpc/service_config/service_config.proto#L205
+[OrcaLoadReportProto]: https://github.com/cncf/xds/blob/0feb69152e9f7e8a45c8a3cfe8c7dd93bca3512f/xds/data/orca/v3/orca_load_report.proto#L15
--- a/A114-wrr-metric-names-for-computing-utilization.md
+++ b/A114-wrr-metric-names-for-computing-utilization.md
@@ -34,17 +34,17 @@ We will add a new field `metric_names_for_computing_utilization` to the [`Weight
 message WeightedRoundRobinLbConfig {
   // ... existing fields ...
 
-  // A list of metric names to use for computing utilization.
+  // A list of metrics to be considered for overriding the default utilization
+  // computation behavior.
   //
   // For map fields in the ORCA proto, the string will be of the form
   // "<map_field_name>.<map_key>". For example, the string "named_metrics.foo"
   // will mean to look for the key "foo" in the ORCA "named_metrics" field.
   //
-  // By default, utilization is computed by taking the max of the values of the
-  // metrics specified here. If none of the specified metrics are present in the
-  // load report (or if this field is empty, or all values are less than or
-  // equal to 0), the policy will fall back to using 'application_utilization'
-  // and then 'cpu_utilization'.
+  // Utilization is computed by taking the max of the values of the
+  // metrics specified here. In the absence of this field or absence of
+  // valid values for the specified metrics, the policy will fall back to the
+  // existing default behavior.
   repeated string metric_names_for_computing_utilization = 7;
 }
 ```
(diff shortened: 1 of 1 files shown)


A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,165 @@
...
+## Abstract
+
+This proposal updates the client-side `weighted_round_robin` (WRR) load balancing policy to support customizable utilization metrics. It adds a new configuration field `metric_names_for_computing_utilization` to the WRR LB policy config. This allows users to specify which backend metrics should be used to compute endpoint weights, enabling the use of custom metrics (via ORCA named metrics) instead of relying solely on the default `application_utilization` or `cpu_utilization`.
+
+## Background
+
+The existing `weighted_round_robin` policy (defined in [gRFC A58][A58]) calculates endpoint weights based on standard metrics provided by the backend via ORCA (Open Request Cost Aggregation) load reports. Specifically, it uses `application_utilization` if available, and falls back to `cpu_utilization`.
+
+However, services may want to drive load balancing decisions based on other resources, such as memory utilization, queue depth, or custom application-defined metrics. The [xDS Custom Backend Metrics][A51] specification (ORCA) supports reporting arbitrary named metrics, and xDS has updated its WRR implementation to allow selecting these metrics for utilization calculation.
```
markdroth: A51 has nothing to do with xDS.  The link to A51 should say something like "The existing gRPC custom backend metric support (defined in gRFC A51)".

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,165 @@
...
+### Weight Calculation Logic
+
+The weight calculation logic in the WRR policy will be updated to determine the `utilization` value as follows from the [`OrcaLoadReport`][OrcaLoadReportProto]
+
+1.  **Check `application_utilization`**: If the backend reports `application_utilization` (value > 0), use it.
+2.  **Check Custom Metrics**: If `application_utilization` is not reported (or is 0), and `metric_names_for_computing_utilization` is configured:
+    - Iterate through the specified metric names.
+    - **Resolve Metric Value**:
+      - If the name is of the format `field.key` (e.g., `named_metrics.foo`), look up the map field `field` and retrieve the value for `key`.
+      - If the name is a simple field name (e.g., `cpu_utilization`, `mem_utilization`), look up the `field`.
```
markdroth, line 62: I don't think we want to use proto reflection here, so we're going to wind up hard-coding a specific set of fields to support.  Let's explicitly document those fields.

I think we should support the following:
- cpu_utilization
- mem_utilization
- application_utilization
- named_metrics.*

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,165 @@
...
+  # 3. Fallback
+  return report.cpu_utilization
+```
+
+#### Implementation Notes
+
+Since `OrcaLoadReport` is often exposed as a language-specific proxy object rather than a raw Protobuf message (e.g., in Java and C++), implementations are **not** expected to use Protobuf reflection to look up arbitrary fields. Instead, implementations should manually handle:
+
+- **Map Fields**: Lookups in the `named_metrics` map (e.g., `named_metrics.foo`). The string is split on the first dot: `foo.bar.baz` looks up key `bar.baz` in map `foo`.
+- **Standard Fields**: Explicit lookups for known fields (e.g., `cpu_utilization`).
```
markdroth: As above, please document the specific set of fields that we will support.

markdroth (commented):
This looks really good!  I have just a few minor comments.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: I've specified the exact list, but would like to understand why do we want to limit ourselves to these particular fields instead of supporting all the intersection of fields currently present in the orcaloadreport proxy objects across languages?  Initial guess would be that we currently only need these, and people can workaround/extend by moving the other fields manually into `named_metrics` ?

markdroth: WRR can only use utilization values, so any field that doesn't represent utilization doesn't make sense here.  This includes `rps_fractional`, `eps`, and `request_cost`.

The `utilization` map does represent utilization, but it was defined such that values in the map must be no greater than 1.0, and it turns out that there are over-subscription cases where we need values higher than that, which is one of the reasons why we added the `named_metrics` map to replace it.  But that having been said, I'd be okay with supporting the `utilization` map here in case someone wants it for some reason.

markdroth (approved):
I'd like @ejona86 and @dfawley to take a quick pass on this before we merge it.

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,167 @@
...
+  // 'cpu_utilization' is used instead.
+  repeated string metric_names_for_computing_utilization = 7;
+}
+```
+
+### Weight Calculation Logic
+
+The weight calculation logic in the WRR policy will be updated to determine the `utilization` value as follows from the [`OrcaLoadReport`][OrcaLoadReportProto]
+
+1.  **Check `application_utilization`**: If the backend reports `application_utilization` (value > 0), use it.
```
dfawley: So if the real `application_utilization` field truly means zero utilization for the server, then we will ignore it instead?  That doesn't seem ideal.

A58 doesn't say this -- it just says "if set".  But Go's implementation checks for 0, matching this wording: https://github.com/grpc/grpc-go/blob/7136e99ee323c26984174eb3cec85c201fef9946/balancer/weightedroundrobin/balancer.go#L551-L552

Presumably because it's not `optional`:

https://github.com/cncf/xds/blob/dba9d589def2cd10099a3a64887d859188c2f57a/xds/data/orca/v3/orca_load_report.proto#L57

Should we update the documentation in the ORCA proto to say something like "A non-zero (epsilon) value should be used if the application utilization is truly zero, to prevent consumers of this field from interpreting 0 as unset"?

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,167 @@
...
+2.  **Check Custom Metrics**: If `application_utilization` is not reported (or is 0), and `metric_names_for_computing_utilization` is configured:
+    - Iterate through the specified metric names.
+    - **Resolve Metric Value**:
+      - If the name is of the format `field.key` (e.g., `named_metrics.foo`), look up the map field `field` and retrieve the value for `key`.
+      - If the name is a simple field name (e.g., `cpu_utilization`, `mem_utilization`), look up the `field`.
+      - **Only the following fields are supported:**
+        - `application_utilization`
+        - `cpu_utilization`
```
dfawley, line 65: Surely there's no reason to include `application_utilization` since we know it must be zero in order for us to be following this path.  I guess `cpu_utilization` has an argument for being useful, but it seems at least questionable due to the next fallback case.

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,167 @@
...
+        - `mem_utilization`
+        - `named_metrics.*`
+    - **Compute Max**: Track the maximum value among all successfully resolved, positive ( > 0), finite metrics.
+    - If a max value is found, use it as the `utilization`.
+3.  **Fallback to `cpu_utilization`**: If neither of the above yields a value, use `cpu_utilization` (utilizing the existing logic).
```
dfawley: So here again a zero for the computation above implies "no value" and not "zero utilization"?

At least with custom metrics in the map we can distinguish presence vs set-to-zero.

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,167 @@
...
+### Related Proposals
+
+- [A58: Client-Side Weighted Round Robin LB Policy][A58]
+- [A51: Custom Backend Metrics][A51]
+
+## Proposal
+
+### Service Config Update
+
+We will add a new field `metric_names_for_computing_utilization` to the [`WeightedRoundRobinLbConfig`][WeightedRoundRobinLbConfigProto] message in the Service Config.
```
dfawley, line 31: Is it too late for bikeshedding on the field name?  Maybe `utilization_metrics` or `fallback_utilization_metrics` instead would be more concise?

sauravzg: I don't mind. Are you okay with `utilization_metric_names` ?

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: See the other comment, 0 utilization breaks the math.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Agreed, I had the same question and didn't feel it was ideal.

A58 to some degree does say this at https://github.com/grpc/proposal/blob/86990145a7cef9e5473a132709b2556fec00c4c6/A58-client-side-weighted-round-robin-lb-policy.md?plain=1#L137 where it requires utilisation to be > 0 

I am guessing there are two reasons
1. The field is not optional , so there's no way to distinguish between 0 and a not present value.
2. Weight computation will return `dividebyzero` exception on zero utilisation.

That is also a fair call, although I do think the field itself needs to be optional to allow 0 being a valid value, but I don't truly understand how to handle zero as a valid value and send more traffic to it without sizable changes in the implementation across all languages. 

So, maybe handling `0` as a valid value needs to be its own RFC.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Agreed on both to some degree.
That leaves us with essentially supporting `mem_utilization` and `named_metrics` , and maybe utilisation map.
I'll close off this comment and move it off to Mark's comment to centralise this.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Joining the conversation here to allign on the fields that we'd like to support.

@dfawley raised concerns about removing `application_utilization` and `cpu_utilization` . I tend to agree that supporting `application_utilization` is probably moot given that if we ever reach custom metrics it's guaranteed to be zero.

There is value in supporting `cpu_utilization` since it allows a max(utilization) paradigm across different types of utilizations like cpu and memory, picking whichever is worst as the true utilization.

The problem about `utilization` map not going over 1.0 may also apply to `memory_utilization` since that can also go over 1.0. 

So, to summarize , is it okay to allign on the following metric list 
- cpu_utilization
- mem_utilization
- utilization.*
- named_metrics.*

with the caveat that grpc is not responsible for validating the quality of values beyond undefined behaviors like nan, 0.

A114-wrr-metric-names-for-computing-utilization.md
dfawley: Making it optional would be a breaking change (at least, Go's API will change in a non-backward compatible way).

That probably means `0` can't ever be a valid value.  So I think the best we can probably do here is document it in the envoy protos like I suggested above.

A114-wrr-metric-names-for-computing-utilization.md
dfawley: What does `0` result in? Does it not use the endpoint at all if that happens?

A114-wrr-metric-names-for-computing-utilization.md
dfawley: That's fine by me.

A114-wrr-metric-names-for-computing-utilization.md
dfawley: I'm fine with this.

A114-wrr-metric-names-for-computing-utilization.md
markdroth: I'd prefer to keep the same name as used in the xDS protos to avoid confusion.

A114-wrr-metric-names-for-computing-utilization.md
markdroth: Good point.  Unfortunately, it looks like this is a pre-existing problem -- the same thing applies for the existing fallback between application_utilization and cpu_utilization.  So I don't think there's anything we can do about this without breaking existing users.

This isn't really a new issue with this specific design, but I'm fine with adding a note about it somewhere, just to call out the issue.

A114-wrr-metric-names-for-computing-utilization.md
markdroth: I'm not aware of any cases where systems are doing oversubscription on memory, because memory usage is not spiky the way that CPU utilization is.  So the fact that `mem_utilization` can't go above 1.0 isn't really a concern.

With regard to `application_utilization`, this makes me wonder why Envoy defined the behavior such that `application_utilization` takes precedence over this field.  It would probably make more sense to say that if this field is set and the fields it indicates exist, then we use those values first.

@efimki, was there a reason for this precedence order in Envoy?  How hard would it be for Envoy to change this?  If we can agree that this field should come first, then maybe we can change Envoy to match.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Noted. Resolving.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: I do recall that borg allows configuring jobs to go over limit using some flags  , although I am not sure if that bubbles up to display a number greater than 1.0 though, but if it did , we do have a very valid case for `mem_utilization` going above 1.0.

A114-wrr-metric-names-for-computing-utilization.md
dfawley: Envoy is already using this 12 syllable field name? 😞

sauravzg: Yep https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/client_side_weighted_round_robin/v3/client_side_weighted_round_robin.proto#client-side-weighted-round-robin-load-balancing-policy-proto

A114-wrr-metric-names-for-computing-utilization.md
markdroth: Let's not talk about specific internal systems in a public forum like this.

Speaking more generally, I think that CPU usage has two key properties that memory does not:
1. If the CPU you need is not available at the exact moment you need it, the work gets queued, so your application slows down a bit, but it continues to work.
2. It tends to be very spiky at very short timescales based on moment-to-moment load, which means that if the CPU you need is not available immediately, the odds are that it will be available very soon.

Memory usage is exactly the opposite of that: it does not tend to fluctuate as quickly as CPU usage, and if the memory you need is not available at the moment you need it, the application doesn't just slow down; it generally crashes with an OOM error.  As a result, oversubscription for memory really doesn't make any sense.

I am not aware of any compute systems that allow over-subscription for memory.

efimki: > I'm not aware of any cases where systems are doing oversubscription on memory, because memory usage is not spiky the way that CPU utilization is. So the fact that `mem_utilization` can't go above 1.0 isn't really a concern.
> 
> With regard to `application_utilization`, this makes me wonder why Envoy defined the behavior such that `application_utilization` takes precedence over this field. It would probably make more sense to say that if this field is set and the fields it indicates exist, then we use those values first.
> 
> @efimki, was there a reason for this precedence order in Envoy? How hard would it be for Envoy to change this? If we can agree that this field should come first, then maybe we can change Envoy to match.

@markdroth - Looking at Envoy implementation I see the following order of precedence:

1. `application_utilization`
2. Most constrained metric listed in the `metric_names_for_computing_utilization` WRR config field.
3. `cpu_utilization`.

This matches Envoy [doc](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/load_balancing_policies/client_side_weighted_round_robin/v3/client_side_weighted_round_robin.proto#envoy-v3-api-msg-extensions-load-balancing-policies-client-side-weighted-round-robin-v3-clientsideweightedroundrobin).

IIRC this order was chosen for compatibility with GRPC, is this not the case?

markdroth: Currently (prior to the design proposed in this PR), gRPC supports only `application_utilization` and `cpu_utilization`, in that order.  In this design, we are adding the new `metric_names_for_computing_utilization` field, so it's a new feature -- there is no existing behavior for where this new field fits into the precedence order in gRPC, so gRPC's behavior certainly wasn't a reason for Envoy to choose any particular behavior here.

I think it makes more sense to put `metric_names_for_computing_utilization` in prioritization order *before* `application_utilization` instead of afterward, for a couple of reasons:
- If the list of metrics is explicitly configured, it seems like that should take precedence over any default.
- If `application_utilization` comes first, then there's no way to set `metric_names_for_computing_utilization` to include both `application_utilization` and other metrics, because as soon as `application_utilization` is non-zero, we'll never honor `metric_names_for_computing_utilization`.

So my preferred order would be:
1. `metric_names_for_computing_utilization`
2. `application_utilization`
3. `cpu_utilization`

How hard would it be to change the precedence order in Envoy?

efimki: > Currently (prior to the design proposed in this PR), gRPC supports only `application_utilization` and `cpu_utilization`, in that order. In this design, we are adding the new `metric_names_for_computing_utilization` field, so it's a new feature -- there is no existing behavior for where this new field fits into the precedence order in gRPC, so gRPC's behavior certainly wasn't a reason for Envoy to choose any particular behavior here.
> 
> I think it makes more sense to put `metric_names_for_computing_utilization` in prioritization order _before_ `application_utilization` instead of afterward, for a couple of reasons:
> 
> * If the list of metrics is explicitly configured, it seems like that should take precedence over any default.
> * If `application_utilization` comes first, then there's no way to set `metric_names_for_computing_utilization` to include both `application_utilization` and other metrics, because as soon as `application_utilization` is non-zero, we'll never honor `metric_names_for_computing_utilization`.

I see, good point.

> So my preferred order would be:
> 
> 1. `metric_names_for_computing_utilization`
> 2. `application_utilization`
> 3. `cpu_utilization`

Sounds reasonable.

> How hard would it be to change the precedence order in Envoy?

The code change is trivial. This will be a user visible change in behavior, but I *think* that should be fine.

markdroth: Okay, then let's go ahead and make this change.  Thanks!

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Added details to the `Validity and Edge Cases` section. Proceeding to resolve. Please re-open if needed.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: @markdroth  Made the changes in the GRFC. PTAL at the latest commit and resolve this comment.

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,162 @@
...
+message WeightedRoundRobinLbConfig {
+  // ... existing fields ...
+
+  // A list of metric names to use for computing utilization.
+  //
+  // For map fields in the ORCA proto, the string will be of the form
+  // "<map_field_name>.<map_key>". For example, the string "named_metrics.foo"
+  // will mean to look for the key "foo" in the ORCA "named_metrics" field.
+  //
+  // By default, utilization is computed by taking the max of the values of the
```
ejona86: "By default". How are other computations possible?

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,162 @@
...
+  // metrics specified here. If none of the specified metrics are present in the
+  // load report (or if this field is empty, or all values are less than or
+  // equal to 0), the policy will fall back to using 'application_utilization'
```
ejona86: What do these two strings mean? It wasn't actually defined in this comment. If this comment describes the behavior to our users, then it needs to say something about it.

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,162 @@
...
+      - If the name is of the format `field.key` (e.g., `named_metrics.foo`), look up the map field `field` and retrieve the value for `key`.
+      - If the name is a simple field name (e.g., `cpu_utilization`, `mem_utilization`), look up the `field`.
+      - **Only the following fields are supported:**
+        - `application_utilization`
+        - `cpu_utilization`
+        - `mem_utilization`
+        - `named_metrics.*`
+    - **Compute Max**: Track the maximum value among all successfully resolved, positive ( > 0), finite metrics.
+    - If a max value is found, use it as the `utilization`.
+2.  **Fallback**: If no custom metric is found (or if `metric_names_for_computing_utilization` is not configured), fall back to the existing WRR utilization behavior defined in [gRFC A58][A58].
```
ejona86: This is saying "if no custom metric is found," but it could have been resolved and just discarded. Envoy uses the same language. Does that mean we just use the average weight in that case, or that we fallback? Should this be "If checking custom metrics did not determine a utilization" (to align with the bullet text)?

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,162 @@
...
+      map_name, key = split_on_first_dot(metric_name)
+      if report has map field map_name:
+         value = report[map_name][key]
+    else:
+      # Root field lookup (e.g. "mem_utilization") via Reflection
+      if report has field metric_name:
+         value = report[metric_name]
+
+    # Only consider valid, non-nan, positive values
+    if value is not null and !is_nan(value) and value > 0:
```
ejona86, line 94: Based on the "finite" mention above, I thought `Inf` was supposed to be ignored. Here it is being used.

ejona86 (commented):
Seems pretty close

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Rephrased to convey that this is an "overriding field" with fallback to default behavior. 
This was an artifact of the recent changes to the computation and fallback order. Keeping this open. Please resolve if this looks okay.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Rephrased to hide implementation details by referring to "existing default behavior". Resolving this comment in favor of the previous one.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Rephrase to mention "did not determine a valid utilization".
Depending on how we define "successfully resolved" ( I meant valid values), I don't think there's a case for discarding it at all. 
I don't think we use average weight in any case. Proceeding to resolve.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Removed the "finite" to "valid" above. This was an interesting case.

Envoy's implementation currently doesn't have any different handling for "nans" and "inf".  It works reasonably well with math for the weights formula to reasonable behaviors i.e. inf -> 0 weights, nan -> discarded since `nan <0`. 

I think `nan` creates some annoying issues , since the behavior of max(....) is language dependent and since grpc needs to be language consistent , it seems better to discard it.

Roughly,
c++ : max(1.0, nan) -> 1.0 , max(nan,1.0) -> nan 
java(and most languages): max(1.0, nan) and max(nan, 1.0)  are the same i.e. `nan`.

Envoy's current implementation (by choice or accident) is be closer to  `discard all nans`, since  it initializes `max` with 0 and iterates over entries , resulting in all nans being ignored. 

This is slightly  mentioned  at a very high level as undefined behavior(which should be renamed to inconsistent behavior IMO) in the section below about corner cases with a section about nan. 

I'll keep this open for now. Please resolve if no followups are needed.

sauravzg: LMK, if you'd like some part of this explanation to be added to the grfc somewhere, seemed pretty low level to be mentioned explicitly to me.

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -56,25 +53,25 @@ message WeightedRoundRobinLbConfig {
...
+1.  **Check Custom Metrics**: If `metric_names_for_computing_utilization` is configured:
     - Iterate through the specified metric names.
     - **Resolve Metric Value**:
       - If the name is of the format `field.key` (e.g., `named_metrics.foo`), look up the map field `field` and retrieve the value for `key`.
       - If the name is a simple field name (e.g., `cpu_utilization`, `mem_utilization`), look up the `field`.
+      - **Only the following fields are supported:**
+        - `application_utilization`
+        - `cpu_utilization`
+        - `mem_utilization`
+        - `named_metrics.*`
```
markdroth, line 67: Let's also support `utilization.*`, just so we don't have to add it later if someone needs it for some reason.

A114-wrr-metric-names-for-computing-utilization.md
markdroth: @efimki Please follow up on making the change on the Envoy side.

Resolving this comment.

markdroth (commented):
Just one minor nit remaining, otherwise looks good from my end!

dfawley (approved):

markdroth (approved):

A114-wrr-metric-names-for-computing-utilization.md
ejona86: I was actually arguing that we should define what the strings here mean. Anyone who sets it will need to know that. So changing "metric names" to "metrics" is a step in the wrong direction.

"Implementation details" is a weird way of putting it, as users are expected to know what that behavior is. I can understand not wanting to define it _here_ but it is not a goal to hide that information.

ejona86 (commented):
Sending what I have.

A114-wrr-metric-names-for-computing-utilization.md
efimki: I've opened https://github.com/envoyproxy/envoy/pull/44196 on the Envoy side.

ejona86:
@sauravzg, I don't know why you marked me to review, as you still haven't handled by last comment.

sauravzg:
@ejona86 I replied the the last remaining unresolved comment with a question: https://github.com/grpc/proposal/pull/536#discussion_r2880737283  , which is why I marked you for review.

Maybe the confusion stems from github not handling the comment chain well and showing them as individual comments?

ejona86:
@sauravzg and I responded to your comment saying the change was worse than before. I don't understand how you are linking to the correct comment thread but aren't seeing my comment dated March 17th is outstanding.

sauravzg:
> @sauravzg and I responded to your comment saying the change was worse than before. I don't understand how you are linking to the correct comment thread but aren't seeing my comment dated March 17th is outstanding.

NVM. I realized my comments were in "pending" . Sorry for the noise. I'll send the fix.

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Updated to be close to envoy according to our discussion offline. PTAL

A114-wrr-metric-names-for-computing-utilization.md
```diff
@@ -0,0 +1,157 @@
...
+### Service Config Update
+
+We will add a new field `metric_names_for_computing_utilization` to the [`WeightedRoundRobinLbConfig`][WeightedRoundRobinLbConfigProto] message in the Service Config.
+
+```protobuf
+message WeightedRoundRobinLbConfig {
+  // ... existing fields ...
+
+  // By default, Endpoint weight is computed by taking the max of the values of the metric names specified here from the `OrcaLoadReport` proto.
+  // For map fields in the ORCA proto, the string will be of the form ``<map_field_name>.<map_key>``. For example, the string ``named_metrics.foo`` will mean to look for the key ``foo`` in the ORCA `named_metrics` field.
```
ejona86: This displays pretty poorly. Can you word wrap it to some degree?

ejona86 (approved):

A114-wrr-metric-names-for-computing-utilization.md
sauravzg: Done. Wrapped to 80.

PR #541, opened 2026-03-17
A115: disable priority LB child policy retention cache

apolcyn:
Update on https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md#child-lifetime-management


A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,59 @@
...
+
+Additionally, priority LB child names are generated with strictly increasing numbers
+(once a priority LB child name is unconfigured, it will never be configured again). As such,
+the cache is not providing us value.
+
+## Proposal
+
+Priority LB should disable the child policy retention cache.
+
+### Temporary environment variable protection
```
easwars, line 53: I remember @ejona86 mentioning about the use of an env var to protect this change, so that it can be reverted by the user if required. Does it make sense to make that part of this update?

apolcyn: noted

A56-update-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,62 @@
+A56: disable Priority LB policy child policy retention cache
```
markdroth: This is a new gRFC and needs a new number.  This will be A115.

Please also add an "Updated by:" header to A56 that points to this gRFC.

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,62 @@
...
+* Author(s): @apolcyn
+* Approvers: @markdroth, @ejona86, @dfawley, @easwars
+* Status: {Draft}
+* Implemented in: C-core, Java, Go, Node
+* Last updated: 2026-03-17
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+[A56](https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md) describes
```
markdroth: This statement is misleading.  A56 specifies that the caching mechanism will be used in two cases:

1. When a higher priority child becomes reachable, we deactive the lower-priority children and eventually remove them.
2. When a child is removed from the LB policy config.

This gRFC is removing case (2) only.  We did not discuss removing case (1), and I don't think it makes sense to do so.

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,62 @@
...
+sequence of locality updates. On each child name update, previous policies are added
+to the retention cache.
+
+```
+[[AA,BB],[CC, DD]] => [priority-0-0 priority-0-1]
+[[AA, BB, CC], [DD]] => [priority-0-1 priority-0-2]
+[[AA, BB], [CC, DD]] => [priority-0-1 priority-0-2]
+[[AA, BB, CC], [DD]] => [priority-0-2 priority-0-3]
+[[AA, BB], [CC, DD]] => [priority-0-2 priority-0-3]
+[[AA, BB, CC], [DD]] => [priority-0-3 priority-0-4]
```
markdroth: I don't think that's the result we should see with this sequence of changes.  If CC is moving between priority 0 and priority 1, but DD is always in priority 1, we should continue to reuse the same two child names.

I think an example like this would better demonstrate the problem:

1. [[AA, BB], [CC, DD]]
2. [[CC], [DD, EE]]
3. [[AA, BB], [CC, DD]]

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,62 @@
...
+Priority LB should disable the child policy retention cache.
+
+### Temporary environment variable protection
+
+Implementations should provide an environment variable to revert
+to the previous behavior (child policy cache enabled with 15-minute timer).
+
+Env var name: `GRPC_EXPERIMENTAL_ENABLE_PRIORITY_LB_CHILD_POLICY_CACHE`.
+
+## Rationale
```
markdroth, line 62: I think this section needs to make a few points:

- Caching the child when it gets removed from the config does not actually accomplish anything useful in the case of choosing a priority within an xDS cluster (which is the primary case where this policy is used), because the hueristic in the cds policy that assigns the child names will never reuse a child name once it has been removed from the config.
- We have seen cases where retaining the children has used up a lot of memory and file descriptors, which has caused problems for users.
- In the long run, we want a better solution involving a separate layer for caching subchannels rather than LB policies, but that will be a separate project to be undertaken later.

markdroth (commented):
Thanks for writing this up!

A56-update-remove-priority-lb-child-policy-cache.md
apolcyn: done, updated other file

A115-remove-priority-lb-child-policy-cache.md
apolcyn: added these points

A115-remove-priority-lb-child-policy-cache.md
apolcyn: I made the example slightly more subtle, based on what I was reproducing with grpc-go

Tagged you internally for the real example. Maybe grpc-go is doing something unique here but it seemed to iterate localities in order and reuse the priority name of a locality if it already existed -- so if the failover locality showed up earlier in the list we'd get this result (my previous example didn't capture that).

A115-remove-priority-lb-child-policy-cache.md
apolcyn: added this text

A115-remove-priority-lb-child-policy-cache.md
easwars: Removing the caching only for case (2) is not going to as simple as setting an integer (at least in Go), as discussed today.

To give a little bit more context, we use a `balancergroup`, which is a helper that manages a bunch of LB policies and one can add and remove policies from the `balancergroup` and it allows the caller to control aggregating state updates from these policies. Our caching is implemented in the `balancergroup`. And this `balancergroup` is used by a bunch of LB policies like `priority`, `rls`, `weightedtarget` and `clustermanager`.

We *could* add a flavor of the `Remove` method that supports caching and one that doesn't. But I'm wondering how the other languages implement this caching and how easy/hard it is going to be to support this **only** for case (2).

apolcyn: cc @markdroth about this

markdroth: I don't think we want to disable caching for case (1).  If Go needs more work to do that, then we need to do that work.

FWIW, in C-core, we don't have any shared code for the policies you listed.  The priority policy deactivates the child [here for case 1](https://github.com/grpc/grpc/blob/c0728bdae796f214a9fc56a617b21075c2ed3b40/src/core/load_balancing/priority/priority.cc#L482) and [here for case 2](https://github.com/grpc/grpc/blob/c0728bdae796f214a9fc56a617b21075c2ed3b40/src/core/load_balancing/priority/priority.cc#L344).  So all we would need to do to remove this functionality for case 2 is to change the latter to immediately remove the child instead of deactivating it.

I can understand why Go might have used shared code for the priority, weighted_target, and xds_cluster_manager LB policies, since they all have this caching behavior.  But I will note that the RLS policy was never supposed to have this caching behavior, so it seems strange to me that you'd also use it there.

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -10,10 +10,15 @@ A56: disable Priority LB policy child policy retention cache
...
+[mechanisms](https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md#child-lifetime-management)
+whereby priority LB child policies are cached. There are two cases:
 
-This proposal removes that cache.
+1) When a higher priority child becomes reachable, we deactive
+the lower-priority children, and remove them only after an expiry.
+
+2) When a child is removed from the LB policy config.
+
+This proposal removes the usage of a cache for case 2).
```
markdroth: Unbalanced paren.

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -26,12 +31,9 @@ sequence of locality updates. On each child name update, previous policies are a
...
 ```
-[[AA,BB],[CC, DD]] => [priority-0-0 priority-0-1]
-[[AA, BB, CC], [DD]] => [priority-0-1 priority-0-2]
-[[AA, BB], [CC, DD]] => [priority-0-1 priority-0-2]
-[[AA, BB, CC], [DD]] => [priority-0-2 priority-0-3]
-[[AA, BB], [CC, DD]] => [priority-0-2 priority-0-3]
-[[AA, BB, CC], [DD]] => [priority-0-3 priority-0-4]
+[[BB, CC], [AA, DD]] => [priority-0-0 priority-0-1]
+[[AA, BB, CC], [DD, EE]] => [priority-0-1 priority-0-2]
+[[BB, CC], [AA, DD]] => [priority-0-2 priority-0-3]
```
markdroth: I think the algorithm should be sticking with child numbers 1 and 2 in this last step.

I've sent you https://github.com/grpc/grpc/pull/41896 to add tests demonstrating the algorithm's behavior in C-core.

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+Additionally, priority LB child names are generated with strictly increasing numbers
+(once a priority LB child name is unconfigured, it will never be configured again). As such,
+the cache is not providing us value.
+
+## Proposal
+
+Priority LB should disable the child policy retention cache.
```
markdroth: ...when a child is removed from its config (i.e., case 2 only).

markdroth: We also need to note here that we are proposing this change for Java and Go only.  For C-core, we are actually potentially getting benefit from this behavior due to subchannel sharing, so we're not planning to drop it there until we have the longer-term solution ready.

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,75 @@
...
+* Author(s): @apolcyn, @markdroth
+* Approvers: @markdroth, @ejona86, @dfawley, @easwars
+* Status: {Draft}
+* Implemented in: C-core, Java, Go, Node
+* Last updated: 2026-03-17
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+[A56](https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md) describes
```
ejona86: Use relative links: `[A56](A56-priority-lb-policy.md)`

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,75 @@
...
+the cache is not providing us value.
+
+## Proposal
+
+Priority LB should disable the child policy retention cache.
+
+### Temporary environment variable protection
+
+Implementations should provide an environment variable to revert
+to the previous behavior (child policy cache enabled with 15-minute timer).
```
ejona86, line 56: "for a period of time"? Or "for a few releases"? This makes it sound like we'll keep tihs around forever.

ejona86 (commented):
Do we (all of us) have a plan for cluster_manager? Are we going to make a separate gRFC for that?

A115-remove-priority-lb-child-policy-cache.md
danielduhh: Why not implement subchannel sharing in all languages?

markdroth: That's essentially what we'd wind up doing via the longer-term solution described in the "Rationale" section.

A115-remove-priority-lb-child-policy-cache.md
apolcyn: removed on line 21

A115-remove-priority-lb-child-policy-cache.md
apolcyn: hmm the original text is what Go was doing (i tagged you on an internal doc to illustrate)

but maybe that was specific to Go

Changed to your original suggestion

A115-remove-priority-lb-child-policy-cache.md
apolcyn: updated

A115-remove-priority-lb-child-policy-cache.md
apolcyn: noted for a few releases

A115-remove-priority-lb-child-policy-cache.md
apolcyn: done

A115-remove-priority-lb-child-policy-cache.md
markdroth: In the case you have now in the gRFC, in the third update, the algorithm will use child number 3 for [AA, BB] and reuse child number 2 for [CC, DD], because it was already using that number for [CC] in the second update.

If what you're seeing in Go matches what you currently have in the doc, then I suspect Go's implementation of this algorithm isn't quite right.  @easwars

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -42,13 +42,22 @@ the cache is not providing us value.
...
 ## Proposal
 
-Priority LB should disable the child policy retention cache.
+Priority LB should disable the child policy retention cache, when a
+child is removed from its config (i.e., case 2 only).
+
+Note this should be done for Java and Go only.
+
+For C-core, we are actually potentially getting benefit from this
+behavior due to subchannel sharing, so we're not planning to drop
```
markdroth: s/sharing/pooling/, just to be precise.

apolcyn: done

A115-remove-priority-lb-child-policy-cache.md
easwars: Will take a look and get back asap.

A115-remove-priority-lb-child-policy-cache.md
easwars: Yeah, we use the same `balancergroup` in `rls` as well, but we don't enable caching for it.

I ended up making a PR that adds a new flavor of `Remove` and use it from `priority`.

A115-remove-priority-lb-child-policy-cache.md
easwars: I don't see the actual algorithm described in the gRFC and I think this is where the algorithm was originally described: https://docs.google.com/document/d/13L2YVTGI2OrLpeCAPbgI4qEs0Oz02P7X0DKhk6Y9LMk/edit?tab=t.0#heading=h.r3twn1e6r8ct

Based on my reading of this section and my reading of C++, Java and Go, none of the languages actually implement the algorithm correctly. Specifically, none of the languages seem to implement this sentence:
- If there is at least one locality in the priority that was present in the same priority in the old config, copy the name from the same priority in the old config.

All languages seem to use a greedy approach where the moment they find a locality (from the new config) being present in the old config, they end up using that name. They don't actually check for localities in the same priority.

And I do see that in C++, when a map contains locality names as keys, they are ordered using the following: https://github.com/grpc/grpc/blob/2c62d7b7a927db2572b50277ba56eb4a754c43b7/src/core/xds/xds_client/xds_locality.h#L38. And I think this probably helps because in most of our examples, we seem to use very ordered locality names like "L0", "L1" ... or "aa", "bb". In Go, map key ordering is not deterministic.

I do have the changes ready in Go to actually implement the algorithm specified at the top of this comment. But, before sending them out, wanted to check here if there was a reason why none of the languages ended up implementing the logic for looking the locality in the same priority.

easwars: I ended up making the change in the name generation logic to give preference to localities in the same priority in the old config with the assumption that if we really don't want to be doing that, I can always revert the change back.

https://github.com/grpc/grpc-go/pull/8997

markdroth: Yeah, unfortunately, I didn't write a gRFC for that refactoring, just the internal doc.  I've regretted that several times since then, since we've had to patch holes in several subsequent gRFCs as a result (and this PR is yet another example of that).  And if I had done it as a gRFC, that would likely have prompted me to be more detailed about the intended algorithm here.

In any case, you're right, it looks like C-core isn't implementing that bullet correctly.  But I'm not sure it actually matters that much, because regardless of whether or not we do that, I think there will be cases where the algorithm will do the wrong thing, since it doesn't know anything about which localities are actually reachable or which ones have actually been created by the priority policy.  I'm not sure that the order of localities really matters at all -- although I would at least argue for deterministic behavior, because that will be much easier to debug.

(As an aside, it's worth noting that Java and Go also do not correctly implement the "all gRPC implementations must ensure that subchannels for locality 1 are retained through this transition instead of being destroyed and recreated" part, because that requires subchannel pooling.  Only C-core and Node do this correctly.)

BTW, why are Go's child names of the form "priority-{X}-{Y}"?  I see that Y is the child number here, but what is X?  If X is the priority, then that should not actually be part of the name at all, because we want a specific child to be able to move between priorities.  In C-core, the child name is just "child{N}", which is what the doc specifies.

easwars: > BTW, why are Go's child names of the form "priority-{X}-{Y}"? I see that Y is the child number here, but what is X?

Yes, X is the priority. That is also unfortunate how it ended up being implemented. I'll clean it up soonish.

> because that requires subchannel pooling. Only C-core and Node do this correctly

Yes, yes, yes :)

ejona86 (approved):

A115-remove-priority-lb-child-policy-cache.md
```diff
@@ -0,0 +1,84 @@
...
+This is especially the case when a locality is flapping between failover and primary
+priorities. For example, notice how priority LB child names increase in the following
+sequence of locality updates. On each child name update, previous policies are added
+to the retention cache.
+
+```
+[[AA, BB], [CC, DD]] => [priority-0-0 priority-0-1]
+[[CC], [DD, EE]] => [priority-0-1 priority-0-2]
+[[AA, BB], [CC, DD]] => [priority-0-3 priority-0-2]
+```
```
markdroth: Let's make this example reflect what should ideally be happening, rather than the weirdness that was Go's original implementation:

```suggestion
For example, consider the following sequence of updates:
1. P0=[AA, BB], P1=[CC, DD]: P0 will be assigned "child0", P1 will be assigned "child1".
2. P0=[CC], P1=[DD, EE]: P0 will be assigned "child1" (reusing the child name containing "CC" from the previous update), P1 will be assigned "child2" (new child number).
3. P0=[AA, BB], P1=[CC, DD]: P0 will be assigned "child3" (new child number), P1 will be assigned "child1" (reusing the child name containing "CC" from the previous update).
```

A115-remove-priority-lb-child-policy-cache.md
markdroth: Okay.  Note that having the priority number in the child name will cause more connection churn in some cases.  For example:

1. Update: P0=[AA,BB], P1=[CC,DD].  P0 gets `priority-0-0`, P1 gets `priority-1-1`.
2. Update: P0=[CC,DD], P1=[AA,BB].  P0 gets `priority-0-1`, P1 gets `priority-1-0`.

In this situation, Go will wind up creating two completely new children, thus briefly doubling its number of connections.  What it *should* do here is to simply swap the two:

1. Update: P0=[AA,BB], P1=[CC,DD].  P0 gets `child0`, P1 gets `child1`.
2. Update: P0=[CC,DD], P1=[AA,BB].  P0 gets `child1`, P1 gets `child0`.

easwars: Filed an issue in the grpc-go repo. Will have that fixed soon.

easwars (approved):

markdroth (approved):
Looks great, thanks!

PR #547, opened 2026-05-05
L73 update: Add a status code/description to the transport shutdown txn

jdcormie:


ejona86 (approved):
When do we want to merge this?

jdcormie:
> When do we want to merge this?

I was planning on merging this update to the wire format as soon as you approved it. Since both the existing shutdown flags field and the new status field are optional, the current grpc-java implementation is still compliant. LMK if you want to wait for some reason.

I don't yet have a PR against grpc-java to actually implement this but probably will soon.

PR #550, opened 2026-05-18
A118: TLS Telemetry

gtcooke94:
The gRFC to support authentication telemetry in the TLS stack for handshake and TLS offload operation monitoring.


A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
+A118: Authentication Telemetry
+----
+* Author(s): @gtcooke94
+* Approver: @dfawley, @easwars, @ejona86, @mattstev, @markdroth
```
matthewstevenson88: nit: My Github username is matthewstevenson88 :)

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+* Status: In Review
+* Implemented in:
+* Last updated: 2026-05-18
+* Discussion at: <google group thread> (filled after thread exists)
+
+## Abstract
+
+gRPC's authentication stack has no telemetry. This document details adding
+non-per-call metrics to the TLS authentication stack.
```
matthewstevenson88: micro-nit: s/the TLS/gRPC's SSL/TLS

Rationale: Make it unambiguous that we are adding this support for both SslCreds and TlsCreds.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+* [A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient](https://github.com/grpc/proposal/blob/master/A78-grpc-metrics-wrr-pf-xds.md)
+* [A79: OpenTelemetry Non-Per-Call Metrics Architecture](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)
+* [A89: Backend Service Metric Label](https://github.com/grpc/proposal/blob/master/A89-backend-service-metric-label.md)
+* [A107: TLS Private Key Offloading](https://github.com/grpc/proposal/blob/master/A107-tls-private-key-offloading.md)
+
+## Proposal
+
+All metrics will be scoped to TLS exclusively.
+
+### TLS Telemetry Result Enum
```
matthewstevenson88: I made a similar comment in https://github.com/grpc/grpc/pull/42446, but should we call this "TLS Handshake Result Enum" instead?

Rationale: I don't think we should include "Telemetry" in the enum name, since that is referencing what the enum is used for rather than what it represents. For example, below we use the terminology "TLS Resumption Type Enum", so using "TLS Handshake Type Enum" would be consistent.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+The handshaker status will be represented by an enum that indicates success or
```
matthewstevenson88: nit: s/handshaker/TLS handshake

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+### Metrics Definitions
+
+The following metrics are count metrics, with the primary information coming
+from the labels.
+
+* `grpc.client.tls.handshakes`
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryResult` enum indicating success or the reason for handshake failure |
```
matthewstevenson88: micro-nit: Missing period at the end of the sentence.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+| `grpc.target` | Required | The target string (as defined in A66) passed to the channel. |
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum |
```
matthewstevenson88: micro-nit: Missing period at the end of the sentence. Also consider matching the language from the handshake.result field, e.g. "The `TlsResumptionType` enum indicating if and how the handshake was resumed."

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in A78). TODO - is this actually possible to get in every language |
```
matthewstevenson88: Please remove the TODO.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in A89). TODO - is this actually possible to get in every language |
```
matthewstevenson88: Please remove the TODO.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+* `grpc.server.tls.handshakes`
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryResult` enum indicating success or the reason for handshake failure |
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum |
```
matthewstevenson88: micro-nit: Missing period at the end of the sentence. Also consider matching the language from the handshake.result field, e.g. "The `TlsResumptionType` enum indicating if and how the handshake was resumed."

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+| `grpc.target` | Required | The target string (as defined in A66) passed to the channel. |
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum |
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in A78). TODO - is this actually possible to get in every language |
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in A89). TODO - is this actually possible to get in every language |
+
+* `grpc.server.tls.handshakes`
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryResult` enum indicating success or the reason for handshake failure |
```
matthewstevenson88: micro-nit: Missing period at the end of the sentence.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.status` | Required | Result of the certificate selection offloading, in the format of a gRPC status code (as defined in A66). |
+
+Note - there is no associated client certificate selection metric. This is a
+server specific feature.
+
+For the offloaded private key metrics, we specify signing because that is the
+only offloaded private key operation supported by gRPC. Older TLS versions have
```
matthewstevenson88: Can we add an example in a parenthetical? For example:

(for example, signature offload using an EC or RSA key)

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+the concept of other private key operations that gRPC does not support. See
```
matthewstevenson88: Can we add an example in a parenthetical? For example:

(for example, decryption offload using an RSA key)

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+* `grpc.client.tls.offload_private_key_signing_duration` (unit: float64, type: histogram - latency buckets defined in A66)
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.status` | Required | Result of the certificate selection offloading, in the format of a gRPC status code (as defined in A66). |
+| `grpc.target` | Required | The target string (as defined in A66) passed to the channel. |
+| `grpc.tls.private_key_algorithm` | Optional | An algorithm enum indicating how the offloaded private key signing was done, e.g. “RsaPkcs1Sha256”. |
+| `grpc.tls.private_key.implementation` | Required | A string identifying the private key signer implementation. |
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in A78). TODO - is this actually possible to get in every language |
```
matthewstevenson88: Please remove the TODO.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in A89). TODO - is this actually possible to get in every language |
```
matthewstevenson88: Please remove the TODO.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.status` | Required | Result of the certificate selection offloading, in the format of a gRPC status code (as defined in A66). |
+| `grpc.tls.private_key_algorithm` | Optional | An algorithm enum indicating how the offloaded private key signing was done, e.g. “RsaPkcs1Sha256”. |
+| `grpc.tls.private_key.implementation` | Required | A string identifying the private key signer implementation. |
+
+### Temporary environment variable protection
+
+This feature will be explicitly configured by users, thus no environment
```
matthewstevenson88, line 161: How will a user configure this?

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+variable protection is needed. If a user does not configure TLS metrics or
+offloading, telemetry won't be collected under this mechanism unless general
+telemetry/stats plugins are active.
+
+## Rationale
+
+The alternative of splitting generic handshaker metrics and TLS-specific metrics
+was considered, but opened many rabbit holes as to what would qualify as a
+handshaker (e.g. are HTTP Connect, TCP, RPC Switch, etc. all in scope, or just
```
matthewstevenson88: Please remove "RPC Switch" from this list - it is a concept that is specific to certain ALTS deployments, and does not apply to TLS.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+```diff
+ tsi_result tsi_ssl_client_handshaker_factory_create_handshaker(
+     tsi_ssl_client_handshaker_factory* factory,
+     const char* server_name_indication, size_t network_bio_buf_size,
+     size_t ssl_bio_buf_size,
+     std::optional<std::string> alpn_preferred_protocol_list,
+-    tsi_handshaker** handshaker);
++    tsi_handshaker** handshaker,
++    std::shared_ptr<grpc_core::GlobalStatsPluginRegistry::StatsPluginGroup> stats_plugin_group = nullptr);
```
matthewstevenson88: I made a similar comment in https://github.com/grpc/grpc/pull/42430, but we should consider leaving `handshaker` as the last argument in the list as it is an output variable. Similarly below.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+In the C++ implementation, the transport security interface (TSI) has
+historically been decoupled from gRPC. However, this design choice has been
+broken over time, and the two have been coupled for years now. The reasons
+behind decoupling TSI and gRPC are no longer relevant, therefore we will fully
+accept this coupling. Thus, the TSI code can contain gRPC monitoring specifics.
+In the few use-cases where TSI is not called via gRPC, we will ensure that
+metric incrementation is not performed.
+
+We will add the Channel's `StatsPluginGroup` as an optional argument to TSI
```
matthewstevenson88: nit: s/TSI handshaker/SSL TSI handshaker

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+     tsi_ssl_server_handshaker_factory* factory, size_t network_bio_buf_size,
+-    size_t ssl_bio_buf_size, tsi_handshaker** handshaker);
++    size_t ssl_bio_buf_size, tsi_handshaker** handshaker,
++ std::shared_ptr<grpc_core::GlobalStatsPluginRegistry::StatsPluginGroup> stats_plugin_group = nullptr);
+```
+### Golang
+
+For the implementation of the general handshake metric, we leverage gRPC-Go's existing transport-level architecture.
+
+The abstraction layer for general handshake information is the Transport
```
matthewstevenson88: nit: Several words in this paragraph are capitalized but probably should not be, e.g. "Transport Connection Layer" and "Handshakes".

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
++     <increment metric>
++		}
++
+ 		if err != nil {
+ 			return nil, connectionErrorf(isTemporary(err), err, "transport: authentication handshake failed: %v", err)
+ 		}
+
+```
+
+To pass out resumption information, we will need to augment the [`TLSInfo`](https://github.com/grpc/grpc-go/blob/660208049b96ff6232e8c7212905b3e357b5bf42/credentials/tls.go#L41) to include [`ConnectionState.DidResume`](https://github.com/golang/go/blob/e62d3e6e897175a07aa44a7b2c7f99700072f22f/src/crypto/tls/common.go#L257).
```
matthewstevenson88: micro-nit: s/pass out/extract

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+1. Define a new `InternalSecurityMetrics.java` class in the api/ module under the `io.grpc` package.  
+2. Expose `MetricRecorder` from `GrpcHttp2ConnectionHandler` in the netty/ module.  
+3. Instrument the Netty `ClientTlsHandler` and `ServerTlsHandler` in `ProtocolNegotiators.java` to track handshake duration and record it.  
+
+The TLS specific offload metrics will go with their implementation. This feature is not yet written in Java, so we cannot discuss specific metric implementation details.
+
+### Wrapped Languages
+
+Wrapped languages that support non-per-call metrics should be able to get this "for free" from the Core implementation. However, Python for example, does not currently support non-per-call metrics.
```
matthewstevenson88: micro-nit for clarity: s/should be able to get this/will get the authentication telemetry feature

A118-tls-telemetry.md
```diff
@@ -0,0 +1,258 @@
...
+
+The alternative of splitting generic handshaker metrics and TLS-specific metrics
+was considered, but opened many rabbit holes as to what would qualify as a
+handshaker (e.g. are HTTP Connect, TCP, RPC Switch, etc. all in scope, or just
+alternative protocols to TLS such as ALTS). Thus, the decision was made to scope
+the metrics to TLS.
+
+## Implementation
+
+### C++
```
matthewstevenson88: Since we're referencing the wrapped languages later, should this section be retitled to "Core"?

If yes, please also s/C++/Core in the first sentence of the paragraph below.

A118-tls-telemetry.md
gtcooke94: Fixed

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Disccused in the other, going with `TlsTelemetryHandshakeResult`

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: This is from discussion with Eric and Mark - we want this if possible, but if it's not readily available in any language we will remove it, so I need to see if it's readily available

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done, thanks!

A118-tls-telemetry.md
gtcooke94: Done, thanks!

A118-tls-telemetry.md
gtcooke94: See other thread

A118-tls-telemetry.md
gtcooke94: See other thread

A118-tls-telemetry.md
gtcooke94: Users configure the telemetry they want to consume - I don't think we need to spell that out in this gRFC as it's the content for other gRFCs

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Renamed to C/C++ (I think core is somewhat an internal code structuring thing)

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: I believe I've fixed this

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
matthewstevenson88: Please also update the section heading.

A118-tls-telemetry.md
matthewstevenson88: Does this block us proceeding with the gRFC?

A118-tls-telemetry.md
matthewstevenson88: Ah it's following the same model as in the other gRFCs. Thanks!

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+
+The following metrics are count metrics, with the primary information coming
+from the labels.
+
+* `grpc.client.tls.handshakes`
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryHandshakeResult` enum indicating success or the reason for handshake failure. |
+| `grpc.target` | Required | The target string (as defined in [A66]) passed to the channel. |
```
matthewstevenson88, line 109: nit: Missing link on [A66]

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum indicating if and how the handshake was resumed. |
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in [A78]). |
```
matthewstevenson88: nit: Missing link on [A78]

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in [A89]). |
```
matthewstevenson88: nit: Missing link on [A89]

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryHandshakeResult` enum indicating success or the reason for handshake failure. |
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum  indicating if and how the handshake was resumed. |
+
+### TLS Offload Specific Metrics
+
+The following metrics are non-per-call bucketed latency metrics that report the duration of offloaded cryptographic operations.
+
+* `grpc.server.tls.offload_certificate_selection_duration` (unit: float64, type: histogram - latency buckets defined in [A66])
```
matthewstevenson88: nit: Missing link on [A66]

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.status` | Required | Result of the certificate selection offloading, in the format of a gRPC status code (as defined in [A66]). |
```
matthewstevenson88, line 129: nit: Missing link on [A66]

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+
+Note - there is no associated client certificate selection metric. This is a
+server specific feature.
+
+For the offloaded private key metrics, we specify signing because that is the
+only offloaded private key operation supported by gRPC (for example, signature
+offload using an EC or RSA key). Older TLS versions have the concept of other
+private key operations that gRPC does not support (for example, decryption
+offload using an RSA key). See [A107] for more detail on private key signers.
```
matthewstevenson88, line 138: nit: Missing link on [A107]

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+
+* `grpc.client.tls.offload_private_key_signing_duration` (unit: float64, type: histogram - latency buckets defined in A66)
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.status` | Required | Result of the certificate selection offloading, in the format of a gRPC status code (as defined in A66). |
+| `grpc.target` | Required | The target string (as defined in A66) passed to the channel. |
+| `grpc.tls.private_key_algorithm` | Optional | An algorithm enum indicating how the offloaded private key signing was done, e.g. “RsaPkcs1Sha256”. |
+| `grpc.tls.private_key.offloader_name` | Required | A string identifying the private key signer implementation, e.g. "HSM" or "private_key_signer_service". This must be low-cardinality |
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in [A78]). |
```
matthewstevenson88: nit: Missing link on [A78]

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in [A89]). |
```
matthewstevenson88: nit: Missing link on [A89]

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+
+* `grpc.server.tls.offload_private_key_signing_duration` (unit: float64, type: histogram - latency buckets defined in A66)
```
matthewstevenson88: Should we add a link on A66, as we do in the other references?

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.status` | Required | Result of the certificate selection offloading, in the format of a gRPC status code (as defined in A66). |
```
matthewstevenson88: Should we add a link on A66, as we do in the other references?

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+without a structured mechanism for aggregation and analysis.
+
+### Related Proposals:
+* [A66: OpenTelemetry Metrics/Stats](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md)
+* [A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient](https://github.com/grpc/proposal/blob/master/A78-grpc-metrics-wrr-pf-xds.md)
+* [A79: OpenTelemetry Non-Per-Call Metrics Architecture](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)
+* [A89: Backend Service Metric Label](https://github.com/grpc/proposal/blob/master/A89-backend-service-metric-label.md)
+* [A107: TLS Private Key Offloading](https://github.com/grpc/proposal/blob/master/A107-tls-private-key-offloading.md)
+
+[A66]: A66-otel-stats.md
```
matthewstevenson88, line 36: Is duplicating the list above intentional?

gtcooke94: Yes, IIUC correctly these don't show up in the final markdown, but they allow for links like `[A66]` in the text

matthewstevenson88 (approved):

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+### Metrics Definitions
+
+The following metrics are count metrics, with the primary information coming
+from the labels.
+
+* `grpc.client.tls.handshakes`
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryHandshakeResult` enum indicating success or the reason for handshake failure. |
```
ejona86: Let's make it clear this is the name as a string ("SUCCESS", not "1"). Ditto elsewhere.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+};
+```
+
+### Metrics Definitions
+
+The following metrics are count metrics, with the primary information coming
+from the labels.
+
+* `grpc.client.tls.handshakes`
+
```
ejona86: Missing the instrument type (counter), the unit ({handshake}), and a description. The description should say it is experimental.

The histograms have the instrument type, but are still missing the unit (`s`) and description.

ejona86 (commented):
This looks good. I'll delay giving my Approval until others have, and then I'll Approve+merge when it seems there's concensus.

A118-tls-telemetry.md
gtcooke94: Swapped to `enum string`

A118-tls-telemetry.md
gtcooke94: Added

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
+A118: Authentication Telemetry
+----
+* Author(s): @gtcooke94
+* Approver: @dfawley, @easwars, @ejona86, @matthewstevenson88, @markdroth
+* Status: In Review
+* Implemented in:
+* Last updated: 2026-05-22
+* Discussion at: <google group thread> (filled after thread exists)
```
markdroth: Please create the mailing list thread and add a link to it here.  Note that this starts the 10-day comment period before this can be merged.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+balance of low-cardinality while being fine-grained enough to be useful;
+therefore, an enum containing subdomains of authentication errors will be
+created. This is presented as a C++ enum below, but will be identical in all
+languages. In cases where we cannot categorize an error or cannot get enough
+granularity in a given implementation and/or language, `UNKNOWN_FAILURE` will be
+the catch-all error code.
+
+```c++
+enum class TlsTelemetryHandshakeResult {
+  UNKNOWN_FAILURE,
```
markdroth: As I mentioned in the implementation PR, this isn't actually the right naming style for C++ enum names: https://google.github.io/styleguide/cppguide.html#Enumerator_Names

Given that the names we're going to use in C++ are not the same as the actual strings that we want in the label values, it might be a good idea for this gRFC to just list the strings we're going to use, not the C++ enum definition.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+existing telemetry infrastructure defined in
+[A66](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md) and the
+specific architecture for non-per-call metrics established in
+[A79](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md).
+Because authentication and handshakers are connection-level abstractions, they
+inherently require non-per-call instrumentation. Currently, application owners
+face challenges in diagnosing handshake failures, relying on verbose logging
+without a structured mechanism for aggregation and analysis.
+
+### Related Proposals:
```
markdroth, line 28: I suggest also including a reference to A94, since it includes details about how we plumb the locality and backend_service labels into the subchannel.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+
+* `grpc.client.tls.handshakes` (unit: handshakes type: counter. This is experimental.)
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryHandshakeResult` enum string indicating success or the reason for handshake failure. |
+| `grpc.target` | Required | The target string (as defined in [A66]) passed to the channel. |
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum string indicating if and how the handshake was resumed. |
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in [A78]). |
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in [A89]). |
```
markdroth: Both of these labels should also reference A94.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+* `grpc.client.tls.offload_private_key_signing_duration` (unit: float64 seconds, type: histogram - latency buckets defined in [A66]. This is experimental.)
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.status` | Required | Result of the certificate selection offloading, in the format of a gRPC status code (as defined in [A66]). |
+| `grpc.target` | Required | The target string (as defined in [A66]) passed to the channel. |
+| `grpc.tls.private_key.offloader_name` | Required | A string identifying the private key signer implementation, e.g. "HSM" or "private_key_signer_service". This must be low-cardinality |
+| `grpc.tls.private_key_algorithm` | Optional | An algorithm enum indicating how the offloaded private key signing was done, e.g. “RsaPkcs1Sha256”. |
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in [A78]). |
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in [A89]). |
```
markdroth: Both of these labels should also reference A94.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,269 @@
...
+
+In the C/C++ implementation, the transport security interface (TSI) has
+historically been decoupled from gRPC. However, this design choice has been
+broken over time, and the two have been coupled for years now. The reasons
+behind decoupling TSI and gRPC are no longer relevant, therefore we will fully
+accept this coupling. Thus, the TSI code can contain gRPC monitoring specifics.
+In the few use-cases where TSI is not called via gRPC, we will ensure that
+metric incrementation is not performed.
+
+We will add the Channel's `StatsPluginGroup` as an optional argument to the SSL
```
markdroth: See my comment in the implementation PR.  I think we actually want to pass in the collection scope here, not the entire stats plugin group.

markdroth (commented):
This looks good overall!  My comments are all relatively minor.

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Changed the language to just be a set of strings.

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: I figured we'd pass the whole plugin group for future proofing in case we need to do something more complicated in the future, but I'm happy either way.

Added a TODO here for me to update to whatever we decide in the impl PR, updating the text and code diffs respectively.

A118-tls-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+ 	if transportCreds != nil {
++		isTLS := transportCreds.Info().SecurityProtocol == "tls"
++		var startTime time.Time
++		if isTLS {
++			startTime = time.Now()
++		}
+ 		conn, authInfo, err = transportCreds.ClientHandshake(connectCtx, addr.ServerName, conn)
++		if isTLS {
++			duration := time.Since(startTime).Seconds()
++     <increment metric>
```
dfawley: Which metric is this, specifically?  The ones defined above all seem to be finer-grained than this.  And `grpc.client.tls.handshakes` is just a counter, so doesn't need a duration?

A118-tls-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+only offloaded private key operation supported by gRPC (for example, signature
+offload using an EC or RSA key). Older TLS versions have the concept of other
+private key operations that gRPC does not support (for example, decryption
+offload using an RSA key). See [A107] for more detail on private key signers.
+
+* `grpc.client.tls.offload_private_key_signing_duration` (unit: float64 seconds, type: histogram - latency buckets defined in [A66]. This is experimental.)
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.status` | Required | Result of the certificate selection offloading, in the format of a gRPC status code (as defined in [A66]). |
```
dfawley: Copy/paste-o from above.  "Result of the private key signing...."

A118-tls-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+"UNEXPECTED_MESSAGE",
+"HANDSHAKE_TIMEOUT",
+"PEER_CONNECTION_CLOSED"
+```
+
+### TLS Resumption Type Enum
+
+Further, whether the handshake is resumed or not is also critical for
+understanding authentication behavior. An enum describing the type of resumption
+used (or none) will be created - it can be extended in the future to make
```
dfawley, line 88: This says "or none" but there is no corresponding NONE enum value below?

A118-tls-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+### Golang
+
+For the implementation of the general handshake metric, we leverage gRPC-Go's existing transport-level architecture.
+
+The abstraction layer for general handshake information is the transport
+connection layer (`internal/transport`). Client handshakes are performed in a
+single place: `internal/transport/http2_client.go` inside `NewHTTP2Client` via
+`transportCreds.ClientHandshake`. Server handshakes are also performed in a
+single place: `internal/transport/http2_server.go` inside `NewServerTransport`
+via `config.Credentials.ServerHandshake`. We can incremement the metrics here
```
dfawley: *increment

A118-tls-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum string indicating if and how the handshake was resumed. |
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in [A78], [A94]). |
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in [A89], [A94]). |
+
+* `grpc.server.tls.handshakes` (unit: handshakes type: counter. This is experimental.)
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryHandshakeResult` enum string indicating success or the reason for handshake failure. |
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType`  enum string  indicating if and how the handshake was resumed. |
```
dfawley: Nit:
```suggestion
| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum string  indicating if and how the handshake was resumed. |
```

A118-tls-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+  RESUMED_HANDSHAKE,
+};
+```
+
+### Metrics Definitions
+
+The following metrics are count metrics, with the primary information coming
+from the labels.
+
+* `grpc.client.tls.handshakes` (unit: handshakes type: counter. This is experimental.)
```
dfawley: The unit here should be "`{handshake}`" (the braces and everything are important for otel).  And "`int64 counter`".

A118-tls-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryHandshakeResult` enum string indicating success or the reason for handshake failure. |
+| `grpc.target` | Required | The target string (as defined in [A66]) passed to the channel. |
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType` enum string indicating if and how the handshake was resumed. |
+| `grpc.lb.locality` | Optional | The locality to which the traffic is being sent (as defined in [A78], [A94]). |
+| `grpc.lb.backend_service` | Optional | The backend service to which the traffic is being sent (as defined in [A89], [A94]). |
+
+* `grpc.server.tls.handshakes` (unit: handshakes type: counter. This is experimental.)
```
dfawley: (As above)

A118-tls-telemetry.md
```diff
@@ -0,0 +1,270 @@
...
+| Label Name | Required/Optional | Description |
+| :--- | :--- | :--- |
+| `grpc.tls.handshake.result` | Required | The `TlsTelemetryHandshakeResult` enum string indicating success or the reason for handshake failure. |
+| `grpc.tls.handshake.resumed` | Optional | The `TlsResumptionType`  enum string  indicating if and how the handshake was resumed. |
+
+### TLS Offload Specific Metrics
+
+The following metrics are non-per-call bucketed latency metrics that report the duration of offloaded cryptographic operations.
+
+* `grpc.server.tls.offload_certificate_selection_duration` (unit: float64 seconds, type: histogram - latency buckets defined in [A66]. This is experimental.)
```
dfawley: "float64" isn't a unit.  I think you want "`unit: s`" and "`type: float64 histogram`".

dfawley (commented):
Did we want to find someone from otel to review the propsed metrics before we implement this?  I don't think any of us are experts in this area.

A118-tls-telemetry.md
gtcooke94: `FULL_HANDSHAKE` means no resumption was done - should we rename to `NO_RESUMPTION`?

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
gtcooke94: My bad, this was from when we still had the handshake at a latency metric. Updated the code snippet.

A118-tls-telemetry.md
gtcooke94: Whoops, missed this in the review, fixed

A118-tls-telemetry.md
dfawley: I don't feel too strongly about the names, but it might want a comment.  But what exactly does "RESUMED_HANDSHAKE" actually mean, relative to the other future things you suggested could be added in the future like "ticket-based resumption vs. session-based resumption"?  Is "RESUMED_HANDSHAKE" actually "SESSION_RESUMPTION"?

A118-tls-telemetry.md
```diff
@@ -0,0 +1,265 @@
...
+only in the case where TLS is the protocol being used.
+
+```diff
+--- a/internal/transport/http2_client.go
++++ b/internal/transport/http2_client.go
+@@ -294,7 +294,23 @@
+ 	if transportCreds != nil {
++		isTLS := transportCreds.Info().SecurityProtocol == "tls"
+ 		conn, authInfo, err = transportCreds.ClientHandshake(connectCtx, addr.ServerName, conn)
++		if isTLS {
```
dfawley, line 231: Other thought: do we want to read the `authInfo` returned by the handshake to determine whether TLS was involved (I guess cast it to `TLSInfo`?) or trust the creds string directly?  I don't know that it matters at all, but it's something to ponder, perhaps.

A118-tls-telemetry.md
dfawley: OK thanks. One other nit here is that you're mixing tabs and spaces on the different lines, so the alignment is off for me.

dfawley (approved):

ejona86:
@mbissa, can you look at this from the metrics definition perspective (did we define everything we needed to define)?

A118-tls-telemetry.md
```diff
@@ -0,0 +1,265 @@
...
+[A94]: A94-subchannel-otel-metrics.md
+[A107]: A107-tls-private-key-offloading.md
+
+## Proposal
+
+All metrics will be scoped to TLS exclusively.
+
+### TLS Telemetry Handshake Result Enum
+
+The TLS handshake result will be represented by an enum that indicates success
```
easwars, line 49: The TLS creds implementation in Go ends up calling the `Handshake` method on the wrapped `tls.Conn`. From what I see, the `tls` package provides this generic error type: https://pkg.go.dev/crypto/tls#CertificateVerificationError that doesn't seem to contain any sort of granular information about the error. How do we plan to get any granular information in Go?

A118-tls-telemetry.md
```diff
@@ -0,0 +1,265 @@
...
+or provides information on why the handshake failed. This value must manage a
+balance of low-cardinality while being fine-grained enough to be useful;
+therefore, a set consisting of subdomains of authentication errors will be
+created. These are strings that will be a label value in the handshake metrics.
+These strings will be identical in all languages. In cases where we cannot
+categorize an error or cannot get enough granularity in a given implementation
+and/or language, `UNKNOWN_FAILURE` will be the catch-all error code.
+
+```
```
easwars, line 58: This enum contains sub-groups. Would it make sense to have a catch-all per sub-group apart from the global catch-all `UNKNOWN_FAILURE`?

A118-tls-telemetry.md
easwars: FYI: Both the tls creds and xds creds today set this field to "tls". Similarly, both of them set `AuthInfo` to `TLSInfo`. Maybe the idea was always to include `xds` creds as well, in which case we are good.

A118-tls-telemetry.md
gtcooke94: I think we sort of do
If something in certificate verification fails, the catch all is `CERTIFICATE_VERIFICATION_FAILED`.

If TLS negotiation fails, we should get one of those, otherwise it's unknown. I think if we _don't_ get one of those, we wouldn't be confident in some generic `NegotiationMismatch` failure.

`SIGNATURE_VERIFICATION_FAILED` is a catch-all for the cryptographic failures.

A118-tls-telemetry.md
gtcooke94: The CertificateVerificationError wraps underlying X509 errors, I think we will reach into that.

Code snippets seem to use matching via error.As() to get the wrapped underlying error.

A118-tls-telemetry.md
gtcooke94: RESUMED_HANDSHAKE means that it was resumed, with no specificity on session or ticket

A118-tls-telemetry.md
gtcooke94: Either way - I don't see a reason using XDS creds should preclude folks from monitoring - if it's doing TLS and we have TLS monitoring, the user can get monitoring?

As for how we want to see if it's TLS - can we discuss on the implementation PR since it's more of an impl detail? This is meant more as a demonstrative snippet, not a code source of truth.

A118-tls-telemetry.md
markdroth: The implementation PR was changed to use `CollectionScope`.

markdroth (approved):
This looks good, modulo updating the C++ implementation details.

A118-tls-telemetry.md
gtcooke94: Done

A118-tls-telemetry.md
ejona86: It is still missing a description.

A118-tls-telemetry.md
```diff
@@ -185,7 +185,7 @@ In the few use-cases where TSI is not called via gRPC, we will ensure that
 metric incrementation is not performed.
 
 // TODO: Update this comment to be more accurate based on the impl PR
```
markdroth: Please remove this line.

easwars (approved):

A118-tls-telemetry.md
gtcooke94: I think you might be looking at an old diff - at least I have it in my view

ejona86: This is what I see:

> `grpc.client.tls.handshakes` (unit: {handshake} type: int64 counter. This is experimental.)

A118-tls-telemetry.md
ejona86: AHH! My eyes! I was completely fine, but then I saw the second double-space in `string  indicating` in Doug's suggestion to fix a double-space :smile:.

A118-tls-telemetry.md
gtcooke94: Ah, do you want a larger description than "This is experimental."

I wasn't sure if something like "The count of the TLS handshakes done" was actually useful, or if the name is descriptive enough

A118-tls-telemetry.md
gtcooke94: I wish your eyes the quickest recovery possible 😆

A118-tls-telemetry.md
ejona86: It's more that every language will [need to define a description string](https://github.com/grpc/grpc-java/blob/66c6ab1da6def61556850cc408b726a559f0a118/xds/src/main/java/io/grpc/xds/WeightedRoundRobinLoadBalancer.java#L116) when they go to implement it, and that would not just be "experimental." And once this is implemented we'll be updating the guide. See these for some examples: https://grpc.io/docs/guides/opentelemetry-metrics/#lb-policy-instruments  (These specific metrics I linked to aren't more interesting than the others, other than you can see how experimental mixes in with the description.) See how those docs are 1:1 with the API we use to register metrics? (I'm not concerned about the different formatting; having the labels listed like this works fine. We're only missing the description.)

gtcooke94: Ah thank you for the reference - I descriptions for all of the metrics

ejona86 (approved):

PR #556, opened 2026-06-23
A121: RPC Delay Observability

mbissa:


A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,433 @@
+# A121: RPC Delay Observability
+----
+* Author(s): Madhav Bissa (@mbissa)
+* Approver: @markdroth, @ejona86, @dfawley, @easwars
+* Implemented in: Go, Java, C++
+* Last updated: 2026-06-19
```
AgraVator: Shouldn't this be today ??

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,433 @@
...
+    *   **Metric**: `grpc.client.call.delay.duration` (Float64 Histogram).
+    *   **Tracing Span**: A child span named **`"Call Delay"`**, created as a child of the main RPC Call span.
+2.  **Attempt Level Observability (Attempt-level Delays)**:
+    *   **Scope**: Measures delays that occur within the lifecycle of a specific RPC attempt, primarily during the load balancing pick loop and connection establishment.
+    *   **Metric**: `grpc.client.attempt.delay.duration` (Float64 Histogram).
+    *   **Tracing Span**: A child span named **`"Attempt Delay"`**, created as a child of the individual Attempt span.
+
+#### Span Attributes and Events
+
+For both layers, a child span is created for each distinct `grpc.delay_type`, and transitions within that type are recorded as span events:
```
AgraVator, line 49: For each consecutively distinct right (i.e. whenever the delay_type value changes) ?

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,433 @@
...
+| **Default Enabled** | `false` (experimental, opt-in) |
+
+### 2. Delay Types & Reasons
+
+To ensure consistency across implementations, we define the taxonomy of metric and tracing labels. 
+
+All connection-related delays are consolidated into a single low-cardinality metric label value (`"connecting"`), while their detailed reasons are recorded inside the `"Delay state transition"` span event (as the `grpc.delay_reason` event attribute).
+
+#### 1. Metric Delay Types (`grpc.delay_type`)
+The `grpc.delay_type` label is a low-cardinality, restricted set of values. To ensure maximum clarity, these types are explicitly partitioned into Call-Level and Attempt-Level scopes, corresponding to their respective duration histograms. Structural container policies may compose the attempt-level values by prepending logical prefixes.
```
AgraVator: "Structural container policies may compose the attempt-level values by prepending logical prefixes."
Does this depend upon the implementor ?

AgraVator: Nevermind, Saw the relevant section below

A121-rpc-delay-observability.md
AgraVator: Nevermind, saw the relevant section below

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,433 @@
...
+Structural container policies prepend their logical prefixes to the base attempt-level types using a colon separator:
+*   **Priority Policy**: Prepends the active priority tier index, resulting in composed types like:
+    *   `p0:connecting`
+    *   `p1:subchannel_state_mismatch`
+    *   `p0:picker_failing_with_wait_for_ready`
+*   **Pass-Through Container Policies**: Policies like `xds_cluster_manager`, `weighted_target`, and `rls` **do not prepend any prefix or wrap the delay type**. They simply bubble up the child's `grpc.delay_type` (e.g., `"connecting"`) directly as-is.
+
+##### Metadata Propagation & Composition
+The prepending and wrapping logic is handled entirely inside the Balancer Picker tree hierarchy, keeping the channel's attempt-routing wrapper and tracer plugins completely decoupled and simple:
+1.  **Leaf Pickers** (such as `pick_first` or `round_robin`): Generate the base leaf types (e.g., `delay_type = "connecting"`) and the initial, descriptive `delay_reason` (e.g., `"subchannel connecting: TCP handshake in progress"`).
```
AgraVator: "Generate the base leaf types" -> "Generate the base delay types"

A121-rpc-delay-observability.md
mbissa: Updated.

A121-rpc-delay-observability.md
```diff
@@ -125,24 +125,24 @@ Recorded on the `grpc.client.attempt.delay.duration` histogram. These represent
...
 *   **Priority Policy**: Prepends the active priority tier index, resulting in composed types like:
     *   `p0:connecting`
-    *   `p1:subchannel_state_mismatch`
-    *   `p0:picker_failing_with_wait_for_ready`
-*   **Pass-Through Container Policies**: Policies like `xds_cluster_manager`, `weighted_target`, and `rls` **do not prepend any prefix or wrap the delay type**. They simply bubble up the child's `grpc.delay_type` (e.g., `"connecting"`) directly as-is.
+*   **Pass-Through Container Policies**: Policies like `xds_cluster_manager`, `weighted_target`, and `rls` **do not prepend any prefix or wrap the delay type**. They bubble up the child's `grpc.delay_type` (e.g., `"connecting"`) as-is.
+
+Because only the priority policy contributes a prefix and it contributes exactly one (its active tier index), the resulting `grpc.delay_type` cardinality stays bounded; prefixes do not stack across nested pass-through containers. Implementations should keep the prefix a small bounded token (the tier index) so metric cardinality remains low; detailed per-container structure belongs in the `grpc.delay_reason` span event, not the metric label.
```
AgraVator, line 130: We are not gonna limit the nesting for priority policies right(as in px:py:pz.....:connecting ) ??

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
+# A121: RPC Delay Observability
```
markdroth, line 1: Please remove the `#` here, since this results in (a) a double underline in the formatted view and (b) adding the title to the table of contents, which isn't useful.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+Before an RPC attempt can be sent over the network, the client channel must perform several operations, including resolving the target name, parsing service configurations, instantiating load balancing policies, and obtaining a connectivity picker. If any of these phases stall—such as during a slow DNS lookup, a Route Lookup Service (RLS) control-plane query, or a Cluster Discovery Service (CDS) metadata fetch—the RPC is delayed. To the application, this appears as high latency or a timeout. However, because current telemetry lacks visibility into these resolution and routing states, developers cannot distinguish between a slow network, a slow backend, or a channel initialization delay. This is particularly challenging for clients using the Route Lookup Service (RLS), where diagnosing RLS-related hangs or isolating whether delays stem from pending route lookups requires manual debugging.
```
markdroth, line 18: A few comments here:
- Parsing the service config is synchronous and does not impose any delay.  Same thing for instantiating LB policies.  I don't think we should mention either one here, since they're not relevant.
- RLS is not actually a public API and therefore shouldn't be covered too much here.  There should be only one place where we discuss it, which is where we talk about the changes in the RLS LB policy, and we should have a note on that section that says something like "(not a public API, not covered in any gRFC)".

```suggestion
When a gRPC channel is first created, it is in IDLE state, meaning that it has not done name resolution or attempted to connect to any endpoints.  When the application sends an RPC on the channel, the channel will leave IDLE state and attempt to connect.  Any RPCs sent on the channel while it is attempting to connect will be queued until the channel becomes connected, which results in increased latency for these RPCs.  However, because current telemetry lacks visibility into these connecting delays, developers cannot distinguish between a slow network, a slow backend, or a channel initialization delay.
```

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+While there are many potential sources of delay along the client-side pipeline (including interceptor execution, credential fetching, and filter evaluation), this proposal focuses specifically on introducing observability for the two most common bottlenecks: **name resolution** and **load balancing pick** delays. This proposal establishes a generic telemetry framework extensible to other client-side delays in the future.
```
markdroth, line 20: Interceptor execution and filter evaluation are not relevant.  We care only about things that do network I/O.

Credential fetching is a good example to cite here.  Another example is retry delays.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+When the delay resolves or transitions to a different delay type, the active child span is closed.
+
+#### Span Event Schema Specification
+To ensure tracing backends can programmatically parse `"Delay state transition"` events, the event MUST conform to the following structural schema:
+```json
+{
+  "name": "Delay state transition",
+  "attributes": {
+    "grpc.delay_type": "string (the active delay type, e.g., 'connecting')",
```
markdroth, line 67: Why do we need this field?  The purpose of this event is to record the delay reason, but the delay type should be constant for the entire span, so it seems like we shouldn't need to record the delay type in each event.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+##### 1. Call Delay Duration Histogram
+
+| Field | Value |
+|---|---|
+| **Name** | `grpc.client.call.delay.duration` |
+| **Type** | Float64 Histogram |
+| **Unit** | `s` (seconds) |
+| **Description** | EXPERIMENTAL. Time an RPC spent waiting at the call level before an attempt was initiated, such as waiting for name resolution. |
+| **Labels** | `grpc.target`, `grpc.delay_type` |
+| **Optional Labels** | `grpc.method` |
```
markdroth, line 86: Why is the method label optional?  I think it's a required label on all of the A66 metrics.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+##### 2. Attempt Delay Duration Histogram
+
+| Field | Value |
+|---|---|
+| **Name** | `grpc.client.attempt.delay.duration` |
+| **Type** | Float64 Histogram |
+| **Unit** | `s` (seconds) |
+| **Description** | EXPERIMENTAL. Time an RPC attempt spent waiting for a load balancing pick or connection establishment. |
+| **Labels** | `grpc.target`, `grpc.delay_type` |
+| **Optional Labels** | `grpc.method` |
```
markdroth, line 99: Same question here: Why is this label optional?

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+##### 1. Call-Level Delay Types
+Recorded on the `grpc.client.call.delay.duration` histogram. These represent delays that occur before an individual attempt is created:
+*   `"resolving"`: The channel is delayed waiting for name resolution.
+
+##### 2. Attempt-Level Delay Types
+Recorded on the `grpc.client.attempt.delay.duration` histogram. These represent delays that occur during a specific RPC attempt:
+*   `"connecting"`: The attempt is delayed waiting for subchannel connection establishment or picker initialization.
+*   `"rls_lookup_pending"`: Specifically for Route Lookup Service (RLS) control-plane cache-miss query lookups.
+*   `"cds_dynamic_discovery"`: Specifically for xDS Cluster Discovery Service (CDS) dynamic metadata resource fetches.
```
markdroth, line 120: This should reference A61.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+##### 3. Composed Attempt-Level Types
+Structural container policies prepend their logical prefixes to the base attempt-level types using a colon separator:
+*   **Priority Policy**: Prepends the active priority tier index, resulting in composed types like:
+    *   `p0:connecting`
+*   **Pass-Through Container Policies**: Policies like `xds_cluster_manager`, `weighted_target`, and `rls` **do not prepend any prefix or wrap the delay type**. They bubble up the child's `grpc.delay_type` (e.g., `"connecting"`) as-is.
+
+Because only the priority policy contributes a prefix and it contributes exactly one (its active tier index), the resulting `grpc.delay_type` cardinality stays bounded; prefixes do not stack across nested pass-through containers. Implementations should keep the prefix a small bounded token (the tier index) so metric cardinality remains low; detailed per-container structure belongs in the `grpc.delay_reason` span event, not the metric label.
```
markdroth, line 130: Instead of "tier index", please say "numeric priority".

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+Recorded on the `grpc.client.attempt.delay.duration` histogram. These represent delays that occur during a specific RPC attempt:
+*   `"connecting"`: The attempt is delayed waiting for subchannel connection establishment or picker initialization.
+*   `"rls_lookup_pending"`: Specifically for Route Lookup Service (RLS) control-plane cache-miss query lookups.
+*   `"cds_dynamic_discovery"`: Specifically for xDS Cluster Discovery Service (CDS) dynamic metadata resource fetches.
+*   `"subchannel_state_mismatch"`: The target subchannel transitioned out of `READY` (e.g., disconnected or went to `TRANSIENT_FAILURE`), but the active channel picker has not yet been updated with a new picker. *Note: This delay type is generated by the channel intercepting a stale pick, not by the picker itself.*
+*   `"picker_failing_with_wait_for_ready"`: A `wait_for_ready` RPC is buffered/queued because the picker is in `TRANSIENT_FAILURE`, waiting for the next connection attempt to succeed. *Note: This delay type is generated by the channel wrapper detecting a `wait_for_ready` RPC when the picker returns an error, keeping pickers ignorant of `wait_for_ready` semantics.*
```
markdroth, line 122: In several places, the document uses the term "channel wrapper".  I'm not sure what that means.  I think it should just say "channel" instead.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+1.  **Leaf Pickers** (such as `pick_first` or `round_robin`): Generate the base delay types (e.g., `delay_type = "connecting"`) and the initial `delay_reason` (e.g., `"subchannel connecting: TCP handshake in progress"`).
+2.  **Container Pickers**: Intercept the child's deferred pick result as it bubbles up the picker tree. The `priority` picker prepends its active tier index to the type (e.g., producing `"p0:connecting"`). A pass-through picker (such as `xds_cluster_manager`) forwards the child's `delay_type` unmodified, but can add its own structural details to the `delay_reason`.
+3.  **The Channel Wrapper**: Receives the composed `delay_type` and `delay_reason` strings from the root picker and passes them to the tracer (`recordAttemptDelayStart`) for a single delay_type segment (which is identified by a change in the delay_type value). The channel wrapper is also responsible for intercepting specific channel-level states (e.g., generating `"picker_failing_with_wait_for_ready"` when a picker returns an error for a `wait_for_ready` RPC, or `"subchannel_state_mismatch"` when a transport disconnects before the picker is updated).
+
+---
```
markdroth, line 138: Please remove.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+### Related Proposals:
+* **[gRPC A66: OpenTelemetry Metrics][A66]**: Establishes base OpenTelemetry metrics.
+* **[gRPC A72: OpenTelemetry Tracing][A72]**: Establishes tracing spans and events.
+* **[gRPC A56: Priority LB Policy][A56]**: Establishes the priority load balancing policy and child failover mechanics.
+* **[gRPC A28: xDS Traffic Splitting and Routing][A28]**: Establishes the `xds_cluster_manager` and `weighted_target` container policies.
+
+
+## Proposal
+
+### 1. Telemetry Schema
```
markdroth: General comment: I think there are way too many headings and sub-headings in this doc, to the point where the outline is so noisy that it's very hard to understand the structure of the document.  The organization makes it hard to follow, and there's also a lot of duplication of information across various sections.

I think we should restructure this into a smaller set of headings that describe the design in an order that can be easily understood by the reader, as follows:

- Anatomy of a Delay (describes call vs. attempt, delay type vs delay reason, and general rules for determining when a delay starts and ends)
- Metric Schema (defines the metrics and labels)
- Tracing Schema (defines the trace span and events)
- Call Tracer API changes (describes the new API methods and explains what the caller is responsible for vs. what the call tracer impl is responsible for; can refer back to the metric schema and tracing schema sections above)
- Resolver Delays (covers the channel's behavior for resolver delays, how it interacts with the resolver and the call tracer)
- LB Pick Delays
  - LB Picker API changes
    - (subsections for each individual LB policy, describing what delay types and reasons will be generated by that policy)
  - Channel Behavior for LB pick delays (covers the channel's interaction with the picker and the call tracer, and channel-specific cases like `subchannel_state_mismatch` and `picker_failing_with_wait_for_ready`)

Also, please remove numbers from all of the section and subsection headers.  The outline already shows the nesting, we don't need a number to do this.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+    *   **Scope**: Measures delays that occur within the lifecycle of a specific RPC attempt, primarily during the load balancing pick loop and connection establishment.
+    *   **Metric**: `grpc.client.attempt.delay.duration` (Float64 Histogram).
+    *   **Tracing Span**: A child span named **`"Attempt Delay"`**, created as a child of the individual Attempt span.
+
+#### Span Attributes and Events
+
+For both layers, a child span is created for each distinct `grpc.delay_type`, and transitions within that type are recorded as span events:
+
+*   **`grpc.delay_type` (Span Attribute & Metric Label)**: A low-cardinality, composed string representing the type of the delay. The value format is `[parent_prefix:]base_type` (e.g., `connecting`, `p0:connecting`, or `rls_lookup_pending`). It is recorded as a **span attribute** on the child span and as a **metric label** on the corresponding duration histogram.
+*   **`grpc.delay_reason` (Span Event Attribute)**: A high-cardinality string capturing runtime details (such as target names, subchannel IP addresses, or connection error messages).
```
markdroth, line 52:
```suggestion
Each delay will result in a data point being added to the relevant histogram and a new trace span.  Every delay has two attributes:

*   **`grpc.delay_type`**: A low-cardinality string representing the type of the delay. See below for supported values. This value is constant over the lifetime of an individual delay; if the value changes, that implies that the original delay has ended and a new delay has started.  For metrics, this value is used in a metric label.  For tracing, this value is an attribute on the trace span.
*   **`grpc.delay_reason`**: A high-cardinality string capturing runtime details (such as target names, subchannel IP addresses, or connection error messages).  This value can change over the lifetime of a delay.  This value is not used in metrics.  In tracing, this value will be recorded in an event on the trace span.
```

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+*   **`grpc.delay_reason` (Span Event Attribute)**: A high-cardinality string capturing runtime details (such as target names, subchannel IP addresses, or connection error messages). 
+
+When a delay begins, the tracer creates the child span (`"Call Delay"` or `"Attempt Delay"`) carrying the `grpc.delay_type` attribute. It also records the initial reason as a span event named `"Delay state transition"` containing the `grpc.delay_reason` string.
+
+If the state changes but the `grpc.delay_type` remains the same (e.g., a priority policy fails over or RLS changes backend targets), the tracer emits a new `"Delay state transition"` span event on the active child span with the updated `grpc.delay_reason` string, without recreating the span.
+
+
+When the delay resolves or transitions to a different delay type, the active child span is closed.
+
+#### Span Event Schema Specification
```
markdroth, line 61: This section shows the event schema but not the schema for the span itself.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+### 1. Telemetry Schema
+
+To measure client-side delays, we introduce a duration histogram and a dedicated tracing span at the **Call Level** (for name resolution delays), and another histogram and span at the **Attempt Level** (for load balancing pick delays).
+
+We split observability data between low-cardinality metrics and high-cardinality debug tracing across two layers:
+
+1.  **Call Level Observability (Channel-level Delays)**:
+    *   **Scope**: Measures delays that occur at the channel level before an individual network attempt is initiated. This is primarily used to observe name resolution and configuration resolution delays.
+    *   **Metric**: `grpc.client.call.delay.duration` (Float64 Histogram).
+    *   **Tracing Span**: A child span named **`"Call Delay"`**, created as a child of the main RPC Call span.
```
markdroth, line 41: Why does the span name need to be different for call-level vs. attempt-level delays?  It should be obvious whether a given delay is call-level or attempt-level based on whether the delay span is nested under the call span or the attempt span.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+| **Name** | `grpc.client.attempt.delay.duration` |
+| **Type** | Float64 Histogram |
+| **Unit** | `s` (seconds) |
+| **Description** | EXPERIMENTAL. Time an RPC attempt spent waiting for a load balancing pick or connection establishment. |
+| **Labels** | `grpc.target`, `grpc.delay_type` |
+| **Optional Labels** | `grpc.method` |
+| **Bucket Boundaries** | Same as A66 latency buckets: 0, 0.00001, 0.00005, 0.0001, 0.0003, 0.0006, 0.0008, 0.001, 0.002, 0.003, 0.004, 0.005, 0.006, 0.008, 0.01, 0.013, 0.016, 0.02, 0.025, 0.03, 0.04, 0.05, 0.065, 0.08, 0.1, 0.13, 0.16, 0.2, 0.25, 0.3, 0.4, 0.5, 0.65, 0.8, 1, 2, 5, 10, 20, 50, 100 |
+| **Default Enabled** | `false` (experimental, opt-in) |
+
+### 2. Delay Types & Reasons
```
markdroth, line 103: Instead of defining all delay types and reasons in a single section like this and dividing them up into fairly arbitrary categories, I think we should instead have a section for each LB policy and for the channel, indicating which delay types will be reported by each piece of code, and showing examples of delay reasons for each one.

See my comment above for suggested organization.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+##### Category C: Subchannel Connection Scenarios (Type: `"connecting"`)
+Recorded when an RPC attempt is queued waiting for a subchannel to establish a transport:
+*   **Idle Subchannel Trigger**: The picker selects a subchannel that is in `IDLE` state, triggering a new connection attempt. The reason string captures this transition (e.g., `"subchannel was idle, triggering connection attempt"`).
+*   **Active TCP/TLS Handshake**: The subchannel is actively connecting. The reason string describes that the TCP or TLS handshake is in progress, and can include the target subchannel's remote address (e.g., `"subchannel connecting: TCP handshake in progress to 192.168.1.50:8080"`).
+*   **A105 Connection Scaling**: Per gRFC A105, an established subchannel connection is scaling up to add additional connections to satisfy high concurrent streams. The reason string captures this scaling state (e.g., `"scaling up additional subchannel connections per A105 limits"`).
```
markdroth, line 158: This design includes only resolver and LB pick delays, so subchannel connection scaling delays are not in scope.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+*   **Waiting for Health Report**: The physical transport connection is established, but the picker is blocked waiting for the initial Out-of-Band (OOB) health check stream to return `SERVING` (e.g., `"transport connected, waiting for initial health check report"`).
+*   **Subchannel Non-Existence**: The load balancer has not yet created or resolved the target subchannel. The reason string describes the missing subchannel state.
+*   **Round Robin / WRR Scenarios**: The policy is connecting because it is waiting for any of its configured subchannels to become ready. The reason string describes the policy's connectivity attempt and can list the subchannels being attempted.
+*   **Ring Hash Scenarios**: The hash ring is waiting for subchannels in the ring to connect. The reason string describes that the ring is connecting and can list the specific ring nodes being attempted.
+*   **xDS Override Host Scenarios**: The policy is attempting to connect to a specific overridden host (subchannel) but the host is not yet connected. The reason string explains the overridden host state.
```
markdroth, line 163: This should reference A55, A60, and A75.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+*   **Scope Resolution**: The decision of whether a delay segment is Call-Level (recorded to `grpc.client.call.delay.duration`) or Attempt-Level (recorded to `grpc.client.attempt.delay.duration`) is determined entirely by the scope of the tracer object on which the callbacks are invoked. Call-level delays (such as `"resolving"`) are invoked strictly on the call-scoped tracer object, while attempt-level delays (such as `"connecting"`) are invoked strictly on the attempt-scoped tracer object. This mapping is enforced statically by the respective language API signatures.
```
markdroth, line 190: Rather than calling this out as a general principle, I think we should just cover this in the section that talks about the channel's behavior for both resolution delays and LB pick delays.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+
+### 3. Telemetry API
+
+#### Lifecycle & State Machine
+
+The client channel, load balancer, and telemetry tracer coordinate synchronously to record delays during routing.
+
+##### 1. Timer Orchestration
+The channel owns the **lifecycle** of a delay segment (deciding when it starts, transitions, and ends) while the telemetry plugin owns the **timing** of that segment (it timestamps the start on `recordDelayStart`, computes the elapsed duration on `recordDelayEnd`, and emits the histogram). The channel therefore does not maintain a separate duration clock; "logical timer" below refers to this plugin-side measurement, delimited by the channel's start/end signals.
```
markdroth, line 187: This can be covered in the section about the call tracer API.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+*   **Scope Resolution**: The decision of whether a delay segment is Call-Level (recorded to `grpc.client.call.delay.duration`) or Attempt-Level (recorded to `grpc.client.attempt.delay.duration`) is determined entirely by the scope of the tracer object on which the callbacks are invoked. Call-level delays (such as `"resolving"`) are invoked strictly on the call-scoped tracer object, while attempt-level delays (such as `"connecting"`) are invoked strictly on the attempt-scoped tracer object. This mapping is enforced statically by the respective language API signatures.
+
+##### 2. Emission
```
markdroth, line 192: The info here should be moved to the section defining the call tracer API changes.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+*   **Simultaneous Metric & Span Close**: Upon receiving the `recordDelayEnd` signal, the telemetry plugin **simultaneously ends the child trace span and emits the final duration metric** to the corresponding duration histogram (`grpc.client.call.delay.duration` or `grpc.client.attempt.delay.duration`), measuring the elapsed time of the logical timer.
+
+##### 3. Cancellation & Timeout
+If an active RPC call or attempt is cancelled (due to client cancellation, `DEADLINE_EXCEEDED` timeouts, or channel shutdown) while a delay is logically active:
+*   The core channel **MUST** invoke the tracer's end callback (`recordCallDelayEnd()` or `recordAttemptDelayEnd()`), which finalizes the plugin-side timing for the open segment. 
+*   This ensures the telemetry plugin can capture the elapsed duration up to the point of failure, close the trace span, and emit the partial duration metric, guaranteeing that bottlenecks preceding a failure remain observable.
+
+##### 4. Retries & Hedging
```
markdroth, line 202: I think this section can just be removed.  The structure it describes already exists and isn't something new in this design.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+##### 2. Emission
+Both metric duration recording and trace span management are delegated to the registered telemetry plugin (e.g., OpenTelemetry) via the tracer API:
+*   **Span Lifecycle**: Upon receiving the `recordDelayStart` signal, the telemetry plugin instantiates the child trace span. Subsequent `recordDelayReasonChanged` calls append `"Delay state transition"` events to the active span.
+*   **Simultaneous Metric & Span Close**: Upon receiving the `recordDelayEnd` signal, the telemetry plugin **simultaneously ends the child trace span and emits the final duration metric** to the corresponding duration histogram (`grpc.client.call.delay.duration` or `grpc.client.attempt.delay.duration`), measuring the elapsed time of the logical timer.
+
+##### 3. Cancellation & Timeout
+If an active RPC call or attempt is cancelled (due to client cancellation, `DEADLINE_EXCEEDED` timeouts, or channel shutdown) while a delay is logically active:
+*   The core channel **MUST** invoke the tracer's end callback (`recordCallDelayEnd()` or `recordAttemptDelayEnd()`), which finalizes the plugin-side timing for the open segment.
```
markdroth, line 199: Do we actually need to explicitly close the span in this case?  Note that when the RPC is cancelled or hits its deadline, we are already notifying the call tracer about that.  We could just say that if there's an open delay when the call gets cancelled, the call tracer will automatically terminate it, updating both metrics and tracing accordingly.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+*   **Go**: introduces a **new** call-scoped and attempt-scoped tracer API (gRPC-Go's V2 stats handler). A new API is required because the V1 `stats.Handler` is attempt-scoped only and cannot host a call-level hook. The broader V2 API is out of scope here beyond the delay hooks it must expose.
+*   **Java**: **reuses the existing tracer types** but adds new no-op-default methods to them — attempt hooks on `ClientStreamTracer`, call hooks on `ClientStreamTracer.Factory` — and reuses the existing name-resolution-delay plumbing (`ClientStreamTracer.NAME_RESOLUTION_DELAYED` and the `createPendingStream()` callback) for the `"resolving"` segment. The attempt-level `"connecting"` delay is recorded directly on the attempt-scoped `ClientStreamTracer` instance.
+*   **C++ (Core)**: **reuses the existing annotation framework**, adding only a new `DelayAnnotation` subtype that replaces the current free-form `"Delayed name resolution complete."` / `"Delayed LB pick complete."` string annotations.
```
markdroth, line 255: I don't think we can do this via simple annotations.  I think we'll instead add new methods to the call tracer API.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+}  // namespace grpc_core
+```
+A new `AnnotationType::kDelay` is added to the annotation-type enum (before the trailing `kDoNotUse_MustBeLast` sentinel). The OpenTelemetry plugin reads `stage()`/`type()`/`reason()` to open, update, and close the delay child span.
+
+### Feature Flag
+
+
+All delay metrics, tracing, and API hooks will be guarded by a feature flag:
+* **Go/Java Env Var**: `GRPC_EXPERIMENTAL_ENABLE_DELAY_OBSERVABILITY` (Default: `false`)
+* **C++ Core Experiment**: register `client_delay_observability` in `experiments.yaml`; gate via the generated `IsClientDelayObservabilityEnabled()` accessor (core uses generated per-experiment accessors, not string-keyed lookups).
```
markdroth, line 297: I don't think we need any env var guard for this, since the new functionality is going to be configured locally instead of being triggered by remote input.

(Also, if we did need an env var guard, we'd use the same approach in C-core rather than using our experiment framework, so that interop tests have a single way to enable them across all languages.)

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Metric Registration and Recording
+
+Metric registration and recording will follow the established architectural patterns defined in [gRPC A66][A66] and [gRPC A94][A94]. The channel and attempts will delegate duration measurement directly to the registered telemetry plugin (e.g., OpenTelemetry) to avoid code duplication across the core stack. The specific plugin API registrations are omitted here for brevity but will conform to language-idiomatic OTel APIs (e.g., `RegisterFloat64Histo` in Go, `registerDoubleHistogram` in Java, and `RegisterDoubleHistogram` in C++).
```
markdroth, line 301: I'm not sure what this is actually saying.  Isn't this already implicit in the design described above?

I'm also not sure why this is referencing A94, since that doesn't seem relevant here.

markdroth (commented):
The design looks good overall.  I have one or two minor questions about design details, but they don't affect the overall shape of the design.  Most of my comments are about the organization of the document -- the content is good but is not organized in a way that makes it easy to understand.

A121-rpc-delay-observability.md
mbissa: correct, fixed.

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
mbissa: Done — rewrote Background using your suggested text and dropped service-config parsing and LB instantiation. RLS is now discussed only in the rls policy section, with the "(not a public API, not covered by any gRFC)" note.

A121-rpc-delay-observability.md
mbissa: Done — removed interceptor/filter; now cites credential fetching and retry delays as the other network-I/O sources.

A121-rpc-delay-observability.md
mbissa: Agreed — removed. The event carries only grpc.delay_reason; delay_type is a span attribute (constant for the span).

A121-rpc-delay-observability.md
mbissa: Fixed — grpc.method is now a required label on both histograms.

A121-rpc-delay-observability.md
mbissa: Fixed — grpc.method is now a required label on both histograms.

A121-rpc-delay-observability.md
mbissa: Added. I placed it in the pick_first/connecting section, since A61 (IPv4/IPv6 dualstack) is about the connection attempts

A121-rpc-delay-observability.md
mbissa: Done

A121-rpc-delay-observability.md
mbissa: Done.

A121-rpc-delay-observability.md
mbissa: Removed — the composition-mechanics section is gone; that behavior is now folded into the priority section and "Channel Behavior for LB Pick Delays."

A121-rpc-delay-observability.md
mbissa: Done — restructured into your outline (Anatomy of a Delay / Metric Schema / Tracing Schema / Call Tracer API Changes / Resolver Delays / LB Pick Delays → Picker API + per-policy + Channel Behavior) and removed all section numbers.

A121-rpc-delay-observability.md
mbissa: Done — adopted your suggested text in "Anatomy of a Delay."

A121-rpc-delay-observability.md
mbissa: Added — Tracing Schema now describes the span itself (name Delay, nested under the call or attempt span, with the grpc.delay_type attribute) alongside the event schema.

A121-rpc-delay-observability.md
mbissa: Agreed — there's now a single span name (Delay); call-level vs attempt-level is distinguished by whether it's nested under the call span or the attempt span.

A121-rpc-delay-observability.md
mbissa: Done — delay types and example reasons are now defined per policy (and for the channel). I also added a small summary table of the base delay_type values for reference, which points back to those sections.

A121-rpc-delay-observability.md
mbissa: Removed — connection scaling is out of scope.

A121-rpc-delay-observability.md
mbissa: Added — the xds_override_host section now references A55, A60, and A75.

A121-rpc-delay-observability.md
mbissa: Done — dropped the general-principle framing; the channel's behavior is covered directly in "Resolver Delays" and "Channel Behavior for LB Pick Delays."

A121-rpc-delay-observability.md
mbissa: Moved — now under "Call Tracer API Changes" (caller vs. tracer-impl responsibilities).

A121-rpc-delay-observability.md
mbissa: Moved — the scope-determines-histogram rule is now in "Call Tracer API Changes."

A121-rpc-delay-observability.md
mbissa: Removed

A121-rpc-delay-observability.md
mbissa: Good point — no. The doc now says that if a delay is open when the RPC is cancelled or times out, the call tracer terminates it automatically (updating metrics and tracing); the channel doesn't explicitly close it.

A121-rpc-delay-observability.md
mbissa: Done — C++ now adds new methods to the call and attempt tracer interfaces instead of using annotations.

A121-rpc-delay-observability.md
mbissa: Done — removed the env-var guard. The metrics remain experimental/off-by-default via the standard OTel per-metric enablement.

A121-rpc-delay-observability.md
mbissa: Removed the "Metric Registration and Recording" section (and the A94 reference) — agreed it was already implicit in the design above.

jggj-c (approved):
...

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
-*   **`grpc.delay_reason` (Span Event Attribute)**: A high-cardinality string capturing runtime details (such as target names, subchannel IP addresses, or connection error messages). 
+The base `grpc.delay_type` values are summarized below for reference; each is defined in full — with example reasons — in the per-policy sections that follow ([Resolver Delays](#resolver-delays), [LB Pick Delays](#lb-pick-delays)):
 
-When a delay begins, the tracer creates the child span (`"Call Delay"` or `"Attempt Delay"`) carrying the `grpc.delay_type` attribute. It also records the initial reason as a span event named `"Delay state transition"` containing the `grpc.delay_reason` string.
+| `grpc.delay_type` | Scope | Meaning | Generated by |
+|---|---|---|---|
+| `resolving` | call | waiting on name resolution | channel (resolver) |
+| `connecting` | attempt | waiting on a subchannel connection | leaf pickers (`pick_first`, `round_robin`, `ring_hash`, `weighted_round_robin`, `xds_override_host`) |
```
markdroth, line 48: "leaf picker" is the wrong term here, because the xds_override_host policy is not a leaf policy -- and technically, every other policy here except pick_first is a petiole policy, not a leaf policy.

I suggest instead saying something like "LB policies that trigger connection attempts".

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
+| `rls_lookup_pending` | attempt | waiting on an RLS control-plane lookup | `rls` |
+| `cds_dynamic_discovery` | attempt | waiting on a CDS resource | `cds` |
+| `subchannel_state_mismatch` | attempt | picked subchannel left `READY` before it could be used | channel |
+| `picker_failing_with_wait_for_ready` | attempt | `wait_for_ready` RPC queued on a failing picker | channel |
 
-If the state changes but the `grpc.delay_type` remains the same (e.g., a priority policy fails over or RLS changes backend targets), the tracer emits a new `"Delay state transition"` span event on the active child span with the updated `grpc.delay_reason` string, without recreating the span.
+The `priority` policy composes the attempt-level values by prepending its numeric priority (e.g. `"0:connecting"`), and these prefixes stack when priority policies are nested (e.g. `"0:1:connecting"`), so the composed set is not statically enumerable.
```
markdroth, line 54: The phrase "attempt-level value" seems confusing here, since this has nothing to do with the question of call vs. attempt.

I suggest instead saying something like "The priority policy prepends the numeric priority to the delay type returned by the child policy".

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
-##### 3. Composed Attempt-Level Types
-Structural container policies prepend their logical prefixes to the base attempt-level types using a colon separator:
-*   **Priority Policy**: Prepends the active priority tier index, resulting in composed types like:
-    *   `p0:connecting`
-*   **Pass-Through Container Policies**: Policies like `xds_cluster_manager`, `weighted_target`, and `rls` **do not prepend any prefix or wrap the delay type**. They bubble up the child's `grpc.delay_type` (e.g., `"connecting"`) as-is.
-
-Because only the priority policy contributes a prefix and it contributes exactly one (its active tier index), the resulting `grpc.delay_type` cardinality stays bounded; prefixes do not stack across nested pass-through containers. Implementations should keep the prefix a small bounded token (the tier index) so metric cardinality remains low; detailed per-container structure belongs in the `grpc.delay_reason` span event, not the metric label.
+```json
+{
+  "name": "Delay state transition",
```
markdroth, line 96: This isn't actually a state transition if it's the initial delay reason recorded when the span starts.  Maybe this should instead just say something like "Delay triggered".

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
+Whether a delay is call-level or attempt-level is reflected by which histogram it is recorded to and by whether its trace span is nested under the call span or the attempt span.
 
-#### Span Attributes and Events
+Each delay results in a data point being added to the relevant histogram and a new trace span. Every delay has two attributes:
 
-For both layers, a child span is created for each distinct `grpc.delay_type`, and transitions within that type are recorded as span events:
+*   **`grpc.delay_type`**: A low-cardinality string representing the type of the delay. See below for supported values. This value is constant over the lifetime of an individual delay; if the value changes, that implies that the original delay has ended and a new delay has started. For metrics, this value is used in a metric label. For tracing, this value is an attribute on the trace span.
+*   **`grpc.delay_reason`**: A high-cardinality string capturing runtime details (such as target names, subchannel IP addresses, or connection error messages). This value can change over the lifetime of a delay. This value is not used in metrics. In tracing, this value is recorded in an event on the trace span.
```
markdroth, line 41: It might be useful to explicitly say that this field is intended primarily for human understanding, not for automated interpretation.

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
+**Caller (channel) responsibilities.** The channel detects when a delay begins and ends, chooses the `delay_type` and `delay_reason`, and calls `Start` when a `delay_type` first appears or changes, `ReasonChanged` when only the reason changes, and `End` when the wait resolves. The scope — and therefore which histogram the delay is recorded to (see [Metric Schema](#metric-schema)) — is chosen by whether the channel calls the method on the call-scoped or the attempt-scoped tracer.
 
-##### Category A: Resolver Scenarios (Type: `"resolving"`)
-*   **DNS Resolver Pending**: The channel is waiting for the initial name resolution query to complete. The reason string should describe the pending resolver query (e.g., `"waiting for DNS query to complete for target example.com"`).
+**Call tracer (telemetry plugin) responsibilities.** On `Start`, the plugin opens the `Delay` span (see [Tracing Schema](#tracing-schema)) and begins timing. On `ReasonChanged`, it adds a `Delay state transition` event. On `End`, it closes the span and records the elapsed duration to the corresponding histogram. The plugin owns timing, so the `End` methods carry no duration argument.
```
markdroth, line 118: Instead of "begins timing", I suggest saying "records the delay type and start time of the delay".

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
+### Resolver Delays
 
-##### Category F: Pass-Through Container Policy Scenarios (Type: `"connecting"` bubbled up directly)
-Policies like `xds_cluster_manager`, `weighted_target`, and `rls` do not modify the metric `grpc.delay_type` (it remains strictly `"connecting"` or whatever leaf type is bubbled up from the child).
-However, to provide visibility in tracing, the parent container policy's structural details (such as the target cluster name, RLS route shard, or weight group) are recorded inside the trace event's free-form description or as span attributes (e.g., `"waiting for child cluster 'cluster-abc' to connect"`).
+The channel records a call-level delay of type **`resolving`** while an RPC is blocked waiting for name resolution. When an RPC is sent while the channel has not yet received its first resolver result, the channel calls `recordCallDelayStart("resolving", reason)` on the call tracer; when the resolver delivers its first result and the RPC can proceed, the channel calls `recordCallDelayEnd()`. If resolution has already completed by the time the RPC is sent, no delay is recorded.
```
markdroth, line 130: Note that if the RPC has wait_for_ready=true, then the delay reason might be the resolver having reported a failure rather than it not yet having reported any result.  That failure should be reported as the delay reason.  And if the resolver returns a second failure, the channel will also call `recordCallDelayReasonChanged()` to indicate the new failure reason.

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
-*   **Scope Resolution**: The decision of whether a delay segment is Call-Level (recorded to `grpc.client.call.delay.duration`) or Attempt-Level (recorded to `grpc.client.attempt.delay.duration`) is determined entirely by the scope of the tracer object on which the callbacks are invoked. Call-level delays (such as `"resolving"`) are invoked strictly on the call-scoped tracer object, while attempt-level delays (such as `"connecting"`) are invoked strictly on the attempt-scoped tracer object. This mapping is enforced statically by the respective language API signatures.
+When a picker cannot return a ready connection, it defers the pick. Two values are added to the pick-result type each runtime already returns, populated on the deferred path:
```
markdroth, line 140: Please use the word "queue" instead of "defer" for delayed LB picks, since that's the term we use for this.

markdroth, line 140: Please explicitly state that these are both string values.

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
-| `recordCallDelayEnd()` | call | close the span; emit `grpc.client.call.delay.duration` |
-| `recordAttemptDelayStart(delay_type, reason)` | attempt | open the `"Attempt Delay"` span; begin timing |
-| `recordAttemptDelayReasonChanged(reason)` | attempt | append a `"Delay state transition"` event |
-| `recordAttemptDelayEnd()` | attempt | close the span; emit `grpc.client.attempt.delay.duration` |
+Generate `delay_type = "connecting"`. Example reasons:
 
-The **scope of the receiving tracer object** (call-scoped vs attempt-scoped) statically determines which histogram a segment is recorded to ([3.1](#lifecycle--state-machine), Scope Resolution). Each runtime binds these operations onto its client-side telemetry API as follows:
+*   Subchannel is `IDLE`, triggering a new connection attempt (e.g. `"subchannel was idle, triggering connection attempt"`).
+*   TCP/TLS handshake in progress, optionally with the remote address (e.g. `"subchannel connecting: TCP handshake in progress to 192.168.1.50:8080"`). For dualstack targets, the connection attempts across address families follow [gRFC A61][A61].
```
markdroth: Why are we referencing A61 here?  I'm not sure it's relevant.

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
 *   **Go**: `balancer.PickResult` (deferred pick = the `ErrNoSubConnAvailable` return).
 *   **Java**: `PickResult` (deferred pick = `PickResult.withNoResult()`).
-*   **C++ (Core)**: the `PickResult::Queue` variant (deferred pick = `Queue`).
+*   **C++ (Core)**: the `PickResult::Queue` variant.
 
-##### 2. Tracer-Side API
+The delay types and example reasons generated by each LB policy follow. These examples are non-exhaustive; implementations are encouraged to append additional debug details to the reason.
 
-The channel drives the telemetry plugin through six logical operations, split across the two existing telemetry scopes. The end callbacks carry no duration argument because the plugin owns timing ([3.1](#lifecycle--state-machine)):
+##### `pick_first` and `round_robin`
```
markdroth, line 153: I think we probably don't want to combine PF and RR in the same section, because they're actually two different cases.

Conversely, it seems a bit redundant to have other petiole policies like WRR and ring_hash in their own sections, since basically every petiole policy is the same thing.

So I suggest one section for pick_first, and another for all of the petiole policies (RR, WRR, and ring_hash).  It should probably also include least_request (and reference gRFC A48).

A121-rpc-delay-observability.md
markdroth, line 158: I don't think we need to differentiate between IDLE and CONNECTING here.  In both cases, we can just say that we're waiting for a connection.

For pick_first, the delay reason will want to be something like "waiting for subchannel to connect to ${address}".

Have we talked about how much detail we want to see in the delay reason for the petiole policies?  They will often be connecting to multiple endpoints in parallel, so I'm thinking that they should just delegate to one of the connecting endpoints at random and wrap the delay reason reported by the PF child policy, so they'd say something like "waiting for connections to endpoints (example: waiting for subchannel to connect to ${address})".

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
-*   **`grpc.delay_reason` (Span Event Attribute)**: A high-cardinality string capturing runtime details (such as target names, subchannel IP addresses, or connection error messages). 
+The base `grpc.delay_type` values are summarized below for reference; each is defined in full — with example reasons — in the per-policy sections that follow ([Resolver Delays](#resolver-delays), [LB Pick Delays](#lb-pick-delays)):
 
-When a delay begins, the tracer creates the child span (`"Call Delay"` or `"Attempt Delay"`) carrying the `grpc.delay_type` attribute. It also records the initial reason as a span event named `"Delay state transition"` containing the `grpc.delay_reason` string.
+| `grpc.delay_type` | Scope | Meaning | Generated by |
+|---|---|---|---|
+| `resolving` | call | waiting on name resolution | channel (resolver) |
+| `connecting` | attempt | waiting on a subchannel connection | leaf pickers (`pick_first`, `round_robin`, `ring_hash`, `weighted_round_robin`, `xds_override_host`) |
+| `rls_lookup_pending` | attempt | waiting on an RLS control-plane lookup | `rls` |
+| `cds_dynamic_discovery` | attempt | waiting on a CDS resource | `cds` |
```
markdroth, line 50: Suggest referencing A74 here.

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
 
-**Channel-Intercepted Delays:** We implement specific delay types like `picker_failing_with_wait_for_ready` and `subchannel_state_mismatch` as channel-level interceptions rather than picker-generated types. This keeps leaf pickers ignorant of `wait_for_ready` semantics and transport-level race conditions, enforcing a separation of concerns where the picker only reports its current state, and the channel dictates queueing behavior.
+##### `cds`
 
-**Delegated Recording:** We implement duration recording at the client channel wrapper level rather than inside specific load balancing policies because only the client channel manages the buffering, queueing, and context cancellation lifecycles of RPC calls. This decouples policy-level state reporting from duration measurement.
+Generates `delay_type = "cds_dynamic_discovery"` while waiting for a dynamic CDS cluster resource definition. The reason can include the targeted cluster name (e.g. `"waiting for CDS resource definition for cluster cluster_abc"`).
```
markdroth: This can also reference A74.

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
+| `subchannel_state_mismatch` | attempt | picked subchannel left `READY` before it could be used | channel |
+| `picker_failing_with_wait_for_ready` | attempt | `wait_for_ready` RPC queued on a failing picker | channel |
 
-If the state changes but the `grpc.delay_type` remains the same (e.g., a priority policy fails over or RLS changes backend targets), the tracer emits a new `"Delay state transition"` span event on the active child span with the updated `grpc.delay_reason` string, without recreating the span.
+The `priority` policy composes the attempt-level values by prepending its numeric priority (e.g. `"0:connecting"`), and these prefixes stack when priority policies are nested (e.g. `"0:1:connecting"`), so the composed set is not statically enumerable.
 
+A delay begins when the channel starts waiting and ends when the wait resolves, when the `grpc.delay_type` changes (which ends the current delay and starts a new one), or when the RPC is cancelled or reaches its deadline.
 
-When the delay resolves or transitions to a different delay type, the active child span is closed.
+### Metric Schema
```
markdroth, line 58: This needs to include a statement about metric stability, similar to this:

https://github.com/mbissa/proposal/blob/0baf481e1964bd6fc4725e85357a61b7fe5ed0a1/A78-grpc-metrics-wrr-pf-xds.md#metric-stability

A121-rpc-delay-observability.md
markdroth: Sorry, I steered you wrong here: I meant that it should reference A74, not A61.  The reference should be where we discuss the `cds_dynamic_discovery` delay type, not where we discuss pick_first.

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
-*   It is written as a **human-readable string** (e.g. `"subchannel is connecting"` or `"waiting for DNS query to complete"`), **not** a `snake_case` token or closed enum.
-*   It is designed to contain **high-cardinality metadata** (such as specific subchannel IP addresses, target names, cache keys, or raw connection error messages) to provide diagnostics inside the trace span.
+| Method | Scope | Called when |
+|---|---|---|
+| `recordCallDelayStart(delay_type, reason)` | call | a call-level delay begins (or the `delay_type` changes) |
+| `recordCallDelayReasonChanged(reason)` | call | the reason changes within the same `delay_type` |
+| `recordCallDelayEnd()` | call | the call-level delay resolves |
+| `recordAttemptDelayStart(delay_type, reason)` | attempt | an attempt-level delay begins (or the `delay_type` changes) |
+| `recordAttemptDelayReasonChanged(reason)` | attempt | the reason changes within the same `delay_type` |
+| `recordAttemptDelayEnd()` | attempt | the attempt-level delay resolves |
```
markdroth, line 114: For LB pick delays, I think this API will require both the channel and the call tracer impl to store the current delay type.  The channel will need to store it to know whether it should call `recordAttemptDelayStart()` or `recordAttemptDelayReasonChanged()` on each LB pick, and the call tracer will need to store it to know what label value to record for the metric when the delay ends.

I think it might be better to structure this API such that we need to store the delay type in only one place, either the channel or the call tracer impl.  It would probably make more sense to do it in the channel, because there could be multiple call tracers, and it would use more memory if each one has to do this itself.

To do it in the channel, we could just add the delay_type parameter to the `recordAttemptDelayReasonChanged()` and `recordAttemptDelayEnd()` methods.

markdroth, line 114: I don't think we need different method names for call-level and attempt-level delays.  We'll already differentiate them based on whether they are called on the call-level tracer or the attempt-level tracer.  I think both objects can have the same methods.  So they can just be called `RecordDelayStart()`, `RecordDelayReasonChanged()`, and `RecordDelayEnd()`.

A121-rpc-delay-observability.md
```diff
@@ -1,324 +1,222 @@
...
 
+*   **`subchannel_state_mismatch`**: When the picker returns a ready subchannel but it has transitioned out of `READY` before the RPC can use it (a race before the picker is updated), the channel re-queues the RPC on the same attempt and records this delay type. It waits for the next picker and re-picks; no new attempt is created. The delay ends when a re-pick assigns a subchannel that is actually ready.
+
+## Rationale
+
+**Call vs. attempt metrics.** We split metrics into two histograms to differentiate channel-initialization bottlenecks (name resolution) from per-attempt routing bottlenecks (LB pick and connection). Combining them would obscure whether a delay occurred before or during connection establishment.
```
markdroth, line 203: I'm not sure this bullet is useful.  It's not really about whethre the delay occurred before or during connection establishment -- that in and of itself wouldn't necessarily have been a reason to split the metrics.

Given that both metrics and tracing already differentiate calls from attempts, and that hedging means that there can be multiple attempts in parallel, I think this split was just the most natural approach.  I don't think it's something we need to explicitly justify.

markdroth (commented):
This looks much better!  Remaining comments are mostly minor.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,222 @@
...
+| `recordAttemptDelayStart(delay_type, reason)` | attempt | an attempt-level delay begins (or the `delay_type` changes) |
+| `recordAttemptDelayReasonChanged(reason)` | attempt | the reason changes within the same `delay_type` |
+| `recordAttemptDelayEnd()` | attempt | the attempt-level delay resolves |
+
+**Caller (channel) responsibilities.** The channel detects when a delay begins and ends, chooses the `delay_type` and `delay_reason`, and calls `Start` when a `delay_type` first appears or changes, `ReasonChanged` when only the reason changes, and `End` when the wait resolves. The scope — and therefore which histogram the delay is recorded to (see [Metric Schema](#metric-schema)) — is chosen by whether the channel calls the method on the call-scoped or the attempt-scoped tracer.
+
+**Call tracer (telemetry plugin) responsibilities.** On `Start`, the plugin opens the `Delay` span (see [Tracing Schema](#tracing-schema)) and begins timing. On `ReasonChanged`, it adds a `Delay state transition` event. On `End`, it closes the span and records the elapsed duration to the corresponding histogram. The plugin owns timing, so the `End` methods carry no duration argument.
```
AgraVator, line 118: How are we getting the initial delay_reason in tracing since Span will only contain delay_type ? Shouldn't an event with `Delay state transition` be fired at the start as well ?

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
mbissa: done.

mbissa: done.

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
mbissa: I have simplified the examples under petiole policies, I think we can leave that detail for implementations to populate based on the specific policy.

A121-rpc-delay-observability.md
mbissa: We do mention that they are experimental in the table, but I have added it nevertheless.

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
mbissa: Done.

mbissa: done.

A121-rpc-delay-observability.md
mbissa: removed.

A121-rpc-delay-observability.md
mbissa: Updated.

A121-rpc-delay-observability.md
mbissa: done.

A121-rpc-delay-observability.md
```diff
@@ -38,26 +38,27 @@ Whether a delay is call-level or attempt-level is reflected by which histogram i
...
 
-The `priority` policy composes the attempt-level values by prepending its numeric priority (e.g. `"0:connecting"`), and these prefixes stack when priority policies are nested (e.g. `"0:1:connecting"`), so the composed set is not statically enumerable.
+The `priority` policy prepends the numeric priority to the delay type returned by the child policy (e.g. `"0:connecting"`); these prefixes stack when priority policies are nested (e.g. `"0:1:connecting"`), so the composed set is not statically enumerable.
 
 A delay begins when the channel starts waiting and ends when the wait resolves, when the `grpc.delay_type` changes (which ends the current delay and starts a new one), or when the RPC is cancelled or reaches its deadline.
 
 ### Metric Schema
 
-The following client-side per-call metrics are registered, extending the instrumentation framework defined in [gRPC A66][A66].
+All metrics added in this proposal will start as experimental and therefore off by default. The long term goal will be to de-experimentalize them and have them be on by default, but the exact criteria for that change are TBD.
```
markdroth, line 60: You added a section on this below, so I don't think it needs to be stated here as well.  I would just remove this line.

A121-rpc-delay-observability.md
```diff
@@ -102,111 +107,109 @@ Each value of `grpc.delay_reason` (the initial reason when the span opens, and e
...
-|---|---|---|
-| `recordCallDelayStart(delay_type, reason)` | call | a call-level delay begins (or the `delay_type` changes) |
-| `recordCallDelayReasonChanged(reason)` | call | the reason changes within the same `delay_type` |
-| `recordCallDelayEnd()` | call | the call-level delay resolves |
-| `recordAttemptDelayStart(delay_type, reason)` | attempt | an attempt-level delay begins (or the `delay_type` changes) |
-| `recordAttemptDelayReasonChanged(reason)` | attempt | the reason changes within the same `delay_type` |
-| `recordAttemptDelayEnd()` | attempt | the attempt-level delay resolves |
+| Method | Called when |
+|---|---|
+| `RecordDelayStart(delay_type, reason)` | a delay begins, or the `delay_type` changes (which ends the previous delay and starts a new one) |
```
markdroth, line 114: I don't think the "(which ends the previous delay and starts a new one)" part makes sense.  If the channel is the only thing storing the current delay type, then the call tracer cannot automatically end the previous delay, because it doesn't know what the delay type was previously.

I think we need to require that if the delay type changes, the channel must call `RecordDelayEnd()` with the old delay type before calling `RecordDelayStart()` with the new delay type.

A121-rpc-delay-observability.md
```diff
@@ -102,111 +107,109 @@ Each value of `grpc.delay_reason` (the initial reason when the span opens, and e
...
+| `RecordDelayEnd(delay_type)` | the delay resolves |
 
-**Caller (channel) responsibilities.** The channel detects when a delay begins and ends, chooses the `delay_type` and `delay_reason`, and calls `Start` when a `delay_type` first appears or changes, `ReasonChanged` when only the reason changes, and `End` when the wait resolves. The scope — and therefore which histogram the delay is recorded to (see [Metric Schema](#metric-schema)) — is chosen by whether the channel calls the method on the call-scoped or the attempt-scoped tracer.
+**Caller (channel) responsibilities.** The channel is the single owner of the current `delay_type`: it stores it, chooses the `delay_type` and `delay_reason`, and passes the `delay_type` on every call so the call tracer never has to store it. It calls `RecordDelayStart` when a `delay_type` first appears or changes, `RecordDelayReasonChanged` when only the reason changes, and `RecordDelayEnd` when the delay resolves. The scope — and therefore which histogram the delay is recorded to (see [Metric Schema](#metric-schema)) — is chosen by whether the channel calls the method on the call-scoped or the attempt-scoped tracer.
```
markdroth, line 118: As per my comment above, please change "first appears or changes" to just "first appears".

A121-rpc-delay-observability.md
```diff
@@ -102,111 +107,109 @@ Each value of `grpc.delay_reason` (the initial reason when the span opens, and e
...
 
-**Call tracer (telemetry plugin) responsibilities.** On `Start`, the plugin opens the `Delay` span (see [Tracing Schema](#tracing-schema)) and begins timing. On `ReasonChanged`, it adds a `Delay state transition` event. On `End`, it closes the span and records the elapsed duration to the corresponding histogram. The plugin owns timing, so the `End` methods carry no duration argument.
+**Call tracer (telemetry plugin) responsibilities.** On `RecordDelayStart`, the plugin opens the `Delay` span (see [Tracing Schema](#tracing-schema)) and records the delay type and start time of the delay. On `RecordDelayReasonChanged`, it adds a `Delay triggered` event. On `RecordDelayEnd`, it closes the span and records the elapsed duration to the histogram identified by the supplied `delay_type`. Because `delay_type` is supplied on every call, the plugin does not need to store it, and the plugin owns timing (the methods carry no duration argument).
```
markdroth, line 120: The call tracer will also add a `Delay triggered` event in `RecordDelayStart()`.

A121-rpc-delay-observability.md
```diff
@@ -102,111 +107,109 @@ Each value of `grpc.delay_reason` (the initial reason when the span opens, and e
...
 
 **Cancellation and deadlines.** The channel already notifies the call tracer when an RPC is cancelled or reaches its deadline. If a delay is open at that point, the call tracer automatically terminates it — closing the span and recording the partial duration — so the channel does not need to explicitly end open delays on these paths.
```
markdroth, line 122: This can no longer be the case if we're saying that only the channel will store the current delay type.  We must require that the channel will call `RecordDelayEnd()` in this case.

A121-rpc-delay-observability.md
```diff
@@ -102,111 +107,109 @@ Each value of `grpc.delay_reason` (the initial reason when the span opens, and e
...
 *   **Java**: new methods on the existing `ClientStreamTracer` (attempt scope) and `ClientStreamTracer.Factory` (call scope).
 *   **C++ (Core)**: new methods on the call and attempt tracer interfaces.
 
 ### Resolver Delays
 
-The channel records a call-level delay of type **`resolving`** while an RPC is blocked waiting for name resolution. When an RPC is sent while the channel has not yet received its first resolver result, the channel calls `recordCallDelayStart("resolving", reason)` on the call tracer; when the resolver delivers its first result and the RPC can proceed, the channel calls `recordCallDelayEnd()`. If resolution has already completed by the time the RPC is sent, no delay is recorded.
+The channel records a call-level delay of type **`resolving`** while an RPC is blocked waiting for name resolution. When an RPC is sent while the channel has not yet received its first resolver result, the channel calls `RecordDelayStart("resolving", reason)` on the call tracer; when the resolver delivers its first result and the RPC can proceed, the channel calls `RecordDelayEnd("resolving")`. If resolution has already completed by the time the RPC is sent, no delay is recorded.
```
markdroth, line 132: s/delivers/delivers/

A121-rpc-delay-observability.md
```diff
@@ -102,111 +107,109 @@ Each value of `grpc.delay_reason` (the initial reason when the span opens, and e
...
 
 ##### `xds_cluster_manager` and `weighted_target`
 
 These pass-through container policies do not modify the metric `delay_type`; they forward the child's value. They may add their own structural details (such as the target cluster name or weight group) to the `delay_reason`.
 
 #### Channel Behavior for LB Pick Delays
 
-For picker-generated delays, the channel reads the deferred pick's `delay_type` and `delay_reason` and drives the call tracer per the [Call Tracer API](#call-tracer-api-changes): `Start` when the `delay_type` first appears or changes, `ReasonChanged` when only the reason changes, and `End` when a pick assigns a ready subchannel.
+For picker-generated delays, the channel reads the queued pick's `delay_type` and `delay_reason` and drives the call tracer per the [Call Tracer API](#call-tracer-api-changes): `RecordDelayStart` when the `delay_type` first appears or changes, `RecordDelayReasonChanged` when only the reason changes, and `RecordDelayEnd` when a pick assigns a ready subchannel.
```
markdroth, line 194: As above, please change "first appears or changes" to just "first appears".

markdroth (commented):
This is getting closer!

A121-rpc-delay-observability.md
AgraVator: How are we planning to gate the delay spans and events ??

markdroth: I don't think we need to gate this functionality at all, because it's not going to be triggered by any input that is external to the process.

AgraVator: So, just to confirm anyone who updates to a branch which has this will get delay spans by default ?

markdroth: Yes.

A121-rpc-delay-observability.md
mbissa: Updaed.

A121-rpc-delay-observability.md
mbissa: removed.

A121-rpc-delay-observability.md
mbissa: we still call it when the delay type changes, only we need to call RecordDelayEnd() too in that case - I have updated it that way. let me know if it works.

A121-rpc-delay-observability.md
mbissa: updated.

A121-rpc-delay-observability.md
mbissa: ?

A121-rpc-delay-observability.md
mbissa: Updated same as above.

A121-rpc-delay-observability.md
mbissa: Updated the cancellation section.

A121-rpc-delay-observability.md
```diff
@@ -0,0 +1,226 @@
...
+
+| Method | Called when |
+|---|---|
+| `RecordDelayStart(delay_type, reason)` | a delay begins, or the `delay_type` changes (after calling `RecordDelayEnd(delay_type)` for the previous delay type) |
+| `RecordDelayReasonChanged(delay_type, reason)` | the reason changes within the same `delay_type` |
+| `RecordDelayEnd(delay_type)` | the delay resolves or the `delay_type` changes (followed by RecordDelayStart) |
+
+**Caller (channel) responsibilities.** The channel is the single owner of the current `delay_type`: it stores it, chooses the `delay_type` and `delay_reason`, and passes the `delay_type` on every call so the call tracer never has to store it. It calls `RecordDelayStart` when a `delay_type` first appears or changes (in which case the channel ends the previous delay first with a call to `RecordDelayEnd`), `RecordDelayReasonChanged` when only the reason changes, and `RecordDelayEnd` when the delay resolves. The scope — and therefore which histogram the delay is recorded to (see [Metric Schema](#metric-schema)) — is chosen by whether the channel calls the method on the call-scoped or the attempt-scoped tracer.
```
markdroth, line 117: I think it's clearer to describe this purely in terms of the lifecycle of a delay, and mention in only one place that `delay_type` changing means the end of one delay and the start of a new one.

```suggestion
| `RecordDelayStart(delay_type, reason)` | a delay begins |
| `RecordDelayReasonChanged(delay_type, reason)` | the reason changes within the same `delay_type` |
| `RecordDelayEnd(delay_type)` | a delay ends |

**Caller (channel) responsibilities.** The channel is the single owner of the current `delay_type`: it stores it, chooses the `delay_type` and `delay_reason`, and passes the `delay_type` on every call so the call tracer never has to store it. It must call `RecordDelayStart` with the appropriate `delay_type` when a delay starts and `RecordDelayEnd` with the same `delay_type` when the delay ends. If the `delay_type` changes, that indicates that the original delay has ended and a new delay has started, so the channel must first call `RecordDelayEnd` with the old `delay_type` and then `RecordDelayStart` with the new `delay_type`. The channel must call `RecordDelayReasonChanged` when the `delay_type` does not change but the `delay_reason` does. The scope — and therefore which histogram the delay is recorded to (see [Metric Schema](#metric-schema)) — is chosen by whether the channel calls the method on the call-scoped or the attempt-scoped tracer.
```

A121-rpc-delay-observability.md
markdroth: Sorry, no idea what I was thinking there -- I must have misread something.  You can ignore. :)

markdroth (commented):
For future reference, please do not force-push to a branch once review has started, because that makes it basically impossible for the reviewer to see what's changed since the last review pass.  Thanks!

PR #558, opened 2026-08-01
A50: Exclude locally-initiated client cancellations from Call Counter (#12923)

AgraVator:
## Summary
Clarifies gRFC A50 (`A50-xds-outlier-detection.md`) to explicitly specify that locally-initiated client-side cancellations (hedging cancellations, application cancellation, and deadline exceeded) are excluded from the Outlier Detection Call Counter.

## Key Clarifications
1. **Picker**: Updates the counter in the map entry on request completion.
2. **Call Counter**: Specifies that the active bucket tracks successes and failures while excluding locally-initiated client-side cancellations (hedging cancellations, application cancellation, and deadline exceeded etc).
3. **Rationale**:
   - **Hedging & Application Cancellations**: Align with Envoy's `resetStream()` behavior (silently resetting the stream without recording endpoint errors), preventing false-positive ejections during hedging.
   - **Deadline Exceeded**: Differs from Envoy (which counts timeouts as failures for outlier detection). In gRPC, client-side deadline expiration cannot be reliably distinguished from other client cancellations cross-language, and client timeouts cannot be definitively attributed to an endpoint failure rather than network or client-side delays.

## Related Work
- gRPC-Java PR #12923: https://github.com/grpc/grpc-java/pull/12923/changes
- Envoy Outlier Detection: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier

### DIFF
--- a/A50-xds-outlier-detection.md
+++ b/A50-xds-outlier-detection.md
@@ -284,7 +284,7 @@ Envoy defines some errors as "external" and some as "local origin", and their sp
 
 ### Excluding Locally-Initiated Client-Side Cancellations
 
-Locally-initiated client-side cancellations (`hedging cancellations`, application `CANCELLED`, and `DEADLINE_EXCEEDED`) are excluded from outlier detection counting (`successCount` and `failureCount`). For locally-initiated cancellations (`hedging cancellations` of non-winning sibling attempts and application-initiated cancellations), this aligns with Envoy's `resetStream()` behavior and prevents false-positive ejections during hedging. For `DEADLINE_EXCEEDED`, this differs from Envoy (which counts deadline expirations as failures for outlier detection); however, because deadline expiration semantics vary between Envoy and gRPC, and implementation-wise it is much simpler to handle `DEADLINE_EXCEEDED` identically to cancellation, gRPC treats it as a client-side cancellation.
+Locally-initiated client-side cancellations (`hedging cancellations`, application `CANCELLED`, and `DEADLINE_EXCEEDED`) are excluded from outlier detection counting (`successCount` and `failureCount`). For locally-initiated cancellations (`hedging cancellations` of non-winning sibling attempts and application-initiated cancellations), this aligns with Envoy's `resetStream()` behavior and prevents false-positive ejections during hedging. For `DEADLINE_EXCEEDED`, this differs from Envoy (which counts deadline expirations as failures for outlier detection); however, because deadline expiration semantics vary between Envoy and gRPC, and because a deadline expiration cannot be definitively attributed to a server failure rather than a network delay, gRPC treats it as a client-side cancellation.
 
 ### Map Entry Source
 
--- a/A50-xds-outlier-detection.md
+++ b/A50-xds-outlier-detection.md
@@ -137,7 +137,7 @@ When the `outlier_detection` LB policy receives an address update, it will creat
 
 When the child policy asks for a subchannel, the `outlier_detection` will wrap the subchannel with a wrapper (see [Subchannel Wrapper section](#subchannel-wrapper)). Then, the subchannel wrapper will be added to the list in the map entry for its address, if that map entry exists. If there is no map entry, or if the subchannel is created with multiple addresses, the subchannel will be ignored for outlier detection. If that address is currently ejected, that subchannel wrapper's `eject` method will be called.
 
-The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, increment the corresponding counter in the map entry referenced by the subchannel wrapper that was picked. Locally-initiated client-side cancellations (hedging cancellations, application `CANCELLED`, and `DEADLINE_EXCEEDED`) MUST be excluded from `successCount` and `failureCount`. If both the `success_rate_ejection` and `failure_percentage_ejection` fields are unset in the configuration, the picker should not do that counting.
+The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, updates the counter in the map entry referenced by the subchannel wrapper that was picked. If both the `success_rate_ejection` and `failure_percentage_ejection` fields are unset in the configuration, the picker should not do that counting.
 
 The `outlier_detection` LB policy will have a timer that triggers on a period determined by the `interval` config option, and does the following:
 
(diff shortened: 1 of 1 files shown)


A50-xds-outlier-detection.md
```diff
@@ -282,6 +282,10 @@ Envoy's specification of outlier detection includes the ejection criteria Consec
 
 Envoy defines some errors as "external" and some as "local origin", and their specification of outlier detection allows separate configurations for handling each of them. gRPC does not separate errors that way, so there is no way to split them like that and handle those two categories separately.
 
+### Excluding Locally-Initiated Client-Side Cancellations
+
+Locally-initiated client-side cancellations (`hedging cancellations`, application `CANCELLED`, and `DEADLINE_EXCEEDED`) are excluded from outlier detection counting (`successCount` and `failureCount`). Although `DEADLINE_EXCEEDED` could be argued as an endpoint or network issue, implementation-wise it is treated as a client-side cancellation. This aligns with Envoy's `resetStream()` behavior and prevents false-positive ejections during hedging.
```
kannanjgithub: Envoy's `resetStream` behavior is for locally initiated cancellations, not for deadlines. We may in fact be differing from Envoy's deadline expiration counting as Envoy does count them as failures but the semantics vary for deadline expiration between Envoy and gRPC, that Mark mentioned in the chat discussion.

AgraVator: Thanks for the clarification! I have updated the Rationale section in `A50-xds-outlier-detection.md` (commit 0ec0d2c) to explicitly distinguish between locally-initiated cancellations and deadline expirations:
- For **hedging cancellations and application `CANCELLED`**, excluding them aligns with Envoy's `resetStream()` behavior (which silently resets the stream without incrementing outlier detection error counters).
- For **`DEADLINE_EXCEEDED`**, excluding it differs from Envoy (which counts timeouts as failures for outlier detection). We note that because deadline expiration semantics vary between Envoy and gRPC (e.g., gRPC lacks Envoy's per-try / response timeouts), and implementation-wise it is much simpler to handle `DEADLINE_EXCEEDED` identically to cancellation, gRPC treats it as a client-side cancellation.

kannanjgithub: Update the PR description as well, which currently mentions Envoy's `resetStream` behavior for both hedged cancellations and deadline expiry.

kannanjgithub (approved):

A50-xds-outlier-detection.md
```diff
@@ -137,7 +137,7 @@ When the `outlier_detection` LB policy receives an address update, it will creat
...
 
-The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, increment the corresponding counter in the map entry referenced by the subchannel wrapper that was picked. If both the `success_rate_ejection` and `failure_percentage_ejection` fields are unset in the configuration, the picker should not do that counting.
+The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, increment the corresponding counter in the map entry referenced by the subchannel wrapper that was picked. Locally-initiated client-side cancellations (`hedging cancellations`, application `CANCELLED`, and `DEADLINE_EXCEEDED`) MUST be excluded from `successCount` and `failureCount`. If both the `success_rate_ejection` and `failure_percentage_ejection` fields are unset in the configuration, the picker should not do that counting.
```
ejona86: application `CANCELLED` and `DEADLINE_EXCEEDED` shouldn't be in all-upper-case, as it makes it seem like this is based on the status code. Just describe them. Ditto for `hedging cancellations`; dunno why it is within backticks.

A50-xds-outlier-detection.md
```diff
@@ -137,7 +137,7 @@ When the `outlier_detection` LB policy receives an address update, it will creat
...
 
-The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, increment the corresponding counter in the map entry referenced by the subchannel wrapper that was picked. If both the `success_rate_ejection` and `failure_percentage_ejection` fields are unset in the configuration, the picker should not do that counting.
+The `outlier_detection` LB policy will provide a picker that delegates to the child policy's picker, and when the request finishes, increment the corresponding counter in the map entry referenced by the subchannel wrapper that was picked. Locally-initiated client-side cancellations (hedging cancellations, application `CANCELLED`, and `DEADLINE_EXCEEDED`) MUST be excluded from `successCount` and `failureCount`. If both the `success_rate_ejection` and `failure_percentage_ejection` fields are unset in the configuration, the picker should not do that counting.
```
ejona86, line 140: application CANCELLED and DEADLINE_EXCEEDED shouldn't be in all-upper-case, as it makes it seem like this is based on the status code. Just describe them. ("application cancellation, and deadline exceeded")

ejona86, line 140: Don't do MUST. This document is not using that style.

ejona86, line 140: Referencing `successCount` and `failureCount` before they are defined is confusing.

I think we could avoid the change in this paragraph. The Call Counter section is the one that looks to be really defining this. Maybe we can make this section more generic so it won't be assumed to be thorough by `s/increment the corresponding counter/updates the counter/`. Michael may have a preference.

At that point I think we remove defining `successCount` and `failureCount` as variables, and remove the reference to the variable names in the rationale.

A50-xds-outlier-detection.md
```diff
@@ -282,6 +282,10 @@ Envoy's specification of outlier detection includes the ejection criteria Consec
 
 Envoy defines some errors as "external" and some as "local origin", and their specification of outlier detection allows separate configurations for handling each of them. gRPC does not separate errors that way, so there is no way to split them like that and handle those two categories separately.
 
+### Excluding Locally-Initiated Client-Side Cancellations
+
+Locally-initiated client-side cancellations (hedging cancellations, application `CANCELLED`, and `DEADLINE_EXCEEDED`) are excluded from outlier detection counting (`successCount` and `failureCount`). For locally-initiated cancellations (hedging cancellations of non-winning sibling attempts and application-initiated cancellations), this aligns with Envoy's `resetStream()` behavior and prevents false-positive ejections during hedging. For `DEADLINE_EXCEEDED`, this differs from Envoy (which counts deadline expirations as failures for outlier detection); however, because deadline expiration semantics vary between Envoy and gRPC, and because a deadline expiration cannot be definitively attributed to a server failure rather than a network delay, gRPC treats it as a client-side cancellation.
```
ejona86, line 287: "semantics vary" is pretty vague. "gRPC can't reliably distinguish between cancellation and deadline exceeded cross-language" seems relevant.

kishorekasi: A race or network delay can result in client trigger `CANCELLED` due to deadline before server responds with `DEADLINE_EXCEEDED`. This can result in outlier detector not counting it. Aren't both scenarios same and counted equally? Ref: https://github.com/grpc/grpc-java/issues/13001

AgraVator: Hi, 
Client timeouts can stem from client-side queuing or transit delays rather than endpoint unhealthiness which should not be counted as failures.
cc @ejona86

jggj-c (approved):

A50-xds-outlier-detection.md
kishorekasi: Please let me know if my understanding is correct: once a physical attempt has been assigned to an endpoint, the client deadline can race with the server’s `DEADLINE_EXCEEDED` response. The same attempt may therefore be excluded as a client-side timeout or counted as a server response depending on which signal arrives first. Is the proposal intentionally excluding client/parent deadline expiry while counting a server-returned `DEADLINE_EXCEEDED`? We are hoping the API can expose enough information for a policy to optionally count deadline expiry for an active endpoint attempt, while still excluding explicit cancellations and non-winning retry or hedging attempts.

PR #559, opened 2026-08-18
A36 update: allow listener address to match any port

markdroth:
This will allow control planes to send the same listener resource to all server instances even if their ports are assigned dynamically.


ejona86 (approved):

ejona86:
CC @murgatroid99

dfawley (commented):
LGTM.  Should the title of the PR be:

A36 update: allow listener address to match any ~~address and~~ port

("A36 update: allow listener address to match any port")

?

markdroth:
Yeah, updated, thanks.  (Too late to change the merge commit, though.)

